PR #5, opened 2016-02-22
Add assembly task

ravwojdyla:


ravwojdyla:
One can run `assembly` on any project. `assembly` on `root` does almost nothing - propagation is disabled via

    aggregate in assembly := false

To build complete fat jar:

    sbt 'project scio-assembly' assembly

Issue #39, opened 2016-03-14
AvroHadoopFileSource does not allow specifying a reader schema


rijuk:
By default AvroKeyInputFormat uses the writer schema as the reader schema. AvroHadoopFileSource uses this default behavior.
While this is ok when reading a set of Avro files which are all written using the same schema, it'll break down when reading a set of files that do not share the exact same schema.

PR #40, opened 2016-03-14
Issue #39: Use AvroCoder's schema as the reader schema while reading …

rijuk:
…Avro files


nevillelyh:
Great, thanks!

nevillelyh:
@rijuk not sure if you're using Scio but it'd be great if you can add something to this: https://github.com/spotify/scio/wiki/Powered-By

PR #50, opened 2016-03-18
Add some extra logging to the it script

ravwojdyla:


ravwojdyla:
just here to get travis runs

PR #55, opened 2016-03-22
Add countMinSketch

ravwojdyla:


nevillelyh:
Let's leave these approximate algos to out unless there's a strong need? Spark only implements approx count, distinct, etc. for example.

A complete list of Algebird examples here and most can be implemented easily with `sum` or `aggregate`.
https://github.com/spotify/scio/blob/master/scio-examples/src/test/scala/com/spotify/scio/examples/extra/AlgebirdSpec.scala#L162

PR #59, opened 2016-03-23
HDFS user for Simple Auth

ravwojdyla:


ravwojdyla:
@nevillelyh - tried:
 * HADOOP_USER_NAME
 * UserGroupInformation.doAs

1st one seems less complex.

rgruener:
:+1: Awesome

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopUserUtils.java
```diff
@@ -0,0 +1,19 @@
+package com.google.cloud.dataflow.contrib.hadoop;
```
nevillelyh: License header?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -101,6 +101,15 @@ package object hdfs {
       self.context.makeFuture(HdfsTextTap(path))
     }
 
+    def saveAsHdfsTextFile(path: String, user: String): Future[Tap[String]] = {
```
nevillelyh: Instead of overloading the method with an extra argument, add it to the previous one with default `null`?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -111,7 +120,29 @@ package object hdfs {
...
+                                    job.getConfiguration)
+      self
+        .map(x => KV.of(new AvroKey(x), NullWritable.get()))
+        .applyInternal(Write.to(sink))
+      self.context.makeFuture(HdfsAvroTap[T](path, schema))
+    }
+
+    /** Save this SCollection as an Avro file on HDFS. */
+    // TODO: numShards
+    def saveAsHdfsAvroFile(path: String, schema: Schema, user: String): Future[Tap[T]] = {
```
nevillelyh: Same here, add `user: String = null` to the previous one. Also `schema` needs to be nullable too for `SpecificRecord`.

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/SimpleAuthHadoopFileSink.java
```diff
@@ -0,0 +1,89 @@
+package com.google.cloud.dataflow.contrib.hadoop;
```
nevillelyh: License header?

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopUserUtils.java
ravwojdyla: solved

PR #68, opened 2016-04-06
add HDFS support in Taps

nevillelyh:


scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -150,6 +150,27 @@ package object hdfs {
...
+    /** Get a `Future[Tap[T]]` for a text file on HDFS. */
+    def hdfsTextFile(path: String): Future[Tap[String]] =
+      self.mkTap(s"Text: $path", () => isPathDone(path), () => HdfsTextTap(path))
+
+    /** Get a `Future[Tap[T]]` for a Avro file on HDFS. */
+    def hdfsAvroFile[T: ClassTag](path: String, schema: Schema = null): Future[Tap[T]] =
+      self.mkTap(s"Avro: $path", () => isPathDone(path), () => HdfsAvroTap(path, schema))
+
+    private def isPathDone(path: String): Boolean = {
+      val job = Job.getInstance()
```
ravwojdyla: Is this job here created only to get configuration object out of it? then `new Configuration()` works too

nevillelyh: There're a few more occurrences, fix later?

ravwojdyla: I remember in Source/Sink `job` object is actually useful here and there. i'm ok with both now or later (as long as there is issue/todo open to clean it up)?

PR #69, opened 2016-04-06
Fix race conditions in DistCache

ravwojdyla:
There can be multiple DoFns/Threads trying to fetch the same data/files on the same
worker. To prevent from situation where some workers see incomplete data, and keep the
solution simple - let's synchronize on FileDistCache companion object (which is a singleton).
There is a downside - more specifically we might have to wait a bit longer then in more
optimal solution, but, simplicity > performance.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/DistCache.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/DistCache.scala
@@ -35,6 +35,8 @@ sealed trait DistCache[F] extends Serializable {
   def apply(): F
 }
 
+private[scio] object FileDistCache
+
 private[scio] abstract class FileDistCache[F](options: GcsOptions) extends DistCache[F] {
 
   override def apply(): F = data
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/DistCache.scala
```diff
@@ -35,6 +35,8 @@ sealed trait DistCache[F] extends Serializable {
   def apply(): F
 }
 
+private[scio] object FileDistCache {}
```
nevillelyh: you don't need the trailing `{}`?

scio-core/src/main/scala/com/spotify/scio/values/DistCache.scala
```diff
@@ -52,26 +54,40 @@ private[scio] abstract class FileDistCache[F](options: GcsOptions) extends DistC
     .as(classOf[GcsOptions])
 
   private def fetchFromGCS(uri: URI, prefix: String): File = {
-    val path = prefix + uri.getPath.split("/").last
-    val file = new File(path)
-
-    if (!file.exists()) {
+      val path = prefix + uri.getPath.split("/").last
```
nevillelyh: indentation is off?

PR #72, opened 2016-04-11
add IoCommands

nevillelyh:


nevillelyh:
@ravwojdyla

scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
```diff
@@ -227,13 +227,22 @@ class ScioILoop(scioClassLoader: ScioReplClassLoader,
...
+  }
+
   override def createInterpreter(): Unit = {
     super.createInterpreter()
     welcome()
     intp.beQuietDuring {
       addImports()
       createBigQueryClient()
       newScioCmdImpl("sc")
+      loadIoCommands()
```
ravwojdyla: I wonder if we should think about doing some of this stuff async to save on startup time?

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
```diff
@@ -0,0 +1,151 @@
...
+    if (path.startsWith("hdfs://")) {
+      fs.open(new Path(path))
+    } else if (path.startsWith("gs://")) {
+      Channels.newInputStream(gcsUtil.open(GcsPath.fromUri(path)))
+    } else {
+      new FileInputStream(path)
+    }
+
+  private def outputStream(path: String, contentType: String): OutputStream =
+    if (path.startsWith("hdfs://")) {
```
ravwojdyla: I have seen a couple of these - here and there - should we have a function that checks if paths is `hdfs` or `gs`?

scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
nevillelyh: Shouldn't be a huge difference? Let's not over optimize for now?

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
```diff
@@ -0,0 +1,151 @@
...
+    val fileWriter = new DataFileWriter[T](writer).create(schema, outputStream(path, BINARY))
+    data.foreach(fileWriter.append)
+    fileWriter.close()
+    println(s"${data.size} record${plural(data)} written to $path")
+  }
+
+  /** Write to a text file on local filesystem, GCS, or HDFS. */
+  def writeText(path: String, data: Seq[String]): Unit = {
+    IOUtils.writeLines(data.asJava, IOUtils.LINE_SEPARATOR, outputStream(path, TEXT))
+    println(s"${data.size} line${plural(data)} written to $path")
```
ravwojdyla: Should this be a log?

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
```diff
@@ -0,0 +1,151 @@
...
+    val cls = ScioUtil.classOf[T]
+    val (writer, schema) = if (classOf[SpecificRecordBase] isAssignableFrom cls) {
+      (new SpecificDatumWriter[T](cls), data.head.asInstanceOf[SpecificRecordBase].getSchema)
+    } else {
+      (new GenericDatumWriter[T](), data.head.asInstanceOf[GenericRecord].getSchema)
+    }
+    val fileWriter = new DataFileWriter[T](writer).create(schema, outputStream(path, BINARY))
+    data.foreach(fileWriter.append)
+    fileWriter.close()
+    println(s"${data.size} record${plural(data)} written to $path")
```
ravwojdyla: Log?

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
```diff
@@ -0,0 +1,151 @@
...
+
+/** Commands for simple file I/O in the REPL. */
+// scalastyle:off regex
+class IoCommands(options: PipelineOptions) {
+
+  private val TEXT = "text/plain"
+  private val BINARY = "application/octet-stream"
+
+  private val gcsUtil: GcsUtil = new GcsUtilFactory().create(options)
+  private val fs = FileSystem.get(new Configuration())
```
ravwojdyla: Something to keep in mind - probably not a problem here - but if someone was to use this functions in a long running service, configuration (on disk/resources) may change - thus you may need to refresh configuration. Just something to remember.

scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
```diff
@@ -227,13 +227,22 @@ class ScioILoop(scioClassLoader: ScioReplClassLoader,
     }
   }
 
+  private def loadIoCommands(): IR.Result = {
+    intp.interpret(
+      """
+        |val _ioCommands = new com.spotify.scio.repl.IoCommands(sc.options)
```
ravwojdyla: So if I would like to use IOCommands with different opts - i have to manually create it myself right? And by default you get the IOCommands from your startup opts?

build.sbt
```diff
@@ -319,7 +321,8 @@ lazy val scioRepl: Project = Project(
 ).settings(
   assemblyJarName in assembly := s"scio-repl-${version.value}.jar"
 ).dependsOn(
-  scioCore
+  scioCore,
+  scioHdfs
```
ravwojdyla: I kinda on the edge on this one - let's merge but - I think we will remove it next time we have an issue with repl and all the HDFS deps.

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
nevillelyh: Log will also print out a bunch of junk we don't need.

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
nevillelyh: Not a problem here since it's in scio-repl, not intended for a pipeline job.

nevillelyh:
Let's keep it open now until there's a real need for this.

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
sisidra: Any reason not to support all hadoop supported schemas via https://hadoop.apache.org/docs/r2.7.2/api/org/apache/hadoop/fs/FileSystem.html#get(java.net.URI, org.apache.hadoop.conf.Configuration) ?

nevillelyh: Yeah I should do that but I'm putting this PR on hold since adding hdfs dep may cause other problems.

ravwojdyla: @sisidra the only problem with that is that then you need to depend on hadoop deps - and that is something we mostly want to stay away from. But again if a job needs HDFS (or repl in this case, at the moment depends on Hadoop), then we already have hadoop, but (what @nevillelyh said).

scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
ravwojdyla: But log you can control, can't see a reason for `println`.

ravwojdyla:
👍

PR #85, opened 2016-04-15
Add compression to hdfs output

ravwojdyla:


scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -110,11 +110,28 @@ package object hdfs {
 
     /** Save this SCollection as a text file on HDFS. */
     // TODO: numShards
-    def saveAsHdfsTextFile(path: String, username: String = null): Future[Tap[String]] = {
+    def saveAsHdfsTextFile(path: String,
+                           username: String = null,
+                           conf: Configuration = null): Future[Tap[String]] = {
+      val _conf = Option(conf).getOrElse {
```
nevillelyh: no need to go full monadic here.
`val _conf = if (conf != null) { conf } else {` is less cryptic?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -110,11 +110,28 @@ package object hdfs {
...
+        newConf.set(FileOutputFormat.COMPRESS_CODEC, classOf[DefaultCodec].getName)
+        newConf.set(FileOutputFormat.COMPRESS_TYPE, "BLOCK")
+        newConf.setInt(AvroOutputFormat.DEFLATE_LEVEL_KEY, 6)
+        newConf
+      }
+
       val sink = if (username != null) {
-        new SimpleAuthHadoopFileSink(path, classOf[TextOutputFormat[NullWritable, Text]], username)
+        new SimpleAuthHadoopFileSink(path,
+                                     classOf[TextOutputFormat[NullWritable, Text]],
```
nevillelyh: maybe extract `classOf[TextOutputFormat[NullWritable, Text]]` in both if/else into a `val` and this will fit on 1 line.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -126,21 +143,31 @@ package object hdfs {
...
       val s = if (schema == null) {
         ScioUtil.classOf[T].getMethod("getClassSchema").invoke(null).asInstanceOf[Schema]
       } else {
         schema
       }
       AvroJob.setOutputKeySchema(job, s)
       val sink = if (username != null) {
-        new SimpleAuthHadoopFileSink(
-          path, classOf[AvroKeyOutputFormat[T]], conf, username)
+        new SimpleAuthHadoopFileSink(path, classOf[AvroKeyOutputFormat[T]], jobConf, username)
```
nevillelyh: same here with `classOf[AvroKeyOutputFormat[T]]`

scio-hdfs/src/test/scala/com/spotify/scio/hdfs/HdfsTapTest.scala
```diff
@@ -79,6 +114,32 @@ class HdfsTapTest extends PipelineSpec {
     FileUtils.deleteDirectory(dir)
   }
 
+  it should "support default compression with saveAsHdfsTextFile" in {
```
nevillelyh: Do you also want a `it should "support user conf with saveAsHdfsTextFile"` test?

Issue #84, opened 2016-04-14
BigQueryType.fromQuery inconsistent with the schema of destination table


nevillelyh:
```
@BigQueryType.fromQuery("""
SELECT
  es.user_id user_id,
  CONCAT(STRING(YEAR(MSEC_TO_TIMESTAMP(es.time*1000))),STRING(MONTH(MSEC_TO_TIMESTAMP(es.time*1000)))) time_window,
  es.play_track track_gid,
  te.vector.vector
FROM (
  SELECT
    user_id,
    time,
    play_track
  FROM
    [di-bigquery-data-commons:end_song_cleaned_decrypted.end_song_cleaned_decrypted_20160413]
  WHERE
    ms_played > 30000 ) es
INNER JOIN EACH [di-bigquery-data-commons:track_entity.track_entity_20160124] te
ON
  es.play_track==te.track_gid
WHERE
  te.vector.vector IS NOT NULL
LIMIT
  1000
""") class Row
```

Temp result copied to `gabocontinuousintegration:neville_test.nested_debug`

nevillelyh:
So `@BigQueryType.fromTable` works on the temp result table, but the view schema is different from the actual result table schema.

nevillelyh:
Also related: https://code.google.com/p/google-bigquery/issues/detail?id=251

nevillelyh:
There's a workaround being discussed in #24 that should also address this.

andrewsmartin:
This won't be an issue with SQL 2011 when views are supported.

nevillelyh:
We support SQL-2011 syntax now. Closing this.

PR #86, opened 2016-04-19
allow flatten results and destination table in BigQuery

nevillelyh:


nevillelyh:
This should provide a hackish workaround for #84

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -197,8 +197,29 @@ class BigQueryClient private (private val projectId: String,
...
+                     flattenResults: Boolean = false): Job =
+    queryIntoTable(sqlQuery, BigQueryIO.parseTableSpec(destinationTable), flattenResults)
+
+  /**
+   * Create a new query job and save results into a temporary table.
+   *
+   * No query will be executed if a cached table exists. Query execution is delayed until
+   * `waitForResult` is called on the result `QueryJob`.
+   */
+  def newQueryJob(sqlQuery: String, flattenResults: Boolean): QueryJob = {
```
nevillelyh: Not sure about the name and semantic of this method though, Any better idea?

scio-core/src/main/scala/com/spotify/scio/io/Taps.scala
```diff
@@ -70,9 +73,9 @@ trait Taps {
   private def tableExists(table: TableReference): Boolean =
     Try(BigQueryClient.defaultInstance().getTableSchema(table)).isSuccess
 
-  private def bigQueryTap(sqlQuery: String): BigQueryTap = {
+  private def bigQueryTap(sqlQuery: String, flattenResults: Boolean): BigQueryTap = {
```
ravwojdyla: I wonder if it makes sense to expose JobConfigurationQuery instead - what if we have to expose more options?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -197,8 +197,29 @@ class BigQueryClient private (private val projectId: String,
       getTable(table).getSchema
     }
 
-  /** Execute a query and save results into a temporary table. */
-  def queryIntoTable(sqlQuery: String): QueryJob = {
+  /** Make a query and save results to a destination table. */
+  def queryIntoTable(sqlQuery: String,
+                     destinationTable: TableReference,
+                     flattenResults: Boolean): Job = {
```
ravwojdyla: Why no default `false` value?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
ravwojdyla: Why no default `false` value?

ravwojdyla: I guess if this is the "user facing" method - maybe just `query`?

scio-core/src/main/scala/com/spotify/scio/io/Taps.scala
ravwojdyla: Should there be default `false` for `flattenResults`?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -314,15 +335,16 @@ class BigQueryClient private (private val projectId: String,
   }
 
   private def makeQueryJob(sqlQuery: String,
-                           destinationTable: TableReference): QueryJob = new QueryJob {
+                           destinationTable: TableReference,
+                           flattenResults: Boolean): QueryJob = new QueryJob {
```
ravwojdyla: Should there be default `false` for `flattenResults`?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
nevillelyh: This should probably not be user facing. Make it private and keep the name?
No need for default if not user facing.

Also maybe I should create a user facing version `def query` that blocks and return `Job` instead?

scio-core/src/main/scala/com/spotify/scio/io/Taps.scala
nevillelyh: Not sure about exposing `JobConfigurationQuery` since all other options must be set in a specific way in our case.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
nevillelyh: Again don't want default for private method, may lead to bugs.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
nevillelyh: `queryIntoTable` is overloaded and 2 versions cannot have the same number of arguments with optional `flattenResults`, or it'll cause ambiguous overloading error.

nevillelyh:
Simplified everything into one `def query` and limited the delayed version to inside `ScioContext` only. What do you think?

nevillelyh:
Close and re-open to test Trello integration.

PR #88, opened 2016-04-19
allow flatten results and dest table in BQ #84

nevillelyh:


nevillelyh:
@ravwojdyla what do you think?

ravwojdyla:
👍

PR #100, opened 2016-04-21
fix build issues

nevillelyh:
- scoverage resets scalaVersion to 2.11
- kantan.csv build error in 2.10
- GcsUtil credential error in integration build


scio-repl/src/main/scala/com/spotify/scio/repl/IoCommands.scala
```diff
@@ -45,7 +45,7 @@ class IoCommands(options: PipelineOptions) {
   private val BINARY = "application/octet-stream"
 
   // TODO: figure out how to support HDFS without messing up dependencies
-  private val gcsUtil: GcsUtil = new GcsUtilFactory().create(options)
+  private lazy val gcsUtil: GcsUtil = new GcsUtilFactory().create(options)
```
ravwojdyla: What what the problem here with `lazy`?

nevillelyh: Without it the REPL tries to load credentials on startup and throws an exception in Travis.

ravwojdyla:
👍

Issue #24, opened 2016-03-04
Macro generated BigQuery classes not working IntelliJ IDEA


nevillelyh:
Auto-completion and syntax highlighting doesn't work in IDEA.
https://youtrack.jetbrains.com/oauth?state=%2Fissue%2FSCL-8834

yonromai:
:+1:

nevillelyh:
http://blog.jetbrains.com/scala/2015/10/14/intellij-api-to-build-scala-macros-support/

nevillelyh:
Right now the closest workaround, with #111, would be:

Make a query in the REPL, fetch result schema with `@BigQueryType.fromTable()`, and get case class definitions with `toPrettyString()`.

```
scala> val table = bq.query("SELECT repository.* FROM [bigquery-public-data:samples.github_nested] LIMIT 1000")
...
table: com.google.api.services.bigquery.model.TableReference = {datasetId=scio_bigquery_staging_us, projectId=gabocontinuousintegration, tableId=scio_query_20160426204308_37560664}

scala> table.getProjectId + ":" + table.getDatasetId + "." + table.getTableId
res1: String = gabocontinuousintegration:scio_bigquery_staging_us.scio_query_20160426204308_37560664

scala> @BigQueryType.fromTable("gabocontinuousintegration:scio_bigquery_staging_us.scio_query_20160426204308_37560664") class Row
defined class Row
defined object Row
defined class Repository$

scala> Row.toPrettyString()
res2: String =
@BigQueryType.toTable
case class Row(repository: Option[Repository$2])
case class Repository$2(url: Option[String], has_downloads: Option[Boolean], created_at: Option[String], has_issues: Option[Boolean], description: Option[String], forks: Option[Int], fork: Option[Boolean], has_wiki: Option[Boolean], homepage: Option[String], integrate_branch: Option[String], master_branch: Option[String], size: Option[Int], private: Option[Boolean], name: Option[String], organization: Option[String], owner: Option[String], open_issues: Option[Int], watchers: Option[Int], pushed_at: Option[String], language: Option[String])
```

And then copy these case class definitions to the job code and use `sc.typedBigQuery[Row](query)`.

nevillelyh:
Documented in https://github.com/spotify/scio/wiki/BigQueryType-and-IntelliJ-IDEA
Doesn't look like it'll be fixed in IntelliJ any time soon so closing this for now.

PR #111, opened 2016-04-26
make @BigQueryType.toTable usable for input #24

nevillelyh:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaUtil.scala
```diff
@@ -23,15 +23,18 @@ import com.google.api.services.bigquery.model.{TableFieldSchema, TableSchema}
...
+  /** Convert schema to case class definitions. */
   def toPrettyString(schema: TableSchema, name: String, indent: Int): String = {
+    "@BigQueryType.toTable\n" +
     getCaseClass(schema.getFields, name, indent)
   }
 
   private def getRawType(tfs: TableFieldSchema, indent: Int): (String, Seq[String]) = {
     val name = tfs.getType match {
-      case "INTEGER" => "Int"
+      case "INTEGER" => "Long"
```
ravwojdyla: Why is this needed, I probably missed something.

nevillelyh: It's a bug in the original code. INTEGER in BQ actually means Long, and their FLOAT really is Double.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -77,7 +77,12 @@ package object experimental {
         val query = if (newSource != null) newSource else bqt.query.get
         self.bigQuerySelect(query).map(bqt.fromTableRow)
       } else {
-        throw new IllegalArgumentException(s"Missing table or query field in companion")
+        if (newSource != null) {
+          self.bigQuerySelect(newSource).map(bqt.fromTableRow)
```
ravwojdyla: `bqt` would not be derived from `table` nor `query` if it was manually created case class, and that is the use case this is solving - correct? In that case maybe a comment inside the code?

Should the doc above be fixed from:
```
sc.typedBigQuery[Row]("myproject:samples.gsod")
```
to
```
sc.typedBigQuery[Row]("select field from myproject:samples.gsod limit 1000")
```?

File: scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaUtil.scala (hunk 1 above)
Comment (ravwojdyla) [inline scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaUtil.scala]: Cool 👍

File: scio-core/src/main/scala/com/spotify/scio/experimental/package.scala [hunk 3]
```diff
@@ -176,7 +181,12 @@ package object experimental {
         val query = if (newSource != null) newSource else bqt.query.get
         self.getQueryRows(query).map(bqt.fromTableRow)
       } else {
-        throw new IllegalArgumentException(s"Missing table or query field in companion")
+        if (newSource != null) {
```
ravwojdyla: Same as above doc might need to be updated to include query instead of table.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
nevillelyh: Good point. I updated the docs and behavior so `newSource` can be either table or query if `query` or `table` is missing form companion object.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
nevillelyh: Updated doc and behavior.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -164,20 +176,33 @@ package object experimental {
      *
      * // Read from [myproject:samples.gsod] instead.
      * bq.getTypedRows[Row]("myproject:samples.gsod")
+     *
+     * // Read from a query instead.
+     * sc.getTypedRows[Row]("SELECT * FROM [publicdata:samples.gsod] LIMIT 1000")
```
ravwojdyla: Everything looks 👍 the only thing is that the example should really be a table if it does `SELECT * ...`, I know it's tiny, but even BQ doc says in most cases you should not do `SELECT * ...`

ravwojdyla:
👍

PR #115, opened 2016-04-27
Add hadoopDistCache - support all Hadoop filesystems

ravwojdyla:
hadoopDistCache will first upload artifacts from source to GCS and use
those via standard GCS based cache mechanism.

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/HadoopDistCacheExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/HadoopDistCacheExample.scala
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio._
+import com.spotify.scio.bigquery._
+import com.spotify.scio.hdfs._
+import com.spotify.scio.examples.common.ExampleData
+import org.joda.time.Instant
+
+/*
+SBT
+runMain
+  com.spotify.scio.examples.extra.DistCacheExample
+  --project=[PROJECT] --runner=DataflowPipelineRunner --zone=[ZONE]
+  --stagingLocation=gs://[BUCKET]/path/to/staging
+  --input=gs://dataflow-samples/wikipedia_edits/wiki_data-*.json
+  --output=gs://[BUCKET]/[PATH]/dist_cache_example
+*/
+
+// Use distributed cache inside a job
+object HadoopDistCacheExample {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    // declare a HDFS file to be distributed to all workers and logic to load the file
+    val dc = sc.hadoopDistCache(args.getOrElse("months",
+      "hdfs://dataflow-samples/samples/misc/months.txt")) { f =>
+        scala.io.Source.fromFile(f).getLines().map { s =>
+          val t = s.split(" ")
+          (t(0).toInt, t(1))
+        }.toMap
+    }
+
+    sc
+      .tableRowJsonFile(args.getOrElse("input", ExampleData.EXPORTED_WIKI_TABLE))
+      .map(row => new Instant(row.getLong("timestamp") * 1000L).toDateTime.getMonthOfYear)
+      .countByValue
+      // distributed cache available inside a transform
+      .map(kv => dc().getOrElse(kv._1, "unknown") + " " + kv._2)
+      .saveAsTextFile(args("output"))
+
+    sc.close()
+  }
+}
--- a/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
+++ b/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
@@ -17,18 +17,25 @@
 
 package com.spotify.scio
 
-import java.io.{InputStream, SequenceInputStream}
+import java.io.{File, InputStream, SequenceInputStream}
+import java.net.URI
+import java.nio.channels.Channels
+import java.security.PrivilegedAction
 import java.util.Collections
 
+import com.google.api.client.util.ByteStreams
 import com.google.cloud.dataflow.contrib.hadoop._
 import com.google.cloud.dataflow.contrib.hadoop.simpleauth._
 import com.google.cloud.dataflow.sdk.coders.AvroCoder
 import com.google.cloud.dataflow.sdk.io.{Read, Write}
+import com.google.cloud.dataflow.sdk.util.MimeTypes
+import com.google.cloud.dataflow.sdk.util.gcsfs.GcsPath
 import com.google.cloud.dataflow.sdk.values.KV
 import com.google.common.base.Charsets
+import com.google.common.hash.Hashing
 import com.spotify.scio.io.{Tap, Taps}
 import com.spotify.scio.util.ScioUtil
-import com.spotify.scio.values.SCollection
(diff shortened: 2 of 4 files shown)


ravwojdyla:
pls do not merge yet - i need to test `local` mode. Do comment if there is anything to change.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +118,61 @@ package object hdfs {
...
+        self.distCache(path)(initFn)
+      } else {
+        require(self.options.getStagingLocation != null,
+          "Staging directory not set - use `--stagingLocation`!")
+        require(path != null, "Artifact path can't be null")
+
+        logger.debug(s"Add '$path' to dist cache")
+
+        val _conf = Option(conf).getOrElse(new Configuration())
+        val target = new Path(self.options.getStagingLocation, path.split("/").last)
```
nevillelyh: Maybe "$stagingLocation/distcache/$file" for better organization?
Also the filename part should include a random hash to avoid collision?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -60,10 +68,21 @@ import scala.reflect.ClassTag
  */
 package object hdfs {
 
+  implicit class ScioHdfsUtil(val self: ScioUtil) {
```
nevillelyh: Where's `isHdfs` called?
No need to use implicit class if the method is only used within this file?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +118,61 @@ package object hdfs {
         .map(_.getKey.datum())
     }
 
+    /**
+     * Hadoop source flavoured distributed cache - any source supported by Hadoop Filesystem
```
nevillelyh: "flavored", this is murica!

```
Create a new [[com.spotify.scio.values.DistCache DistCache]] instance for a file on HDFS.
```

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +118,61 @@ package object hdfs {
...
+     * @param username optional Hadoop Simple Authentication remote username
+     */
+    def hadoopDistCache[F](path: String,
+                           conf: Configuration = null,
+                           username: String = null)
+                          (initFn: File => F): DistCache[F] = self.pipelineOp {
+      if (self.isTest) {
+        //TODO: if test - can the path be remote HDFS/S3 etc?
+        self.distCache(path)(initFn)
+      } else {
```
nevillelyh: This `else` block basically assumes the file is on HDFS, so it doesn't really support S3/local/GCS unless they are configured in Hadoop conf and the required jars are loaded?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +118,61 @@ package object hdfs {
         .map(_.getKey.datum())
     }
 
+    /**
+     * Hadoop source flavoured distributed cache - any source supported by Hadoop Filesystem
+     * abstraction can be used as a source for distributed cache.
+     *
+     * @param path to the artifact - it can be local/HDFS/S3/GCS ...
```
nevillelyh: Technically local/S3/GCS can be supported if Hadoop conf is set properly? I'd prefer to not mention them to avoid confusion.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
ravwojdyla: Good idea about extra `dir`. initially I was thinking we could just overwrite files, but hash is probably better.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
ravwojdyla: After `else` if can be anything supported by Hadoop - for example `S3` usually comes out of the box, but you have to add config for S3 to the `Configuration` object. So it's not only `HDFS`.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
ravwojdyla: 👍

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +109,71 @@ package object hdfs {
...
+      } else {
+        //TODO: should upload be asynchronous, blocking on context close
+        require(self.options.getStagingLocation != null,
+          "Staging directory not set - use `--stagingLocation`!")
+        require(path != null, "Artifact path can't be null")
+
+        val _conf = Option(conf).getOrElse(new Configuration())
+
+        //TODO: should we add checksums on both src and GCS to reuse uploaded artifacts?
+        val path_hash = Hashing.sha1().hashString(path, Charsets.UTF_8)
```
nevillelyh: camelCase, i.e. pathHash, targetHash, targetDistCache, etc.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +109,71 @@ package object hdfs {
...
+        val target_hash = path_hash.toString.substring(0, 8)
+
+        logger.debug(s"Add '$path' (hash: '$path_hash') to dist cache")
+
+        val target_distcache =  new Path("distcache", s"$target_hash-${path.split("/").last}")
```
nevillelyh: extra space after `=`

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +109,71 @@ package object hdfs {
...
+
+        val target = new Path(self.options.getStagingLocation, target_distcache)
```
nevillelyh: Check if target already exists to avoid extra work?
DF does that for staging jars too.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +109,71 @@ package object hdfs {
...
+     * @param conf optional custom Hadoop configuration
+     * @param username optional Hadoop Simple Authentication remote username
+     */
+    def hadoopDistCache[F](path: String,
+                           conf: Configuration = null,
+                           username: String = null)
+                          (initFn: File => F): DistCache[F] = self.pipelineOp {
+      if (self.isTest) {
+        //TODO: if test - can the path be remote HDFS/S3 etc?
+        self.distCache(path)(initFn)
```
nevillelyh: It doesn't matter since the no file IO ever happens in test.

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -99,6 +109,71 @@ package object hdfs {
...
+      }
+    }
+
+    private[scio] def hadoopDistCacheCopy(src: Path, target: URI, conf: Configuration): Unit = {
+      logger.debug(s"Will copy ${src.toUri}, to $target")
+
+      val fs = src.getFileSystem(conf)
+      val inStream = fs.open(src)
+
+      // TODO: Should we attempt to detect the Mime type rather tha always using MimeTypes.BINARY?
```
nevillelyh: "rather than"

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
ravwojdyla: Yea - tested it.

andrewsmartin:
👍 Can we merge?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/HadoopDistCacheExample.scala
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+// Use distributed cache inside a job
+object HadoopDistCacheExample {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    // declare a HDFS file to be distributed to all workers and logic to load the file
+    val dc = sc.hadoopDistCache(args.getOrElse("months",
+      "hdfs://dataflow-samples/samples/misc/months.txt")) { f =>
```
nevillelyh: Use `val MONTHS` in ExampleData.scala?

PR #123, opened 2016-04-28
WIP BigQuery SQL-2011

nevillelyh:


nevillelyh:
Do we need a flag for legacySql? @ravwojdyla thoughts?

ravwojdyla:
@nevillelyh need to read about the update first. pending thoughts.

PR #128, opened 2016-05-07
PoC for partition in side input

ravwojdyla:


ravwojdyla:
do not merge. currently side in/out is kinda messy - would be nice to clean it up first. where would `partition` like this fit in - it takes side inputs and returns multiple side outputs. Such `partition` could be useful for example to partition `KVSCollection` based on some side input computed as part of the pipeline (I want to use it for skewed join).

ravwojdyla:
example:
```
val s = sc.parallelize(List(1,2,3,4,5))
val sInput = sc.parallelize(Map("foo" -> 0, "bar" -> 3)).asMapSideInput
val sIS = s.withSideInputs(sInput)
import com.spotify.scio.values._
val s1 = SideOutput[Int]()
val s2 = SideOutput[Int]()
val part = sIS.partition(Seq(s1,s2), (i: Int, partitions: Int, SIC: SideInputContext[Int]) => SIC(sInput)("foo"))
val s1M = part(s1).materialize
sc.close
s1M.waitForResult().value.toList
```

andrewsmartin:
👍

ravwojdyla:
@andrewsmartin thanks! @nevillelyh what do you think?

ravwojdyla:
We decided that it's too much, and:
 * if we need it, we shall not expose it to users
 * rework `*WithSide*` too allow for multi input output and less confusion around current abstractions

PR #129, opened 2016-05-09
Added methods for reading multiple hdfs files

andrewsmartin:


nevillelyh:
Do we really need this? It's pretty easy to `sc.hdfsAvroFile` and `unionAll` later.
Plus this will make the HDFS Avro API inconsistent with other file based sources.

andrewsmartin:
Yeah, not really much boilerplate to begin with I guess. Was thinking it would be nice to support out of the box directly from the cmd line args but I agree -- consistency probably more important

andrewsmartin:
@nevillelyh @ravwojdyla Revisiting this - what if we just use varargs and always unionAll (even if just one input)?

nevillelyh:
Varargs only works if it's the last arg and there're no other args with default value. So doing it will still make this API inconsistent with everything else.

PR #132, opened 2016-05-12
Cmsjoin

ravwojdyla:


ravwojdyla:
related to #9

ravwojdyla:
this is rfc - do not merge - still need to add some comments and fix style here and there, so please comment on the idea/implementation.

ravwojdyla:
ready for review @andrewsmartin @nevillelyh

scio-core/src/main/scala/com/spotify/scio/util/FunctionsWithSideInput.scala
```diff
@@ -54,4 +54,24 @@ private[scio] object FunctionsWithSideInput {
       c.output(g(c.element(), sideInputContext(c)))
   }
 
+  /**
```
nevillelyh: Now that we are not exposing this, maybe move this logic to the same place in PairSCollectionFunctions so it's more coherent?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -193,6 +194,108 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+  : SCollection[(K, (V, Option[W]))] = self.transform { in =>
+    val side = that.asMultiMapSideInput
+    in.withSideInputs(side).flatMap[(K, (V, Option[W]))] { (kv, s) =>
+      val (k, v) = kv
+      val m = s(side)
+      if (m.contains(k)) m(k).map(w => (k, (v, Some(w)))) else Seq((k, (v, None)))
+    }.toSCollection
+  }
+
+  /* Skew operations */
```
nevillelyh: Don't really need this since there's only one method in this section?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -193,6 +194,108 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+                                delta: Double = 1E-10,
+                                seed: Int = 42)(implicit hasher: CMSHasher[K])
+  : SCollection[(K, (V, W))] = {
+    import com.twitter.algebird._
+
+    // Count k->#values:
+    val keyAggregator = CMS.aggregator[K](eps, delta, seed)
+
+    // Use asIterableSideInput as workaround for:
+    // http://stackoverflow.com/questions/37126729/ismsinkwriter-expects-keys-to-be-written-in-strictly-increasing-order
```
nevillelyh: This will freak scalastyle out. Suppress?

scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideInput.scala
```diff
@@ -67,6 +68,28 @@ class SCollectionWithSideInput[T: ClassTag] private[values] (val internal: PColl
     new SCollectionWithSideInput[U](o, context, sides)
   }
 
+  /**
```
nevillelyh: Same here, move the logic to `skewedN1Join`?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -193,6 +194,108 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+    val keyCMS = self.keys.aggregate(keyAggregator).asIterableSideInput
+
+    val (hotSelf, chillSelf) = (SideOutput[(K,V)](), SideOutput[(K,V)]())
+    val partitionedSelf = self
```
nevillelyh: without re-implementing private `partition`, maybe you can do `self.cross(keyCMS)` and then normal withSideOutputs?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -193,6 +194,108 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+   * @param delta A bound on the probability that a query estimate does not lie within some small
+   *              interval (an interval that depends on `eps`) around the truth.
+   * @param seed A seed to initialize the random number generator used to create the pairwise
+   *             independent hash functions.
+   */
+  def skewedN1Join[W: ClassTag](that: SCollection[(K, W)],
+                                hotKeyThreshold: Long = 9000,
+                                eps: Double = 0.0001,
+                                delta: Double = 1E-10,
+                                seed: Int = 42)(implicit hasher: CMSHasher[K])
```
nevillelyh: Seed should default to something random and hard coded in a test?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -193,6 +194,108 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+      val (k, v) = kv
+      val m = s(side)
+      if (m.contains(k)) m(k).map(w => (k, (v, Some(w)))) else Seq((k, (v, None)))
+    }.toSCollection
+  }
+
+  /* Skew operations */
+
+  /**
+   * N to 1 skewproof flavor of [[PairSCollectionFunctions.join()]].
```
nevillelyh: Do we really need N1 in the name? Maybe hard to explain in layman's them. `skewedJoin` sounds clear enough?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ravwojdyla: done.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ravwojdyla: Isn't this gonna blow up left site? If I understand it correctly it will duplicate whole CMS for each record of the left side. While partition is basically a filter on CMS?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ravwojdyla: Something `random` - you mean more random than `42`, or remove all default values and give example in the doc? After thinking about it a bit - maybe it's better to remove the default all together and give example in doc. ?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ravwojdyla: Name changed to `skewedJoin`.

scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideInput.scala
ravwojdyla: Would prefer to leave it here to have access to `parDo` etc

PR #136, opened 2016-05-16
Andrew/hdfs username fix

andrewsmartin:
@ravwojdyla Thoughts? At the bottom there are commented out methods that show how we can maybe use reflection instead for a bit more boilerplate removal, but it feels rather hacky. I think if we wanted to completely remove all the boilerplate (considering there are other classes that we have to do the same thing for) we'd use some kind of AOP framework, but I think that'd be overkill for the scope of SimpleAuth.


andrewsmartin:
This is RFC, not ready to merge yet. Still need to add tests and make changes for simple auth sources. Just want some feedback on the implementation.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/BigQueryClientTest.scala
```diff
@@ -47,7 +47,7 @@ class BigQueryClientTest extends FlatSpec with Matchers {
                      |JOIN `bigquery-public-data.hacker_news.comments` AS c
                      |ON s.`by` = c.`by`
                    """.stripMargin
-    bq.getQueryTables(sqlQuery).map(BigQueryIO.toTableSpec) should equal (Seq(
+    bq.getQueryTables(sqlQuery).map(BigQueryIO.toTableSpec).toSet should equal (Set(
```
nevillelyh: Submit a separate PR for this?

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSink.java
```diff
@@ -46,7 +46,6 @@
  * @param <V> The type of values to be written to the sink.
  */
 public class HadoopFileSink<K, V> extends Sink<KV<K, V>> {
-
```
nevillelyh: There're no other logic change in this file other than the 3 blank line deletions. I'd prefer to not let them in.

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/simpleauth/SimpleAuthHadoopFileSink.java
```diff
@@ -18,31 +18,154 @@
...
+      ugi = UserGroupInformation.createRemoteUser(username);
+    }
+
+    @Override
+    public void open(String uId) throws Exception {
+      final String uid = uId;
+      ugi.doAs(new PrivilegedExceptionAction<Void>() {
+        @Override
+        public Void run() throws Exception {
+          _open(uid);
```
nevillelyh: Just call `super.open()` instead of defining a `_open()`?

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/simpleauth/SimpleAuthHadoopFileSink.java
```diff
@@ -18,31 +18,154 @@
...
+    private void _open(String uId) throws Exception {
+      super.open(uId);
+    }
+
+    @Override
+    public String close() throws Exception {
+      return ugi.doAs(new PrivilegedExceptionAction<String>() {
+        @Override
+        public String run() throws Exception {
+          return _close();
```
nevillelyh: Call `super.close()` directly here too?

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/simpleauth/SimpleAuthHadoopFileSink.java
andrewsmartin: Can't do this in anonymous function scope (since super would refer to `super` of `PrivilegedActionExceptionAction`), so I had to define _open

nevillelyh: How about `HadoopWriter<K, V> parent = super;` right before the scope?

andrewsmartin: I don't think that would work either because of dynamic method dispatch, right? Need to explicitly have the method in scope where child calls parent method

scio-hdfs/src/main/java/com/google/cloud/dataflow/contrib/hadoop/simpleauth/SimpleAuthHadoopFileSink.java
```diff
@@ -18,31 +18,154 @@
...
+//    @Override
+//    public String close() throws Exception {
+//      return ugi.doAs(new SimpleAuthHadoopAction<String>(
+//              MethodHandles.lookup().findSpecial(HadoopWriter.class, "close", MethodType.methodType(Void.class).unwrap(), this.getClass()),
+//              null));
+//    }
+
   }
+
+  public static class SimpleAuthHadoopAction<T> implements PrivilegedExceptionAction<T> {
```
nevillelyh: I guess you don't need this class either.

nevillelyh:
LGTM feel free to merge.

PR #145, opened 2016-05-22
Improve implementation of saveAsMultipleTable

rgruener:

After meeting with the Bigtable team, the implementation of writeToMultipleTables is suboptimal (for a lot of use cases) and only batches the mutations within the iterable (which in itself is converted into a list thus needing to be small enough to fit in memory).  This adds the optimization of batching writes to a single table across the tuples, making it much more performant to have a dataflow job which writes ~100M rows to 2 separate tables.

I thought of changing it to take a `KV<String, Mutation>` instead of `KV<String, Iterable<Mutation>>` but backward compatible is better I supposed. I just don't like that its misleading since the Iterable really just gets read into memory as a list (might cause people to do a groupByKey which will not work).

I am unsure whether this will/should be implemented in the bigtable-hbase-dataflow itself perhaps @garye would have an opinion but for now this gives up 1000x increase in write throughput when writing across regions and will be great to have within scio.

### DIFF
--- a/scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableMultiTableWrite.java
+++ b/scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableMultiTableWrite.java
@@ -1,0 +1,146 @@
+package com.spotify.scio.bigtable;
+
+import com.google.api.client.util.Lists;
+import com.google.bigtable.repackaged.com.google.cloud.config.BulkOptions;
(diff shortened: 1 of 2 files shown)


garye:
We want bigtable-hbase-dataflow to support the most straightforward dataflow + bigtable use cases as well as provide building blocks for specific use cases, which you've used here. The fact that we have a (sub-optimal) existing version of writeToMultipleTables is an exception so we'll have to take a look at whether or not migrating the current implementation to this one makes sense for existing users.

PR #152, opened 2016-05-25
Fix NullPointerException in getting buffered mutator

nikhil-tibrewal:
This should fix the NullPointerException in getting the mutator for a `tableName`. Under the previous implementation, the `transient` variable `mutators` was not serialized to the workers, and hence was `null`. Instantiating it under `startBundle` makes sure the variable exists on the workers.


rgruener:
:+1: Thanks for finding/fixing this

PR #153, opened 2016-05-26
Do not override df appName if set by user

rouzwawi:
I ran into problems with this setting when running multiple jobs from the same jvm instance. Even though I passed the `--appName` argument, it got overridden by this line.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -125,8 +125,10 @@ class ScioContext private[scio] (val options: DataflowPipelineOptions,
 
   import Implicits._
 
-  // Set default name
-  this.setName(CallSites.getAppName)
+  // Set default name if no app name specified by user
+  if (options.getAppName == null) {
+    this.setName(CallSites.getAppName)
+  }
 
   /** Dataflow pipeline. */
   def pipeline: Pipeline = {


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -211,7 +211,7 @@ class ScioContext private[scio] (val options: DataflowPipelineOptions,
       throw new RuntimeException("Cannot set name once pipeline is initialized")
     }
     // override app name and job name
-    options.setAppName(name)
+    if (options.getAppName == null) options.setAppName(name)
```
nevillelyh: But this will make `setName` a no-op if app name is already set.

`setName` is actually used only once in `ScioContext` constructor, so maybe do the `== null` check there?
https://github.com/rouzwawi/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L129

rouzwawi: Right. I'm not completely sure when the `@InstanceFactory(DataflowPipelineOptions.JobNameFactory.class)` thing is executed normally. Is it on each property read, or just after the underlying proxy is craeted (after parse).

I'll look into it and change to your suggestion if it works as expected.

rouzwawi:
@nevillelyh fixed

PR #154, opened 2016-05-26
fix AvroCoder usage in scio-hdfs

nevillelyh:


nevillelyh:
@andrewsmartin this should fix the problem you saw awhile ago.

andrewsmartin:
👍

PR #157, opened 2016-05-27
fix app name detection

nevillelyh:


nevillelyh:
Fixes `sciocontext0-$user-$jobId` caused by 3523983b8c64a7a3cbcbccd1aa59de526c9cb26b.
@rouzwawi FYI

Issue #144, opened 2016-05-20
hadoopDistCache should support multiple files


nevillelyh:
So it's consistent with regular `distCache`.

PR #161, opened 2016-05-30
WIP support for multiple paths for Hadoop dist cache #144

andrewsmartin:
@nevillelyh @ravwojdyla RFC - would like to get your input on this. Since Scala doesn't allow overridden methods with default parameters, I saw 2 options:

1) remove the default parameters and instead have a slew of overloaded methods, or
2) use a builder. This is currently what is in place but unfortunately this would be a breaking change.

Thoughts?

### DIFF
--- a/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
+++ b/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
@@ -121,41 +121,58 @@ package object hdfs {
     def hadoopDistCache[F](path: String,
                            conf: Configuration = null,
                            username: String = null)
-                          (initFn: File => F): DistCache[F] = self.pipelineOp {
+                          (initFn: File => F): DistCache[F] =
+      hadoopDistCacheMulti(Seq(path), conf, username) {fs: Seq[File] => initFn(fs.head)}
+
+    /**
+     * Create a new [[com.spotify.scio.values.DistCache DistCache]] instance for a list of
+     * files on Hadoop/HDFS.
+     *
+     * @param paths Sequence of paths to the Hadoop/HDFS artifacts
+     * @param conf optional custom Hadoop configuration
+     * @param username optional Hadoop Simple Authentication remote username
+     */
+    def hadoopDistCacheMulti[F](paths: Seq[String],
+                                conf: Configuration = null,
+                                username: String = null)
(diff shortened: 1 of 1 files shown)


nevillelyh:
Sorry just got a chance to look at this.
How about just use a different name this time, e.g. `def hadoopDistCacheMulti` and maintain argument consistency with the core version?

andrewsmartin:
@nevillelyh Updated with your suggestion

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -121,41 +121,58 @@ package object hdfs {
...
+    /**
+     * Create a new [[com.spotify.scio.values.DistCache DistCache]] instance for a list of
+     * files on Hadoop/HDFS.
+     *
+     * @param paths Sequence of paths to the Hadoop/HDFS artifacts
+     * @param conf optional custom Hadoop configuration
+     * @param username optional Hadoop Simple Authentication remote username
+     */
+    def hadoopDistCacheMulti[F](paths: Seq[String],
+                                   conf: Configuration = null,
```
nevillelyh: fix indentation?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -121,41 +121,58 @@ package object hdfs {
...
+        require(!paths.contains(null), "Artifact path can't be null")
 
         val _conf = Option(conf).getOrElse(new Configuration())
 
-        //TODO: should we add checksums on both src and GCS to reuse uploaded artifacts?
-        // to keep it simple, for now we upload each artifact, thus artifacts should be
-        // relatively small.
-        val pathHash = Hashing.sha1().hashString(path, Charsets.UTF_8)
-        val targetHash = pathHash.toString.substring(0, 8)
+        val targetPaths = paths.foldLeft(Seq[Path]()) { (targets, path) =>
```
nevillelyh: looks like this can be done with a `paths.map`?

andrewsmartin: Woops. Not entirely sure what I was thinking

nevillelyh:
LGTM, merge when green.

PR #165, opened 2016-06-10
fix repl from sbt

ravwojdyla:
After these fixes - scio shell can be started from both:
 * assembly jar via `java -jar <assembly-jar>`
 * sbt via `sbt run` or `sbt scio-repl/run`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -166,6 +166,7 @@ lazy val root: Project = Project(
 ).settings(
   unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject
     -- inProjects(scioRepl) -- inProjects(scioSchemas) -- inProjects(scioExamples),
+  run <<= run in Compile in scioRepl,
   aggregate in assembly := false
 ).aggregate(
   scioCore,
--- a/build.sbt
+++ b/build.sbt
@@ -346,6 +346,16 @@ lazy val scioRepl: Project = Project(
     )
   )
 ).settings(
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -128,6 +128,7 @@ val commonSettings = Project.defaultSettings ++ Sonatype.sonatypeSettings ++ ass
   }
 )
 
+lazy val printClasspath = taskKey[Unit]("Dump classpath")
```
nevillelyh: Is this necessary or debugging code?

ravwojdyla: Only for debugging purposes.

ravwojdyla:
The seems to be some issue with `2.10` - let me double check.

ravwojdyla:
Tests are fine and both `2.10` and `2.11` is fine for assembly jar. Sbt launch works with `2.11` but NOT with `2.10` - seems to be related changes between `2.10` and `2.11`. Not sure if it's worth to investigate `2.10` more (?)

nevillelyh:
Not worth supporting 2.10 except the potential zeppelin use case but we're not quite there yet.
LGTM except the debugging lines. Remove them?

bzz:
Sorry for digging this out, but have just came across this: just FYI, as of apache/zeppelin#747 Apache Zeppelin supports Scala 2.11 starting from upcoming 0.6.1 release.

nevillelyh:
@bzz thanks that's good to know!

PR #171, opened 2016-06-17
decouple DataflowPipelineOptions #139

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -117,23 +117,36 @@ object ScioContext {
...
+  if (options.isInstanceOf[DataflowPipelineOptions]) {
+    if (dfOptions.getAppName == null || dfOptions.getAppName.startsWith("ScioContext$")) {
+      this.setName(CallSites.getAppName)
+    }
   }
 
   /** Dataflow pipeline. */
   def pipeline: Pipeline = {
     if (_pipeline == null) {
-      options.setFilesToStage(getFilesToStage(artifacts).asJava)
```
ravwojdyla: `else` warning for artifacts?

ravwojdyla:
Except the one comment - 👍

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: Right now the only other case is direct pipeline with no artifacts, so no need to warn?

ravwojdyla: I just don't like the silent failure - so if someone adds artifacts, and if there is more runners some time from now, will someone remember to look in this piece of code to make sure artifacts are handled.

PR #177, opened 2016-06-23
use Semigroup[Array[T]] in DoubleSCollectionFunctions#histogram

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/package.scala
```diff
@@ -83,4 +85,47 @@ package object scio {
...
+  implicit val floatArraySg: Semigroup[Array[Float]] = new ArraySemigroup[Float]
+
+  /** [[com.twitter.algebird.Semigroup Semigroup]] for Array[Double]. */
+  implicit val doubleArraySg: Semigroup[Array[Double]] = new ArraySemigroup[Double]
+
+  class ArraySemigroup[@specialized(Int, Long, Float, Double) T : ClassTag : Numeric]
+    extends Semigroup[Array[T]]{
+    private val num = implicitly[Numeric[T]]
+    private def plusI(l: Array[T], r: Array[T]): Array[T] = {
+      require(l.length == r.length, "array lengths do not match")
```
ravwojdyla: Breeze will just use the smallest size - what about this idea?

nevillelyh: There's no breeze dependency in `scio-core`, only in `scio-extra` and I kinda want to keep it that way.
Although rare, throwing an error is still better than silently truncating arrays?

PR #182, opened 2016-06-27
Add Protobuf File IO #179

ravwojdyla:


ravwojdyla:
wip. to do:
- [ ] is this the way we want to go? (specifically avro/objectFile)
- [ ] clean up code/comments
- [ ] add tests

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -367,6 +368,15 @@ class ScioContext private[scio] (val options: PipelineOptions,
   }
 
   /**
+   * Get an SCollection for an Protobuf file.
+   * @group input
+   */
+  def protobufFile[T: ClassTag](path: String)(implicit ev: T <:< Message): SCollection[T]
+  = pipelineOp {
+    this.objectFile(path)
```
nevillelyh: Better do a `.map()` first to convert `T` into `Array[Byte]` using `ProtoCoder` instead of default Kryo.

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -59,11 +61,18 @@ case class TextTap(path: String) extends Tap[String] {
...
 /** Tap for Avro files on local file system or GCS. */
+//TODO (rav): what if schema is null?
 case class AvroTap[T: ClassTag](path: String, schema: Schema = null) extends Tap[T] {
   override def value: Iterator[T] = FileStorage(path).avroFile(schema)
   override def open(sc: ScioContext): SCollection[T] = sc.avroFile[T](path, schema)
 }
 
+/** Tap for Protobuf files on local file system or GCS. */
+case class ProtobufTap[T: ClassTag](path: String)(implicit ev: T <:< Message) extends Tap[T] {
+  override def value: Iterator[T] = ObjectFileTap.apply(path).value
```
nevillelyh: Use `ProtoCoder` explicitly here too.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -861,6 +862,20 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   * Save this SCollection as an Protobuf file.
+   * @group output
+   */
+  def saveAsProtobufFile(path: String, numShards: Int = 0)(implicit ev: T <:< Message)
+  : Future[Tap[T]] =
+    if(context.isTest) {
+      context.testOut(ProtobufIO(path))(this)
+      saveAsInMemoryTap
+    } else {
+      this.asInstanceOf[SCollection[Message]].saveAsObjectFile(path, numShards)
```
nevillelyh: Use `ProtoCoder` here.

nevillelyh:
Avro + objectFile still uses `KryoAtomicCoder` internally which uses `chill-protobuf`.
https://github.com/twitter/chill/blob/develop/chill-protobuf/src/main/java/com/twitter/chill/protobuf/ProtobufSerializer.java

It looks like `chill-protobuf` also writes array length which is redundant if we wrap bytes in Avro.
So it's better we do the `Message` <-> `Array[Byte]` <-> Avro coding explicitly.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -367,6 +369,21 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+   * Get an SCollection for a Protobuf file.
+   * @group input
+   */
+  def protobufFile[T: ClassTag](path: String)
+                               (implicit ev: T <:< GeneratedMessage with Message): SCollection[T] =
+    pipelineOp {
+      // protobuf parser is not serializable - defeat via twitter's externalizer
+      val serializableParser = Externalizer(ScioUtil.getProtobufParser[T])
+      this.objectFile[Array[Byte]](path).map{ bytes =>
+        lazy val parser = serializableParser.get
```
ravwojdyla: is there a better way to do this? we could potentially add something like AvroIO (ProtobufIO) that would handle parser and process elements as they come. @nevillelyh ?

scio-repl/src/main/scala/com/spotify/scio/repl/ReplScioContext.scala
```diff
@@ -17,21 +17,13 @@
 
 package com.spotify.scio.repl
 
-import java.io.{OutputStream, PrintStream}
```
nevillelyh: Changes in this file aren't related, move to a different PR?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -861,6 +862,22 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   */
+  def saveAsProtobufFile(path: String, numShards: Int = 0)
+                        (implicit ev: T <:< GeneratedMessage with Message): Future[Tap[T]] =
+    if(context.isTest) {
+      context.testOut(ProtobufIO(path))(this)
+      saveAsInMemoryTap
+    } else {
+      this.asInstanceOf[SCollection[Message]]
+        .map(_.toByteArray)
+        .saveAsObjectFile(path, numShards)
```
nevillelyh: `.saveAsObjectFile` will still serialize the `Array[Byte]` using Kryo, is that transparent?
Or we can add private `ScioContext#AvroBytesFile` & `SCollection#saveAsAvroBytesFile` and use them for both object file and proto file?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ravwojdyla: NVM - I now see the `ProtoCoder` idea ...

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -367,6 +369,21 @@ class ScioContext private[scio] (val options: PipelineOptions,
   }
 
   /**
+   * Get an SCollection for a Protobuf file.
+   * @group input
+   */
+  def protobufFile[T: ClassTag](path: String)
+                               (implicit ev: T <:< GeneratedMessage with Message): SCollection[T] =
```
nevillelyh: `GeneratedMessage` already implements `Message` right? Is `with Message` necessary here?

nevillelyh:
Also add a test case in `TapTest.scala`?

ravwojdyla:
For the reference:

```
// two messages serialized via ProtoCoder:
{"bytes":"\n\u0003ola\u0010\u0002"}
{"bytes":"\n\u0003ala\u0010\u0001"}


// two messages converted to byte[] and serialized via Kryo:
{"bytes":"\u0001\u0000[Â\u0001\b\n\u0003ala\u0010\u0001"}
{"bytes":"\u0001\u0000[Â\u0001\b\n\u0003ola\u0010\u0002"}


// two messages directly serialized via Kryo:
{"bytes":"\u0001\u0000sh.rav.scio.test.proto.Simple$SimplePÂ\u0001\u0006\u0000\u0001\u000E\u0001\u0002\u0001\u0001com.google.protobuf.LiteralByteStrinç\u0001\u0001\u0004ala\u0000\u0001\u0001\u0002java.util.Collections$EmptyMað\u0001\u0005"}
{"bytes":"\u0001\u0000sh.rav.scio.test.proto.Simple$SimplePÂ\u0001\u0006\u0000\u0001\u000E\u0001\u0004\u0001\u0001com.google.protobuf.LiteralByteStrinç\u0001\u0001\u0004ola\u0000\u0001\u0001\u0002java.util.Collections$EmptyMað\u0001\u0005"}
```

nevillelyh:
Yeah just what I suspected. Kryo on bytes adds array length and kryo on object adds class name.

ravwojdyla:
Yes. For the reference, case class with PB inside:

```
// two instances of case class with PB inside (no PB kryo serializer registered):
{"bytes":"\u0001\u0000Ä\u0001$line22.TestS\u0001\u0001\u0006\u0000\u0001\u000E\u0001\u0002\u0003\u0001alá\u0001\u0001\u0001java.util.Collections$EmptyMað\u0001\u0005\u0002"}
{"bytes":"\u0001\u0000Ä\u0001$line22.TestS\u0001\u0001\u0006\u0000\u0001\u000E\u0001\u0004\u0003\u0001olá\u0001\u0001\u0001java.util.Collections$EmptyMað\u0001\u0005\u0002"}


// two instances of case class with PB inside (chill's PB kryo serializer registered):
{"bytes":"\u0001\u0000Ä\u0001$line10.TestS\u0001\u0001\u0007\n\u0003ola\u0010\u0002\u0002"}
{"bytes":"\u0001\u0000Ä\u0001$line10.TestS\u0001\u0001\u0007\n\u0003ala\u0010\u0001\u0002"}
```

build.sbt
```diff
@@ -45,6 +46,7 @@ val scalaCheckVersion = "1.13.0"
 val scalaMacrosVersion = "2.1.0"
 val scalaTestVersion = "2.2.6"
 val slf4jVersion = "1.7.21"
+val scalapbVersion = "0.5.19" // inner protobuf-java version must match beam/dataflow-sdk one
```
nevillelyh: keep these `val *Version` sorted? The only exception is the first `dataflowSdkVersion` or `beamVersion` to make code porting easier.

project/plugins.sbt
```diff
@@ -7,3 +7,4 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.5.4")
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "0.8.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.3.5")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
+addSbtPlugin("com.trueaccord.scalapb" % "sbt-scalapb" % "0.5.19")
```
nevillelyh: Same here, nice to keep them sorted.

project/scalapb.sbt
```diff
@@ -0,0 +1,3 @@
+libraryDependencies ++= Seq(
+  "com.github.os72" % "protoc-jar" % "3.0.0-b1"
```
nevillelyh: Any reason this is in a separate file, not `scio-schemas` section in `build.sbt`? This here applies to all sub-modules right?

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -77,11 +79,12 @@ case class BigQueryTap(table: TableReference) extends Tap[TableRow] {
 }
 
 /** Tap for object files on local file system or GCS. */
-case class ObjectFileTap[T: ClassTag](path: String) extends Tap[T] {
+case class ObjectFileTap[T: ClassTag](path: String, coder: Coder[T] = null) extends Tap[T] {
```
nevillelyh: We're not overriding `coder` yet and it's not exposed in `ScioContext` or `SCollection`? Remove the optional arg for now?

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -77,11 +79,12 @@ case class BigQueryTap(table: TableReference) extends Tap[TableRow] {
...
   override def value: Iterator[T] = {
-    val coder = KryoAtomicCoder[T]
+    import com.spotify.scio.Implicits.RichCoderRegistry
```
nevillelyh: Objects like `Implicits` are intended to be imported as `import com.spotify.scio.Implicits._`, so we don't expose the internal name `RichCoderRegistry` here.

scio-core/src/main/scala/com/spotify/scio/io/Taps.scala
```diff
@@ -61,6 +63,11 @@ trait Taps {
   def textFile(path: String): Future[Tap[String]] =
     mkTap(s"Text: $path", () => isPathDone(path), () => TextTap(path))
 
+  /** Get a `Future[Tap[T]]` of a Protobuf file. */
+  def protobufFile[T: ClassTag](path: String)(implicit ev: T <:< Message)
+  : Future[Tap[T]] =
+    mkTap(s"Protobuf: $path", () => isPathDone(path), () => ObjectFileTap[T](path, null))
```
nevillelyh: Get rid of the `null` optional arg for now?

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
nevillelyh: On second thought, add a `ScioUtil.getScalaCoder[T]` method and create new `CoderRegistry` plus implicits inside?

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
nevillelyh: Scratch that, just saw this is used later.

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -29,4 +31,13 @@ private[scio] object ScioUtil {
 
   def classOf[T: ClassTag]: Class[T] = implicitly[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]]
 
+  def getNewCoderRegistry: CoderRegistry = {
```
nevillelyh: Put this and implicit stuff inside a `ScioUtil.getScalaCoder[T]`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -806,15 +807,15 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
         .parDo(new DoFn[T, GenericRecord] {
-          private val coder = KryoAtomicCoder[T]
           override def processElement(c: DoFn[T, GenericRecord]#ProcessContext): Unit = {
-            c.output(AvroBytesUtil.encode(coder, c.element()))
+            c.output(AvroBytesUtil.encode(elemCoder, c.element()))
           }
         })
         .saveAsAvroFile(path, numShards, AvroBytesUtil.schema)
-      context.makeFuture(ObjectFileTap[T](path + "/part-*"))
+      context.makeFuture(ObjectFileTap[T](path + "/part-*", elemCoder))
```
nevillelyh: You don't need the optional arg `elemCoder` here since `ObjectFileTap` can figure out itself?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ProtobufExample.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.spotify.scio.proto.SimpleV2.SimplePB
+import com.spotify.scio.proto.Track.TrackPB
+
+// Read protobuf on input, and write another protobuf on output
+object ProtobufExample {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    sc.protobufFile[SimplePB](args("input"))
+    .map(p => TrackPB.newBuilder().setTrackId(p.getTrackId))
```
nevillelyh: Indentation

scio-schemas/src/main/protobuf/simple_v2.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+
+option java_package = "com.spotify.scio.proto";
+option optimize_for = SPEED;
+
+message SimplePB {
```
nevillelyh: Just curious, why `SimpleV2.SimplePB` and `Track.TrackPB`, not just something like `TrackV1` & `TrackV2`?

nevillelyh:
LGTM overall except the minor style/naming stuff, coder handling in `ScioUtil` and optional coder arg.

project/scalapb.sbt
ravwojdyla: This will not propagate too the submodules. I verified that it does not end up in submodule's `libraryDependencies`. Initially I wanted to put it in `scioSchema` but then I get failure:

```
scio/build.sbt:313: error: object os72 is not a member of package com.github
    com.github.os72.protocjar.Protoc.runProtoc("-v300" +: args.toArray)
```

when I try to reference it in subsequent `settings` section.

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
ravwojdyla: In `SCollection` and `ScioContext` we have `CoderRegistry` available - `Tap`s are totally separated - thus user might want to decide, also I later use it to to pass coder from `saveAsObject` to point to the same coder without the need to create another `CoderRegistry` instance just to find something that is already available.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
ravwojdyla: Since it's at hand - why create another `CoderRegistry` if we can just pass to the `Tap`?

project/scalapb.sbt
ravwojdyla: For now I have removed the explicit version, and moved the lib to `scioSchema`.

PR #185, opened 2016-06-28
Fix parsing of timestamp strings with missing fractional seconds; add unit test

mchalek:
I have encountered a failure due to BigQuery returning timestamps with missing fractional seconds.  In these instances, the timestamp parser throws an IllegalArgumentException because the expected pattern is "too short", e.g.: 
```
[info]   java.lang.IllegalArgumentException: Invalid format: "2016-06-28 22:13:29" is too short
[info]   at org.joda.time.format.DateTimeFormatter.parseDateTime(DateTimeFormatter.java:945)
[info]   at com.spotify.scio.bigquery.package$Timestamp$.parse(package.scala:95)
[info]   at com.spotify.scio.bigquery.TimestampTest$$anonfun$3.apply$mcV$sp(TimestampTest.scala:40)
[info]   at com.spotify.scio.bigquery.TimestampTest$$anonfun$3.apply(TimestampTest.scala:36)
[info]   at com.spotify.scio.bigquery.TimestampTest$$anonfun$3.apply(TimestampTest.scala:36)
```

Note that I received this error when querying tables that I generated using `scio-bigquery`, with input data that originally did not contain any fractional seconds.  I'm not quite sure why BigQuery does not return zeros in the fractional second columns when they are queried, but this revision ensures that the parser does not break when this happens.

I added a unit test for this condition, which will illustrate the failure when run without the changes to package.scala in place.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
@@ -22,6 +22,7 @@ import org.joda.time.Instant
 import org.joda.time.format.DateTimeFormat
 
 import scala.collection.JavaConverters._
+import scala.util.{Try, Failure}
 
 /**
  * Main package for BigQuery APIs. Import all.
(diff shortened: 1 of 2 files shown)


nevillelyh:
Thanks for fixing this!

mchalek:
@nevillelyh no problem!  Thanks for putting this library together, it's already saved me far more time than it took me to make this little patch.

cheers!

kevin

nevillelyh:
@mchalek good to hear! It'd be great if you can add something to this: https://github.com/spotify/scio/wiki/Powered-By

mchalek:
@nevillelyh we're currently in experimentation mode so we're not sure what we'll end up putting into production, but if we do end up using scio I'll try to remember to add us to the list!

PR #186, opened 2016-06-29
Add infix to save methods

ravwojdyla:


ravwojdyla:
Makes it more user friendly naming like `gs://<bucket>/data/path/part-0.protobuf.avro` etc

nevillelyh:
LGTM but infix usually means infix notation though? Can you think of a better name?

ravwojdyla:
infix comes from https://en.wikipedia.org/wiki/Infix. Options: `affix` `subsuffix` `innersuffix` or just `suffix`?

nevillelyh:
Maybe just `suffix`? Same semantic as `.txt.deflate` where `.deflate` is added by the platform?

PR #187, opened 2016-06-29
upgrade bigtable dependency to 0.9.0

rgruener:
@nevillelyh @ravwojdyla This upgrade starts to use v2 of the Bigtable api (transparent within the hbase connector)


ravwojdyla:
LGTM. Pending tests on some (semi) production use case.

rgruener:
Tests on production use case were successful. This seems good to go

PR #188, opened 2016-06-30
add function to update the number of bigtable nodes

rgruener:
@nevillelyh @ravwojdyla Should save a lot of money to be able to update the number of nodes before and after a job. 

Tested and this does work. Perhaps I should add it to the examples?

### DIFF
--- a/scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
+++ b/scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigtable;
+
+import com.google.bigtable.repackaged.com.google.cloud.config.BigtableOptions;
+import com.google.bigtable.repackaged.com.google.cloud.grpc.io.ChannelPool;
+import com.google.bigtable.repackaged.com.google.com.google.bigtable.admin.v2.Cluster;
+import com.google.bigtable.repackaged.com.google.com.google.bigtable.admin.v2.ListClustersRequest;
+import com.google.bigtable.repackaged.com.google.com.google.bigtable.admin.v2.BigtableInstanceAdminGrpc;
+import com.google.bigtable.repackaged.com.google.com.google.bigtable.admin.v2.BigtableInstanceAdminGrpc.BigtableInstanceAdminBlockingStub;
+import com.google.bigtable.repackaged.com.google.com.google.bigtable.admin.v2.ListClustersResponse;
+import com.google.cloud.bigtable.dataflow.CloudBigtableConfiguration;
+import org.joda.time.Duration;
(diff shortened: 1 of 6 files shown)


scio-bigtable/src/main/scala/com/spotify/scio/bigtable/Bigtable.scala
```diff
@@ -27,4 +33,42 @@ object Bigtable {
   def parseOptions(args: Array[String]): CloudBigtableOptions =
     ScioContext.parseArguments[CloudBigtableOptions](args)._1
 
+  /**
+    * Updates all clusters within the specified Bigtable Instance to have a specified number
```
nevillelyh: Can you fix doc to JavaDoc style? See style guide on the wiki.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/Bigtable.scala
```diff
@@ -17,8 +17,14 @@
...
+
+import com.google.bigtable.repackaged.com.google.com.google.bigtable.admin.v2.{BigtableInstanceAdminGrpc, Cluster, ListClustersRequest}
+import com.google.cloud.bigtable.dataflow.{CloudBigtableConfiguration, CloudBigtableOptions}
 import com.spotify.scio.ScioContext
+import org.joda.time.Duration
+
+import scala.collection.JavaConverters._
 
 /** Utilities for Bigtable. */
 object Bigtable {
```
nevillelyh: Possible to write this in Java? Easier to port to Beam Java SDK in the future.

nevillelyh:
Examples sounds great. You can include it in the existing ones?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/Bigtable.scala
rgruener: Yeah sure. I hope the Bigtable team will incorporate this themselves at some point as I feel its pretty important.

rgruener:
Hmm adding it to the examples makes the test fail. I will need to figure out how to properly stub this stuff.

nevillelyh:
Maybe add wrapper a method to [BigtableScioContext](https://github.com/spotify/scio/blob/master/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala#L42) and bypass this if `self.isTest`?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/Bigtable.scala
```diff
@@ -26,5 +26,4 @@ object Bigtable {
   /** Parse CloudBigtableOptions from command line arguments. */
   def parseOptions(args: Array[String]): CloudBigtableOptions =
     ScioContext.parseArguments[CloudBigtableOptions](args)._1
-
```
nevillelyh: No actual change in this file, skip this?

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
```diff
@@ -0,0 +1,85 @@
...
+   * it at the end to lower costs yet still get high throughput during bulk ingests/dumps.
+   *
+   * @param cloudBigtableConfiguration Configuration of Bigtable Instance
+   * @param numberOfNodes Number of nodes to make cluster
+   * @param sleepDuration How long to sleep after updating the number of nodes. Google recommends
+   *                      at least 20 minutes before the new nodes are fully functional
+   * @throws IOException If setting up channel pool fails
+   * @throws InterruptedException If sleep fails
+   */
+  public static void updateNumberOfBigtableNodes(final CloudBigtableConfiguration cloudBigtableConfiguration,
```
rgruener: This now has a wrapper in the BigtableScioContext which allows this to be ignored in tests. However this method is still public so it is confusing to know which one to use. Package private doesn't work here. Any suggestions @nevillelyh ?

nevillelyh: It's OK, Java people can still use this one and create their own wiring.

PR #193, opened 2016-07-07
Mitigate DataflowSDK 330 bug

ravwojdyla:
Once fix for #330 is merged, and available in upstream SDK, we can
revert this commit.


nevillelyh:
LGTM merge when green.

PR #195, opened 2016-07-11
tweak BigQuery default project detection

nevillelyh:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -498,12 +499,17 @@ object BigQueryClient {
    */
   def defaultInstance(): BigQueryClient = {
     if (instance == null) {
-      val project = sys.props(PROJECT_KEY)
-      if (project == null) {
-        throw new RuntimeException(
-          s"Property $PROJECT_KEY not set. Use -D$PROJECT_KEY=<BILLING_PROJECT>")
+      instance = if (sys.props(PROJECT_KEY) != null) {
+        BigQueryClient(sys.props(PROJECT_KEY))
```
ravwojdyla: the only reason this is still here is for backwards compatibility, correct?

nevillelyh: Or in environments with no default, i.e. Travis.

ravwojdyla:
👍

Issue #198, opened 2016-07-21
Upgrade Bigtable client to 0.9.1


nevillelyh:
@rgruener FYI

https://github.com/apache/incubator-beam/pull/701

nevillelyh:
Fixed in #199

PR #199, opened 2016-07-21
update bigtable version #198

rgruener:


nevillelyh:
There're grpc, netty and tcnative version changes in the Beam PR.
https://github.com/apache/incubator-beam/pull/701/files

Do we need them too?

nevillelyh:
This is the tcnative change I mentioned.
https://github.com/apache/incubator-beam/pull/701/files#diff-e75a52e9af6675b31f5d14b7eba2bbaeR201

rgruener:
Tcnative change was needed, added in the upgraded version.

PR #201, opened 2016-07-22
Adding the ability to skip hdfs validation

znichols:


nevillelyh:
We want to avoid creating too many variations of these classes and should add a withoutValidation method like other IO classes. Can you keep it in your repo for now?

scio-hdfs/src/main/java/com/google/cloud/dataflow/sdk/io/hdfs/skipvalidation/SkipValidationHDFSFileSink.java
```diff
@@ -0,0 +1,46 @@
...
+ * A {@code Sink} for writing records to a Hadoop filesystem using a Hadoop file-based output
+ * format with Simple Authentication.
+ *
+ * Skips validation for hdfs location (in case the permissions don't exist to read it on the
+ * calling machine)
+ *
+ * @param <K> The type of keys to be written to the sink.
+ * @param <V> The type of values to be written to the sink.
+ */
+public class SkipValidationHDFSFileSink<K, V> extends HDFSFileSink<K, V> {
```
ravwojdyla: Why is this not part of `HDFSFileSink` itself?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -207,7 +208,8 @@ package object hdfs {
     // TODO: numShards
     def saveAsHdfsTextFile(path: String,
                            username: String = null,
-                           conf: Configuration = null): Future[Tap[String]] = {
+                           conf: Configuration = null,
+                           skipValidation: Boolean = false): Future[Tap[String]] = {
```
ravwojdyla: `validate: Boolean = true` ?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -207,7 +208,8 @@ package object hdfs {
     // TODO: numShards
     def saveAsHdfsTextFile(path: String,
```
ravwojdyla: What about `Avro`?

ravwojdyla:
@nevillelyh technically Dataflow SDK has validation option - scio currently does not expose it for sinks, maybe we should in fact.

ravwojdyla:
@nevillelyh actually after some thought - I agree - 👎

nevillelyh:
@znichols closing this for now but I'll work on refactoring HDFS code soon.

znichols:
SGTM, thanks guys

PR #202, opened 2016-07-24
Added functions for join-filtering.

andrewsmartin:
@nevillelyh @rav RFC please. I often see code to filter an SCollection[(K, V)] by another one of type SCollection[K], where (k, v) pairs are filtered out if k is not present on the right hand side. It usually looks like this:

```
val s1: SCollection[(K, V)] = getIt()
val s2: SCollection[K] = getThat()

val filtered: SCollection[(K, V)] = 
    s1.join(s2.distinct.map(_, ())).map{ case(k, (v, _)) => (k, v)}
```

I find this pattern doesn't really speak to the high level intent of the operation, and personally it always takes me a bit of staring to understand what's happening when I see it. So I thought it might be useful to include in the core API. The equivalent of the above would look like this:

```
val s1: SCollection[(K, V)] = getIt()
val s2: SCollection[K] = getThat()

val filtered: SCollection[(K, V)] = s1.filterJoin(s2)
```

Do you consider this useful at all? Is there a more obvious way of doing this that I'm not thinking about? If you do think it's useful, any suggestions on naming? `filterJoin` is the first thing that came to mind since I'm implementing this on top of join but there's probably a better name. Similar functionality is achieved in a plain `SCollection[K]` via `intersect`, which internally converts both collections to a dummy `SCollection[(K, V)]`, and inner join is the equivalent for `SCollection[(K, V)], SCollection[(K, W)]`. Maybe `keyIntersect` is a better name.

### DIFF
--- a/scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
@@ -474,7 +474,7 @@ class PairSCollectionFunctionsTest extends PipelineSpec {
     }
   }
 
-  it should "support skewJoin() with empty key count (no hash join)" in {
+  it should "support skewedJoin() with empty key count (no hash join)" in {
     import com.twitter.algebird.CMSHasherImplicits._
     runWithContext { sc =>
       val p1 = sc.parallelize(Seq(("a", 1), ("a", 2)))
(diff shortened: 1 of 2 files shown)


ravwojdyla:
Would not use `join` in the name (the implementation should most likely use cogroup instead of joins), some variation of `intersect` (`intersectByKey`) makes more sense. Also since you are going through left side twice anyway - maybe keeping HLL and decide using side input or not?

scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
```diff
@@ -496,4 +496,22 @@ class PairSCollectionFunctionsTest extends PipelineSpec {
...
+  it should "support intersectByKey()" in {
+    runWithContext { sc =>
+      val p1 = sc.parallelize(Seq(("a", 1), ("b", 2), ("c", 3)))
+      val p2 = sc.parallelize(Seq("a", "b", "d"))
+      val p = p1.intersectByKey(p2)
+      p should containInAnyOrder (Seq(("a", 1), ("b", 2)))
+    }
+  }
+
+  it should "support intersectByKey() with duplicate keys" in {
```
ravwojdyla: Does it make sense to test on empty left&right hand side?

ravwojdyla:
+1 pls squash commits

PR #203, opened 2016-07-25
Add jars from jar manifest Class-Path attribute

ravwojdyla:


ravwojdyla:
review pls - @nevillelyh @andrewsmartin

andrewsmartin:
👍

PR #205, opened 2016-07-28
Adding the ability to specify the compression type for a textFile

ElPicador:


nevillelyh:
@ElPicador thanks for the contribution!
Feel free to edit https://github.com/spotify/scio/wiki/Powered-By if you're using Scio for anything interesting.

ElPicador:
@nevillelyh We are only experimenting a bit. If we use it we'll update the page 😄

PR #213, opened 2016-08-12
Save mulitple objects in the same bigtable table

ElPicador:
I didn't found tests for saving into BigTable. So I'm not sure how to make one for this


rgruener:
Hey, I don't really see the benefit of this change as everything is using a buffered mutator under the hood which will batch multiple objects already. So it would actually be best to just flatten and use the regular saveAsBigtable

ElPicador:
Ok 👍, closing it then

PR #214, opened 2016-08-16
Add assert message to haveSize assertion

ElPicador:


scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatcher.scala
```diff
@@ -57,7 +57,8 @@ private[scio] trait SCollectionMatcher {
       val g = new SerializableFunction[java.lang.Iterable[Any], Void] {
         val s = size  // defeat closure
         override def apply(input: JIterable[Any]): Void = {
-          assert(input.asScala.size == s)
+          val inputSize = input.asScala.size
+          assert(inputSize == s, s"expected collection to have size of $s, but was $inputSize")
```
nevillelyh: Can you use the message "SCollection had size $x instead of expected size $y" so it's consistent with ScalaTest's "should have size" matcher?

ElPicador:
Changed!

Issue #216, opened 2016-08-17
Support for both legacy and SQL 2011 BigQuery syntax


nevillelyh:
It's probably better to support both in parallel for easy transitioning. There're 2 options as I see it.

1 Feature flag. To do this we need

- Some annotation e.g. `@BigQueryOptions.legacySql` to be used with `@BigQueryType.fromQuery`
- Additional Boolean arguments to `ScioContext#bigQuerySelect`, `BigQueryClient#getQueryRows` and `BigQueryClient#getQuerySchema`, etc.

2 Automatically detect query syntax

- Either with comments in the query string i.e. `-- legacy.sql=true`
- Or by patterns in the query string, e.g. `[project:dataset.table]` vs `` `project.dataset.table` ``

I'm leaning towards 2 but not sure which solution. Comments is more predictable but seems hacky. Automatically detection sounds nice but there might be edge cases?

andrewsmartin:
@nevillelyh +1 on automatic detection - AFAIK just the presence of a `:` in the query string should be enough because to the best of my knowledge it is not allowed in field names and isn't part of the language itself, only the project name in legacy SQL. Will have to double check this though.

andrewsmartin:
The only downside is that there isn't an explicit way for users to switch in the future - as in it's not so obvious that just changing the query syntax is all that's needed. But as long as we document it I guess it's fine?

nevillelyh:
There're a few catch though. In legacy SQL project name can be omitted, i.e. `[dataset.table]` and it'll default to the billing project. Also there could be extreme edge cases like `WHERE field LIKE "[abc:def.ghi]"`. Just wanna make sure we cover all the grounds.

nevillelyh:
http://stackoverflow.com/questions/39003334/programmatically-detemine-if-a-query-is-legacy-or-sql-2011-syntax

nevillelyh:
Got auto detection working in https://github.com/spotify/scio/tree/neville/bq
Need to polish it up and add integration tests before submitting PR

ravwojdyla:
Is it really a problem for users to decide which syntax to use, and which deps to select? I would prefer to be explicit over implicit, and just stick with the simple and _working_ solution - separate deps, with going forward default to sql-2011 starting from scio `0.2.0`? I see how this is a fancy etc - but why complicate things (less code -> less bugs)?

nevillelyh:
Fixed in #218

PR #218, opened 2016-08-18
support both legacy and SQL syntanx for BigQuery

nevillelyh:
This addresses #216

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
@@ -86,6 +86,7 @@ private[scio] trait QueryJob {
 }
 
 /** A simple BigQuery client. */
+// scalastyle:off number.of.methods
 class BigQueryClient private (private val projectId: String,
                               credential: Credential = null) { self =>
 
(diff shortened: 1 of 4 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -390,6 +417,42 @@ class BigQueryClient private (private val projectId: String,
...
+      val jobConfig = new JobConfiguration().setQuery(queryConfig).setDryRun(true)
+      val job = new Job().setConfiguration(jobConfig)
+      bigquery.jobs().insert(projectId, job).execute()
+    }
+    try {
+      run(false)
+      false
+    } catch {
+      case e: GoogleJsonResponseException =>
+        if (e.getDetails.getErrors.get(0).getReason == "invalidQuery") {
```
ravwojdyla: User `errorExtractor`

nevillelyh: Where's that from?

ravwojdyla: `com.google.cloud.hadoop.util.ApiErrorExtractor`

nevillelyh: That class doesn't have a "invalidQuery" reason and `getErrorInfo` is protected :(

PR #219, opened 2016-08-24
Add skewing to timestampBy

ElPicador:
This skewing thing is only useful when using streaming jobs, and a custom timestamp.

As I'm not 100% about the difference between `DoFn.ProcessContext.outputWithTimestamp` and `WithTimestamps` I preferred to use the latest.

Further more, the skew is [disabled by default](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/DoFn.java#L292-L311) as the one in `WithTimestamps` [is not](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/WithTimestamps.java#L124-L127).

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -774,10 +774,12 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
 
   /**
    * Assign timestamps to values.
+   * With a optional skew
    * @group window
    */
-  def timestampBy(f: T => Instant): SCollection[T] =
-    this.parDo(FunctionsWithWindowedValue.timestampFn(f))
+  def timestampBy(f: T => Instant, allowedTimestampSkew: Duration = Duration.ZERO): SCollection[T] =
+    this.applyTransform(WithTimestamps.of(Functions.serializableFn(f))
+      .withAllowedTimestampSkew(allowedTimestampSkew))
 
   // =======================================================================
   // Write operations
--- a/scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
@@ -443,4 +443,13 @@ class SCollectionTest extends PipelineSpec {
     }
   }
 
+  it should "support timestampBy() with skew"  in {
+    runWithContext { sc =>
+      val p = sc.parallelize(Seq(1, 2, 3))
+      val r = p.timestampBy(new Instant(_), Duration.millis(1))
+        .withTimestamp.map(kv => (kv._1, kv._2.getMillis))
+      r should containInAnyOrder (Seq((1, 1L), (2, 2L), (3, 3L)))
+    }
+  }
+
 }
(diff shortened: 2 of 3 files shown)


ElPicador:
I'm not sure to understand the error in travis:
```
The command "openssl aes-256-cbc -K $encrypted_1d86b5bcde73_key -iv $encrypted_1d86b5bcde73_iv -in scripts/data-integration-test-421ab9f796e8.json.enc -out scripts/data-integration-test-421ab9f796e8.json -d" failed and exited with 1 during .
```
anyone could help me on this?

nevillelyh:
That's the credential for integration tests but apparently it doesn't work for PR. I'll take a look later.

Out of office now so I'll take a look at this PR later.

ElPicador:
Ok, I repushed something to rerun travis, worked this time

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -779,6 +779,10 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   def timestampBy(f: T => Instant): SCollection[T] =
     this.parDo(FunctionsWithWindowedValue.timestampFn(f))
 
+  def timestampByWithSkew(f: T => Instant, skew: Duration): SCollection[T] =
+    this.applyTransform(WithTimestamps.of(Functions.serializableFn(f))
```
nevillelyh: `WithTimestamps.of` is new to me. It wasn't available when I wrote `timestampBy`.

Do you mind do the following:

- add an optional `allowedTimestampSkew` argument (that defaults to `Duration.ZERO`) to `timestampBy` instead of creating a new method
- remove the `FunctionsWithWindowedValue.timestampFn` method
- update ScalaDocs and tests accordingly.

nevillelyh:
Sorry for the reply but this is a perfect case for removing our own implementation and adapt to Dataflow SDK updates. Thanks!

ElPicador:
Are the changes what you expected?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -774,10 +774,12 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
 
   /**
    * Assign timestamps to values.
+   * With a optional skew
    * @group window
    */
-  def timestampBy(f: T => Instant): SCollection[T] =
-    this.parDo(FunctionsWithWindowedValue.timestampFn(f))
+  def timestampBy(f: T => Instant, skew: Duration = Duration.ZERO): SCollection[T] =
```
nevillelyh: I'd name the 2nd argument `allowedTimestampSkew` to match DF API. Plus it's clearer when used as a named arg.

Would be nice to remove `FunctionsWithWindowedValue.timestampFn` which I believe is unused now.

scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
```diff
@@ -443,4 +443,13 @@ class SCollectionTest extends PipelineSpec {
     }
   }
 
+  it should "support timestampBy() with skew"  in {
```
nevillelyh: This test doesn't seem to verify the skew logic. Is there an easy way to do so?

ElPicador: A part from an integration test, I don't know how to test it:/

PR #221, opened 2016-08-25
Refactor cache key

ravwojdyla:
* happy path - no if
 * use murmur instead of sha1 - no security requirements


nevillelyh:
"happy path - no if" what does that mean?
Also can you explain the "no security requirements" part?

ravwojdyla:
* remove `if` statement and use proper API to create parent dirs.
* murmur3 should be faster than sha1, mostly given the fact that sha was optimized for security, while murmur was not.

Issue #220, opened 2016-08-25
Support BQ macros in Intellij


ravwojdyla:
Currently users need a warkaround to make intellij happy, as documented here:
https://github.com/spotify/scio/wiki/FAQ#how-to-make-intellij-idea-work-with-type-safe-bigquery-classes

Add support for our macros.

ravwojdyla:
Plugin is available in Intellij plugin repository. Code is [here](https://github.com/spotify/scio-idea-plugin). Plugin can be used together with next release of scio (0.2.2). Closing.

PR #222, opened 2016-08-25
Dump pretty code of BQ macros if running in IntelliJ

ravwojdyla:
If we dump macro expanded code, we can then use scala plugin
synthetic members injector to support code completion etc.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
@@ -21,6 +21,7 @@ import java.util.{List => JList}
 
 import com.google.api.services.bigquery.model.{TableFieldSchema, TableSchema}
 import com.google.common.base.Charsets
+import com.google.common.hash.{HashCode, Hashing}
 import com.google.common.io.Files
 import com.spotify.scio.bigquery.types.MacroUtil._
 import com.spotify.scio.bigquery.{BigQueryClient, BigQueryUtil}
@@ -84,19 +85,13 @@ private[types] object TypeProvider {
         val caseClassTree = q"""${caseClass(c)(name, fields, body)}"""
         (q"""$caseClassTree
             ${companion(c)(name, Nil, Seq(defSchema, defToPrettyString), fields.asInstanceOf[Seq[Tree]].size)}
-        """, caseClassTree, name)
+        """, caseClassTree, name.toString())
       case t => c.abort(c.enclosingPosition, s"Invalid annotation $t")
     }
     debug(s"TypeProvider.toTableImpl:")
     debug(r)
 
-    if(isRunningInIntelliJ) {
-      // TODO scala 2.11
-      // import c.universe._
-      // showCode(r)
-      val prettyCode = pShowCode(c)(Seq.empty, caseClassTree).mkString("\n")
-      dumpCode(prettyCode, s"$name.scala")
-    }
+    if(isRunningInIntelliJ) dumpCodeForScalaPlugin(c)(Seq.empty, caseClassTree, name)
 
     c.Expr[Any](r)
   }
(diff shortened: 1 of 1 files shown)


ravwojdyla:
This is WIP. This is part of #220 
It can be used together with https://github.com/ravwojdyla/scio-idea-plugin

ravwojdyla:
You can try out the plugin [here](https://github.com/ravwojdyla/scio-idea-plugin/releases/download/0.1.0-SNAPSHOT/scio-idea-plugin.zip):
1. download zip
2. `Preferences` -> `Plugins` -> `Install plugin from disk` 

Tested in Mac and for now only `BQT.fromQuery`. Could you @nevillelyh @andrewsmartin please test.

ravwojdyla:
Current plugin implementation -> https://github.com/ravwojdyla/scio-idea-plugin : 
1. looks for macro expanded dumps of code in well known location (this can technically be a problem if there are the same class names, I need to figure something better - that said, it will never cause incorrect code to compile)
2. parses scala files -> it's looking for case classes mostly
3. injects case classes and companion object into the context (containing class/object)

For macros to be expanded in IntelliJ you need to make/compile the project - for example via `Build` -> `Make Project`. Also every time you change BQ macros you have to recompile the project, obviously.

If you get the location (EU/US) error - you need to add location setting to scala compiler in your IntelliJ, the easiest way is to click on the weird looking "clock" in the bottom-left corner - it's a shortcut to scala compiler - if it's already running - stop it, then press `Configure...` and in `JVM parameters` add `-Dbigquery.staging_dataset.location=EU`.

ravwojdyla:
Also - technically I think it's possible for the plugin to fetch schema from BigQuery by itself or reuse schema cache, and build case classes - this process might be more expensive and complicated thus decided to give the approach above a try.

The (happy path) plan would be:
- [ ] this review - decide if we want to move forward and how
- [ ] implement the rest of the macros (don't see any blockers there - "yet")
- [ ] find some users to test
- [ ] publish the plugin in public plugin repo and/or submit injector to scala-plugin itself

ravwojdyla:
Unfortunately - there is no `showCode` in scala `2.10.x`, `showCode` originally comes from https://github.com/VladimirNik/sprinter - unfortunately it seems like it's no longer published as artifact.

ravwojdyla:
Most recent plugin release with support for:
 * `fromQuery`
 * `fromTable`
 * `toTable`
 * `fromSchema`
is available to download [here](https://github.com/ravwojdyla/scio-idea-plugin/releases/download/0.1.0-SNAPSHOT-alpha.1/scio-idea-plugin.zip)

ravwojdyla:
Most recent plugin release with support for hash (up to date with current head) -> [here](https://github.com/ravwojdyla/scio-idea-plugin/releases/download/0.1.0-RC2/scio-idea-plugin.zip).

ravwojdyla:
@andrewsmartin @nevillelyh could you please review/comment. I would consider it ready to test with a couple of users.

andrewsmartin:
:+1:, this is awesome Rav!

ravwojdyla:
@nevillelyh could you pls test on linux?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -73,17 +77,21 @@ private[types] object TypeProvider {
...
+        (q"""$caseClassTree
             ${companion(c)(name, Nil, Seq(defSchema, defToPrettyString), fields.asInstanceOf[Seq[Tree]].size)}
-        """
+        """, caseClassTree, name.toString())
       case t => c.abort(c.enclosingPosition, s"Invalid annotation $t")
     }
     debug(s"TypeProvider.toTableImpl:")
     debug(r)
+
+    if(isRunningInIntelliJ) dumpCodeForScalaPlugin(c)(Seq.empty, caseClassTree, name)
```
nevillelyh: Space after if and `{}` around code.
`if () {}`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -137,19 +145,23 @@ private[types] object TypeProvider {
...
             ${companion(c)(name, traits, Seq(defSchema, defToPrettyString) ++ overrides, fields.size)}
             ..$records
-        """
+        """, caseClassTree, name.toString())
       case t => c.abort(c.enclosingPosition, s"Invalid annotation $t")
     }
     debug(s"TypeProvider.schemaToType[$schema]:")
     debug(r)
+
+    if(isRunningInIntelliJ) dumpCodeForScalaPlugin(c)(records, caseClassTree, name)
```
nevillelyh: `if (cond) { code }`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -220,6 +232,74 @@ private[types] object TypeProvider {
...
   }
 
+  /**
+   * Check if code is running in IntelliJ IDEA. It may return false positives.
+   *
+   * This is used to mitigate lack of support for Scala macros in IntelliJ.
+   */
+  private def isRunningInIntelliJ = {
+    val classPath = sys.props("java.class.path")
+    classPath.contains("IntelliJ IDEA")
```
nevillelyh: `idea-IC` for community edition and `idea-IU` for ultimate edition on Linux.

PR #223, opened 2016-08-25
bump bigtable version

rgruener:
@nevillelyh @andrewsmartin @ravwojdyla This fixes the bug which caused duplicate rows when performing a bigtable scan. I would recommend a new scio release with this change.


andrewsmartin:
Cool, I'll merge this now. @nevillelyh should we release 0.2.1 right away or could it wait?

PR #225, opened 2016-08-26
feat(scollection): Add support for saveAsText with custom parameters

ElPicador:
Writing to a files can have a lot of parameter so I just decided to pass a function to add any parameter.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -28,6 +28,7 @@ import com.google.api.services.bigquery.model.{TableReference, TableRow, TableSc
 import com.google.api.services.datastore.DatastoreV1.Entity
 import com.google.cloud.dataflow.sdk.coders.{Coder, TableRowJsonCoder}
 import com.google.cloud.dataflow.sdk.io.BigQueryIO.Write.{CreateDisposition, WriteDisposition}
+import com.google.cloud.dataflow.sdk.io.TextIO.Write.Bound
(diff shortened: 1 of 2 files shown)


nevillelyh:
Do we really need this? I prefer to not overload the API with too many options.
The `numShards` option is already exposed in `saveAsText`. And there's always `applyTransform` for people who need to drop down to DF API.

ElPicador:
I don't know if you need this, be we do 😄.
And not having to reimplement a custom `applyTransform` just to pass one more option to the `TextIO` would be pretty cool.

I understand the need to hide the underlying Dataflow API, but for some parameters it might be interesting to just let the user do whatever he wishes. `TextIO.Bound` as quite a lot of parameters that are usefull, [see](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/2dd0fabf16ec06869f0c591e4156e07ed958f22b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/TextIO.java#L408-L474).

For example, in our use case, we need the `withoutSharding` and `withCoder`.

ravwojdyla:
Agree with @nevillelyh here - would prefer to keep the API as simple as possible while still easy to drop down to DF, or extend yourself.

I get the point about the `with*` you need, that said about `withoutSharding` - unless you care about sharding template, you might consider setting number of shards to 1. On the other hand would dropping to DF be good enough (?):

```
TextIO.Write.to(pathWithShards(path)).withoutSharding().withSuffix(suffix).withCoder([CODER])
```

ElPicador:
Ok, so for this we'll have to recode `saveAsText` just for setting a param :/

PR #227, opened 2016-08-29
Enforce macro is not enclosed by package.

ravwojdyla:
Therefore enforce macro being declared inside class/object etc

### DIFF
--- a/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/BigQueryTypeIT.scala
+++ b/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/BigQueryTypeIT.scala
@@ -23,21 +23,25 @@ import org.scalatest.{FlatSpec, Matchers}
 
 import scala.collection.JavaConverters._
 
-@BigQueryType.fromQuery(
-  "SELECT word, word_count FROM [bigquery-public-data:samples.shakespeare] WHERE word = 'Romeo'")
-class LegacyT
+object BigQueryTypeIT {
+  @BigQueryType.fromQuery(
+    "SELECT word, word_count FROM [bigquery-public-data:samples.shakespeare] WHERE word = 'Romeo'")
+  class LegacyT
 
-@BigQueryType.fromQuery(
-  "SELECT word, word_count FROM `bigquery-public-data.samples.shakespeare` WHERE word = 'Romeo'")
-class SqlT
+  @BigQueryType.fromQuery(
+    "SELECT word, word_count FROM `bigquery-public-data.samples.shakespeare` WHERE word = 'Romeo'")
+  class SqlT
 
-@BigQueryType.fromTable("bigquery-public-data:samples.shakespeare")
-class FromTableT
+  @BigQueryType.fromTable("bigquery-public-data:samples.shakespeare")
+  class FromTableT
 
-@BigQueryType.toTable
-case class ToTableT(word: String, word_count: Int)
+  @BigQueryType.toTable
+  case class ToTableT(word: String, word_count: Int)
+}
 
 class BigQueryTypeIT extends FlatSpec with Matchers {
+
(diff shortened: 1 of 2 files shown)


ravwojdyla:
This is to provide a user with more meaningful error, and also to enforce better policy -> macro should be used inside class/object.

Works fine in REPL.

The problem:

* `toTable` and `fromSchema` as top level will compile just fine.

* `fromQuery` and `fromTable` won't compile due compilation error:
```
top-level class without companion can only expand either into an eponymous class or into a block consisting in eponymous companions
```

The fact that compiler acts differently on different macros, might also be misleading to users.

Also this will simplify scio plugin implementation - see ravwojdyla/scio-idea-plugin#1

ravwojdyla:
So the it-tests had macros including `fromQuery` in the top level yet - they were not failing (which actually makes sense - cause there should be companion object generated), with:
```
top-level class without companion can on ...
```
Yet when compiling in IntelliJ I get the error, in `sbt compile` there is no error (explains why travis is fine). I need check where is this difference coming from.

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/BigQueryTypeIT.scala
```diff
@@ -47,7 +49,7 @@ class BigQueryTypeIT extends FlatSpec with Matchers {
     "SELECT word, word_count FROM `bigquery-public-data.samples.shakespeare` WHERE word = 'Romeo'"
 
   "fromQuery" should "work with legacy syntax" in {
-    val bqt= BigQueryType[LegacyT]
+    val bqt= BigQueryType[BigQueryTypeIT.LegacyT]
```
nevillelyh: do a `import BigQueryTypeIT._` at the beginning of the class to avoid these lengthy names?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -220,6 +222,18 @@ private[types] object TypeProvider {
       q"override def toTableRow: ($name => ${p(c, GModel)}.TableRow) = ${p(c, SType)}.toTableRow[$name]")
   }
 
+  /** Enforce macro is not enclosed by package, therefore should be enclosed by class/object etc. */
```
nevillelyh: "Enforce that the macro is not enclosed by a package, but a class or object instead."

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -220,6 +222,18 @@ private[types] object TypeProvider {
...
+  private def checkMacroEnclosed(c: Context): Unit = {
+    // TODO: scala 2.11
+    //if (!c.internal.enclosingOwner.isClass) {
+    // c.abort(c.enclosingPosition,
+    //  "Invalid macro application - must be applied within class/object.")
+    // }
+    if (c.enclosingClass.isEmpty) {
+      c.abort(c.enclosingPosition,
+        "Invalid macro application - must be applied within class/object.")
```
nevillelyh: This can still be confusing for average users. "@BigQueryType.${name} declaration must be inside a class or object."?

ravwojdyla: 👍 that said will skip the name, that we get from `enclosingPosition`. So the error is gonna be (for example - you click on the error and gets you to the exact line etc):
```
Error:(5, 2) @BigQueryType declaration must be inside a class or object.
@BigQueryType.fromQuery("""|SELECT activity.value.value.value.reporting_country
```

nevillelyh:
👍

PR #228, opened 2016-08-29
fix(FileStorage): Catch and handle IOException gracefully

aseure:
When an Avro file is read, a `java.io.IOException` is thrown if the file
doesn't exist.

This commit makes `scio.io.FileStorage` return an `Iterator[T]()` instead of
letting the Java exception propagates.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
@@ -62,7 +62,9 @@ private trait FileStorage {
     } else {
       new GenericDatumReader[T](schema)
     }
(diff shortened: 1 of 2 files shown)


andrewsmartin:
Hey, thanks for the contribution. In my opinion it's probably better to not silently hide this error - if an input file is not found, I think failing fast is the better choice because it's likely a mistake setting the input, or something else that's critical, which means the job shouldn't run.

@nevillelyh @ravwojdyla do you have any thoughts?

nevillelyh:
+1, throwing exception for non-existent file should be the expected behavior.

aseure:
Understood. Thank you for the quick review.

PR #230, opened 2016-08-31
Added namespace option to datastore query method.

sinaso:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -457,13 +457,26 @@ class ScioContext private[scio] (val options: PipelineOptions,
    * Get an SCollection for a Datastore query.
    * @group input
    */
-  def datastore(datasetId: String, query: Query): SCollection[Entity] = pipelineOp {
-    if (this.isTest) {
-      this.getTestInput(DatastoreIO(datasetId, query))
-    } else {
-      wrap(this.applyInternal(gio.DatastoreIO.readFrom(datasetId, query)))
+  def datastore(datasetId: String, query: Query): SCollection[Entity] =
```
nevillelyh: Instead of overloading, just add a third argument `name: String = null`?

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -101,8 +101,8 @@ case class AvroIO[T](path: String) extends TestIO(path)
 
 case class BigQueryIO(tableSpecOrQuery: String) extends TestIO[TableRow](tableSpecOrQuery)
 
-case class DatastoreIO(datasetId: String, query: Query = null)
-  extends TestIO[Entity](s"$datasetId\t$query")
+case class DatastoreIO(datasetId: String, namespace: String = null, query: Query = null)
+  extends TestIO[Entity](s"$datasetId:$namespace\t$query")
```
nevillelyh: Use `\t` instead of `:` here?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
sinaso: I thought (datasetId, namespace, query) looks like a more natural ordering for this. Yet I can make the change if you insist. Can we keep it with this order?

nevillelyh: We prefer `(datasetId, query, namespace=null)` to keep it simple.

sinaso:
Done. Please take a look.

nevillelyh:
Thanks!

PR #234, opened 2016-09-02
Add function trait to companion object from BigQuerytype.toTable

andrewsmartin:
This is so that the companion object can be passed around as a function, consistent with normal companion behaviour.

### DIFF
--- a/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
+++ b/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
@@ -272,6 +272,9 @@ class TypeProviderTest extends FlatSpec with Matchers {
   @BigQueryType.toTable
   case class ToTable(f1: Long, f2: Double, f3: Boolean, f4: String, f5: Instant)
 
+  @BigQueryType.toTable
+  case class Record(f1: Int, f2: String)
+
   "BigQueryType.toTable" should "support .tupled in companion object" in {
     val r1 = ToTable(1L, 1.5, true, "hello", NOW)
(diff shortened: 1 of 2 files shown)


scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -21,6 +21,8 @@ import com.google.api.services.bigquery.model.TableRow
 import org.joda.time.Instant
 import org.scalatest.{FlatSpec, Matchers}
 
+import scala.runtime.{AbstractFunction4, AbstractFunction5}
```
nevillelyh: Unused imports?

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -290,4 +295,16 @@ class TypeProviderTest extends FlatSpec with Matchers {
     (classOf[(ToTable => TableRow)] isAssignableFrom ToTable.toTableRow.getClass) shouldBe true
   }
 
+  it should "create companion object that is a Function subtype" in {
+    (classOf[Function5[Long, Double, Boolean, String, Instant, ToTable]] isAssignableFrom ToTable.getClass) shouldBe true
+    (classOf[Function2[Int, String, Record]] isAssignableFrom Record.getClass) shouldBe true
+  }
+
+  it should "create companion object that is functionally equal to its apply method" in {
+    def apply$(f: (Int, String) => Record)(x: (Int, String)): Record = f(x._1, x._2)
```
nevillelyh: Avoid non-alpha-numeric characters?
`def doApply` maybe?

nevillelyh:
:+1:

PR #235, opened 2016-09-02
Test schema compatibility in typedBigQuery

ravwojdyla:


ravwojdyla:
A couple of changes here:
 * 7d2655c - `bigquery.project` if set, trumps Dataflow's `GcpOptions` - which I think makes sense
 * 7d2655c - validate that `GcpOptions.getProject` is not null, before we pass it to BigQuery client
 * d068055 - initial tests for schema compatibility in `typedBigQuery` and integration tests for Scio pipeline with BigQuery
 * f4c0a20 - BigQuery scala files are not needed on travis

ravwojdyla:
@nevillelyh @andrewsmartin review pls?

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -233,4 +244,10 @@ package object experimental {
 
   }
 
+  private def areSchemasCompatible(from: TableSchema, to: TableSchema): Boolean = {
+    import scala.collection.JavaConverters._
+    val fields2 = to.getFields.asScala
+    from.getFields.asScala.forall(fields2.contains(_))
```
nevillelyh: What about nested RECORDs?

andrewsmartin:
👍 LGTM

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -85,8 +86,18 @@ package object experimental {
...
         val table = scala.util.Try(BigQueryIO.parseTableSpec(newSource)).toOption
         if (table.isDefined) {
+          val newTableSchema = bigQueryClient.getTableSchema(table.get)
+          require(areSchemasCompatible(bqt.schema, newTableSchema),
+            s"""|New source $newSource has incompatible schema.
+                |New source schema ${newTableSchema.toString},
+                |Original schema ${bqt.schema.toString}!""".stripMargin)
           self.bigQueryTable(table.get)
         } else {
+          val newQuerySchema = bigQueryClient.getQuerySchema(newSource)
```
nevillelyh: This `newQuerySchema` may still be from view if query is legacy syntax. How do you plan to fix this?

ravwojdyla:
@nevillelyh could you give reproducible example that current logic breaks.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
ravwojdyla: Add some extra tests.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
ravwojdyla: Out of scope for this change.

ravwojdyla:
There must be something with the integration tests and this PR, cause other PR are relatively fine +/- the unpredictable sampling algorithms.

nevillelyh:
This has been sitting for a long time. IMHO I prefer to abandon this since it's not critical and diverges further from upstream BigQuery handling mechanism.

ravwojdyla:
There is nothing wrong with the logic here - only that travis is timing out. I would like to find out why, cause this PR also introduces some IT tests.

PR #239, opened 2016-09-06
fix ScalaCheck tests and race condition in JobTest

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -86,7 +86,7 @@ object ScioContext {
   /** Create a new [[ScioContext]] instance for testing. */
   def forTest(): ScioContext = {
```
ravwojdyla: Would it make sense to change the signature here and expect `Class name` here. There would be only one way to get `testId`?

andrewsmartin:
👍

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: Well class name doesn't really matter in a test so I'd like to keep it this way.

PR #248, opened 2016-09-13
Closes #247. Add debug to SCollection.

### LINKED ISSUE #247: Add `debug` transformation
Add `debug` transformation to dump content of the SCollection. Useful for local debugging of tests or logic.

Technically one could use `materialize` or `textFile` etc, but in both cases it requires changes to the pipeline logic, and are not as simple as a drop in `debug` transformation known from for example scalding.

ravwojdyla:


nevillelyh:
This will only work with in process runner right?

ravwojdyla:
@nevillelyh this will work on any runner. On Cloud Dataflow runners, `Console.out` goes to worker logs as `INFO`, `Console.err` goes to worker logs as `ERROR`. But again I see this as local debugging, if you plan to log anything in Cloud Dataflow, a proper Logger (+ sampling) is a better idea.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -511,6 +511,19 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       .toSCollection
   }
 
+  /**
+   * Print content of a SCollection to `out()`.
+   * @group debug
+   */
+  def debug(out:() => PrintStream = () => Console.out, prefix: String = ""): SCollection[T] =
```
nevillelyh: Space after :

nevillelyh:
:+1:

Issue #250, opened 2016-09-14
BigQueryType class name collision


nevillelyh:
Looks like a name collision. Top level class is `Track` while there's a nested `Track$`.

```
scio> @BigQueryType.fromTable("di-bigquery-data-commons:track_entity.track_entity_20160901") class Track
[main] INFO com.google.cloud.dataflow.sdk.options.GcpOptions$DefaultProjectFactory - Inferred default GCP project 'gabocontinuousintegration' from gcloud. If this is the incorrect project, please cancel this Pipeline and specify the command-line argument --project.
<console>:29: warning: Class Track$ differs only in case from Track$. Such classes will overwrite one another on case-insensitive filesystems.
       @BigQueryType.fromTable("di-bigquery-data-commons:track_entity.track_entity_20160901") class Track
        ^
defined class Track
defined object Track
defined class MasterMetadata$
defined class Track$
...

scio> bq.getTypedRows[Track]()
java.lang.NoSuchFieldException: MODULE$
  at java.lang.Class.getField(Class.java:1690)
  at scala.reflect.runtime.ReflectionUtils$.staticSingletonInstance(ReflectionUtils.scala:68)
  at scala.reflect.runtime.JavaMirrors$JavaMirror$JavaModuleMirror.instance(JavaMirrors.scala:530)
  at com.spotify.scio.bigquery.types.BigQueryType.<init>(BigQueryType.scala:199)
  at com.spotify.scio.bigquery.types.BigQueryType$.apply(BigQueryType.scala:182)
  at com.spotify.scio.experimental.package$TypedBigQueryClient.getTypedRows(package.scala:186)
  ... 32 elided
```

ravwojdyla:
Need to fix tests and add more tests - but seems like repl doesn't like the `$`, enforcing number suffix from the very beginning helps (as in the `closing_250` branch). If looks 👍 will fix tests and open PR.

nevillelyh:
I was thinking of wrapping them in an object but that's a much bigger change.

ravwojdyla:
I see - yea that would be more code changes. So you are ok with the counting from 0, correct?

PR #251, opened 2016-09-14
Closing #250. Start counting from 1

ravwojdyla:


ravwojdyla:
@nevillelyh review pls?

nevillelyh:
👍 

We can think about wrapping in `object X {}` later.

PR #253, opened 2016-09-14
Bump sdk to 1.7.0

ravwojdyla:


ravwojdyla:
Let's see what needs to be fixed.

mpavlov:
@ravwojdyla Looks like 1.7.0 renames the datastore client path and some methods, and changes slightly the workflow of adding properties to datastore entities. The old path still works, but that gets you v1beta2 of the datastore API which [shuts down in 2 weeks](https://cloud.google.com/datastore/docs/apis#cloud-datastore-api-v1beta2-deprecated).

A naive update to the v1 client looks like this: https://github.com/spotify/scio/compare/update_sdk...mpavlov:update_sdk?expand=1

I'll run it through the tests, the examples, and also our team's pipeline which uses scio to write to datastore. Let me know if there are additional verification steps I should perform. If all works out I'll submit a new PR (since I can't push to this one).

scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
```diff
@@ -33,19 +34,33 @@ import scala.util.control.NonFatal
 trait SCollectionMatchers {
 
   private def m(f: () => Any): MatchResult = {
-    val r = try { f(); true } catch { case NonFatal(_) => false }
```
ravwojdyla: obsolete.

nevillelyh:
👍

PR #255, opened 2016-09-14
Fix #254 not found: type Function23

### LINKED ISSUE #254: "error: not found: type Function23" for BigQueryType
Probably a regression from https://github.com/spotify/scio/commit/ed78cfa0e631149803626d5288d26d1613cfcc02

Should be easily fixable by not extending the trait if arity > 22.

```
scio> @BigQueryType.toTable
     | case class Test (
     | a1: Int, a2: Int, a3: Int, a4: Int, a5: Int, a6: Int, a7: Int, a8: Int,
     | a9: Int, a10: Int, a11: Int, a12: Int, a13: Int, a14: Int, a15: Int,
     | a16: Int, a17: Int, a18: Int, a19: Int, a20: Int, a21: Int, a22: Int,
     | a23: Int
     | )
[main] INFO com.spotify.scio.bigquery.types.TypeProvider$ - Will dump generated Test of $line9.$read.$iw.$iw.$iw.$iw.$iw.$iw.$iw.$iw.$iw.$iw.$iw.$iw from <console> to /var/folders/hd/f_wgpmsj3h1cv071xszfmyzh0000gn/T/bigquery-classes/Test-5f84b0d6.scala
<console>:29: error: not found: type Function23
       @BigQueryType.toTable
        ^
```

andrewsmartin:


scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -305,4 +305,10 @@ class TypeProviderTest extends FlatSpec with Matchers {
     doApply(Record)((3, "a")) shouldEqual Record(3, "a")
   }
 
+  // This should compile
+  @BigQueryType.toTable
+  case class TwentyThree(a1:Int,a2:Int,a3:Int,a4:Int,a5:Int,a6:Int,a7:Int,a8:Int,a9:Int,a10:Int,
```
ravwojdyla: Isn't this gonna fail on 2.10 here?

andrewsmartin: Yeah you're right. Is there an easy way to flag this for 2.11 only?

ravwojdyla: There are tests for scala-2.11 in `test/scala-2.11`.

andrewsmartin: Nice. Moved it over there.

scio-bigquery/src/test/scala-2.11/com/spotify/scio/bigquery/types/TypeProviderTestScala211.scala
```diff
@@ -64,4 +64,11 @@ class TypeProviderTestScala211 extends FlatSpec with Matchers {
   it should "not provide .tupled in companion object with more than 22 fields" in {
     ArtisanalMoreThan22Fields.getClass.getMethods.map(_.getName) should not contain "tupled"
   }
+  
+  // This should compile
+  @BigQueryType.toTable
```
ravwojdyla: Would it make sense to add some tests on the `TwentyThree` type - similar to the `ArtisanalMoreThan22Fields` tests?

andrewsmartin: I was wondering this and initially decided not to, because technically speaking we don't really care whether the type -say- has a .`tupled` method (at least for the scope of this issue). All we care about is whether it compiles, which for all intents and purposes is the same as not having `.tupled`, but it's not directly the thing we care about. Maybe we should though - am definitely not 100% on this.

ravwojdyla: I would be more on having tests end.

andrewsmartin: @ravwojdyla OK - I added tests.

PR #260, opened 2016-09-17
AvroIO compression support

ravwojdyla:


ravwojdyla:
Do not merge. Pending on #253 #259 

Todo:
 - [ ] add tests
 - [ ] try to expose CodecFactory instead of String for Codec (otherwise not easy to pass Codec Options)
 - [ ] contribute back to DataflowSDK?

nevillelyh:
Is `CoderFactory` serializable? Otherwise it's hard to pass it. Don't wanna go full on Hadoop `Configuration` style for this. I'm happy with just `String codec` for now with some sane defaults.

build.sbt
```diff
@@ -206,6 +207,7 @@ lazy val scioTest: Project = Project(
       "org.scalatest" %% "scalatest" % scalaTestVersion,
       // DataFlow testing requires junit and hamcrest
       "junit" % "junit" % junitVersion,
+      "com.novocode" % "junit-interface" % junitInterfaceVersion,
```
nevillelyh: What's this for?

ravwojdyla: Without it, I believe:
 * no JUnit tests are actually run via SBT
 * there is no feedback from JUnit tests

nevillelyh:
LGTM. Should there be a non-null default though?

Issue #256, opened 2016-09-15
Upgrade to the new Datastore API


ravwojdyla:
Pending upgrade of SDK to `1.7.0` via #253 , let's upgrade [deprecated](https://cloud.google.com/datastore/docs/apis#cloud-datastore-api-v1beta2-deprecated) datastore API.

mpavlov:
I'll work on this

nevillelyh:
There's a PR recently merged. Can we close this?

ravwojdyla:
@nevillelyh I believe this is still pending some changes @mpavlov is working on to move to the new API.

PR #262, opened 2016-09-19
Update to Datastore API v1

mpavlov:
Dataflow SDK 1.7.0 supports both Datastore API v1beta2 and v1, the latter being the newest and the former scheduled for decommissioning at end of month.

We're currently using v1beta2. This PR moves us to v1. (#256)

Note: There is a known bug in `google-cloud-java` which may manifest itself in scio users getting `java.lang.VerifyError` when trying to build Datastore `Value` objects. The bug, a workaround, and progress on fixing it are tracked in https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1239


ravwojdyla:
👍

Issue #261, opened 2016-09-19
Fix flaky tests


ravwojdyla:
There are flaky tests causing travis to fail, which requires manual work (restarts).
Tests to fix:
- [x] Algebird.aggregate of Double with multiple aggregators [example](https://travis-ci.org/spotify/scio/jobs/160748144)
- [x] Algebird.sum with HyperLogLog: Falsified after 50 passed tests. [example](https://travis-ci.org/spotify/scio/jobs/160598959)
- [x] PoissonValueSampler [example](https://travis-ci.org/spotify/scio/jobs/158783154)

ravwojdyla:
@nevillelyh you have been working on this before - do you have any pointers, ideas why would this still be a problem?

ravwojdyla:
Algebird.sum with HyperLogLog is still problematic [example](https://travis-ci.org/spotify/scio/jobs/161188960)

nevillelyh:
b1f8003fc752fae4a6a3c3010da9407661399ed4 should fix flaky RandomSamplerTest once and for all.
Fail that I didn't catch it all this time...

PR #264, opened 2016-09-19
fix flaky algebird tests #261

nevillelyh:


ravwojdyla:
I think you need to rebase on current `master`. git does not like you today.

ravwojdyla:
👍

Issue #249, opened 2016-09-13
API is missing the 'collect' function.


seacourtaw:

ravwojdyla:
@seacourtaw could you explain what would mean by `collect`? There already is `materialize` and you can use `SideInputs`.

seacourtaw:
Sure. I was thinking of Seq.collect. Example usage: http://ochafik.com/blog/?p=393

nevillelyh:
That was intentionally left out since it might be confused with Spark's RDD `collect()` which does something completely different.

ravwojdyla:
@nevillelyh correct Spark's RDD has `collect()` for materializing but also functional `collect` + scalding has functional `collect`. I'm 👍 for the PR.

seacourtaw:
@ravwojdyla, @nevillelyh: I'll see if I can get you a PR together at some point.

andrewsmartin:
@seacourtaw my bad for stepping on your toes with https://github.com/spotify/scio/pull/268! Didn't see until now that you were considering putting up a PR yourself. Sorry about that.

andrewsmartin:
Added in 	43f5a5f

seacourtaw:
@andrewsmartin: no toe-stepping. I wouldn't have gotten round to this for another week or two. Many thanks.

PR #268, opened 2016-09-21
Add collect() #249

andrewsmartin:


scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
```diff
@@ -140,6 +140,18 @@ class SCollectionTest extends PipelineSpec {
...
   }
 
+  it should "support collect" in {
+    runWithContext { sc =>
+      val records = Seq(
+        ("test1", 1),
+        ("test2", 2),
+        ("test3", 3)
+      )
+      val p = sc.parallelize(records).collect{case ("test2", x) => 2 * x}
```
nevillelyh: Minor style thing, I'd leave space in `{ code }`

nevillelyh:
:+1:

PR #274, opened 2016-10-03
make KryoAtomicCoder thread local

nevillelyh:


ravwojdyla:
This is purely to remove `synchronization` correct?

nevillelyh:
@ravwojdyla yes that's also what Dataflow SDK does to all its coders.
That should also improve performance now that threads are not locked on a single `kryo` instance.

ravwojdyla:
Nice! 👍

Issue #163, opened 2016-06-06
What's the update strategy for streaming jobs?


casidiablo:
When using plain Java, and if you religiously name every step in the job, upgrading a pipeline can be easy as explained in the official documentation:

https://cloud.google.com/dataflow/pipelines/updating-a-pipeline

Using scio, however, there does not seem to be a way to name steps, especially those created by methods in the SCollection. This is rather inconvenient because the job scheduler names the steps with things like `apply@{SCollection.scala:335}` or `groupBy@{MyClass.scala:93}2`; which means that if you update the scio library or, even worse, slightly change your own code, the new pipeline will no longer be compatible.

I've even tried to name the steps using `.applyTransform(ParDo.named("SaveToS3")...)`, but the resulting step still has the same issue (in this case it was named `applyTransform@{MyClass.scala:140}`).

What is the updating strategy used at Spotify to address this kind of scenarios? Creating a new pipeline is, of course, not an option, since data could be loss.

nevillelyh:
Great question! Most internal jobs are batch with a few guys experimenting with streaming so updating streaming job is not a big concern right now.

There are several options I can think of:
- Some flag to disable transform auto naming e.g. `map@{Job.scala:123}` but we still need to name them consistently
- Overload every method with `(name: String, args...)` but that's messy
- Implement a `@name` annotation so one can do `coll.map(@name("Input") fn)`. I'm not an annotation expert though so will have to play around with this idea first.

What do you think?

casidiablo:
The annotations seem to be the best way to go, definitely; not sure how easy it'd be to implement, though.

I think, unfortunately, I'm going to have to go back to a pure Java job in the meantime. I wish I had thought about this before implementing the whole thing in Scala. I will keep my eyes on this issue to see how this evolves.

Thank you guys for the hard work.

nevillelyh:
Looks like one can only annotate methods or arguments when defining them and not on invocation so the 3rd options is out.

nevillelyh:
Igor mentioned that there's an experimental drain feature that might solve this problem.

nevillelyh:
https://cloud.google.com/blog/big-data/2016/08/managing-streaming-pipelines-in-google-cloud-dataflow-just-got-better

@casidiablo Google just announced this "drain" feature. Not exactly the same as update but should solve the problem to some extent?

ElPicador:
Having the update feature in scio will really be a big improvement for us.
[...]

PR #276, opened 2016-10-05
Have deterministic names for transformations

ElPicador:
Ref #163

Sorry for the delay, I had some health issues.

WDYT of it?

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -364,7 +364,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
 
   private[scio] def applyInternal[Output <: POutput](root: PTransform[_ >: PBegin, Output])
   : Output =
-    pipeline.apply(CallSites.getCurrent, root)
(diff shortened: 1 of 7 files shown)


nevillelyh:
Can you add a flag so this is only turned on in streaming mode? Maybe an `object CallSites` and a `var` inside set by `ScioContext` based on `PipelineOptions`. That way you can also keep the `getCurrent` method name.

I'd also keep the file name part since it's even less likely to change than method name, e.g. one replace `map` with `flatMap`.

And more importantly, have you verified that this works?

ElPicador:
Why only enable this on streaming mode?

I don't really agree on the file name part. You can split your pipeline in multiple files without changing it. So adding the file name prevent some (simple) refactoring.

I've tested it locally. But you are right, I'll test it with one of our pipeline

nevillelyh:
It's still nice to have filename and line number, especially in complex pipelines with many similar transforms. So I prefer to not change the behavior for the large number of existing batch users :)

Don't have a strong opinion on filename in streaming, but it can be confusing if you have transforms of the same name spread across multiple files.

ElPicador:
This won't work on the long run, closing it and continuing discussion on the issue #163

PR #286, opened 2016-10-12
Add materialize accumulators

ravwojdyla:


ravwojdyla:
This is WIP - give feedback pls. In the mean time I will add some tests + polish code.

ravwojdyla:
This supports out of the box - local FS and GCS. And looks like this:
```
➜  scio git:(materialize_metrics) ls /tmp/c3*
/tmp/c3_steps.json
/tmp/c3_total.json
```
content:
```
➜  scio git:(materialize_metrics) cat /tmp/c3_total.json| jq .
[
  {
    "name": "foo2",
    "value": 130
  },
  {
    "name": "foo",
    "value": 65
  }
]
```
steps:
```
➜  scio git:(materialize_metrics) cat /tmp/c3_steps.json| jq .
[
  {
    "name": "foo2",
    "steps": [
      {
        "name": "map@{<console>:33}",
        "value": 130
      }
    ]
  },
  {
    "name": "foo",
    "steps": [
      {
        "name": "map@{<console>:33}",
        "value": 65
      }
    ]
  }
]
```

nevillelyh:
What do you think of 1 output file instead of 2?
```json
{"total": [], "steps": []}
```

Also not sure about the `materializeAccumulators` name, something simpler and more straightforward, like `saveAccumulators`?

ravwojdyla:
👍 to `saveAccumulators`.
About single file - i assume most people/graphs will be interested in `total`, why do you suggest a single file - what is the advantage?

nevillelyh:
Because parsing 1 bigger JSON won't be much different than parsing 2 separate ones, but OTOH `save(filename)` is more intuitive than `save(prefix, total_suffix, steps_suffix)`, and we can extend it later without changing API or semantics.

I'm thinking maybe even call it `saveMetrics` and persist other information like time/resources but we can do that later.

ravwojdyla:
There is no different, only if the user actually parses the 2nd file - which i doubt is gonna happen often. I agree about the intuitive `filename` vs `suffixes` tho.

What about having two methods:
 * `saveMetrics` and `saveStepsMetrics` - explicit and intuitive
 * add `version` field to files too allow for different versions (over time)

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -54,7 +59,42 @@ class ScioResult private[scio] (val internal: PipelineResult,
   def accumulatorValuesAtSteps[T](acc: Accumulator[T]): Map[String, T] =
     getAggregatorValues(acc).flatMap(_.getValuesAtSteps.asScala).toMap
 
+  /** Materialize accumulators of finished pipeline to specified file. */
+  def saveMetrics(filename: String): Unit = {
+    require(isCompleted, "Pipeline has to be finished to dump accumulators.")
```
nevillelyh: `to save metrics.` so it's consistent with method name.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -54,7 +59,42 @@ class ScioResult private[scio] (val internal: PipelineResult,
   def accumulatorValuesAtSteps[T](acc: Accumulator[T]): Map[String, T] =
     getAggregatorValues(acc).flatMap(_.getValuesAtSteps.asScala).toMap
 
+  /** Materialize accumulators of finished pipeline to specified file. */
```
nevillelyh: `Save metrics of the finished pipeline to a file.`

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -54,7 +59,42 @@ class ScioResult private[scio] (val internal: PipelineResult,
...
+    require(isCompleted, "Pipeline has to be finished to dump accumulators.")
+
+    val mapper = new ObjectMapper()
+    mapper.registerModule(DefaultScalaModule)
+
+    val out = IOChannelUtils.create(filename, MimeTypes.TEXT)
+
+    try {
+      val totalValues = accumulators
+        .map(acc => (acc, accumulatorTotalValue(acc)))
```
nevillelyh: Why not just fuse them in a single `map`?

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -54,7 +59,42 @@ class ScioResult private[scio] (val internal: PipelineResult,
...
+        .map(e => AccumulatorValue(e._1.name, e._2))
+
+      val stepsValues = accumulators
+        .map(acc => (acc, accumulatorValuesAtSteps(acc)))
```
nevillelyh: Same here, why not single `map`?

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -54,7 +59,42 @@ class ScioResult private[scio] (val internal: PipelineResult,
...
+      }
+    }
+  }
+
   private def getAggregatorValues[T](acc: Accumulator[T]): Iterable[AggregatorValues[T]] =
     aggregators(acc.name).map(a => internal.getAggregatorValues(a.asInstanceOf[Aggregator[_, T]]))
 
 }
+
+private[scio] case class Metrics(version: Int, accumulators: AccumulatorMetrics)
```
nevillelyh: Do we wanna persist some job info as well? Like job name, ID, scio and scala version? And in that case can `scioVersion` replace `version: Int` here?
It'll make batch processing easier if we save all metrics to the same path.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: omg - yes - this was stupid.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: omg - yes - this was stupid.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: Fixed.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: Fixed.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: Fixed.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: Fixed.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: Fixed.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -54,7 +60,49 @@ class ScioResult private[scio] (val internal: PipelineResult,
...
+
+      val stepsValues = accumulators
+        .map(acc => AccumulatorStepsValue(acc.name,
+          accumulatorValuesAtSteps(acc).map(a => AccumulatorStepValue(a._1, a._2))))
+
+      val options = this.pipeline.getOptions
+      val metrics = Metrics(scioVersion,
+                            scalaVersion,
+                            options.as(classOf[ApplicationNameOptions]).getAppName,
+                            options.as(classOf[DataflowPipelineOptions]).getJobName,
```
ravwojdyla: Not sure if this is completely safe. It works fine for local mode (and dataflow), not sure about other runners.

scio-test/src/test/scala/com/spotify/scio/values/AccumulatorTest.scala
```diff
@@ -117,4 +124,55 @@ class AccumulatorTest extends PipelineSpec {
...
+      AccumulatorStepsValue(sumD.name,
+        r.accumulatorValuesAtSteps(sumD).map(e => AccumulatorStepValue(e._1, e._2))),
+      AccumulatorStepsValue(maxI.name,
+        r.accumulatorValuesAtSteps(maxI).map(e => AccumulatorStepValue(e._1, e._2))),
+      AccumulatorStepsValue(sumL.name,
+        r.accumulatorValuesAtSteps(sumL).map(e => AccumulatorStepValue(e._1, e._2))))
+
+    metrics.version should be(scioVersion)
+    metrics.scalaVersion should be(scalaVersion)
+    metrics.jobName should be(sc.pipeline.getOptions.as(classOf[ApplicationNameOptions]).getAppName)
```
ravwojdyla: `sc.optionsAs` report different values  for jobName and Id - need to look into it.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: For now we support only local and dataflow anyway - so safe anyway.

ravwojdyla:
@nevillelyh any other comments?

scio-test/src/test/scala/com/spotify/scio/values/AccumulatorTest.scala
```diff
@@ -17,10 +17,17 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.google.cloud.dataflow.sdk.options.{ApplicationNameOptions, DataflowPipelineOptions}
+import com.spotify.scio._
 import com.spotify.scio.testing.PipelineSpec
+import org.scalatest.Matchers
 
-class AccumulatorTest extends PipelineSpec {
+class AccumulatorTest extends PipelineSpec with Matchers {
```
nevillelyh: I thought `PipelineSpec` already extends `Matchers`?

ravwojdyla: Correct - fixed.

nevillelyh:
👍

PR #288, opened 2016-10-13
upgrade to bigtable 0.9.3

rgruener:


scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -17,6 +17,7 @@
 
 package com.spotify.scio
 
+import com.google.bigtable.repackaged.com.google.cloud.grpc.BigtableClusterUtilities
```
nevillelyh: This is shaded. Is it intended for public use? At least I'd add a wrapper or type alias in https://github.com/spotify/scio/blob/master/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala#L40

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -80,11 +81,13 @@ package object bigtable {
       if (self.isTest) {
         // No need to update the number of nodes in a test
       } else {
-        BigtableUtil.updateNumberOfBigtableNodes(
```
nevillelyh: A wrapper would make this backwards compatible and reduce Java-y boilerplate.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
rgruener: Its in the release notes so it must be intended for public use. It probably is just shaded in the hbase-dataflow-connector for some reason https://github.com/GoogleCloudPlatform/cloud-bigtable-client/releases

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
nevillelyh: @rgruener can you wrap `BigtableClusterUtilities` in a `object BigtableUtil` to maintain backwards compatibility?

nevillelyh:
:+1:

PR #295, opened 2016-10-17
Set avro version used by sbt-avro

andrewsmartin:
@nevillelyh This coordinates the avro versions used by the project and by sbt-avro. Makes it easier to update since we're no longer tied to the default version used by sbt-avro.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -301,6 +301,7 @@ lazy val scioSchemas: Project = Project(
   file("scio-schemas")
 ).settings(
   commonSettings ++ sbtavro.SbtAvro.avroSettings ++ noPublishSettings ++ PB.protobufSettings,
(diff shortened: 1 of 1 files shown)


nevillelyh:
`scio-schemas` is only used as `"test"` or in `scio-examples` so it's never exposed in published artifacts and shouldn't affect Avro version?

The only version that matters in there is avro-compiler but someone it doesn't override compiler version (1.7.3).

Did you see any case where avro version causes problem?

andrewsmartin:
Doesn't cause any problems now, but it could in the future if we upgrade the avro version in the project but not the sbt-avro (i.e. compiler version), and the upgrade has breaking changes right?

nevillelyh:
I was wrong it does fix avro compiler version.

PR #296, opened 2016-10-17
Save metrics on close + close only once + check acc registration

ravwojdyla:


andrewsmartin:
👍 LGTM

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -321,7 +321,12 @@ class ScioContext private[scio] (val options: PipelineOptions,
         Future.successful(result.getState)
     }
 
-    new ScioResult(result, finalState, _accumulators.values.toSeq, pipeline)
+    val scioResult = new ScioResult(result, finalState, _accumulators.values.toSeq, pipeline)
+    val metricsLocation = optionsAs[ScioOptions].getMetricsLocation
+    if (metricsLocation != null && finalState.isCompleted) {
+      scioResult.saveMetrics(metricsLocation)
```
nevillelyh: This only works with blocking runner? Should support non-blocking ones by setting a Future handler?

ravwojdyla: yes. initially thought that supporting non blocking one would be an overkill. but it would be most consistent. need to move things around a bit.

nevillelyh:
👍

Issue #300, opened 2016-10-19
Refactor scio-hdfs


andrewsmartin:
There is some duplication between scio-core, scio-hdfs and scio-test, specifically with Taps and test helper code.

andrewsmartin:
Closed with https://github.com/spotify/scio/pull/301

PR #301, opened 2016-10-19
Refactor Hdfs #300

andrewsmartin:
@ravwojdyla @nevillelyh could you please review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -295,7 +295,7 @@ lazy val scioHdfs: Project = Project(
   )
 ).dependsOn(
   scioCore,
-  scioTest % "test",
+  scioTest % "test->test",
   scioSchemas % "test"
 )
 
--- a/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
+++ b/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
@@ -3,10 +3,12 @@ package com.spotify.scio.hdfs
 import java.io.{File, InputStream}
 import java.net.URI
 import java.nio.file.Path
+
 import com.spotify.scio.io.FileStorage
 import org.apache.avro.file.SeekableInput
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, PathFilter}
+import org.apache.hadoop.io.compress.CompressionCodecFactory
 
 
 object HdfsFileStorage {
@@ -31,9 +33,16 @@ private class HdfsFileStorage(protected val path: String) extends FileStorage {
   }
 
   override protected def getObjectInputStream(path: Path): InputStream = {
+    val hPath = new org.apache.hadoop.fs.Path(path.toString)
     val conf = new Configuration()
+    val factory = new CompressionCodecFactory(conf)
     val fs = FileSystem.get(path.toUri, conf)
-    fs.open(new org.apache.hadoop.fs.Path(path.toString))
+    val codec = factory.getCodec(hPath)
+    if (codec != null) {
+      codec.createInputStream(fs.open(hPath))
+    } else {
+      fs.open(hPath)
+    }
   }
 
   override protected def getAvroSeekableInput(path: Path): SeekableInput =
--- a/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
+++ b/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -295,7 +295,7 @@ lazy val scioHdfs: Project = Project(
   )
 ).dependsOn(
   scioCore,
-  scioTest % "test",
+  scioTest % "test->test",
```
nevillelyh: what does `"test->test"` mean?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
```diff
@@ -0,0 +1,65 @@
+package com.spotify.scio.hdfs
```
nevillelyh: copyright header

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
```diff
@@ -0,0 +1,65 @@
...
+      !path.getName.startsWith("_") && !path.getName.startsWith(".")
+  }
+
+  override protected def listFiles: Seq[Path] = {
+    val conf = new Configuration()
+    val fs = FileSystem.get(new URI(path), conf)
+
+    fs
+      .listStatus(new org.apache.hadoop.fs.Path(path), pathFilter)
+      .map(status => new File(path + "/" + status.getPath.getName).toPath).toSeq
```
nevillelyh: there's a better way than string concat right? `status.getPath` doesn't give you the full path?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  override protected def getObjectInputStream(path: Path): InputStream = {
+    val hPath = new org.apache.hadoop.fs.Path(path.toString)
+    val conf = new Configuration()
+    val factory = new CompressionCodecFactory(conf)
+    val fs = FileSystem.get(path.toUri, conf)
+    val codec = factory.getCodec(hPath)
```
nevillelyh: Does this work for formats like avro?

nevillelyh: Nevermind it's copied from package.scala so should be fine.

build.sbt
andrewsmartin: From http://www.scala-sbt.org/0.12.2/docs/Getting-Started/Multi-Project.html:

"A useful declaration is "test->test" which means Test depends on Test. This allows you to put utility code for testing in bar/src/test/scala and then use that code in foo/src/test/scala, for example."

nevillelyh: Great!

andrewsmartin: "test" by itself is short for "test->compile"

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
andrewsmartin: Added

nevillelyh:
:+1:

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
```diff
@@ -0,0 +1,82 @@
...
+    if (codec != null) {
+      codec.createInputStream(fs.open(hPath))
+    } else {
+      fs.open(hPath)
+    }
+  }
+
+  override protected def getAvroSeekableInput(path: Path): SeekableInput =
+    new SeekableInput {
```
ravwojdyla: There is no support for this in Avro already?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/HdfsFileStorage.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    fs
+      .listStatus(new org.apache.hadoop.fs.Path(path), pathFilter)
+      .map(status => new File(status.getPath.toString).toPath).toSeq
+  }
+
+  override protected def getObjectInputStream(path: Path): InputStream = {
+    val hPath = new org.apache.hadoop.fs.Path(path.toString)
+    val conf = new Configuration()
+    val factory = new CompressionCodecFactory(conf)
```
ravwojdyla: I have a vague memory of something like this being already supported in avro itself?

scio-test/src/test/scala/com/spotify/scio/io/TapTest.scala
```diff
@@ -33,7 +33,34 @@ import org.apache.commons.io.FileUtils
...
+  def runWithFileFuture[T](fn: ScioContext => Future[Tap[T]]): Tap[T] =
+    runWithFuture(ScioContext())(fn)
+
+  def runWithFuture[T](sc: ScioContext)(fn: ScioContext => Future[Tap[T]]): Tap[T] = {
+    val f = fn(sc)
+    sc.close()
+    f.waitForResult()
+  }
+
+  def tmpDir: File = new File(
```
ravwojdyla: There is `Files.createTempDirectory` etc already

build.sbt
ravwojdyla: 👍

Issue #265, opened 2016-09-19
Add support for introspecting avro-wrapped protobufs


nevillelyh:
Avro `DataFileWriter` has a `.setMeta(String, String)` method and we wrap ProtoBuf bytes in Avro files. Maybe set schema there to make ad-hoc file operations easier, e.g. cat, getschema?

nevillelyh:
```scala
scala> TrackPB.getDescriptor.toProto
res1: com.google.protobuf.DescriptorProtos.DescriptorProto =
name: "TrackPB"
field {
  name: "trackId"
  number: 1
  label: LABEL_REQUIRED
  type: TYPE_STRING
}
```

ravwojdyla:
Cool idea! That said - then the user would have to know that even those she is saving data as protobufs, to find schema she needs to use `avro-tools` and then there is this custom `meta` there etc. Isn't this too low level and "scio" specific?

nevillelyh:
I mean we can make a `proto-tools` for quickly eyeballing a file. Not much work and maybe worth it if ProtoBuf becomes popular.

nevillelyh:
https://github.com/nevillelyh/protobuf-generic

This lib can do the following:
- generate schema/JSON for any compiled protobuf type `T`, which we can attach to Avro files as metadata.
- encode/decode binary protobuf as JSON given the above schema.

I'll look into Scio side of things.

nevillelyh:
BEAM
Issue: https://issues.apache.org/jira/browse/BEAM-813
PR: https://github.com/apache/incubator-beam/pull/1177

nevillelyh:
This will be supported in 0.2.6, due for release tomorrow. Closing this now.
https://github.com/spotify/scio/wiki/Protobuf#file-format

PR #310, opened 2016-10-24
save protobuf generic schema as Avro metadata #265

nevillelyh:


scio-core/src/main/java/com/google/cloud/dataflow/sdk/io/PatchedAvroIO.java
```diff
@@ -18,6 +18,7 @@
```
ravwojdyla: I don't really like how to is diverging from SDK/Beam. We essentially don't want to have this file. Could you create a PR/issue for beam to see if they would accept this?

ravwojdyla:
Is this feature really that important? Is dump raw not enough? Is the use case this is solving just eye balling the results? If one wants to process those PB, they should use compiled protobuf and do it the right way.

scio-core/src/main/java/com/google/cloud/dataflow/sdk/io/PatchedAvroIO.java
nevillelyh: Yes but it'll take a while for them to backport to DataflowJavaSDK so this file will exist for a while.

nevillelyh: Sure will do

nevillelyh:
@ravwojdyla this is synced with upstream PR https://github.com/apache/incubator-beam/pull/1177 and ready to merge. It writes output file readable with https://github.com/spotify/gcs-tools/pull/1/

PR #312, opened 2016-10-26
Fix #311 BigQueryType.toTable should only take first child of field to get the type.

### LINKED ISSUE #311: @BigQueryType.toTable does not compile for case class with default args

andrewsmartin:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -82,7 +82,7 @@ private[types] object TypeProvider {
       case List(q"case class $name(..$fields) { ..$body }") =>
         val defSchema = q"override def schema: ${p(c, GModel)}.TableSchema = ${p(c, SType)}.schemaOf[$name]"
         val defToPrettyString = q"override def toPrettyString(indent: Int = 0): String = ${p(c, s"$SBQ.types.SchemaUtil")}.toPrettyString(this.schema, ${name.toString}, indent)"
-        val fnTrait = tq"${newTypeName(s"Function${fields.size}")}[..${fields.flatMap(_.children)}, $name]"
```
nevillelyh: Can you explain the difference here?

andrewsmartin: Yeah, basically each param is a field that has children. For normal (non-default args), there is only one child which is the type. Default arguments have two children, where the first one is the type and the second is the default value. This just makes sure we always take the first child so we only get types.

nevillelyh: Are we throwing away the tail here then? Does that break default value behavior, e.g. does `RecordWithDefault(10)` still work without `y`?

andrewsmartin: @nevillelyh Yeah, because this is only for generating the type of the `Function` trait, it's not replacing the case class definition. If you'd like I can add another test for this?

nevillelyh: Sure that's be great!

andrewsmartin: Added

nevillelyh:
:+1:

PR #314, opened 2016-10-26
Support Dataflow service metrics

ravwojdyla:


nevillelyh:
👍 
but do we really need the `pending-review` label? ;)

Issue #313, opened 2016-10-26
Need a SideOutput example


sfines:
I have a complex data flow that I'm trying to represent in SCIO on Scala. In the java google libraries, I would branch types by using SideOutputs, however, while I can define SideOutputs in SCIO, I do not see an equivalent to PCollection.get(TupleTag<>) which would allow me to work on the relevant SideOutput- the best I see are map() or flatMap(). 

There seem to be no examples using this mechanism, and the only example I can find in the codebase at all are in SCollection.scala and that's a relatively limited use in randomSplit().

jonopwell:
Yes, an example would be helpful.

nevillelyh:
Did you look at the doc for `SCollection#withSideOutput`?
https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L659

nevillelyh:
Also how do you envision it to work? Side output is not a commonly used pattern so we didn't cover all the possible cases. It'd be great if you can show a snippet of what you intend to do.

sfines:
so basically, we receive a heterogenous collection of data types, and we need to be able to split it and process each subtype with a unique subflow- in java it's something like

`  PCollectionTuple output = converted.get(RowToDataPostHandler.dataPost).apply("ProcessRecords",
                ParDo.of(new ParserHandler())
                        .withOutputTags( ParserHandler.glucoseTag,
                                TupleTagList.of(
                                        Lists.newArrayList(ParserHandler.errorRow,
                                                ParserHandler.insertionTag,
                                                ParserHandler.meterTag,
                                                ParserHandler.alertSettingsTag,
                                                ParserHandler.userEventTag,
                                                ParserHandler.deviceSettingsTag,
                                                ParserHandler.senderInformationTag))));


        output.get(ParserHandler.glucoseTag).apply(ParDo.named("GlucoseTableConverter")
                .of(new DoFn<IdentifiedGlucoseRecord, TableRow>(){
                    @Override
                    public void processElement(ProcessContext c) throws Exception {
                        log.info("Processing Glucose Records");
                        c.output(IdentifiedGlucoseConverter.format(c.element()));
                    }
                })
        );

        output.get(ParserHandler.insertionTag).apply(ParDo
                .named("InsertionTableConverter")
                .of(new DoFn<AbstractPublicDataPostData, TableRow>(){
            @Override
            public void processElement(ProcessContext c) throws Exception {
[...]

PR #315, opened 2016-10-27
add SideInOutExample and test #313

nevillelyh:


nevillelyh:
@sfines @jonopwell is this clear enough?

ravwojdyla:
👍 if there is no example of this already (?)

PR #318, opened 2016-10-28
support custom IO in JobTest #317

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1018,6 +1018,20 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     }
   }
 
+  /**
+   * Save this SCollection with a custom output transform.
+   * @group output
+   */
+  def saveAsCustomOutput(name: String, transform: PTransform[PCollection[T], PDone])
```
ravwojdyla: Reuse name from `PTransform` instead of another argument?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -559,6 +559,19 @@ class ScioContext private[scio] (val options: PipelineOptions,
     }
   }
 
+  /**
+   * Get an SCollection with a custom input transform.
+   * @group input
+   */
+  def customInput[T : ClassTag](name: String, transform: PTransform[PInput, PCollection[T]])
```
ravwojdyla: Reuse name from `PTransform` instead of another argument?

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -296,6 +310,26 @@ class JobTestTest extends PipelineSpec {
     an [AssertionError] should be thrownBy { testDistCacheJob("a1", "a2", "b1", "b2", "c3", "d4") }
   }
 
+  def testCustomIOJob(xs: Int*): Unit = {
+    JobTest[CustomIOJob.type]
+      .args("--input=in.txt", "--output=out.txt")
+      .input(CustomIO("TextIn"), Seq(1, 2, 3))
```
ravwojdyla: You probably could move `inputTransform` to field inside `CustomIOJob` (with `named("TextIn ")`) and then here just get name without - thus reduce the duplicate of names etc. The same for `outputTransform`.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
nevillelyh: `PTransform#getName` may return `null` though since it's not strictly enforced.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: Same reason as above.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
ravwojdyla: Let's just enforce it then.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
nevillelyh: I know. It's mostly for aesthetic reason since, i.e. avoid multi-line function argument.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
nevillelyh: Sure.

nevillelyh:
@ravwojdyla look again?

ravwojdyla:
👍

Issue #323, opened 2016-11-02
Drop 2.10.x and support 2.12.0?


nevillelyh:
https://github.com/scala/scala/releases 2.12.0 is out.
We'll have to wait for several dependencies to cross build first though.

Here's a checklist of items:
- [x] remove 2.10 cross build
- [x] fix up deprecated 2.10 macro uses
- [x] fix up `implict class X extends AnyVal` uses
- [x] remove 2.10 or 2.11 specific code (`src/{main,test}/scala-2.1x`)
- [ ] add 2.12 cross build
- [x] target JDK 8 for 2.12 (or deprecate JDK 7 target?)
- [ ] add 2.12 specific code if necessary

Java 8
- [x] Replace joda-time with java.time
- [x] Stream API, lambdas, Java 8 `Optional`

ravwojdyla:
What would be a reason for use to keep `2.10`, is there anyone that requires scala `2.10`? I remember we talked about zeppelin in context of `2.10` - but that one supports `2.11` already, so we are good. Most (if not all) people at Spotify (and use scio) are on `2.11` already.

Also if we move to 2.11/2.12 we could simplify some macros etc and maybe refactor some code (like some repl parts) - not all at once obviously. Also if we decide to go that way - I would suggest for the purpose of spreading the knowledge - split the tasks, and pick the things that we are least familiar with?

nevillelyh:
@ElPicador @mchalek @sinaso @joshreuben456 @torbjornvatn just checking if you're still using Scala 2.10 since we're thinking about deprecating it soon and move on to 2.12.

ElPicador:
We are using 2.11 and thinking about moving to 2.12 asap

mchalek:
We're also using 2.11

torbjornvatn:
Same here!

andrewsmartin:
WIP java 8 branch: https://github.com/spotify/scio/tree/andrew/java8

andrewsmartin:
WIP scala 2.12 branch: https://github.com/spotify/scio/tree/andrew/build-scala-2-12

nevillelyh:
We discussed internally and will do the 2.10 drop and 2.12 support when switching to Beam.
Closing this and tracking it in #279.

PR #324, opened 2016-11-03
Drop scala 2.10 #323

andrewsmartin:


nevillelyh:
The convention is to support the last 2 minor releases, e.g. 2.11 and 2.12. Since we can't do 2.12 yet, Let's keep this in a feature branch for now?

We should also fix a lot of deprecated macro uses (`grep "scala 2.11"`) and make sure it works with 2.12 before merging to master.

Also we should probably deprecate Java 7 once 2.12 is there and bump version to 0.3.0 or something.

andrewsmartin:
Saw you already started with the `scala212` branch, will delete this one.

PR #327, opened 2016-11-04
Build with java 8, move from joda-time to java.time

andrewsmartin:
@nevillelyh @ravwojdyla please review. Mostly moved from joda-time to java-time, although Dataflow SDK still depends on joda so it isn't completely gone. Also the `TIMESTAMP`, `DATETIME` etc fields available in BigQuery are semantically closer to joda so a bit of additional parsing logic was needed to handle some cases.


ravwojdyla:
lgtm.

nevillelyh:
Can you rebase and make two commits, one for SBT/Travis Java 8 and one for joda time? Might make branch management easier.

nevillelyh:
Actually we should still target JDK 1.7 for Scala 2.11 just in case. So maybe get rid of the last commit?

PR #329, opened 2016-11-07
Allow access to unformatted query/table

ravwojdyla:
This is RFC. Let me know what you think and if there is better way to implement it - i went for the simplest possible way here.

This allows to for example:

```
@BigQueryType.fromTable("project:dataset.table_%s", "20160530") class TT

sc.typedBigQuery[TT](TT.unformattedTable.format(args("tbldate")))
```

thus reduce the smell of duplicating the table/query string or getting formatted table/query (`BQT.{query, table}`) and reversing formatting by hand.

- [x] get feedback
- [x] decide on implementation
- [x] add tests

### DIFF
--- a/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/BigQueryTypeIT.scala
+++ b/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/BigQueryTypeIT.scala
@@ -94,11 +94,10 @@ class BigQueryTypeIT extends FlatSpec with Matchers {
 
   "fromTable" should "work" in {
     val bqt = BigQueryType[FromTableT]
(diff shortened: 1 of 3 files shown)


andrewsmartin:
This is already built-in for queries, right?

```
@BigQueryType.fromQuery("SELECT * FROM [project:dataset.table_%s]", "20160530")
class Row

sc.typedBigQuery[Row](Row.query.format("20160630") // Use more recent query at runtime
```

But +1 on doing a similar thing for tables.

nevillelyh:
Do you need to add a field though? For queries it's the same `query` whether formatted our not.

andrewsmartin:
👍 , keeping the string seems more consistent than keeping the `TableReference` and it can be parsed into one anyway.

ravwojdyla:
@andrewsmartin 👍 updated the wiki.

PR #344, opened 2016-11-23
Expose ScioResult metrics

rouzwawi:
This is useful when publishing the job metrics to other services from a Scala context, instead of having to go through the json file.

Not sure what the preferred package layout for the types are, feedback welcome.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
@@ -29,6 +29,7 @@ import com.google.cloud.dataflow.sdk.util._
 import com.google.cloud.dataflow.sdk.{Pipeline, PipelineResult}
 import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.Accumulator
+import com.spotify.scio.metrics._
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/metrics/package.scala
```diff
@@ -0,0 +1,20 @@
+package com.spotify.scio
```
ravwojdyla: No header //scalastyle will fail here.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -81,56 +81,56 @@ class ScioResult private[scio] (val internal: PipelineResult,
         out.close()
       }
     }
+  }
 
-    def getMetrics: MetricSchema.Metrics = {
-      import MetricSchema._
-
-      val totalValues = accumulators
+  // scalastyle:off method.length
```
ravwojdyla: Is the methods still too long?

rouzwawi:
@rav fixed comments

andrewsmartin:
👍 LGTM

rouzwawi:
Missed to align some tests, on it now.

Issue #340, opened 2016-11-17
Support for description fields in .saveAsBigQuery function


ankscircle:
When saving Dataflow output as a BigQuery table, we right now use case class. It would be great to have support to add description fields for columns in the BigQuery table. So something like .saveAsBigQuery(<table-ref>, <schema>) where schema has the ability to setDescriptions on the table fields.

ravwojdyla:
This is already kinda supported, but not very dev friendly:

```
val writeSchema = <BigQueryType>.schema
writeSchema.getFields().asScala.find(_.getName == "fieldName").map(_.setDescription("my new desc"))

sc.<logic>.saveAsBigQuery(<tbl-ref>, writeSchema)
```

Could we place the description inside the macro - which would make it static? In your case @ankscircle how do you obtain description for the fields - would they be static like:

```
@BigQueryType.toTable case class Foo(foo: Int @description("count of stream"), bar: String @description("user name"))
```

or are the descriptions dynamic/parametrized?

nevillelyh:
Can you use annotations like that?
I don't see a use case of description being parameterized, because that would imply type also being parameterized which is impossible?

ravwojdyla:
@ankscircle ping

ankscircle:
@nevillelyh @ravwojdyla That makes sense. The description need not be parametrize. They won't change. So something like `@description` looks great! And would also help maintain the schema and is more readable! 👍

nevillelyh:
Fixed in #349

PR #349, opened 2016-12-05
add @description for type-safe BigQuery #340

nevillelyh:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
```diff
@@ -81,9 +83,36 @@ private[types] object SchemaProvider {
...
+      _.annotations
+        .find(_.tree.tpe.toString == tpe)
+        .map { a =>
+          val q"new $t($v)" = a.tree
+          val Literal(Constant(s)) = v
+          s.toString
+        }
+    }
+    */
+    t.declaration(nme.CONSTRUCTOR).asMethod.paramss.head.map {
```
ravwojdyla: This `head` here is safe right?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
```diff
@@ -81,9 +83,36 @@ private[types] object SchemaProvider {
...
+      _.annotations
+        .find(_.tpe.toString == tpe)
+        .map { a =>
```
ravwojdyla: probably could be a `collect` here?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
nevillelyh: yes unless if there's no public ctors.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/SchemaProviderTest.scala
```diff
@@ -77,4 +77,32 @@ class SchemaProviderTest extends FlatSpec with Matchers {
     SchemaProvider.schemaOf[RepeatedNested] should equal (parseSchema(recordFields("REPEATED")))
   }
 
+  case class User(@description("user name") name: String, @description("user age") age: Int)
+  case class Account(@description("account user") user: User,
+                     @description("in USD") balance: Double)
```
ravwojdyla: I guess it's fine - but another option to test could be mixed - class with and without description fields.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
nevillelyh: not sure how to do that since I can't `case Literal(Constant(s))` on `a` directly.

ravwojdyla: yea - you could:

```
.map(_.scalaArgs.head) // or head is always here right? it should be safe/fine to use headOption here too
.collect{ case Literal(Constant(s)) => s.toString }
```

nevillelyh: That'll make a copy of the list though, not a big deal but no reason to fuss around here?

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/SchemaProviderTest.scala
nevillelyh: Yeah but no-desc cases are covered above. Don't wanna over do it and handcraft exhaustive test cases.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
ravwojdyla: I leave it up to you.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/SchemaProviderTest.scala
ravwojdyla: It could by as simple as remove one of the descriptions here to cover one more test case :)

ravwojdyla:
👍

Issue #343, opened 2016-11-22
Enable custom KryoSerializer and/or Dataflow Coder


ElPicador:
One of our job was taking a huge number of workers, so we tried to understand why.

The job as the following:

```
pub/sub[String] -> 
  parse[MyObject] -> 
  keyBy[(MyCaseClass, Int)] -> 
  Aggregator[(MyCaseClass, Long)] -> 
  pub/sub[(MyCaseClass, Long)]
```

We had ~30k messages/s from pubsub, and the job took ~30 workers to run.
By using a custom Dataflow coder for `MyCaseClass` (a simple mapping of fields) the job now takes ~3 workers to run.

Furthermore, we use the algebird [SpaceSaver](https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/SpaceSaver.scala). 
We tried some micro-benchmarking on coding -> decoding with Kryo for 100k objects:
- Build-in coder/decoder: 2600ms
- Custom : 150ms

It seems that Custom coders is a huge improvement to performances. 

Perhaps having a `log.warning` when scio see a default coder for any class?
Furthermore, it'll be nice to be able to add custom kryo coders easily. Today we had to copy/paste the `RichCoderRegistry` class to add some of ours. 
We also used the [ScalaKryoInstantiator](https://github.com/twitter/chill/blob/develop/chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala) from chill.

nevillelyh:
I agree we should look into enable user defined Kryo `Serializer[T]` or Dataflow `Coder[T]`.
Not sure about warning though since that might be too noisy.

nevillelyh:
@ElPicador I'm surprised that Kyro is so much slower than custom coder for a simple case class. Can you put the case class, custom coder & benchmark code in a gist? I'd prefer to not encourage user defined coders since they're error prone, but we should look into the performance problem.

ElPicador:
The mapping of the case class was made with a custom Dataflow Coder, not a Kryo one.

We only use a specific Kryo coder for SpaceSaver, because it contains scala Maps: https://gist.github.com/ElPicador/74a274bba171eb69b7717a515236ad75.

The benchmark was made on my compute using scalameter: https://gist.github.com/ElPicador/d8ef6195d279ea042bfbb064ba1d4b4e

ElPicador:
> Not sure about warning though since that might be too noisy.

[...]

PR #351, opened 2016-12-07
support custom KryoRegistrar #343

nevillelyh:
WIP DO NOT MERGE

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
@@ -23,14 +23,53 @@ import com.google.cloud.dataflow.sdk.coders.Coder.Context
 import com.google.cloud.dataflow.sdk.coders._
 import com.google.cloud.dataflow.sdk.util.VarInt
 import com.google.common.io.ByteStreams
+import com.google.common.reflect.ClassPath
 import com.google.protobuf.Message
 import com.twitter.chill._
 import com.twitter.chill.algebird.AlgebirdRegistrar
 import com.twitter.chill.protobuf.ProtobufSerializer
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecordBase
+import org.slf4j.LoggerFactory
 
+import scala.collection.JavaConverters._
 import scala.collection.convert.Wrappers.JIterableWrapper
+
+@KryoRegistrar
+private class AlgebirdKryoRegistrar extends IKryoRegistrar {
+  override def apply(k: Kryo): Unit = new AlgebirdRegistrar()(k)
+}
+
+private object KryoRegistrarLoader {
+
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -23,15 +23,54 @@ import com.google.cloud.dataflow.sdk.coders.Coder.Context
...
+  def load(k: Kryo): Unit = {
+    logger.info("Loading KryoRegistrars: " + registrars.mkString(", "))
+    registrars.foreach(_(k))
+  }
+
+  private val registrars: Seq[IKryoRegistrar] = {
+    logger.info("Initializing KryoRegistrars")
+    val classLoader = Thread.currentThread().getContextClassLoader
+    ClassPath.from(classLoader).getAllClasses.asScala.toSeq
+      .filter(_.getName.endsWith("KryoRegistrar"))
```
nevillelyh: The next `flatMap` can be very slow without this fillter, probably due to the cost of `clsInfo.load`. Thoughts?

nevillelyh:
@ElPicador what do you think about this approach with `@KryoRegistrar` annotation? Not super bullet proof but least intrusive and easy to use.

ElPicador:
I'm not a big fan of annotation (it breaks at runtime, not compile time).
But at least it'll work

nevillelyh:
Well there really isn't a way to make it compile time unless you fork the code and monkey patch :)
Even that doesn't guarantee the registrar/serializers will do it right.
This is the least intrusive I can come up with.

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
andrewsmartin: How slow is it? What do you think of having all Kryo registrars declared in a package like `com.spotify.scio.kryo` and do `ClassPath.from(classLoader).getTopLevelClasses("com.spotify.scio.kryo")`? Either approach is a bit of a gotcha for users (even if documented), but with the package thing we wouldn't need the filter and it might be slightly less awkward (but maybe also more annoying?). And in either case maybe we should use an annotation processor to fail compilation (i.e. when class name does not end with KryoRegistrar) to eliminate the gotcha. Thoughts? @ravwojdyla

nevillelyh: ~30s, maybe not a big deal for distributed pipelines but can be annoying for local runner. Good thing is it's done once per JVM. Don't like restricting package since user will likely use their own.

I'll look into name validation.

nevillelyh:
https://github.com/spotify/scio/commit/c66f717539f0e034deb4425e6d8a5f85f297b01f uses a macro-based approach and does validation compile time. Thoughts?

andrewsmartin:
👍

PR #352, opened 2016-12-08
support custom KryoRegistrar #343

nevillelyh:
New macro-based approach. Slightly safer but this would require the compiler plugin dependency on the user side.

@ElPicador thoughts?

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoRegistrar.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoRegistrar.scala
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/KryoRegistrar.scala
```diff
@@ -0,0 +1,56 @@
...
+          c.abort(c.enclosingPosition, s"Registrar class name must end with KryoRegistrar")
+        }
+        c.Expr[Any](q"class $name extends ..$parents with $traitT { ..$body }")
+      case t => c.abort(c.enclosingPosition, s"Invalid annotation $t")
+    }
+  }
+}
+
+/** Trait to be added to Kryo registrar class annotated with `@KryoRegistrar`. */
+trait AnnotatedKryoRegistrar
```
nevillelyh: The only downside is that this trait is exposed.

ElPicador:
I ❤️ it

Issue #335, opened 2016-11-13
Warning/error on old scio version


ravwojdyla:
Warn on old versions of scio, to inform users of new versions and push updates. Preferable on compile time, but runtime is fine if compile is too complex.

nevillelyh:
Fixed in #353

PR #353, opened 2016-12-09
check and warn about Scio version #335

nevillelyh:
This is a run-time approach. Compile-time approach isn't hard but would require some form of macro that I still wanna avoid for now.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio
+
+import com.google.api.client.http.javanet.NetHttpTransport
+import com.google.api.client.http.{GenericUrl, HttpRequest, HttpRequestInitializer}
+import com.google.api.client.json.JsonObjectParser
+import com.google.api.client.json.jackson2.JacksonFactory
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable
+import scala.util.Try
+
+private object VersionUtil {
+
+  case class SemVer(major: Int, minor: Int, rev: Int, snapshot: String) extends Ordered[SemVer] {
+    def compare(that: SemVer): Int = {
+      implicit val revStringOrder = Ordering[String].reverse
+      implicitly[Ordering[(Int, Int, Int, String)]]
+        .compare(SemVer.unapply(this).get, SemVer.unapply(that).get)
+    }
+  }
+
+  private val TIMEOUT = 3000
+  private val url = "https://api.github.com/repos/spotify/scio/releases"
+  private val pattern = """v?(\d+)\.(\d+).(\d+)(-SNAPSHOT)?""".r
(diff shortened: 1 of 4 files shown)


nevillelyh:
@ravwojdyla thoughts?

scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.google.api.client.json.jackson2.JacksonFactory
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable
+import scala.util.Try
+
+private object VersionUtil {
+
+  case class SemVer(major: Int, minor: Int, rev: Int, snapshot: Int) {
+    def toInt: Int = major * 1000000 + minor * 10000 + rev * 100 + snapshot
```
ravwojdyla: I understand how practically this is gonna work - but in theory this breaks. Counterargument:

```
my-version: 0.1.101 (0 + 10000 + 10100 = 20100)
release: 0.2.0 (0 + 20000 + 0 = 20000)
```

I think you could use Algebird's `Min`:

```
import com.twitter.algebird._
import com.twitter.algebird.Operators._
Min((0,2,0,"")) + Min((0,1,101,"")) -> Min((0,1,101,))
Min((0,2,0,"")) + Min((0,2,0,"SNAPSHOT")) -> Min((0,2,0,))
```

scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  private def parseVersion(version: String): SemVer = {
+    val m = pattern.findFirstMatchIn(version).get
+    // higher value for no "-SNAPSHOT"
+    val snapshot = if (m.group(4) == "-SNAPSHOT") 0 else 1
+    SemVer(m.group(1).toInt, m.group(2).toInt, m.group(3).toInt, snapshot)
+  }
+
+  def checkVersion(current: String, latest: Option[String]): Seq[String] = {
+    val b = mutable.Buffer.empty[String]
```
ravwojdyla: You could probably just `++` two `Seq` - and skip the mutable buffer?

scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
```diff
@@ -0,0 +1,82 @@
...
+private object VersionUtil {
+
+  case class SemVer(major: Int, minor: Int, rev: Int, snapshot: Int) {
+    def toInt: Int = major * 1000000 + minor * 10000 + rev * 100 + snapshot
+  }
+
+  private val TIMEOUT = 3000
+  private val url = "https://api.github.com/repos/spotify/scio/releases"
+  private val pattern = """v?(\d+)\.(\d+).(\d+)(-SNAPSHOT)?""".r
+  private val logger = LoggerFactory.getLogger("VersionUtil")
```
nevillelyh: Fixed

scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
nevillelyh: That'd require a tuple, not case class and Algebird dependency. Also we'd never hit 100 in any fields. Would prefer to keep it simple here.

scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
nevillelyh: Yes but that'd require 2 `if () Seq("...") else Nil` blocks which IMO is slightly more cryptic.

scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
ravwojdyla: case class + no algebird:
```
case class Ver(major: Int, minor: Int, rev: Int, classifier: String) extends Ordered[Ver] {
  def compare(that: Ver): Int = {
    implicit val revStringOrder = Ordering[String].reverse
    implicitly[Ordering[(Int, Int, Int, String)]]
      .compare(Ver.unapply(this).get, Ver.unapply(that).get)
  }
}
```

nevillelyh: You win :)

PR #354, opened 2016-12-09
Idea 1 for named transforms

andrewsmartin:
WIP please don't merge. @ElPicador @nevillelyh Here's one idea for implementing named transforms. What do you think? If we're all on board with this we can try to come up with a sexier DSL.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -138,7 +138,7 @@ object ScioContext {
  */
 // scalastyle:off number.of.methods
 class ScioContext private[scio] (val options: PipelineOptions,
-                                 private var artifacts: List[String]) {
+                                 private[scio] var artifacts: List[String]) {
 
   private implicit val context: ScioContext = this
 
--- a/scio-core/src/main/scala/com/spotify/scio/named/NamedScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/named/NamedScioContext.scala
@@ -1,0 +1,32 @@
+package com.spotify.scio.named
+
+import com.google.cloud.dataflow.sdk.values.PCollection
+import com.spotify.scio.values.SCollection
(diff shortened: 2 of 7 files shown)


andrewsmartin:
@ravwojdyla also

ElPicador:
My thoughts:

- I'm not a big fan of weird method names (aka not words)
- I like the fact that it doesn't breaks the "flow" of the code

ElPicador:
But whatever the final decision, thanks for implementing it 👍

andrewsmartin:
@ElPicador So if we were to instead just make it simple like

`.map(...).withName("name")`

would you prefer this over the other solution? We can also just have the regular `SCollection` be lazy by default so it can be built in to the normal API.

ElPicador:
@andrewsmartin I think it's more readable for non scala experts

andrewsmartin:
@ElPicador cool, I can do that then. One last thing, I could also do something similar in the other PR, i.e. `.withName("name").map(...)`, and in either case the `.withName` will always be optional. From a readability standpoint, which would you rather have?

I also did play with an implicit `apply(name: String)` in this PR, so it would look like this: `.filter(...)("name")`, and I think this would have honestly been the best approach, but unfortunately it doesn't work for any transform like `map` which has an implicit class tag parameter. To the best of my knowledge there isn't any way around that but I think would be the best approach if it is somehow possible. Otherwise I will stick with this.

ElPicador:
I would prefer `.map(...).withName("name")` but I don't have any good reason 😄

andrewsmartin:
Cool, I think that's more readable as well.

nevillelyh:
I think we've narrowed it down to `.withName("name").flatMap(fn)` vs `.flatMap(fn).withName("name")` right? Between these 2 I prefer the former because it's clearer (name the following transform instead of default behavior) and probably easier to implement.

andrewsmartin:
Let's go with that then for now - if it turns out later there's a good reason for switching and using lazily-applied transforms instead, we can always revisit. For now #357 will work.

PR #355, opened 2016-12-09
Idea 2 for named transforms

andrewsmartin:
WIP do not merge. @ElPicador @nevillelyh here is idea number 2 for named transforms. Basically the opposite of the other method, and is a lot simpler. Thoughts?


ElPicador:
My thoughts:

- I like the fact that it's clear that we group each step with a name
- I don't like the fact that it breaks the "flow" of the code, so more `a().b().c()`

andrewsmartin:
Closing this, opening a new one ready for review.

PR #357, opened 2016-12-12
Support for named transforms. 

andrewsmartin:
After some discussion in #354 we decided for now this would be the best approach. This is ready for review, but there are a couple other things we will need to address before everything is fully supported

1.  Scio defines several composite transforms (for example in DoubleSCollectionFunctions) that use the current naming strategy (`CallSites.getCurrent`), so even if the user gives a name to the outer transform, the inner transform is still named non-deterministically and would be a problem for updating. We will need to go through the Scio API and identify all the places where we internally use composite transforms, and give them all an appropriate name similar to what the Dataflow SDK does for its composite transforms. 
2. Documentation and more rigorous testing.

I propose we get this out which will solve the problem for the majority of transforms, and open another issue to address the remaining work in the above items.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/accumulators/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/accumulators/package.scala
@@ -60,7 +60,7 @@ package object accumulators {
 
     /** Count elements with an automatically created accumulator. */
     def accumulateCount: SCollection[T] =
-      self.accumulateCount(self.context.sumAccumulator[Long](CallSites.getCurrentName))
+      self.accumulateCount(self.context.sumAccumulator[Long](self.tfName))
 
     /** Count positive and negative results in a filter with the given accumulators. */
     def accumulateCountFilter(accPos: Accumulator[Long],
@@ -79,8 +79,8 @@ package object accumulators {
 
     /** Count positive and negative results in a filter with automatically created accumulators. */
     def accumulateCountFilter(f: T => Boolean): SCollection[T] = {
-      val accPos = self.context.sumAccumulator[Long]("Positive#" + CallSites.getCurrentName)
-      val accNeg = self.context.sumAccumulator[Long]("Negative#" + CallSites.getCurrentName)
+      val accPos = self.context.sumAccumulator[Long]("Positive#" + self.tfName)
+      val accNeg = self.context.sumAccumulator[Long]("Negative#" + self.tfName)
       self.accumulateCountFilter(accPos, accNeg)(f)
     }
 
--- a/scio-core/src/main/scala/com/spotify/scio/accumulators/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/accumulators/package.scala
@@ -60,7 +60,7 @@ package object accumulators {
 
     /** Count elements with an automatically created accumulator. */
     def accumulateCount: SCollection[T] =
-      self.accumulateCount(self.context.sumAccumulator[Long](self.tfName))
+      self.accumulateCount(self.context.sumAccumulator[Long](CallSites.getCurrentName))
 
     /** Count positive and negative results in a filter with the given accumulators. */
     def accumulateCountFilter(accPos: Accumulator[Long],
@@ -79,8 +79,8 @@ package object accumulators {
 
     /** Count positive and negative results in a filter with automatically created accumulators. */
     def accumulateCountFilter(f: T => Boolean): SCollection[T] = {
-      val accPos = self.context.sumAccumulator[Long]("Positive#" + self.tfName)
-      val accNeg = self.context.sumAccumulator[Long]("Negative#" + self.tfName)
+      val accPos = self.context.sumAccumulator[Long]("Positive#" + CallSites.getCurrentName)
+      val accNeg = self.context.sumAccumulator[Long]("Negative#" + CallSites.getCurrentName)
       self.accumulateCountFilter(accPos, accNeg)(f)
     }
 
--- a/scio-core/src/main/scala/com/spotify/scio/values/DoubleSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/DoubleSCollectionFunctions.scala
@@ -39,22 +39,26 @@ class DoubleSCollectionFunctions(self: SCollection[Double]) {
   // def sum: SCollection[Double] = this.stats().map(_.sum)
 
   /** Compute the standard deviation of this SCollection's elements. */
-  def stdev: SCollection[Double] = self.transform(_.stats.map(_.stdev))
+  def stdev: SCollection[Double] = self.transform(_.stats.withName("StdDev").map(_.stdev))
 
   /** Compute the variance of this SCollection's elements. */
-  def variance: SCollection[Double] = self.transform(_.stats.map(_.variance))
+  def variance: SCollection[Double] = self.transform(_.stats.withName("Variance").map(_.variance))
 
   /**
    * Compute the sample standard deviation of this SCollection's elements (which corrects for bias
    * in estimating the standard deviation by dividing by N-1 instead of N).
    */
-  def sampleStdev: SCollection[Double] = self.transform(_.stats.map(_.sampleStdev))
+  def sampleStdev: SCollection[Double] = self.transform {
+    _.stats.withName("SampleStdDev").map(_.sampleStdev)
+  }
 
   /**
    * Compute the sample variance of this SCollection's elements (which corrects for bias in
    * estimating the variance by dividing by N-1 instead of N).
    */
-  def sampleVariance: SCollection[Double] = self.transform(_.stats.map(_.sampleVariance))
+  def sampleVariance: SCollection[Double] = self.transform {
+    _.stats.withName("SampleVariance").map(_.sampleVariance)
+  }
 
   // Ported from org.apache.spark.rdd.DoubleRDDFunctions
 
--- a/scio-core/src/main/scala/com/spotify/scio/values/DoubleSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/DoubleSCollectionFunctions.scala
@@ -39,17 +39,17 @@ class DoubleSCollectionFunctions(self: SCollection[Double]) {
   // def sum: SCollection[Double] = this.stats().map(_.sum)
 
   /** Compute the standard deviation of this SCollection's elements. */
-  def stdev: SCollection[Double] = self.transform(_.stats.withName("StdDev").map(_.stdev))
+  def stdev: SCollection[Double] = self.transform(_.stats.map(_.stdev))
 
   /** Compute the variance of this SCollection's elements. */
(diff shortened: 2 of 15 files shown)


nevillelyh:
I'm OK with the user facing API, but feel like that we should re-work the internals a bit, especially with things carried over in d0e159038672c554e601964feb7d593d775181c5.

I'd remove the optional argument `CallSites.getCurrent`, and replace it with a trait `NameProvider` in `PCollectionWrapper` which either returns current stack or static string depending on the context. E.g.

```scala
trait NameProvider {
  def apply(): String
}
object CallSitesNameProvider extends NameProvider {
  @override def apply(): String = CallSites.getCurrent
}
class StaticNameProvided(private val name: String) extends NameProvider {
  @override def apply(): String = name
}
```

This way we can undo some of the optional args (which are always tricky) in d0e159038672c554e601964feb7d593d775181c5

nevillelyh:
Basically I'm suggesting that we revert d0e159038672c554e601964feb7d593d775181c5 and use the provider approach which is cleaner and more extendable.

andrewsmartin:
+1, do as part of this PR?

scio-core/src/main/scala/com/spotify/scio/package.scala
```diff
@@ -78,6 +78,12 @@ package object scio {
       Await.result(self.flatMap(identity), atMost)
   }
 
+  /** Set the name of the next transform. */
+  implicit class NamedTfSCollection[T: ClassTag](val wrapped: SCollection[T]) {
+    def withName(tfName: String): SCollection[T] =
```
nevillelyh: This should fail if `wrapped` is already a `NamedTfSCollection`. Also I'd name this `NamedSCollection`.

nevillelyh:
Sure. You can do multi-commits PRs. It's clearer if we keep a revert commit (`git revert $hash`).

scio-core/src/main/scala/com/spotify/scio/package.scala
andrewsmartin: True. Easiest way is probably just to have `wrapped` be `SCollectionImpl` instead right?

andrewsmartin: Actually that won't work because user only sees `SCollection[T]`

nevillelyh: Probably true, you might not even need a `NamedSCollectionImpl`, just reflect on `self.nameProvider` should be sufficient.

nevillelyh: I mean you might not need a implicit class, have `def .withName` on `PCollectionWrapper` would be the best/cleanest, if we can get rid of `CallSites.getCurrent` everywhere else and replace them with `this.nameProvider.get()`.

andrewsmartin:
@nevillelyh @ravwojdyla This solution works for all `SCollection` variants, any thoughts?

andrewsmartin:
Should note that this still doesn't not prevent calling `.withName()` twice in a row, would like to get some opinions on the best way to approach. IMO I don't think it's something we need to care about - if a user does it twice before applying a transform, the least-astonishing thing would be to just overwrite the var again.

scio-core/src/main/scala/com/spotify/scio/values/TransformNameable.scala
```diff
@@ -0,0 +1,42 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.values
+
+import com.spotify.scio.util.CallSites
+
+trait TransformNameable[T <: TransformNameable[T]] {
+  this: T =>
```
nevillelyh: I think the convention is to keep `this: T =>` on the previous line.

scio-core/src/main/scala/com/spotify/scio/values/TransformNameable.scala
```diff
@@ -0,0 +1,42 @@
...
+  private var nameProvider: TransformNameProvider = CallSiteNameProvider
+
+  def tfName: String = nameProvider.name
+
+  def withName(name: String): T = {
+    nameProvider = new ConstNameProvider(name)
```
nevillelyh: add an `require()` here against `withName().withName()` ?

nevillelyh:
Also do a `git grep CallSites` in the code base?
There're also some join and multi-join cases.
`MultiJoin` is an object right now but shouldn't be too hard to add `withName` so we can have `MultiJoin.withName("x").join(a, b, c, d)`?

andrewsmartin:
@nevillelyh just pushed two more commits to address your comments.

ravwojdyla:
👍

scio-core/src/main/scala/com/spotify/scio/accumulators/package.scala
```diff
@@ -79,8 +79,8 @@ package object accumulators {
 
     /** Count positive and negative results in a filter with automatically created accumulators. */
     def accumulateCountFilter(f: T => Boolean): SCollection[T] = {
-      val accPos = self.context.sumAccumulator[Long]("Positive#" + CallSites.getCurrentName)
-      val accNeg = self.context.sumAccumulator[Long]("Negative#" + CallSites.getCurrentName)
+      val accPos = self.context.sumAccumulator[Long]("Positive#" + self.tfName)
```
nevillelyh: These 2 are not `PTransform` names but accumulator identifiers. We should keep the original so that they're distinguishable.

scio-core/src/main/scala/com/spotify/scio/accumulators/package.scala
```diff
@@ -60,7 +60,7 @@ package object accumulators {
 
     /** Count elements with an automatically created accumulator. */
     def accumulateCount: SCollection[T] =
-      self.accumulateCount(self.context.sumAccumulator[Long](CallSites.getCurrentName))
```
nevillelyh: This is an accumulator name. Keep the original.

scio-core/src/main/scala/com/spotify/scio/values/DoubleSCollectionFunctions.scala
```diff
@@ -39,22 +39,26 @@ class DoubleSCollectionFunctions(self: SCollection[Double]) {
   // def sum: SCollection[Double] = this.stats().map(_.sum)
 
   /** Compute the standard deviation of this SCollection's elements. */
-  def stdev: SCollection[Double] = self.transform(_.stats.map(_.stdev))
+  def stdev: SCollection[Double] = self.transform(_.stats.withName("StdDev").map(_.stdev))
```
nevillelyh: Let's leave transforms in this file alone for now? There're too many of them and it's impossible to catch all. Like there're a few in `histogram()` that we didn't cover?

Plus I doubt these are super useful in streaming. Better to catch them later as they pop up.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -60,8 +60,8 @@ object SCollection {
   def unionAll[T: ClassTag](scs: Iterable[SCollection[T]]): SCollection[T] = {
     val o = PCollectionList
       .of(scs.map(_.internal).asJava)
-      .apply(CallSites.getCurrent, Flatten.pCollections())
-    new SCollectionImpl(o, scs.head.context)
+      .apply("FlattenList", Flatten.pCollections())
```
nevillelyh: Use `UnionAll` to be more precise?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -163,7 +163,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   def union(that: SCollection[T]): SCollection[T] = {
     val o = PCollectionList
       .of(internal).and(that.internal)
-      .apply(CallSites.getCurrent, Flatten.pCollections())
+      .apply("FlattenList", Flatten.pCollections())
```
nevillelyh: Can use `this.tfName` here? More consistent with other transforms.

scio-core/src/main/scala/com/spotify/scio/values/DoubleSCollectionFunctions.scala
andrewsmartin: I did this up front so that we can have a test for naming transforms on `DoubleSCollectionFunctions`. Since they're composite transforms I wanted to make the internal name deterministic to make an assertion. https://github.com/spotify/scio/blob/playground/scio-test/src/test/scala/com/spotify/scio/values/NamedTransformTest.scala#L45

Maybe instead I can remove these, and just assert the transform name starts with the given name and ignore the inner transform name?

nevillelyh: Yeah maybe we can remove tests for most types derived `SCollection` and assert the part before `/`. Leave the inner ones alone for now and fix when it becomes a problem.

scio-core/src/main/scala/com/spotify/scio/values/TransformNameable.scala
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package com.spotify.scio.values
+
+import com.spotify.scio.util.CallSites
+
+trait TransformNameable {
+  private var nameProvider: TransformNameProvider = CallSiteNameProvider
+
+  def tfName: String = {
```
nevillelyh: make this `private[values]`?

scio-test/src/test/scala/com/spotify/scio/values/NamedTransformTest.scala
```diff
@@ -0,0 +1,150 @@
...
+
+  "Duplicate transform name" should "have number to make unique" in {
+    runWithContext { sc =>
+      val p1 = sc.parallelize(1 to 5)
+        .withName("MyTransform").map(_ * 2)
+      val p2 = p1
+        .withName("MyTransform").map(_ * 3)
+      val p3 = p1
+        .withName("MyTransform").map(_ * 4)
+      assertTransformName(p1, "MyTransform")
```
nevillelyh: Can't find the incremental logic in this PR, where are they from? Is it better to fail duplicate names instead?

andrewsmartin: This logic is built-in to dataflow, I included a test for it because an earlier version of this PR actually broke in this scenario.

scio-core/src/main/scala/com/spotify/scio/values/TransformNameable.scala
andrewsmartin: `ScioContext` also needs this, will make it `private[scio]`

scio-test/src/test/scala/com/spotify/scio/values/NamedTransformTest.scala
nevillelyh: :+1:

Issue #360, opened 2016-12-13
Expose an SCollection with PairSCollection key methods


harveyxia:
For instance, be able to call `groupByKey` on an `SCollection[T]`. The use case is for when we want to group by key, but don't care about the value.

andrewsmartin:
One way to do this might be to have an implicit conversion so the user doesn't have to `.map((_, ()))`, like this:

```
implicit def makeDefaultPairSCollectionFunctions[K: ClassTag](s: SCollection[K])
  : PairSCollectionFunctions[K, Unit] = new PairSCollectionFunctions[K, Unit](s.map((_, ())))
```

I think this is easy enough and doesn't really add any API bloat, since it's just an implicit.

andrewsmartin:
@harveyxia The more I think about this / look at the API, the more I'm wondering about use cases. Most of the functions in `PairSCollectionFunctions` do actually use the values, except for things like counting. What do you want to do after `groupByKey` if you don't care about the values?

harveyxia:
Just discovered that I can use the `.distinct` function, which is essentially what I was trying to do. I think most use cases would be covered by the distinct function, so feel free to close this issue.

andrewsmartin:
Awesome, thanks!

PR #361, opened 2016-12-13
Add default implicit conversion to PairSCollectionFunctions

andrewsmartin:
@nevillelyh @ravwojdyla RFC. To address #360. This allows performing per-key operations on an `SCollection[T]` treating all the elements as keys.


andrewsmartin:
Not needed - see discussion in #360

PR #365, opened 2016-12-15
Make transform public to allow chaining user-defined transforms.

andrewsmartin:
This allows for better flow / readability when users define higher level transforms. For example:

```
def myTransformFn(in: SCollection[T]): SCollection[U] = ...

sc.textFile(...)
  .stuff
  .transform(myTransformFn)
  .etc
```

Instead of 

```
def myTransformFn(in: SCollection[T]): SCollection[U] = ...

val p = sc.textFile(...)
  .stuff

myTransformFn(p)
  .etc
```

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -124,7 +124,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     this.pApply(transform)
 
   /** Apply a transform. */
-  private[values] def transform[U: ClassTag](f: SCollection[T] => SCollection[U])
+  def transform[U: ClassTag](f: SCollection[T] => SCollection[U])
   : SCollection[U] = {
     val o = internal.apply(CallSites.getCurrent, new PTransform[PCollection[T], PCollection[U]]() {
       override def apply(input: PCollection[T]): PCollection[U] = {


ravwojdyla:
👍 is there testing covering this already?

andrewsmartin:
There is a test for `.transform` already in `SCollectionTest`. It was package-private before this change.

ravwojdyla:
👍

nevillelyh:
Is this needed for other subsequent PRs? I intentionally kept it private since it won't always work, especially when the composite transform pulls in non-serializable fields. Would prefer to not leak this as public API.

andrewsmartin:
It is not needed for anything other than method chaining which is to nice to have, so if this is problematic we can revert. That said, composite transforms are a feature in Dataflow (https://cloud.google.com/dataflow/model/composite-transforms), and the problem with serialization applies generally to any closure - even in basic transforms - so there is always gonna have to be some care taken by the user. Or is there another issue that's unique to composites?

nevillelyh:
It's tricker to ser/de Scio composites because they pull in `SCollection` which has many non-serializable fields. That said I'd prefer to undo this until we're more confident about `SCollection` and `ScioContext` etc. being fully serializable.

andrewsmartin:
That's a good point, I'll revert and revisit this later.

nevillelyh:
I force pushed the last 2 commits away so that we don't have to cherry-pick twice into `apache-beam` :)

andrewsmartin:
Thanks, forgot about that!

PR #366, opened 2016-12-15
Fix #364. Support missing BQ source tables

### LINKED ISSUE #364: NPE from extractTables on BQ query without source tables
To reproduce:

```
@BigQueryType.fromQuery("SELECT 5") class TT
sc.typedBigQuery[TT]()
```

or simply

```
BigQueryClient.defaultInstance.extractTables("SELECT 5")
```

ravwojdyla:


andrewsmartin:
👍 , would it also be worth adding a "test" to make sure a case class can still be generated from such queries? I know the test would basically just be "this should compile" (like https://github.com/spotify/scio/blob/master/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala#L308) but still worth having IMO.

andrewsmartin:
Actually not needed

ravwojdyla:
Add `private val DEFAULT_LOCATION = "US"` to not repeat the `"US"` string too over and over again.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -447,22 +450,22 @@ class BigQueryClient private (private val projectId: String,
   }
 
   /** Extract tables to be accessed by a query. */
-  def extractTables(sqlQuery: String): Set[TableReference] = {
+  def extractTables(sqlQuery: String): Option[Set[TableReference]] = {
```
nevillelyh: Instead of `Option[Set[]]`, is it cleaner to return empty set if `tryJob.get.getStatistics.getQuery.getReferencedTables == null`?

Issue #143, opened 2016-05-20
hadoopDistCache should bypass GCS in DirectRunner?


nevillelyh:
Right now it requires `stagingLocation` to be set and copies HDFS file to GCS before pulling it.
https://github.com/spotify/scio/blob/0c90f00a2507cf5c184f1bc6f1941a43836af3bf/scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala#L128

Maybe it should skip that and pull from HDFS directly if `stagingLocation` is not available, or if it's running in local mode?

andrewsmartin:
Closed via c08495b

PR #367, opened 2016-12-15
Hadoop dist cache should skip GCS upload with direct runner #143

andrewsmartin:


scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -158,16 +155,35 @@ package object hdfs {
       }
     }
 
-    private[scio] def hadoopDistCacheCopy(src: Path, target: URI, conf: Configuration): Unit = {
-      logger.debug(s"Will copy ${src.toUri}, to $target")
+    private def getTargetFromOptions() = {
+      if (self.options.getRunner() != classOf[DirectPipelineRunner]) {
+        val dfOptions = self.optionsAs[DataflowPipelineOptions]
+        require(self.optionsAs[DataflowPipelineOptions].getStagingLocation != null,
```
ravwojdyla: I know this code was here before - but is it needed? From doc:
```
   * <p>At least one of {@link #getTempLocation()} or {@link #getStagingLocation()} must be set. If
   * {@link #getTempLocation()} is not set, then the Dataflow pipeline defaults to using
   * {@link #getStagingLocation()}.
```

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -158,16 +155,35 @@ package object hdfs {
...
+          "Staging directory not set - use `--stagingLocation`!")
+        (dfOptions.getStagingLocation, gcsOutputStream _)
+      } else {
+        // should targetDir be specified in options?
```
ravwojdyla: What do you mean by this comment?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
andrewsmartin: Oh nice, so we can remove the require and just manually check `tempLocation` and `stagingLocation`?

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
andrewsmartin: I just mean that the "staging location" here is created with `Files.createTempDirectory` using its default directory - just wondering if we should have this configurable via Options and require it be set. But this is just for testing really so it's probably not an issue to just do this.

ravwojdyla: It seems like this code to lives in `com.google.cloud.dataflow.sdk.runners.DataflowPipelineRunner.fromOptions`, it handles making sure that staging or temp is set, and if needed sets the missing values. Will this code run after or before `fromOptions` ^ ?

ravwojdyla: Actually we shouldn't depend on when this code runs, probably not safe thing to do. So maybe it's just fine to assume that staging has to be set.

ravwojdyla:
👍

scio-hdfs/src/main/scala/com/spotify/scio/hdfs/package.scala
```diff
@@ -28,6 +29,7 @@ import com.google.cloud.dataflow.sdk.coders.AvroCoder
 import com.google.cloud.dataflow.sdk.io.hdfs._
 import com.google.cloud.dataflow.sdk.io.{Read, Write}
 import com.google.cloud.dataflow.sdk.options.DataflowPipelineOptions
+import com.google.cloud.dataflow.sdk.runners.DirectPipelineRunner
```
nevillelyh: I don't see this import being used?

andrewsmartin: Yeah, was checking directly against `DirectPipelineRunner` before switching to use `ScioUtil.isLocalRunner` and this was left in there. Will remove.

PR #369, opened 2016-12-16
Fix #368. Keep parents of Tap after Tap.map

### LINKED ISSUE #368: Tap loses information about parent Tap after Tap.map
After `Tap.map`, newly created `Tap` object loses information about the parent `Tap`. The uses case:

```
val tap = <scol>.saveAsTypedBigQuery[T]
```

`tap` is of type `Tap[T]`, there is no way the user can find out what was the table the data came from because `table` is a field of `BigQueryTap`. It is not possible to cast `tap` to `BigQueryTap`.

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -37,6 +36,9 @@ import scala.reflect.ClassTag
  */
 trait Tap[T] { self =>
 
+  /** Parent Taps this Tap's data came from after [[Tap.map]] */
+  val parent: Seq[Tap[_]] = Nil
```
nevillelyh: You don't really need a `Seq` right? Just `Tap` should be sufficient since its parent also has a `parent` field. Like a linked list.

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -45,6 +47,10 @@ trait Tap[T] { self =>
 
   /** Map items from T to U. */
   def map[U: ClassTag](f: T => U): Tap[U] = new Tap[U] {
+
+    /** Parent Taps this Tap's data came from after [[Tap.map]] */
+    override val parent: Seq[Tap[_]] = self.parent ++ Seq(self)
```
nevillelyh: It can just be `= self` here. Also unrelated but is `seq :+ self` more idiomatic?

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
ravwojdyla: Obviously - I'm stupid :/

ravwojdyla: Fixed

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
ravwojdyla: Fixed

PR #370, opened 2016-12-16
[RFC] Support for user-defined functions in BQ client

andrewsmartin:
@nevillelyh @ravwojdyla This is a WIP, let's not merge yet. Just want to collect some thoughts here. This commit adds (partial) support for UDFs by specifying `bigquery.resource_uri`, which should be a file on GCS containing javascript UDFs as described here: https://cloud.google.com/bigquery/user-defined-functions.

A few things:

1) This only works for legacy SQL. UDFs are supported in standard SQL but it's quite a bit different. I'll look into supporting standard SQL next.

2) This isn't the most flexible approach, because a user can only link to one resource URI per job, which means having to stick all functions together in the same file and is not a good solution for externally shared UDFs. Maybe we should allow linking of multiple resource URIs?

3) I think it would also be really nice to allow users to have UDFs defined in project repos (say in `src/main/js`) and have the `BigQueryClient` upload those to GCS during initialization. Thoughts?

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
@@ -145,6 +145,7 @@ class BigQueryClient private (private val projectId: String,
       // Create temporary table view and get schema
       logger.info(s"Creating temporary view ${BigQueryIO.toTableSpec(temp)}")
       val view = new ViewDefinition().setQuery(sqlQuery)
+      getUdfResource().map(view.setUserDefinedFunctionResources(_))
       val viewTable = new Table().setView(view).setTableReference(temp)
       val schema = bigquery
         .tables().insert(temp.getProjectId, temp.getDatasetId, viewTable)
@@ -409,6 +410,11 @@ class BigQueryClient private (private val projectId: String,
         .setPriority(PRIORITY)
         .setCreateDisposition("CREATE_IF_NEEDED")
         .setWriteDisposition("WRITE_EMPTY")
+
+      if (useLegacySql) {
+        getUdfResource().map(queryConfig.setUserDefinedFunctionResources(_))
+      }
+
       if (!dryRun) {
         queryConfig.setAllowLargeResults(true).setDestinationTable(destinationTable)
       }
@@ -446,6 +452,14 @@ class BigQueryClient private (private val projectId: String,
           case Failure(f) => throw f
         }
       case Failure(e) => throw e
+    }
+  }
+
+  private def getUdfResource(): Option[java.util.List[UserDefinedFunctionResource]] = {
+    BigQueryClient.udfResourceUri.map { uri =>
+      Seq(new UserDefinedFunctionResource()
+          .setResourceUri(uri)
+      ).asJava
     }
   }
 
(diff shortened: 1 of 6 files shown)


andrewsmartin:
Also the UDF used for the integration test right now is not in the project, just a file on GCS which is 👎 . Can fix with item 3

andrewsmartin:
Maybe we can add a parameter to the `fromQuery` annotation which either specifies a fully qualified URI (in which case external function expected to already exist on GCS), or simply the name of a JS file, in which case the file will be in some resource directory in the project and we can inline the JS instead of copying to GCS.

andrewsmartin:
Actually UDFs in standard SQL are built-in to the query itself, so it's pretty much a no-op. Will add another test.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -145,6 +145,7 @@ class BigQueryClient private (private val projectId: String,
       // Create temporary table view and get schema
       logger.info(s"Creating temporary view ${BigQueryIO.toTableSpec(temp)}")
       val view = new ViewDefinition().setQuery(sqlQuery)
+      getUdfResource().map(view.setUserDefinedFunctionResources(_))
```
nevillelyh: The `(_)` at the end should be optional?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -409,6 +410,11 @@ class BigQueryClient private (private val projectId: String,
         .setPriority(PRIORITY)
         .setCreateDisposition("CREATE_IF_NEEDED")
         .setWriteDisposition("WRITE_EMPTY")
+
+      if (useLegacySql) {
+        getUdfResource().map(queryConfig.setUserDefinedFunctionResources(_))
```
nevillelyh: `(_)` optional?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -449,6 +455,14 @@ class BigQueryClient private (private val projectId: String,
     }
   }
 
+  private def getUdfResource(): Option[java.util.List[UserDefinedFunctionResource]] = {
+    BigQueryClient.udfResourceUri.map { uri =>
```
nevillelyh: Not sure about the `sys.props` approach though since it can't handle different resources per query or multiple resources.

To make it work with `fromQuery` we might have to create a new annotation?
Also I wonder how many users actually need this vs doing UDF in native Scio code?
Anyway I'd defer this until we have a better idea.

nevillelyh: How do you pass in the resource URI for standard SQL then? As part of the query string?

nevillelyh: So for standard SQL it'll work out of the box without any Scio modification right? In that case I'd prefer not adding anything awkward to support legacy SQL then.

andrewsmartin:
Closing this - there are no changes needed to support Standard SQL, and we would rather not add anything specifically for legacy support.

PR #374, opened 2016-12-19
add MockBigQuery and integration tests

nevillelyh:
Proof of concept but seems to work well. Wanna get more feedbacks on API usability, edge cases etc.

### DIFF
--- a/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
+++ b/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigquery
+
+import com.spotify.scio.bigquery.types._
+import org.scalatest.Inspectors.forAll
+import org.scalatest._
+
+object BigQueryIT {
+
+  val tableRef = "bigquery-public-data:samples.shakespeare"
+  val legacyQuery =
+    "SELECT word, word_count FROM [bigquery-public-data:samples.shakespeare] LIMIT 10"
+  val sqlQuery =
+    "SELECT word, word_count FROM `bigquery-public-data.samples.shakespeare` LIMIT 10"
+
+  @BigQueryType.fromTable("bigquery-public-data:samples.shakespeare")
+  class Shakespeare
+
+  @BigQueryType.fromQuery(
+    "SELECT word, word_count FROM `bigquery-public-data.samples.shakespeare` LIMIT 10")
+  class WordCount
+
+}
+
+class BigQueryIT extends FlatSpec with Matchers {
+
+  import BigQueryIT._
+
+  // =======================================================================
+  // Integration test with mock data
+  // =======================================================================
+
+  "MockBigQuery" should "support mock data" in {
+    def shakespeare(w: String, wc: Long, c: String, cd: Long): TableRow =
+      TableRow("word" -> w, "word_count" -> wc, "corpus" -> c, "corpus_date" -> cd)
+
+    // BigQuery result TableRow treats integers as strings
+    def wordCount(w: String, wc: Long): TableRow =
+      TableRow("word" -> w, "word_count" -> wc.toString)
+
+    val inData = Seq(
+      shakespeare("i", 10, "kinglear", 1600),
+      shakespeare("thou", 20, "kinglear", 1600),
+      shakespeare("thy", 30, "kinglear", 1600))
+    val expected = Seq(wordCount("i", 10), wordCount("thou", 20), wordCount("thy", 30))
+
+    val mbq = MockBigQuery()
+    mbq.mockTable(tableRef).withData(inData)
+    mbq.queryResult(legacyQuery) should contain theSameElementsAs expected
+    mbq.queryResult(sqlQuery) should contain theSameElementsAs expected
+  }
+
+  // =======================================================================
+  // Integration test with type-safe mock data
+  // =======================================================================
+
+  it should "support typed BigQuery" in {
+    val inData = Seq(
+      Shakespeare("i", 10, "kinglear", 1600),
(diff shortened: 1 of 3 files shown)


scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * BigQuery integration test with mock data and equality check for query result.
+ */
+class BigQueryMockIT extends FlatSpec with Matchers with BigQueryITUtils {
+
+  import BigQueryIT._
+
+  private def shakespeare(w: String, wc: Long, c: String, cd: Long) =
+    TableRow("word" -> w, "word_count" -> wc, "corpus" -> c, "corpus_date" -> cd)
+
```
andrewsmartin: Would it be worth adding some helper so that users don't have to define their own functions / create table rows manually? Realize it's not much overhead really, but even just so that people don't have to work with TableRow directly. Maybe something like this? Would have to define this for up to 23 type params but could easily be generated similar to MultiJoin, and it's slightly "safer" in that users won't be able to pass in TableRows with different arity.

```
mockTable(tableRef).withData(
  Data("word", "word_count", "corpus", "corpus_date")
    .add("i", 10, "kinglear", 1600)
    .add("thou", 20, "kinglear", 1600)
  )
)
```

andrewsmartin: Then again all of this is just solved by using typed data anyway.

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -0,0 +1,147 @@
...
+  it should "fail duplicate mock data" in {
+    val it = new BigQueryITUtils {}
+    val mt = it.mockTable(tableRef)
+    mt.withData(Nil)
+    the[IllegalArgumentException] thrownBy {
+      mt.withData(Nil)
+    } should have message s"requirement failed: Table $tableRef already populated with mock data"
+  }
+
```
andrewsmartin: If we register a mock table but don't supply any data, what's the expected behavior? Treat it as empty table and still allow querying? Should we include a test for this?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
```diff
@@ -0,0 +1,149 @@
...
+    val t = bq.getTable(original)
+    val temp = bq.temporaryTable(t.getLocation)
+    mapping += (original -> temp)
+    new MockTable(bq, t.getSchema, original, temp)
+  }
+
+  /**
+   * Perform a live query against BigQuery service, substituting mocked tables with test data.
+   */
+  def query(sqlQuery: String, flattenResults: Boolean = false): Seq[TableRow] = {
```
andrewsmartin: Just a minor thing (and maybe it's just me) but would `runQuery()` or something be a better name? For me just `query()` makes it sound like a factory function to create some kind of query object instead of actually running one.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
```diff
@@ -0,0 +1,149 @@
...
+    val isLegacy = bq.isLegacySql(sqlQuery, flattenResults)
+    val mockQuery = mapping.foldLeft(sqlQuery){ case (q, (src, dst)) =>
+      q.replace(toTableSpec(src, isLegacy), toTableSpec(dst, isLegacy))
```
andrewsmartin: What do we want to happen if the query hits tables that haven't been mocked? Should we warn if this is the case?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
```diff
@@ -0,0 +1,149 @@
...
+    }
+    bq.getQueryRows(mockQuery, flattenResults).toList
+  }
+
+  /**
+   * Perform a live query against BigQuery service, substituting mocked tables with test data.
+   */
+  def typedQuery[T <: HasAnnotation : ClassTag : TypeTag]
```
andrewsmartin: Same as above, is `runTypedQuery` a better name?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
```diff
@@ -0,0 +1,149 @@
...
+   * Populate the table with mock data.
+   */
+  def withTypedData[T <: HasAnnotation : ClassTag : TypeTag](rows: Seq[T]): Unit = {
+    val bqt = BigQueryType[T]
+    withData(rows.map(bqt.toTableRow))
+  }
+
+  /**
+   * Populate the table with sample data from the original table.
+   * Note that rows are taken from the beginning of the table and not truly random.
```
andrewsmartin: I can imagine this being a problem with a lot of tables - could we use reservoir sampling? Or should we wait until someone asks for it :)

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
andrewsmartin: I guess it is also a verb

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
nevillelyh: The query will fail but we can't ensure that the user actually calls `with*`.

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
nevillelyh: Yeah not a problem really. More API = more confusion.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
nevillelyh: I'm leaning towards a noun since it's used in contexts like `query("...") should contain`.
I'll rename it to `queryResult`?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
nevillelyh: Maybe but we'll have to extract tables using regex which is hacky. I think it's ok to hit real tables (e.g. small lookup ones) but don't wanna go crazy with query parsing.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryITUtils.scala
nevillelyh: Let's wait, think of this as a MVP :)

PR #381, opened 2016-12-28
Add experimental typed BQ Taps

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -233,4 +233,30 @@ package object experimental {
 
   }
 
+
+  /** Enhanced version of [[Taps]] with type-safe features for BigQuery [[Taps]]. */
+  implicit class TypedBigQueryTaps(self: Taps) {
+
+    /** Get a `Future[Tap[T]]` for BigQuery table. */
+    def typedBigQueryTable[T <: HasAnnotation : TypeTag : ClassTag](tableSpec: TableReference)
```
ravwojdyla: Should get table from the provided BQT and rename the argument to new source.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -233,4 +233,30 @@ package object experimental {
...
+    : Future[Tap[T]] = {
+      val bqt = BigQueryType[T]
+      import scala.concurrent.ExecutionContext.Implicits.global
+      self.bigQueryTable(tableSpec).map(_.map(bqt.fromTableRow))
+    }
+
+    /** Get a `Future[Tap[T]]` for BigQuery table. */
+    def typedBigQueryTable[T <: HasAnnotation : TypeTag : ClassTag](tableSpec: String)
```
ravwojdyla: Same as above.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -233,4 +233,30 @@ package object experimental {
 
   }
 
+
```
ravwojdyla: Or maybe i could even consolidate both table/select to one actually.

scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -233,4 +233,36 @@ package object experimental {
...
```
nevillelyh: nitpick, extra blank line?

Issue #382, opened 2017-01-03
Support TensorFlow TFRecord format


nevillelyh:
They're just length prefixed binary strings file. Specs [here](https://www.tensorflow.org/api_docs/python/python_io/data_io__python_functions_).

We need custom [FileBasedSink](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java) and [FileBasedSource](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSource.java) plus Scala hooks.

nevillelyh:
I have a working branch here: https://github.com/spotify/scio/tree/neville/tf

It includes both source and sink and support uncompressed, zlib and gzip formats, although not fully tested.
Had to use wrappers from Apache commons since there're some bugs with the JDK zip library.
And it still doesn't read zlib files properly, something related to a JVM bug 😞

nevillelyh:
https://github.com/apache/beam/pull/1749
https://issues.apache.org/jira/plugins/servlet/mobile#issue/BEAM-1233

Someone is making one for Python.

PR #384, opened 2017-01-04
add TensorFlow TFRecord file support #382

nevillelyh:


nevillelyh:
@ravwojdyla thoughts?

scio-core/src/main/scala/com/spotify/scio/io/TFRecord.scala
```diff
@@ -0,0 +1,283 @@
...
+    case object NONE extends CompressionType
+    case object ZLIB extends CompressionType
+    case object GZIP extends CompressionType
+  }
+
+  /** Default read options. */
+  val readDefault = TFRecordOptions(CompressionType.AUTO)
+
+  /** Default write options. */
+  val writeDefault = TFRecordOptions(CompressionType.NONE)
```
ravwojdyla: Should it be compressed by default?

nevillelyh: Not sure since the Python TF version defaults to NONE.

nevillelyh:
It reads raw/zlib/gzip files generated by Python TF. I can look into TF but it'll be a lot more work like including a TF distribution in the code base or checking in pre-generated binary file.

PR #390, opened 2017-01-09
Add insertId/rowId support for BigQuery de-deplication

nsphung:
This will call a method to trigger best effort de-deduplication on streaming insert in BigQuery. I didn't find any other way to call this from the current version of scio. Please tell me if there's another existing way to do this.

@see https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataconsistency


nevillelyh:
Adding it in `BigQueryClient` won't affect `SCollection#saveAsBigQuery` and it's better to solve that in user space by having an additional `insertId` column in the rows.
So I'd prefer not to add this complication here.

PR #396, opened 2017-01-17
Expose ScioContext#isTest publicly

bcleenders:
@nevillelyh 

Since Mark doesn't have Github, apparently ;-)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -383,7 +383,8 @@ class ScioContext private[scio] (val options: PipelineOptions,
   // Test wiring
   // =======================================================================
 
-  private[scio] def isTest: Boolean = testId.isDefined
+  /**  Whether this is is a test context. */
```
nevillelyh: `is is` :)

PR #398, opened 2017-01-17
Add integration test for DistCache

andrewsmartin:
This is for #373, also did a bit of refactoring to handle `ScioContext` differently in unit vs integration test.


build.sbt
```diff
@@ -15,7 +15,7 @@
  * under the License.
  */
 
-import sbt._
+import sbt.{Defaults, _}
```
nevillelyh: `_` already includes `Defaults`, you shouldn't need this line.

scio-test/src/main/scala/com/spotify/scio/testing/ContextProvider.scala
```diff
@@ -0,0 +1,32 @@
...
+private[testing] trait ContextProvider {
+  def context: ScioContext
+}
+
+private[testing] trait UnitTestContextProvider extends ContextProvider {
+  override def context: ScioContext = ScioContext.forTest()
+}
+
+private[testing] trait IntegrationTestContextProvider extends ContextProvider {
+  override def context: ScioContext = ScioContext()
```
nevillelyh: This is a bit misleading since it's still a local runner and doesn't actually run on any service.

scio-test/src/it/scala/com.spotify.scio.values/DistCacheTest.scala
```diff
@@ -0,0 +1,21 @@
+package com.spotify.scio.values
+
+import com.spotify.scio.testing.ITPipelineSpec
+
+class DistCacheTest extends ITPipelineSpec {
```
nevillelyh: `ITPipelineSpec` and `ContextProvider` classes aren't exposed and only used here. Not sure if it's worth the extra complexity. Move those logic here for now and expand later if there's a need for integration tests?

scio-test/src/it/scala/com.spotify.scio.values/DistCacheTest.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  "GCS DistCache" should "work" in {
+    val distCacheUri = "gs://data-integration-test-us/scio/name-distcache"
+
+    runWithDistCache(Seq("name1", "name2"), distCacheUri) { sc =>
```
nevillelyh: `runWithDistCache` treats `data` as strings but the reading logic is here. This could be a potential cause of confusion.
I'd change the 2nd param from `(fn: ScioContext => T2)` to `(fn: (ScioContext, DistCache[T1]) => T2)` and move reading logic to `runWithDistCache`. That way you can get rid of `distCacheUri` too and use a temp URI inside `runWithDistCache`.

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -131,3 +140,42 @@ trait PipelineTestUtils {
...
+   * @param fn code that tests the components and verifies the result
+   *
+   * {{{
+   * runWithDistCache(Seq("a", "b"), "gs://<bucket>/<file>") { sc =>
+   *   val cache = sc.distCache("gs://<bucket>/<file>") (initFn) // Can use in test
+   *   sc.parallelize(Seq(1, 2, 3)).sum should containSingleValue (6)
+   * }
+   * }}}
+   */
+  def runWithDistCache[T1: ClassTag, T2: ClassTag](data: Iterable[T1], path: String)
```
nevillelyh: `T1` and `T2` aren't logical siblings so I'd use different letters, like `T` and `U`.

scio-test/src/main/scala/com/spotify/scio/testing/ContextProvider.scala
andrewsmartin: Good point, I guess I called it integration because using real data instead of mocking but you're right, it's misleading - any idea on a better name?

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -131,3 +140,42 @@ trait PipelineTestUtils {
...
+                                                  (fn: ScioContext => T2)
+  : ScioResult = {
+    try {
+      val channel = gcsUtil.create(GcsPath.fromUri(path), MimeTypes.BINARY)
+      channel.write(ByteBuffer.wrap(data.mkString("\n").getBytes))
```
nevillelyh: Should be `MimeTypes.TEXT` in this case?
We'll have to make sure `T` round trips instead of treating them as strings if this is public. Possibly using `KryoAtomicCoder`. But for a private test utility it's overkill.

scio-test/src/it/scala/com.spotify.scio.values/DistCacheTest.scala
```diff
@@ -0,0 +1,51 @@
...
+      }
+
+      val p = sc.parallelize(0 to 1)
+        .map(i => (i, names()(i)))
+
+      p should containInAnyOrder(Seq((0, "name1"), (1, "name2")))
+    }
+  }
+
+  def runWithDistCache[T1: ClassTag, T2: ClassTag](data: Iterable[T1], path: String)
```
nevillelyh: `T1` isn't handled generically. Remove it and just use `String`?

scio-test/src/it/scala/com.spotify.scio.values/DistCacheTest.scala
```diff
@@ -0,0 +1,51 @@
...
+import scala.reflect.ClassTag
+
+class DistCacheTest extends PipelineSpec {
+
+  val distCacheUri = "gs://data-integration-test-us/scio/name-distcache"
+  val cacheData = Seq("name1", "name2")
+
+  "GCS DistCache" should "work" in {
+    runWithDistCache(cacheData, distCacheUri) { sc =>
+      val names = sc.distCache(distCacheUri) { f =>
```
nevillelyh: move this to `runWithDistCache` and add a param to `fn` like `fn: (ScioContext, DistCache[String]) => T`?

PR #404, opened 2017-01-18
Fix #394. Add doc for hotKeyThreshold scaling

### LINKED ISSUE #394: Skew join `hotKeyThreshold` should scale with `sampleFraction`?
Right now in `skewedJoin` `hotKeyThreshold` is used as is to check key frequency but frequencies in the `CMS` may scale if a smaller `sampleFraction` is used. Should we scale `hotKeyThreshold` proportionally also or at least document it? @ravwojdyla?

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -248,14 +248,15 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    * @param hotKeyThreshold key with `hotKeyThreshold` values will be considered hot. Some runners
    *                        have inefficient GroupByKey implementation for groups with more than 10K
    *                        values. Thus it is recommended to set `hotKeyThreshold` to below 10K,
-   *                        keep upper estimation error in mind.
+   *                        keep upper estimation error in mind. If you sample input via
+   *                        `sampleFraction` make sure to adjust `hotKeyThreshold` accordingly.
```
nevillelyh: maybe a bit more specific, `accordingly by multiplying the raw frequency estimate with sampleFraction`

PR #405, opened 2017-01-18
Add description to experimental package

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/experimental/package.scala
```diff
@@ -41,6 +41,9 @@ package object experimental {
   /** Typed BigQuery annotations and converters. */
   val BigQueryType = com.spotify.scio.bigquery.types.BigQueryType
 
+  /** description annotation for BigQuery. */
```
nevillelyh: Nitpick: Capitalize, or ```Annotation for BigQuery field `description`.```

nevillelyh:
:+1:

PR #410, opened 2017-01-24
Fix MultiJoin transform name impl to avoid race condition on var

andrewsmartin:
@nevillelyh A tiny concession is that `MultiJoin` doesn't use `TransformNameable` now, but this works and keeps it simple.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -97,7 +97,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def cogroup[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (Iterable[V], Iterable[W]))] =
-    MultiJoin.cogroup(self, that)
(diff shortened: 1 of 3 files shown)


nevillelyh:
That's fine. You don't have to and shouldn't use it here anyway.
Should the joins & cogroups in `PairSCollectionFunctions` be fixed as well?

andrewsmartin:
Sure, added names for them in 54b450f.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -97,7 +97,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def cogroup[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (Iterable[V], Iterable[W]))] =
-    MultiJoin.cogroup(self, that)
+    MultiJoin.withName("CoGroup").cogroup(self, that)
```
nevillelyh: Should these use hard coded string or `self.tfName`?

andrewsmartin: Oops, should be `self.tfName`

nevillelyh:
:+1:

PR #412, opened 2017-01-24
Fix #397 FileStorage#textFile should support compressed files.

### LINKED ISSUE #397: FileStorage#textFile should support compressed files
Text file can be `gz` or `bz2` compressed and we should handle those.

andrewsmartin:
@nevillelyh Thoughts? I'm also looking into refactoring the compression stuff a bit so this can be consolidated with the `TFRecord` stuff, but there are a few differences.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
@@ -17,7 +17,7 @@
 
 package com.spotify.scio.io
 
-import java.io.{File, FileInputStream, InputStream, SequenceInputStream}
+import java.io._
 import java.net.URI
 import java.nio.ByteBuffer
 import java.nio.channels.Channels
@@ -36,6 +36,7 @@ import org.apache.avro.Schema
 import org.apache.avro.file.{DataFileReader, SeekableFileInput, SeekableInput}
 import org.apache.avro.generic.GenericDatumReader
 import org.apache.avro.specific.{SpecificDatumReader, SpecificRecordBase}
+import org.apache.commons.compress.compressors.CompressorStreamFactory
 import org.apache.commons.io.filefilter.WildcardFileFilter
 import org.apache.commons.io.{FileUtils, IOUtils}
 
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -61,9 +62,9 @@ trait Tap[T] { self =>
 }
 
 /** Tap for text files on local file system or GCS. */
-case class TextTap(path: String) extends Tap[String] {
-  override def value: Iterator[String] = FileStorage(path).textFile
-  override def open(sc: ScioContext): SCollection[String] = sc.textFile(path)
+case class TextTap(path: String, compressionType: CompressionType = NONE) extends Tap[String] {
```
nevillelyh: You can auto detect compression type so keep the old interface here?

scio-core/src/main/scala/com/spotify/scio/io/CompressionType.scala
```diff
@@ -0,0 +1,68 @@
...
+import java.io.{InputStream, OutputStream}
+
+import com.google.cloud.dataflow.sdk.io.TextIO.{CompressionType => GCompressionType}
+import org.apache.commons.compress.compressors.CompressorStreamFactory
+
+/**
+ * Wrapper for [[com.google.cloud.dataflow.sdk.io.TextIO.CompressionType]] with
+ * helper methods for reading and writing.
+ */
+sealed trait CompressionType {
```
nevillelyh: Prefer not to add any more public interface right now. So remove this and move the wrap logic to `TapTest`?

scio-test/src/test/scala/com/spotify/scio/io/TapTest.scala
```diff
@@ -168,6 +170,20 @@ class TapTest extends TapSpec {
     FileUtils.deleteDirectory(dir)
   }
 
+  it should "support reading compressed text files" in {
+    val data = Seq.fill(100)(UUID.randomUUID().toString)
+    for ((cType, ext) <- Seq(("gz", "gz"), ("bzip2", "bz2"))) {
+      val file = new File(s"scio-test-data.$ext")
```
nevillelyh: Better use a tmp file instead of hard coded file name in local directory.

scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
```diff
@@ -70,8 +71,11 @@ private[scio] trait FileStorage {
       .map(_.iterator().asScala).reduce(_ ++ _)
   }
 
-  def textFile: Iterator[String] =
-    IOUtils.lineIterator(getDirectoryInputStream(path), Charsets.UTF_8).asScala
+  def textFile: Iterator[String] = {
+    val in = new BufferedInputStream(getDirectoryInputStream(path))
```
nevillelyh: Actually this might not work for a directory of multiple compressed files since concatenated stream of gzip/bz2 != concatenated gzip/bz2 streams. That's why I added https://github.com/spotify/scio/blob/3364b1fcc003541728c42efe827407887fe48702/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala#L127  as part of the `TFRecord` patch.

PR #415, opened 2017-01-25
Add ProtoIO to TestDataManager.

andrewsmartin:
Besides having a named way to deal with test IO for proto (instead of just using `ObjectFileIO`), this is so we get type checking on `T <:< Message`. It's slightly ugly in that, since the non-test implementation delegates to object file IO it will go through the `if (this.isTest)` check twice, but I don't think it's a big deal.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
@@ -112,6 +112,8 @@ case class BigQueryIO(tableSpecOrQuery: String) extends TestIO[TableRow](tableSp
 case class DatastoreIO(projectId: String, query: Query = null, namespace: String = null)
   extends TestIO[Entity](s"$projectId\t$query\t$namespace")
 
+case class ProtobufIO[T](path: String) extends TestIO(path)
+
 case class PubsubIO(topic: String) extends TestIO[String](topic)
 
 case class TableRowJsonIO(path: String) extends TestIO[TableRow](path)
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -112,6 +113,8 @@ case class BigQueryIO(tableSpecOrQuery: String) extends TestIO[TableRow](tableSp
 case class DatastoreIO(projectId: String, query: Query = null, namespace: String = null)
   extends TestIO[Entity](s"$projectId\t$query\t$namespace")
 
+case class ProtobufIO[T](path: String)(implicit ev: T <:< Message) extends TestIO(path)
```
nevillelyh: You shouldn't need the `ev` here. `saveAsProtobufFile` already has that constaint.
Does `[T <: Message]` work?

andrewsmartin: It doesn't, `ScioContext#protobufFile` is `def protobufFile[T: ClassTag](path: String)(implicit ev: T <:< Message)` so `T` alone doesn't have to be a `Message` (at least according to this API) as long as there is evidence in scope.

nevillelyh: But that already has `ev: T <:< Message` so this shouldn't need it. We don't have the constraint for `AvroIO` either.
Assuming user uses the right `T` in `.output[T](ProtobufIO(...))` which is also constrained by the assertion inside.
So no need to go nuts here.

andrewsmartin: Got rid of the type constraint in `ProtoIO`.

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -19,6 +19,7 @@ package com.spotify.scio.testing
 
 import com.google.api.services.bigquery.model.TableRow
 import com.google.datastore.v1.{Entity, Query}
+import com.google.protobuf.Message
```
nevillelyh: unused import

PR #420, opened 2017-01-25
Add examples from extra

andrewsmartin:


scio-examples/src/main/scala/com/spotify/scio/examples/extra/TapExample.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    // fetch tap values directly
+    println(t1.value.mkString(", "))
+    println(t2.value.mkString(", "))
+
+    // second job
+    val (sc2, _) = ContextAndArgs(cmdlineArgs)
+    // re-open taps in new context
+    val s = (t1.open(sc2) ++ t2.open(sc2).map(_.toInt)).sum
+    val result = sc2.close()
```
nevillelyh: use `sc2.close().waitUntilFinish()` here and remove `Await` below?

PR #424, opened 2017-01-26
Fix Pub/Sub subscription in the Scala WindowedWordCount example

vspiewak:
Use sc.pubsubSubscription instead of sc.pubsubTopic.
This will configure PubsubIO with the created subscription, instead of create another one.

ExampleOptions extends ExamplePubsubTopicAndSubscriptionOptions instead of ExamplePubsubTopicOptions.
This allow to use --pubsubSubscription parameter.

More info here: https://gitter.im/spotify/scio?at=5888ec77e836bf70108f0258

Keep Up the good work Spotify Team <3


ravwojdyla:
Thanks @vspiewak

PR #426, opened 2017-01-27
Improve kantan.csv integration.

nrinaudo:
The writeCsv and writeTsv functions were a bit non-idiomatic. By using the dedicated syntax,
you also get slightly better behaviour, in that the CSV will be serialised row by row, rather
than stored entirely in memory and then dump.

Source: kantan.csv author :)


ravwojdyla:
Sweet, thanks @nrinaudo !

nrinaudo:
Quite welcome, happy to help!

PR #427, opened 2017-01-27
switchable caching

sinasohangirsc:
* More efficient isLegacy method based on first line of the sql query.
* Add switch to enable/disable caching

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
@@ -538,7 +538,7 @@ object BigQueryClient {
   val CACHE_DIRECTORY_DEFAULT: String = sys.props("user.dir") + "/.bigquery"
 
   /** System property key for enabling or disabling scio bigquery caching */
-  val CACHE_ENABLED: String = "bigquery.cache.enabled"
+  val CACHE_ENABLED_KEY: String = "bigquery.cache.enabled"
 
   /** Default cache behavior is enabled. */
   val CACHE_ENABLED_DEFAULT: Boolean = true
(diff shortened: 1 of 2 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -528,6 +537,13 @@ object BigQueryClient {
   /** Default cache directory. */
   val CACHE_DIRECTORY_DEFAULT: String = sys.props("user.dir") + "/.bigquery"
 
+  /** System property key for enabling or disabling scio bigquery caching */
+  val CACHE_ENABLED: String = "bigquery.cache.enabled"
```
ravwojdyla: To keep it consistent `CACHE_ENABLED` -> `CACHE_ENABLED_KEY`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -426,27 +428,34 @@ class BigQueryClient private (private val projectId: String,
...
-      case Failure(e: GoogleJsonResponseException) if isInvalidQuery(e) =>
-        // dry run with legacy syntax next
-        dryRunQuery(true) match {
-          case Success(_) =>
-            logger.warn("Legacy syntax is deprecated, use SQL syntax instead. " +
-              "See https://cloud.google.com/bigquery/sql-reference/")
-            logger.warn(s"Legacy query: `$sqlQuery`")
-            true
-          case Failure(f) => throw f
+    sqlQuery.trim.split("\n")(0).trim match {
```
ravwojdyla: `.toLowerCase` ? probably no reason to be case sensitive here.

ravwojdyla:
Interesting idea. Is there any reason to suspect that someone might use those comments without intent to use this feature? We could make it a bit harder to make that mistake by say using `#syntax=legacysql` (but that is "safety" through obscurity anyway). Alternative might be to make it verbose through the API.

What do you think @nevillelyh @andrewsmartin ?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
sinasohangirsc: Let me fix the case sensitivity here. They indeed are not supposed to be case sensitive per:
https://cloud.google.com/bigquery/docs/reference/standard-sql/enabling-standard-sql

sinasohangirsc:
`#syntax=legacysql` would be a deviation from googles syntax though: 
https://cloud.google.com/bigquery/docs/reference/standard-sql/enabling-standard-sql

ravwojdyla:
@sinasohangirsc oh - i wasn't aware of this feature, this is good to know, in this case 👍

andrewsmartin:
👍

ravwojdyla:
Thanks @sinasohangirsc

PR #436, opened 2017-02-03
Bump protobuf-generic

andrewsmartin:


andrewsmartin:
NVM - this is supposed to be done when we switch to 2.12

PR #439, opened 2017-02-08
scala compilation flag to accept long file names

badrock:
this is required for encrypted partitions (linux)


badrock:
@ravwojdyla

ravwojdyla:
Thanks @badrock 

For documentation (and in case there is `NoClassDefFoundError`):
 * http://stackoverflow.com/questions/28565837/filename-too-long-sbt/32862972#32862972
 * http://stackoverflow.com/questions/28565837/filename-too-long-sbt/32862972#comment58076434_32862972

PR #440, opened 2017-02-09
fix(ScioContext): Use the given parameter of setJobName

aseure:
cc @ElPicador

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -305,13 +305,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
     if (_pipeline != null) {
       throw new RuntimeException("Cannot set job name once pipeline is initialized")
     }
-    Try(optionsAs[DataflowPipelineOptions]).foreach { opts =>
-      if (name != "") {
-        opts.setJobName(name)
-      } else {
-        opts.setJobName(new DataflowPipelineOptions.JobNameFactory().create(options))
-      }
-    }
+    Try(optionsAs[DataflowPipelineOptions]).foreach(_.setJobName(name))
   }
 
   /** Close the context. No operation can be performed once the context is closed. */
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -305,8 +305,13 @@ class ScioContext private[scio] (val options: PipelineOptions,
     if (_pipeline != null) {
       throw new RuntimeException("Cannot set job name once pipeline is initialized")
     }
-    Try(optionsAs[DataflowPipelineOptions])
-      .foreach(_.setJobName(new DataflowPipelineOptions.JobNameFactory().create(options)))
+    Try(optionsAs[DataflowPipelineOptions]).foreach { opts =>
+      if (name != "") {
```
ravwojdyla: How about just passing the name, and delegating the checks to the sdk (which happens locally, before the job is submitted)?

scio-test/src/test/scala/com/spotify/scio/ScioContextTest.scala
```diff
@@ -67,14 +67,23 @@ class ScioContextTest extends PipelineSpec {
...
 
+  it should "support user defined job name via context" in {
+    val jobName = "test-job-1"
+    val opts = PipelineOptionsFactory.create().as(classOf[DataflowPipelineOptions])
+    val sc = ScioContext(opts)
+    sc.setJobName(jobName)
+    val pipelineOpts = sc.pipeline.getOptions.as(classOf[DataflowPipelineOptions])
+    pipelineOpts.getJobName shouldBe jobName
+  }
+
```
ravwojdyla: Could you please also add test where:
 * job name is set via options
 * followed by set (with different name) via context
and the end name should be the one provided via  context.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
aseure: Yes indeed. I'll do it.

scio-test/src/test/scala/com/spotify/scio/ScioContextTest.scala
aseure: Will do 👍

ravwojdyla:
Looks good, thanks and good find @aseure !

aseure:
@ravwojdyla When do you plan to create a new `scio` release with this fix?

ravwojdyla:
@aseure `0.2.11` was "just" released, comms/release page will be updated soon.

PR #444, opened 2017-02-13
workaround for BEAM-298

nevillelyh:


andrewsmartin:
👍 for now since it's the easiest thing to do, but in the long term should we think about some way of refactoring / setting up dep injection for test context? So that anything test related could exist in `src/test` rather than `src/main`. Have to think about it more though

nevillelyh:
Yeah prefer not doing it now. Doing it too soon will probably mess more stuff up than fixing it.

PR #446, opened 2017-02-13
add BigtableExample and tests

nevillelyh:


scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -40,7 +41,139 @@ import scala.concurrent.Future
...
 
-  private val DEFAULT_SLEEP_DURATION = Duration.standardMinutes(20);
+  // =======================================================================
+  // Protobuf helpers
+  // =======================================================================
+
+  /** Helper methods for [[Mutation]]. */
+  object Mutations {
+
+    def newSetCell(family: String, column: ByteString, value: ByteString): Mutation =
```
rgruener: I would use `qualifier` or `columnQualifier` since column usually refers to `<columnFamily>:<columnQualifier>`

rgruener: Also without specifying a timestamp it will use `0` (which becomes that date in 1970). It probably makes more sense to help out users by using the current timestamp by default. Its what the hbase api did and will help users switch over.

nevillelyh: There's no default value for `timestampMicros` field in the proto. I guess bigtable service will get 0 as default for `Long` and use that?

rgruener: yeah, thats what it does but I would argue it makes more sense to default to current time since it is more natural and thats what all the higher level apis for bigtable use (the hbase-dataflow api does it if you dont add a timestamp). If somebody wants to explicitly set a timestamp to 0, they still can.

nevillelyh: 👍

Issue #336, opened 2016-11-13
Graduate type safe BigQuery IO to stable API


ravwojdyla:
Graduate type safe BigQuery IO to stable API

- [x] move type safe BQ API out of experimental package
- [ ] properly mark type safe API as the standard IO for BigQuery in scio

nevillelyh:
Make sense if the following conditions are met:

- Legacy syntax is phased out enough that we can deprecate it. Legacy syntax depends on `view` which is not as predictable and users may have to use the unsafe API
- we are confident that type-safe API covers most queries and can verify generated type matches actual query result schema
- We port it to scalameta so it works out of the box in IntelliJ without plugin

PR #450, opened 2017-02-15
graduate type safe BigQuery API from experimental #336

nevillelyh:


andrewsmartin:
👍, is it still possible to use the un-typed API?

nevillelyh:
Yes I just copied stuff from `experimental` package object to right places.

PR #452, opened 2017-02-15
Finish migrating properties to scalatest

andrewsmartin:
Manual `minSuccessful` configuration may be removed if fixed upstream: https://github.com/scalatest/scalatest/issues/1090


nevillelyh:
Feels like overkill. Just copy `implicit override val generatorDrivenConfig...` to each test case? I'd like to avoid too many subprojects.

andrewsmartin:
@nevillelyh removed the sub-project.

Issue #453, opened 2017-02-15
Inefficient side inputs


nevillelyh:
We are potentially doing greedy java -> scala conversion for some side inputs:
ListSideInput: `context.sideInput(view).asScala.toList`
MapSideInput: `context.sideInput(view).asScala.toMap`
MultiMapSideInput: `context.sideInput(view).asScala.map(kv => (kv._1, kv._2.asScala))(breakOut)`

nevillelyh:
Fixed by making list side input a `Seq[T]` and (multi)map side input lazily wrapped.

PR #456, opened 2017-02-16
make side input collections lazy #453

nevillelyh:


dadib:
Ran a crude benchmark for this.
Started two identical jobs on cloud dataflow using a 1.4 million record/~300mb MapSideInput.
Run on 30 n1-standard-8 machines.

**Baseline**
Total vCPU Time: 267.381 vCPU hr

**With this fix**
Total vCPU Time: 94.645 vCPU hr

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -35,6 +34,7 @@ trait SideInput[T] extends Serializable {
 
   private[values] def getCache[I, O](context: DoFn[I, O]#ProcessContext): T = {
     if (cache == null) {
+      // this is called once per DoFn instance, which means once per CPU on Dataflow
```
ravwojdyla: CPU or thread?

nevillelyh: CPU core / thread I mean

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -81,3 +81,35 @@ class SideInputContext[T] private[scio] (val context: DoFn[T, AnyRef]#ProcessCon
   /** Extract the value of a given [[SideInput]]. */
   def apply[S](side: SideInput[S]): S = side.getCache(context)
 }
+
+// Immutable wrapper for j.u.Map[A, B] because .asScala returns a s.c.mutable.Map
+private class SideInputMap[A, B](self: JMap[A, B]) extends Map[A, B] {
+
+  // make eager copies when necessary
+  // scalastyle:off method.name
```
ravwojdyla: Do you think it would make sense to add a log/warning for operations that perform copies?

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
nevillelyh: Fixed

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
nevillelyh: Wanna avoid doing that right now. Side input implies that the map should be treated as a read only look up table. The user should understand that these operations are just as or even more inefficient as regular maps.

ravwojdyla: Sure. But was thinking if that warning should be added only in the methods that make copies (and only if they happen to be invoked), just in case. But this is minor point anyway.

nevillelyh: Yeah gonna keep it as is. It's really silly if someone does `m + (k -> v)` or `m1 ++ m2` in a `fn` no matter how efficient `m` is.

PR #459, opened 2017-02-16
Sparkey side input

andrewsmartin:
@nevillelyh @ravwojdyla Just want to get some initial feedback on the implementation. This works for local files. Still todo: use `tempLocation` for storage, copy to GCS if `gcpTempLocation` defined, further testing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -285,7 +285,7 @@ lazy val scioExtra: Project = Project(
   "scio-extra",
   file("scio-extra")
 ).settings(
-  commonSettings,
+  commonSettings ++ itSettings,
   description := "Scio extra utilities",
   libraryDependencies ++= Seq(
     "com.google.guava" % "guava" % guavaVersion,
@@ -298,8 +298,8 @@ lazy val scioExtra: Project = Project(
   )
 ).dependsOn(
   scioCore,
-  scioTest % "test"
-)
+  scioTest % "it,test"
+).configs(IntegrationTest)
 
 lazy val scioHdfs: Project = Project(
   "scio-hdfs",
--- a/build.sbt
+++ b/build.sbt
@@ -208,7 +208,6 @@ lazy val scioCore: Project = Project(
   description := "Scio - A Scala API for Google Cloud Dataflow",
     libraryDependencies ++= beamDependencies,
   libraryDependencies ++= Seq(
-    "com.spotify.sparkey" % "sparkey" % "2.1.3",
     "com.twitter" %% "algebird-core" % algebirdVersion,
     "com.twitter" %% "chill" % chillVersion,
     "com.twitter" %% "chill-algebird" % chillVersion,
@@ -290,12 +289,16 @@ lazy val scioExtra: Project = Project(
   description := "Scio extra utilities",
   libraryDependencies ++= Seq(
     "com.google.guava" % "guava" % guavaVersion,
+    "com.spotify.sparkey" % "sparkey" % sparkeyVersion,
     "com.twitter" %% "algebird-core" % algebirdVersion,
     "org.scalanlp" %% "breeze" % breezeVersion,
     "info.debatty" % "java-lsh" % javaLshVersion,
     "org.scalatest" %% "scalatest" % scalatestVersion % "test",
     "org.scalacheck" %% "scalacheck" % scalacheckVersion % "test"
   )
+).dependsOn(
+  scioCore,
+  scioTest % "test"
 )
 
 lazy val scioHdfs: Project = Project(
--- a/build.sbt
+++ b/build.sbt
@@ -208,6 +208,7 @@ lazy val scioCore: Project = Project(
   description := "Scio - A Scala API for Google Cloud Dataflow",
     libraryDependencies ++= beamDependencies,
   libraryDependencies ++= Seq(
+    "com.spotify.sparkey" % "sparkey" % "2.1.3",
     "com.twitter" %% "algebird-core" % algebirdVersion,
     "com.twitter" %% "chill" % chillVersion,
     "com.twitter" %% "chill-algebird" % chillVersion,
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -17,6 +17,7 @@
 
 package com.spotify.scio.util
 
+import java.io.{File, FileOutputStream}
 import java.net.URI
 
 import com.fasterxml.jackson.databind.ObjectMapper
@@ -29,6 +30,8 @@ import org.apache.beam.runners.dataflow.options.{DataflowPipelineDebugOptions, D
 import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.services.dataflow.Dataflow
 import com.google.api.services.dataflow.model.JobMetrics
+import org.apache.beam.sdk.util.GcsUtil
+import org.apache.beam.sdk.util.gcsfs.GcsPath
 
 import scala.reflect.ClassTag
 import scala.util.Try
@@ -90,5 +93,37 @@ private[scio] object ScioUtil {
     } else {
       options.getTempLocation
     }
+  }
+
+  /**
+   * Download a file from GCS onto the local file system.
+   *
+   * There can be multiple DoFns/Threads trying to fetch the same data/files on the same
+   * worker. To prevent from situation where some workers see incomplete data, and keep the
+   * solution simple - let's make this method synchronized. There is a downside - more
+   * specifically we might have to wait a bit longer then in more optimal solution, but,
+   * simplicity > performance.
+   */
+  def fetchFromGCS(gcsUtil: GcsUtil, gcsUri: URI, dest: String): File = synchronized {
+    require(isGcsUri(gcsUri), "Invalid GCS URI.")
+    val file = new File(dest)
+    val src = gcsUtil.open(GcsPath.fromUri(gcsUri))
+
+    if (file.exists() && src.size() != file.length()) {
+      // File exists but has different size than source file, most likely there was an issue
+      // on previous thread, let's remove invalid file, and download it again.
+      file.delete()
+    }
+
+    if (!file.exists()) {
+      val fos: FileOutputStream = new FileOutputStream(dest)
+      val dst = fos.getChannel
+      val src = gcsUtil.open(GcsPath.fromUri(gcsUri))
+      val size = dst.transferFrom(src, 0, src.size())
+      dst.close()
+      fos.close()
+    }
+
+    file
   }
 }
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -24,9 +24,8 @@ import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import org.apache.beam.runners.direct.DirectRunner
 import org.apache.beam.sdk.coders.{Coder, CoderRegistry}
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryTableRowIterator
-import org.apache.beam.sdk.options.PipelineOptions
-import org.apache.beam.runners.dataflow.options.{DataflowPipelineDebugOptions,
-                                                 DataflowPipelineOptions}
+import org.apache.beam.sdk.options.{GcpOptions, PipelineOptions}
+import org.apache.beam.runners.dataflow.options.{DataflowPipelineDebugOptions, DataflowPipelineOptions}
 import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.services.dataflow.Dataflow
 import com.google.api.services.dataflow.model.JobMetrics
@@ -80,6 +79,18 @@ private[scio] object ScioUtil {
 
     ScioUtil.executeWithBackOff(getMetrics,
       s"Could not get dataflow metrics of ${getMetrics.getJobId} in ${getMetrics.getProjectId}")
+  }
+
+  def tempLocation(options: PipelineOptions): String = {
+    if (ScioUtil.isLocalRunner(options)) {
+      if (options.getTempLocation == null) {
+        sys.props("java.io.tmpdir")
+      } else {
+        options.getTempLocation
+      }
+    } else {
+      options.as(classOf[GcpOptions]).getGcpTempLocation
+    }
   }
 
 
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -31,6 +31,7 @@ import com.google.api.services.dataflow.Dataflow
 import com.google.api.services.dataflow.model.JobMetrics
 
 import scala.reflect.ClassTag
+import scala.util.Try
 
 private[scio] object ScioUtil {
 
@@ -82,16 +83,12 @@ private[scio] object ScioUtil {
   }
 
   def tempLocation(options: PipelineOptions): String = {
-    if (ScioUtil.isLocalRunner(options)) {
-      if (options.getTempLocation == null) {
-        sys.props("java.io.tmpdir")
-      } else {
-        options.getTempLocation
-      }
+    val location = Try(options.as(classOf[GcpOptions]))
+      .map(_.getGcpTempLocation).getOrElse(options.getTempLocation)
+    if (location == null) {
+      sys.props("java.io.tmpdir")
     } else {
-      options.as(classOf[GcpOptions]).getGcpTempLocation
+      options.getTempLocation
     }
   }
-
-
 }
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -26,7 +26,7 @@ import org.apache.beam.runners.direct.DirectRunner
 import org.apache.beam.sdk.coders.{Coder, CoderRegistry}
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryTableRowIterator
 import org.apache.beam.sdk.options.{GcpOptions, PipelineOptions}
-import org.apache.beam.runners.dataflow.options.{DataflowPipelineDebugOptions, DataflowPipelineOptions}
+import org.apache.beam.runners.dataflow.options._
 import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.services.dataflow.Dataflow
 import com.google.api.services.dataflow.model.JobMetrics
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -34,7 +34,6 @@ import org.apache.beam.sdk.util.GcsUtil
 import org.apache.beam.sdk.util.gcsfs.GcsPath
 
 import scala.reflect.ClassTag
-import scala.util.Try
 
 private[scio] object ScioUtil {
 
@@ -86,12 +85,14 @@ private[scio] object ScioUtil {
   }
 
   def tempLocation(options: PipelineOptions): String = {
-    val location = Try(options.as(classOf[GcpOptions]))
-      .map(_.getGcpTempLocation).getOrElse(options.getTempLocation)
-    if (location == null) {
-      sys.props("java.io.tmpdir")
+    if (ScioUtil.isLocalRunner(options)) {
+      if (options.getTempLocation == null) {
+        sys.props("java.io.tmpdir")
+      } else {
+        options.getTempLocation
+      }
     } else {
-      options.getTempLocation
+      options.as(classOf[GcpOptions]).getGcpTempLocation
     }
   }
 
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -17,14 +17,12 @@
 
 package com.spotify.scio.values
 
-import java.io.File
 import java.lang.{Iterable => JIterable, Long => JLong}
 import java.util.{Map => JMap}
 
 import com.spotify.scio.ScioContext
 import com.spotify.scio.util._
 import com.spotify.scio.util.random.{BernoulliValueSampler, PoissonValueSampler}
-import com.spotify.sparkey.{CompressionType, Sparkey}
 import com.twitter.algebird.{Aggregator, _}
 import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.values.{KV, PCollection, PCollectionView}
@@ -624,30 +622,6 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
         }
       })
     new MapSideInput[K, V](o)
-  }
-
-  def asSparkeySideInput: SideInput[Map[String, String]] = {
-    val f = self.groupBy(_ => ()).values
-      .map(_.map(kv => (kv._1.toString, kv._2.toString))).map { iter =>
-        val indexFile = new File("test.spi")
-        val writer = Sparkey.createNew(indexFile, CompressionType.NONE, 512);
-        iter.foreach { kv =>
-          writer.put(kv._1, kv._2)
-          writer.flush();
-        }
-        writer.writeHash();
-        writer.close();
-      indexFile.toString
-    }
-    val o = f.applyInternal(
-      new PTransform[PCollection[String], PCollectionView[String]]() {
-        override def expand(input: PCollection[String]): PCollectionView[String] = {
-          input.apply(View.asSingleton())
-        }
-      }
-    )
-
-    new SparkeySideInput(o)
   }
 
   /**
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -17,12 +17,14 @@
 
 package com.spotify.scio.values
 
+import java.io.File
 import java.lang.{Iterable => JIterable, Long => JLong}
 import java.util.{Map => JMap}
 
 import com.spotify.scio.ScioContext
 import com.spotify.scio.util._
 import com.spotify.scio.util.random.{BernoulliValueSampler, PoissonValueSampler}
+import com.spotify.sparkey.{CompressionType, Sparkey}
 import com.twitter.algebird.{Aggregator, _}
 import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.values.{KV, PCollection, PCollectionView}
@@ -622,6 +624,30 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
         }
       })
     new MapSideInput[K, V](o)
+  }
+
+  def asSparkeySideInput: SideInput[Map[String, String]] = {
+    val f = self.groupBy(_ => ()).values
+      .map(_.map(kv => (kv._1.toString, kv._2.toString))).map { iter =>
+        val indexFile = new File("test.spi")
+        val writer = Sparkey.createNew(indexFile, CompressionType.NONE, 512);
+        iter.foreach { kv =>
+          writer.put(kv._1, kv._2)
+          writer.flush();
+        }
+        writer.writeHash();
+        writer.close();
+      indexFile.toString
+    }
+    val o = f.applyInternal(
+      new PTransform[PCollection[String], PCollectionView[String]]() {
+        override def expand(input: PCollection[String]): PCollectionView[String] = {
+          input.apply(View.asSingleton())
+        }
+      }
+    )
+
+    new SparkeySideInput(o)
   }
 
   /**
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -821,15 +821,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    */
   def materialize: Future[Tap[T]] = {
     val filename = "scio-materialize-" + UUID.randomUUID().toString
-    val tmpDir = if (ScioUtil.isLocalRunner(context.options)) {
-      if (context.options.getTempLocation == null) {
-        sys.props("java.io.tmpdir")
-      } else {
-        context.options.getTempLocation
-      }
-    } else {
-      context.optionsAs[GcpOptions].getGcpTempLocation
-    }
+    val tmpDir = ScioUtil.tempLocation(context.options)
     val path = tmpDir + (if (tmpDir.endsWith("/")) "" else "/") + filename
     saveAsObjectFile(path)
   }
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -130,7 +130,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     this.pApply(transform)
 
   /** Apply a transform. */
-  private[values] def transform[U: ClassTag](f: SCollection[T] => SCollection[U])
+  private[scio] def transform[U: ClassTag](f: SCollection[T] => SCollection[U])
   : SCollection[U] = {
     val o = internal.apply(this.tfName, new PTransform[PCollection[T], PCollection[U]]() {
       override def expand(input: PCollection[T]): PCollection[U] = {
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -17,9 +17,11 @@
 
 package com.spotify.scio.values
 
+import java.io.File
 import java.lang.{Iterable => JIterable}
 import java.util.{List => JList, Map => JMap}
 
+import com.spotify.sparkey.Sparkey
 import org.apache.beam.sdk.transforms.DoFn
 import org.apache.beam.sdk.values.PCollectionView
 
@@ -67,6 +69,12 @@ private[values] class MapSideInput[K, V](val view: PCollectionView[JMap[K, V]])
     new SideInputMap(context.sideInput(view))
 }
 
+private[values] class SparkeySideInput(val view: PCollectionView[String])
+  extends SideInput[Map[String, String]] {
+  override def get[I, O](context: DoFn[I, O]#ProcessContext): Map[String, String] =
+    new SparkeySideInputMap(context.sideInput(view))
+}
+
 private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K, JIterable[V]]])
   extends SideInput[Map[K, Iterable[V]]] {
   override def get[I, O](context: DoFn[I, O]#ProcessContext): Map[K, Iterable[V]] =
@@ -92,6 +100,29 @@ private class SideInputMap[A, B](self: JMap[A, B]) extends Map[A, B] {
   override def get(key: A): Option[B] = Option(self.get(key))
   override def iterator: Iterator[(A, B)] = self.asScala.iterator
 
+}
+
+private class SparkeySideInputMap(indexFile: String) extends Map[String, String] {
+  val reader = Sparkey.open(new File(indexFile))
+
+  override def get(key: String): Option[String] = Option(reader.getAsString(key))
+
+  override def iterator: Iterator[(String, String)] = new Iterator[(String, String)] {
+    val delegate = reader.iterator()
+
+    override def hasNext: Boolean = delegate.hasNext
+
+    override def next(): (String, String) = {
+      val entry = delegate.next()
+      (entry.getKeyAsString, entry.getValueAsString)
+    }
+  }
+
+  // scalastyle:off method.name
+  override def +[B1 >: String](kv: (String, B1)): Map[String, B1] = ???
+
+  override def -(key: String): Map[String, String] = ???
+  // scalastyle:on method.name
 }
 
 // Immutable wrapper for j.u.Map[A, j.l.terable[B]] because .asScala returns a s.c.mutable.Map
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -17,11 +17,9 @@
 
 package com.spotify.scio.values
 
-import java.io.File
 import java.lang.{Iterable => JIterable}
 import java.util.{List => JList, Map => JMap}
 
-import com.spotify.sparkey.Sparkey
 import org.apache.beam.sdk.transforms.DoFn
 import org.apache.beam.sdk.values.PCollectionView
 
@@ -69,12 +67,6 @@ private[values] class MapSideInput[K, V](val view: PCollectionView[JMap[K, V]])
     new SideInputMap(context.sideInput(view))
 }
 
-private[values] class SparkeySideInput(val view: PCollectionView[String])
-  extends SideInput[Map[String, String]] {
-  override def get[I, O](context: DoFn[I, O]#ProcessContext): Map[String, String] =
-    new SparkeySideInputMap(context.sideInput(view))
-}
-
 private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K, JIterable[V]]])
   extends SideInput[Map[K, Iterable[V]]] {
   override def get[I, O](context: DoFn[I, O]#ProcessContext): Map[K, Iterable[V]] =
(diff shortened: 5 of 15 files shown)


andrewsmartin:
Also this doesn't work for generic types, only strings for now

nevillelyh:
I'd prefer to keep it in `scio-extra` as implicit classes.

build.sbt
```diff
@@ -208,6 +208,7 @@ lazy val scioCore: Project = Project(
   description := "Scio - A Scala API for Google Cloud Dataflow",
     libraryDependencies ++= beamDependencies,
   libraryDependencies ++= Seq(
+    "com.spotify.sparkey" % "sparkey" % "2.1.3",
```
nevillelyh: There's a `sparkeyVersion` variable now?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -624,6 +626,30 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
     new MapSideInput[K, V](o)
   }
 
+  def asSparkeySideInput: SideInput[Map[String, String]] = {
+    val f = self.groupBy(_ => ()).values
+      .map(_.map(kv => (kv._1.toString, kv._2.toString))).map { iter =>
```
nevillelyh: Fuse the `.values.map` together?
Also fuse the `.map.map { iter` together?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -624,6 +626,30 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+        val indexFile = new File("test.spi")
+        val writer = Sparkey.createNew(indexFile, CompressionType.NONE, 512);
+        iter.foreach { kv =>
```
nevillelyh: This is an expensive loop, unroll as `while`?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -624,6 +626,30 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+          writer.put(kv._1, kv._2)
+          writer.flush();
```
nevillelyh: Probably don't want to flush per element? Also trailing `;`

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -624,6 +626,30 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
     new MapSideInput[K, V](o)
   }
 
+  def asSparkeySideInput: SideInput[Map[String, String]] = {
```
nevillelyh: Instead of `Map[String, String]`, expose the underlying `SparkeyReader` directly?

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -94,6 +102,29 @@ private class SideInputMap[A, B](self: JMap[A, B]) extends Map[A, B] {
 
 }
 
+private class SparkeySideInputMap(indexFile: String) extends Map[String, String] {
```
nevillelyh: Don't need this if you expose `SparkeyReader` directly

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
andrewsmartin: I thought about this and decided not to because the users would have to explicitly pull in the Sparkey dependency - but I guess that's no such a big deal?

andrewsmartin: My thinking was that it would be nice for the users to not deal with Sparkey at all

nevillelyh:
Also instead of a single `asSparkeySideInput: SideInput[SparkeyReader]`, maybe split it up to 2 steps?

1. A `SCollection[(K, V)]#asSparkey(url: SparkeyURL = DEFAULT_...): SCollection[SparkeyURL]`, so the URL can be saved to GCS or something and reused in another pipeline
2. A `SCollection[SparkeyURL]#asSparkeySideInput: SideInput[SparkeyReader]` to open it. So you can open existing files on GCS as well.
3. A `SCollection[(String, String)]#asSparkeySideInput` that does 1 and 2 in a single step

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
nevillelyh: Also the 2 `.toString` are scary. I'd rather add view bounds (`T <:< String`) on `K` and `V`

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
nevillelyh: In that case it's better to have a implicit class that from `SparkeyReader` to `Map` or something Map like. Decoupled from any side input logic and can be used standalone.

nevillelyh: No `scio-extra` will depend on `sparkey` and the user doesn't have to explicitly do it.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
andrewsmartin: 👍 although for the `.values.map` though not sure what you mean by fuse? If I do `.mapValues` I would still need to do `.values` at the end anyway to have `SCollection[String]` of the file path, so I figured it makes since to get rid of the key up front.

andrewsmartin: Actually nvm, just realized I can drop the key as part of the map since there is only one.

andrewsmartin:
@nevillelyh @ravwojdyla this is ready for review now.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+object Sparkey {
+
+  implicit class SparkeyScioContext(val self: ScioContext) {
+    def sparkeySideInput(url: SparkeyUri): SideInput[SparkeyReader] = {
+      val view = self.parallelize(Seq(url)).applyInternal(sparkeyTransform)
+      new SparkeySideInput(view)
```
nevillelyh: Do you really need `sparkeyTransform`? `self.internal.apply(View.asSingleton)` should just work?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+
+  implicit class SCollectionWithSparkeyWriter[K, V](val self: SCollection[(K, V)])
+                                                 (implicit ev1: K <:< String, ev2: V <:< String) {
```
nevillelyh: indentation

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+            val kv = it.next()
+            writer.put(kv._1.toString, kv._2.toString)
+          }
+          writer.close()
+          uri
+        }
+    }
+
+    def asSparkey: SCollection[SparkeyUri] = {
+      val tempLocation = Try(self.context.optionsAs[DataflowPipelineOptions].getGcpTempLocation)
```
nevillelyh: There's some temp location logic in `SCollection#materialize` and `ScioContext#pipeline`, consolidate them into something like `ScioUtil.tempLocation(options: Options)`?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  implicit class SparkeySCollection(val self: SCollection[SparkeyUri]) {
+    def asSparkeySideInput(): SideInput[SparkeyReader] = {
+      val view = self.applyInternal(sparkeyTransform)
+      new SparkeySideInput(view)
+    }
+  }
+
+  implicit class SCollectionToSparkeySideInput(val self: SCollection[(String, String)]) {
```
nevillelyh: call it `SparkeyPairSCollectionFunctions`?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+    def asSparkeySideInput(): SideInput[SparkeyReader] = self.asSparkey.asSparkeySideInput()
+  }
+
+  implicit class Reader(val self: SparkeyReader) {
```
nevillelyh: `RichSparkeyReader`

nevillelyh: Is it possible to extend `Map[String, String]` here? That way you get many method impls for free.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+    def get(key: String): Option[String] = Option(self.getAsString(key))
+
+    def toStream(): Stream[(String, String)] = self.iterator().asScala.toStream.map { e =>
```
nevillelyh: Not sure about this. `def iterator: Iterator[(String, String)]` is more intuitive?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -130,7 +130,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     this.pApply(transform)
 
   /** Apply a transform. */
-  private[values] def transform[U: ClassTag](f: SCollection[T] => SCollection[U])
+  private[scio] def transform[U: ClassTag](f: SCollection[T] => SCollection[U])
```
nevillelyh: You don't need this change now that `sparkeyTransform` is gone.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  private[scio] class SparkeySideInput(val view: PCollectionView[SparkeyUri])
+    extends SideInput[SparkeyReader] {
+    override def get[I, O](context: DoFn[I, O]#ProcessContext): SparkeyReader =
+      SparkeyUri(context.sideInput(view).path).getReader()
+  }
+}
+
+trait SparkeyUri extends Serializable {
```
nevillelyh: I'd put the trait and object in `Object Sparkey` so they are imported with `import Sparkey._`.
Also Scala doc for both.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.beam.sdk.options.PipelineOptionsFactory
+import org.apache.beam.sdk.transforms.{DoFn, View}
+import org.apache.beam.sdk.util.GcsUtil.GcsUtilFactory
+import org.apache.beam.sdk.util.gcsfs.GcsPath
+import org.apache.beam.sdk.values.PCollectionView
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+object Sparkey {
```
nevillelyh: Some docs for this, maybe examples of the difference use cases.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
andrewsmartin: Thought about this, but then I'd have to define `def +[B1 >: String](kv: (String, B1)): Map[String, B1]` which doesn't really make sense if we want to consider this as read-only.

nevillelyh: Default is `???` which is sugar for not implemented exception. Or you can override and throw one with a better message?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+}
+
+object SparkeyUri {
+  def apply(path: String): SparkeyUri =
+    if (new URI(path).getScheme == "gs") new GcsSparkeyUri(path) else new LocalSparkeyUri(path)
+}
+
+private case class LocalSparkeyUri(path: String) extends SparkeyUri {
+  override def getReader(): SparkeyReader = JSparkey.open(new File(path + ".spi"))
+
```
nevillelyh: For small classes like this I'll skip the blank line.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+  override def getWriter(): WrappedSparkeyWriter = new LocalSparkeyWriter(path)
+}
+
+private case class GcsSparkeyUri(path: String) extends SparkeyUri {
+  override def getReader(): SparkeyReader = {
```
nevillelyh: This might be executed once per thread? See dist cache impl ensures that the file is downloaded only once.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+    val gcs = new GcsUtilFactory().create(PipelineOptionsFactory.create())
+    val localTmpDir = Files.createTempDir()
+    // Copy .spi and .spl to local files
+    for (ext <- Seq(".spi", ".spl")) {
+      val src = gcs.open(GcsPath.fromUri(s"$path/sparkey$ext"))
```
nevillelyh: This won't work with pipelines with multiple sparkeys. Also multiple sparkey URIs can even have the same filename, e.g. `gs://bucket1/data.spi` and `gs://bucket2/data.spi`. Better hash the URI as a prefix. I think dist cache does something like that?

nevillelyh: You might wanna extra the logic in `https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/DistCache.scala#L64` into `ScioUtil` as well, as a util method that takes `gcsUtil` and `uri`.

nevillelyh: Just saw that you're using `Files.createTempDir()`, but this will download one copy per thread to a different location. See `DistCache` for a stable unique copying mechanism.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+  private[scio] class SparkeySideInput(val view: PCollectionView[SparkeyUri])
+    extends SideInput[SparkeyReader] {
+    override def get[I, O](context: DoFn[I, O]#ProcessContext): SparkeyReader =
+      SparkeyUri(context.sideInput(view).path).getReader()
+  }
+}
+
+trait SparkeyUri extends Serializable {
+  val path: String
+  def getWriter(): WrappedSparkeyWriter
```
nevillelyh: Make `getWriter` private? The user shouldn't use it directly.
Also override `.toString` so `SparkeyUri` can be printed or saved as text file.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+      new SparkeySideInput(view)
+    }
+  }
+
+  implicit class SCollectionWithSparkeyWriter[K, V](val self: SCollection[(K, V)])
+                                                   (implicit ev1: K <:< String, ev2: V <:< String) {
+    def asSparkey(uri: SparkeyUri): SCollection[SparkeyUri] = self.transform { in =>
+      in.groupBy(_ => ())
+        .map { case (_, iter) =>
+          val writer = uri.getWriter()
```
nevillelyh: Actually this is the only place `getWriter` is used. Get rid of the method, `WrappedSparkeyWriter` and 2 subclasses, and put everything here in a utility object? That's a lot cleaner.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
nevillelyh: I'm OK with not extending but maybe in that case get rid of `toStream` and add a `.apply(key: String)` (get or fail).

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
andrewsmartin: I still need it for `SCollectionWithSparkeyWriter.asSparkey`

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+          uri
+        }
+    }
+
+    def asSparkey: SCollection[SparkeyUri] = {
+      val tempLocation = Try(self.context.optionsAs[DataflowPipelineOptions].getGcpTempLocation)
+        .getOrElse(self.context.options.getTempLocation)
+      this.asSparkey(SparkeyUri(tempLocation))
+    }
+  }
```
nevillelyh: Add a `asSparkeySideInput` in this class so that the user doesn't have to do `p.asSparkey.asSparkeySideInput`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
nevillelyh: Right my bad!

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,166 @@
...
+  def getReader(): SparkeyReader
+}
+
+object SparkeyUri {
+  def apply(path: String): SparkeyUri =
+    if (new URI(path).getScheme == "gs") new GcsSparkeyUri(path) else new LocalSparkeyUri(path)
+}
+
+private case class LocalSparkeyUri(path: String) extends SparkeyUri {
+  override def getReader(): SparkeyReader = JSparkey.open(new File(path + ".spi"))
```
rgruener: The GcsSparkeyUri appends `/sparkey.spi` to the path while the Local one only appends `.spi` which is confusing. Perhaps the local one should append the same thing.

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -82,5 +85,45 @@ private[scio] object ScioUtil {
       s"Could not get dataflow metrics of ${getMetrics.getJobId} in ${getMetrics.getProjectId}")
   }
 
+  def tempLocation(options: PipelineOptions): String = {
+    val location = Try(options.as(classOf[GcpOptions]))
+      .map(_.getGcpTempLocation).getOrElse(options.getTempLocation)
+    if (location == null) {
+      sys.props("java.io.tmpdir")
+    } else {
```
ravwojdyla: Is this `else` needed here? Isn't there already `getOrElse(options.getTempLocation)`?

andrewsmartin: It was just there for the return, but you're right, this is more concise:

```
Option(Try(options.as(classOf[GcpOptions]))
      .map(_.getGcpTempLocation).getOrElse(options.getTempLocation))
      .getOrElse(sys.props("java.io.tmpdir")
```

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  implicit class SparkeyScioContext(val self: ScioContext) {
+    def sparkeySideInput(url: SparkeyUri): SideInput[SparkeyReader] = {
+      val view = self.parallelize(Seq(url)).applyInternal(View.asSingleton())
+      new SparkeySideInput(view)
+    }
+  }
+
+  implicit class SCollectionWithSparkeyWriter[K, V](val self: SCollection[(K, V)])
+                                                   (implicit ev1: K <:< String, ev2: V <:< String) {
```
ravwojdyla: `SCollectionWithSparkeyWriter[K <:< String, V <:< String]`?

andrewsmartin: I tried this, but it doesn't work anymore. I think since view bounds were deprecated this has to be done with implicit evidence.

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
andrewsmartin: But the second `.getOrElse` is needed because `tempLocation` can be `null`.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+            writer.put(kv._1.toString, kv._2.toString)
+          }
+          writer.close()
+          uri
+        }
+    }
+
+    /** Write the contents of this SCollection as a Sparkey file using the default uri. */
+    def asSparkey: SCollection[SparkeyUri] = {
+      val uri = ScioUtil.tempLocation(self.context.options) + "/sparkey"
```
ravwojdyla: Would this break if there was multiple `asSparkey`, maybe it should be `/sparkey_<uniq-transform-name>`?

andrewsmartin: You're right, this should be unique. You think it's fine to add a random UUID or something to the prefix?

ravwojdyla: If there is nothing inside of the `self` that we could draw "randomness" from.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
nevillelyh: Actually how about `(val self: SCollection[(String, String)])` instead? That way you can get rid of the implicit args and the method is only available to `(String, String)`.

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -82,5 +85,41 @@ private[scio] object ScioUtil {
       s"Could not get dataflow metrics of ${getMetrics.getJobId} in ${getMetrics.getProjectId}")
   }
 
+  def tempLocation(options: PipelineOptions): String = {
+    Option(Try(options.as(classOf[GcpOptions]))
```
nevillelyh: Not a huge fan of such complex chain. IMO the original `if/else` is more readable.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
nevillelyh: Maybe something like `$tempLocation/sparkey-$uuid.{spi,spl}`?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.beam.sdk.util.gcsfs.GcsPath
+import org.apache.beam.sdk.values.PCollectionView
+
+object Sparkey {
+
+  /**
+   * Represents the base URI for a Sparkey index and log file, either on the local file
+   * system or on GCS. For GCS, the uri should be in the form
+   * 'gs://<bucket>/<path>/<sparkey-prefix>'. For local files, the uri should be in the form
+   * '/<path>/<sparkey-prefix>'. Note that the uri must not be a folder or GCS bucket as the uri is
```
nevillelyh: These are public docs, use `URI` proper capitalization?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
nevillelyh: Also `default uri` is misleading in the doc. "to a temporary location"?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+    def asSparkey: SCollection[SparkeyUri] = {
+      val uri = ScioUtil.tempLocation(self.context.options) + "/sparkey"
+      this.asSparkey(SparkeyUri(uri))
+    }
+
+    def asSparkeySideInput: SideInput[SparkeyReader] = self.asSparkey.asSparkeySideInput()
+  }
+
+  implicit class SparkeySCollection(val self: SCollection[SparkeyUri]) {
+    def asSparkeySideInput(): SideInput[SparkeyReader] = {
```
nevillelyh: I think our convention is no `()` here for no-arg methods.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
nevillelyh: `implicit class SparkelyPairSCollection`?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
andrewsmartin: @nevillelyh It may not be temporary though, since the files written here will persist on GCS. It's just that the location is the default one instead of specified by the user. Maybe it should say "default location"?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+      val view = self.applyInternal(View.asSingleton())
+      new SparkeySideInput(view)
+    }
+  }
+
+  implicit class SparkeyPairSCollectionFunctions(val self: SCollection[(String, String)]) {
```
nevillelyh: `extends AnyVal`

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+    /** Write the contents of this SCollection as a Sparkey file using the default uri. */
+    def asSparkey: SCollection[SparkeyUri] = {
+      val uri = ScioUtil.tempLocation(self.context.options) + "/sparkey"
+      this.asSparkey(SparkeyUri(uri))
+    }
+
+    def asSparkeySideInput: SideInput[SparkeyReader] = self.asSparkey.asSparkeySideInput()
+  }
+
+  implicit class SparkeySCollection(val self: SCollection[SparkeyUri]) {
```
nevillelyh: `extends AnyVal`

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+    def getReader(): SparkeyReader
+    override def toString(): String = basePath
+  }
+
+  object SparkeyUri {
+    def apply(path: String): SparkeyUri =
+      if (ScioUtil.isGcsUri(new URI(path))) new GcsSparkeyUri(path) else new LocalSparkeyUri(path)
+  }
+
+  implicit class SparkeyScioContext(val self: ScioContext) {
```
nevillelyh: `extends AnyVal`

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  implicit class SparkeyPairSCollectionFunctions(val self: SCollection[(String, String)]) {
+    def asSparkeySideInput(): SideInput[SparkeyReader] = self.asSparkey.asSparkeySideInput()
+  }
+
+  implicit class RichSparkeyReader(val self: SparkeyReader) extends Map[String, String] {
+    override def get(key: String): Option[String] = Option(self.getAsString(key))
+
+    override def iterator: Iterator[(String, String)] = new Iterator[(String, String)] {
```
nevillelyh: I think you can just `self.iterator.asJava.map`? `map` on an `Iterator` is lazy.

andrewsmartin: I wanted to convert from `Iterator[Entry]` to `Iterator[(String, String)]` anyway though

andrewsmartin: Nvm, get what you mean

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+
+      override def hasNext: Boolean = delegate.hasNext
+
+      override def next(): (String, String) = {
+        val entry = delegate.next()
+        (entry.getKeyAsString, entry.getValueAsString)
+      }
+    }
+    //scalastyle:off method.name
+    override def +[B1 >: String](kv: (String, B1)): Map[String, B1] = ???
```
nevillelyh: `???` defaults to `new NotImplementedError()`
Maybe add an message "Not implemented. Sparkey-backed map."

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,181 @@
...
+    extends SideInput[SparkeyReader] {
+    override def get[I, O](context: DoFn[I, O]#ProcessContext): SparkeyReader =
+      SparkeyUri(context.sideInput(view).basePath).getReader()
+  }
+
+  private case class LocalSparkeyUri(basePath: String) extends SparkeyUri {
+    override def getReader(): SparkeyReader = JSparkey.open(new File(basePath))
+  }
+
+  private object GcsSparkeyUri
```
nevillelyh: What's the object for?

andrewsmartin: Oops, was using this to synchronize on for downloading from GCS (like in DistCache). That logic is refactored in `ScioUtil` now though so I can remove this.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+    def getReader(): SparkeyReader
+    override def toString(): String = basePath
+  }
+
+  object SparkeyUri {
+    def apply(path: String): SparkeyUri =
+      if (ScioUtil.isGcsUri(new URI(path))) new GcsSparkeyUri(path) else new LocalSparkeyUri(path)
+  }
+
+  implicit class SparkeyScioContext(val self: ScioContext) extends AnyVal {
```
ravwojdyla: What is the `AnyVal` for?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+    def sparkeySideInput(url: SparkeyUri): SideInput[SparkeyReader] = {
+      val view = self.parallelize(Seq(url)).applyInternal(View.asSingleton())
```
ravwojdyla: Why not `asSingletonSideInput`?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+          }
+          writer.close()
+          uri
+        }
+    }
+
+    /** Write the contents of this SCollection as a Sparkey file using the default uri. */
+    def asSparkey: SCollection[SparkeyUri] = {
+      val uid = UUID.randomUUID()
+      val uri = ScioUtil.tempLocation(self.context.options) + s"/sparkey-$uid"
```
ravwojdyla: nitpick, you could add log info here, in case someone wants to connect job execution to sparkey key location.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
andrewsmartin: It's to minimize unnecessary object allocation: http://docs.scala-lang.org/overviews/core/value-classes.html

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
```diff
@@ -0,0 +1,173 @@
...
+      this.asSparkey(SparkeyUri(uri))
+    }
+
+    def asSparkeySideInput: SideInput[SparkeyReader] = self.asSparkey.asSparkeySideInput
+  }
+
+  implicit class SparkeySCollection(val self: SCollection[SparkeyUri]) extends AnyVal {
+    def asSparkeySideInput: SideInput[SparkeyReader] = {
+      val view = self.applyInternal(View.asSingleton())
```
ravwojdyla: Why not asSingletonSideInput?

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
nevillelyh: I don't think you can get away without checking for runner. Falling back to `java.io.tmpdir` will not work for a remote runner. Just duplicate the original logic?

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
nevillelyh: It's a scala 2.11+ thing, `extends AnyVal` make this a value class and optimizes away the `new SparkeyScioContext` constructor step.

scio-extra/src/main/scala/com/spotify/scio/extra/Sparkey.scala
andrewsmartin: `.asSingletonSideInput` would give us a `SideInput[SparkeyUri]`. With the custom `SparkeySideInput` the user gets an open `SparkeyReader` without leaking any details of how it's being read.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -0,0 +1,122 @@
...
+}
+
+private case class LocalSparkeyUri(basePath: String) extends SparkeyUri {
+  override def getReader: SparkeyReader = Sparkey.open(new File(basePath))
+  override def exists: Boolean =
+    (new File(basePath + ".spi").exists || new File(basePath + ".spl").exists)
+}
+
+private case class GcsSparkeyUri(basePath: String,
+                                          @transient options: GcsOptions) extends SparkeyUri {
```
nevillelyh: Indentation.

It might be clearer if it's a simple `private class GcsSparkeyUri` since we don't need the extra stuff from a case class. That way you pass `options` in as an argument and can get rid of `@transient`.

Same goes for `LocalSparkeyUri`, `private class LocalSparkeyUri(val basePath: String)` is clearer IMO.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -0,0 +1,122 @@
...
+  val localBasePath: String = sys.props("java.io.tmpdir") + hashPrefix(basePath)
+
+  private val json: String = new ObjectMapper().writeValueAsString(options)
+  private def opts: GcsOptions = new ObjectMapper()
```
nevillelyh: This is used in `def gcs` only, combine the 2?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -0,0 +1,122 @@
...
+    .readValue(json, classOf[PipelineOptions])
+    .as(classOf[GcsOptions])
+
+  def gcs: GcsUtil = opts.getGcsUtil
```
nevillelyh: This can be private?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  override def exists: Boolean = {
+    val index = GcsPath.fromUri(basePath + ".spi")
+    val log = GcsPath.fromUri(basePath + ".spl")
+    Try(gcs.fileSize(index)).isSuccess || Try(gcs.fileSize(log)).isSuccess
+  }
+
+  private def hashPrefix(path: String): String =
+    Hashing.sha1().hashString(path, Charsets.UTF_8).toString.substring(0, 8) + "-sparkey"
```
nevillelyh: Make it `sparkey-$hash` so it's easier when debugging?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
andrewsmartin: I do pattern matching in several places, so I wanted to keep the case classes. Would you prefer to drop it and do `if (uri.isInstanceOf[])` instead of pattern matching?

nevillelyh: You can still match like `case uri: LocalSparkeyUri` right?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
nevillelyh: Also there's a lot going on inside `getGcsUtil`. I suspect it's not cheap due to auth etc. Make this a `@transient lazy val`?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -0,0 +1,111 @@
...
+              val kv = it.next()
+              writer.put(kv._1.toString, kv._2.toString)
+            }
+            writer.close()
+            uri
+          }
+      }
+    }
+
+    /** Write the contents of this SCollection as a Sparkey file using the default uri. */
```
nevillelyh: nitpick: URI

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
andrewsmartin: True 👍

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -0,0 +1,111 @@
...
+    def asSparkey: SCollection[SparkeyUri] = {
+      val uid = UUID.randomUUID()
+      val uri = ScioUtil.tempLocation(self.context.options) + s"/sparkey-$uid"
+      logger.info(s"Sparkey URI: $uri")
```
nevillelyh: This will show up as "com.spotify... SparkeyPairSCollection - Sparkey URI" right? A bit unclear.
Move the logging to `def asSparkey(uri: SparkeyUri)` instead and make it "Saving as Sparkey: $uri"?

scio-extra/src/it/scala/com/spotify/scio/extra/sparkey/SparkeyIT.scala
```diff
@@ -0,0 +1,71 @@
...
+
+class SparkeyIT extends PipelineSpec {
+  import com.spotify.scio.extra.sparkey._
+
+  val sideData = Seq(("a", "1"), ("b", "2"), ("c", "3"))
+
+  "SCollection" should "support .asSparkeySideInput using GCS tempLocation" in {
+    runWithContext { sc =>
+      val tempLocation = sc.optionsAs[GcpOptions].getGcpTempLocation
+      sc.options.setTempLocation(tempLocation)
```
nevillelyh: Why is this necessary? Because `GcpOptions` creates a temp bucket silently but doesn't set `tempLocation`?

scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
```diff
@@ -0,0 +1,73 @@
...
+
+class SparkeyTest extends PipelineSpec {
+  import com.spotify.scio.extra.sparkey._
+
+  val sideData = Seq(("a", "1"), ("b", "2"), ("c", "3"))
+
+  "SCollection" should "support .asSparkey with default local file" in {
+    val tmpDir = Files.createTempDir().toString
+    val sc = ScioContext()
+    sc.options.setTempLocation(tmpDir)
```
nevillelyh: No need to set this right it should fallback to `java.io.tmpdir`?

nevillelyh:
Squashed and pushed to `apache-beam` branch.

PR #463, opened 2017-02-21
refactor(scio-core): Add setup/teardown to JobTest to be more extensible

### LINKED ISSUE #462: Visibility of classes & methods
We are using scio, and we are pretty happy about it. But we often have issue with going around it/extending it. 

Each time, it's the same issue, we would like to reuse some internal classes/metod but it's private. So we end up with putting our classes in `com.spotify.scio` package.

For exemple, we wanted to put our own Kryo Coders, so instead of subclassing `KryoAtomicCoder` we had to create our own `AtomicCoder` by copy/pasting it.
Same thing with `JobTest`, as we are not using `main` & command line arguments to launch our jobs, so we had to copy/paste `JobTest` to make it work, instead of just subclassing it. Only because `TestDataManager` is private.

We could do a PR for each of those cases, but it's mostly specific to our use case. We also understand the issue of keeping internals closed (easier to move forward).

Any idea on how to tackle this issue?

ElPicador:
Closes #462

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/testing/JobTest.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/testing/JobTest.scala
@@ -80,6 +80,8 @@ object JobTest {
                      outputs: Map[TestIO[_], SCollection[_] => Unit],
                      distCaches: Map[DistCacheIO[_], _]) {
 
+    val testId: String = newTestId(className)
+
     def args(newArgs: String*): Builder =
       this.copy(cmdlineArgs = (this.cmdlineArgs.toSeq ++ newArgs).toArray)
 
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +98,21 @@ object JobTest {
...
+    }
+
+    def teardown(testId: String): Unit = {
+      TestDataManager.unsetInput(testId)
+      TestDataManager.unsetOutput(testId)
+      TestDataManager.unsetDistCache(testId)
+    }
+
+    def run(): Unit = {
+      val testId = newTestId(className)
```
nevillelyh: make `testId` a member val of `Builder`?

scio-core/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +98,21 @@ object JobTest {
       this.copy(distCaches = this.distCaches + (key -> value))
     }
 
-    def run(): Unit = {
-      val testId = newTestId(className)
+    def setup(testId: String): Unit = {
```
nevillelyh: Name it `setUp` just to be consistent with JUnit, etc. And remove the `testId` arg.

scio-core/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +98,21 @@ object JobTest {
...
       TestDataManager.setInput(testId, new TestInput(inputs))
       TestDataManager.setOutput(testId, new TestOutput(outputs))
       TestDataManager.setDistCache(testId, new TestDistCache(distCaches))
+    }
+
+    def teardown(testId: String): Unit = {
```
nevillelyh: Use `tearDown` and remove `testId`?

ElPicador:
Should be fixed

PR #475, opened 2017-02-28
add wait until done method

rgruener:
@nevillelyh 

I probably shouldn't be throwing a regular exception, was hoping there was some exception in beam to use but couldn't find one. Any suggestions for that? I could make a custom one I guess.


nevillelyh:
`PipelineExecutionException` maybe?

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -76,6 +77,16 @@ class ScioResult private[scio] (val internal: PipelineResult,
     this
   }
 
+  /** Wait until the pipeline finishes with the State Done (as opposed to Cancelled). Throws
```
nevillelyh: Nitpick: can you use JavaDoc style as everywhere else? You can set it to default for scala in IntelliJ.
Something like:
```
/**
 * Blah blah ...
 * blah.
 */```

rgruener: Sure.

PR #478, opened 2017-03-05
KryoAtomicCoder and JIterableWrapperSerializer improvements

nevillelyh:


nevillelyh:
Don't think the buffered in/out stream & the JIterableWrapperSerializer changes actually improve performance in our particular case (GroupByKey with large value group) but I feel they're worth having anyway.

andrewsmartin:
👍

scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
```diff
@@ -26,22 +26,46 @@ import com.twitter.chill.KSerializer
 
 import scala.collection.JavaConverters._
 
+/**
+ * Based on [[org.apache.beam.sdk.coders.IterableLikeCoder]] and
+ * [[org.apache.beam.sdk.util.BufferedElementCountingOutputStream]].
+ */
 private class JIterableWrapperSerializer[T] extends KSerializer[Iterable[T]] {
 
+  private val bufferSize = 64 * 1024
```
ravwojdyla: Why?

scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
```diff
@@ -26,22 +26,46 @@ import com.twitter.chill.KSerializer
...
   }
 
   override def read(kser: Kryo, in: Input, cls: Class[Iterable[T]]): Iterable[T] = {
     val list = Lists.newArrayList[T]
-    while (in.readBoolean()) {
-      val item = kser.readClassAndObject(in).asInstanceOf[T]
-      list.add(item)
+    var count = in.readInt()
+    while (count > 0) {
+      var i = 0
```
ravwojdyla: You could reuse `count` instead of another var `i`.

nevillelyh: Yes but it's not gonna make any difference performance wise and `i` feels more natural?

ravwojdyla: Maybe it would feel less natural, because you named the original variable `count`? I got confused here.

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -104,34 +106,156 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
...
 }
+
+/**
+ * Buffered output stream that adds length prefix to each buffer block. Useful for large objects
+ * or collections. Based on [[org.apache.beam.sdk.util.BufferedElementCountingOutputStream]].
+ */
+private class BufferedPrefixOutputStream(private val os: OutputStream)
+  extends OutputStream {
+
+  private val buffer = ByteBuffer.allocate(64 * 1024)
```
ravwojdyla: Repeating 64K in this class twice, does it make sense to refactor this?

ravwojdyla:
Are there any benchmark results for this, both for the sampling and buffering?

scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
nevillelyh: It'd look like this by reusing count. Not sure about it.
```scala
while (count > 0) {
      while (count > 0) {
```

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
nevillelyh: You mean once in input and once in output? I'd have to add an object for it. Again not sure how much value we get since these are not exposed or configurable. It's easier to refactor when the complexity increases and there's a need for it.

nevillelyh:
`GroupByKey` with a single very large (~400MB) key finished in ~10min with this fix and couldn't finish without it. The buffered stream changes don't affect performance much in this case but would if there are many medium sized (larger than buffer) elements as it saves us overhead from GC (copying underlying buffer in `ByteArrayOutputStream`) or 1 byte (boolean) per element in `JIterableWrapperSerializer`.

nevillelyh:
Tested with `hashCode % 1000` as key and either of the last 2 commits (buffered IO streams and counting serializer) improves performance a little bit. Probably more with real workload.

ravwojdyla:
well can't see immediate reason for that extra complexity, but if you believe it is worth it, you have my 👍

Issue #468, opened 2017-02-23
java.lang.BootstrapMethodError: too many bootstrap method arguments


nevillelyh:
This happens on Scala 2.12, from `MultiJoin.scala`:
```
[info]   Cause: java.lang.BootstrapMethodError: too many bootstrap method arguments
[info]   at java.lang.invoke.CallSite.makeSite(CallSite.java:320)
[info]   at java.lang.invoke.MethodHandleNatives.linkCallSiteImpl(MethodHandleNatives.java:307)
[info]   at java.lang.invoke.MethodHandleNatives.linkCallSite(MethodHandleNatives.java:297)
[info]   at com.spotify.scio.util.MultiJoin.$deserializeLambda$(MultiJoin.scala)
[info]   at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[info]   at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[info]   at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[info]   at java.lang.reflect.Method.invoke(Method.java:498)
[info]   at java.lang.invoke.SerializedLambda.readResolve(SerializedLambda.java:230)
[info]   at sun.reflect.GeneratedMethodAccessor23.invoke(Unknown Source)
```

Reducing the multi-join arity fixes the problem.

retronym:
Tracked as https://issues.scala-lang.org/browse/SI-10232

nevillelyh:
Verified that this is fixed by Scala 2.12.3

PR #482, opened 2017-03-10
support Scala 2.12

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/util/ChillClosureCleaner.scala
```diff
@@ -0,0 +1,273 @@
+// scalastyle:off header.matches
```
andrewsmartin: Will this be fixed upstream as well? I don't see an issue tracking it anywhere

scripts/multijoin.py
```diff
@@ -198,7 +198,7 @@ def main(out):
           def toOptions[T](xs: Iterator[T]): Iterator[Option[T]] = if (xs.isEmpty) Iterator(None) else xs.map(Option(_))
         ''').replace('  # NOQA', '').lstrip('\n')
 
-    N = 22
+    N = 11
```
andrewsmartin: I forget, what was the reason for this again?

scio-core/src/main/scala/com/spotify/scio/util/ChillClosureCleaner.scala
nevillelyh: I will.

scripts/multijoin.py
nevillelyh: #468

nevillelyh:
DO NOT MERGE. It's not completely fix. Not working for `object.MODULE$` case.

PR #483, opened 2017-03-13
Support Array[Byte] IO in Sparkey package.

andrewsmartin:
@nevillelyh @rgruener. Note that this does not support auto-encoding from any type, it is on the user to convert the `SCollection` to `(String, String)` or `(Array[Byte], Array[Byte])` before calling `.asSparkey`.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
@@ -183,13 +183,13 @@ package object sparkey {
     def put(w: SparkeyWriter, key: K, value: V)
   }
 
-  implicit object StringSparkeyWritable extends SparkeyWritable[String, String] {
-    override def put(w: SparkeyWriter, key: String, value: String): Unit =
+  implicit val stringSparkeyWritable = new SparkeyWritable[String, String] {
+    def put(w: SparkeyWriter, key: String, value: String): Unit =
       w.put(key, value)
   }
 
-  implicit object ByteArraySparkeyWritable extends SparkeyWritable[Array[Byte], Array[Byte]] {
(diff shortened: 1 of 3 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -178,4 +179,18 @@ package object sparkey {
       context.sideInput(view).getReader
   }
 
+  sealed trait SparkeyWritable[K, V] extends Serializable {
+    def put(w: SparkeyWriter, key: K, value: V)
+  }
+
+  implicit object StringSparkeyWritable extends SparkeyWritable[String, String] {
```
nevillelyh: make them `implicit val`s instead? Both are interchangeable but there's no `implicit object` anywhere else in the repo so consistency = easy grepping.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -128,10 +129,10 @@ package object sparkey {
      *
      * @return A singleton SCollection containing the [[SparkeyUri]] of the saved files.
      */
-    def asSparkey: SCollection[SparkeyUri] = {
+    def asSparkey(implicit w: SparkeyWritable[K, V]): SCollection[SparkeyUri] = {
       val uuid = UUID.randomUUID()
       val basePath = self.context.options.getTempLocation + s"/sparkey-$uuid"
-      this.asSparkey(basePath)
+      this.asSparkey(basePath)(w)
```
nevillelyh: I'm surprised that this is needed. I thought `implicit` params should propagate automatically? Same with `def asSparkeySideInput` down there.

PR #484, opened 2017-03-14
CircleCI build improvements

andrewsmartin:
The changes here help reduce the build time by caching dependencies and build streams. For this to work properly, each sbt task must have `coverageEnabled := true` (set via `sbt coverage`) so that the settings for `libraryDependencies`, `compile` etc remain unchanged and can hence use the cache. 

Also, f1d607d may potentially fix the problem with coverage reports sometimes being incorrect.


andrewsmartin:
Well I guess this didn't fix the coverage report.

scripts/circleci_test.sh
```diff
@@ -20,9 +20,11 @@ DIR_OF_SCRIPT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 if [ "$CI_PULL_REQUEST" = "" ]; then
   echo "Running test for branch: $CIRCLE_BRANCH"
   openssl aes-256-cbc -d -in "$DIR_OF_SCRIPT/data-integration-test-2210ed0f609b.json.enc" -out "$DIR_OF_SCRIPT/data-integration-test-2210ed0f609b.json" -k $ENCRYPTKEY
-  "$DIR_OF_SCRIPT/circleci_parallel_run.sh" 'sbt ++$CI_SCALA_VERSION -Dbigquery.project=data-integration-test '"-Dbigquery.secret=$DIR_OF_SCRIPT/data-integration-test-2210ed0f609b.json"' clean scalastyle coverage test it:test coverageReport coverageAggregate'
+  "$DIR_OF_SCRIPT/circleci_parallel_run.sh" 'sbt ++$CI_SCALA_VERSION -Dbigquery.project=data-integration-test '"-Dbigquery.secret=$DIR_OF_SCRIPT/data-integration-test-2210ed0f609b.json"' scalastyle coverage test it:test coverageReport'
 else
   echo "Running test for PR: $CI_PULL_REQUEST"
   "$DIR_OF_SCRIPT/gen_tornado_schema.sh"
```
ravwojdyla: Is this needed if it is now part of `circleci_compile.sh`?

andrewsmartin: Nope, not needed, good catch.

ravwojdyla:
👍

andrewsmartin:
By the way - this seems to shave a couple mins off PR builds and maybe 4-5 mins on other builds. So it's not a huge improvement but helps a little. I think once we migrate to Circle 2.0 we should expect to see a larger improvement.

PR #486, opened 2017-03-14
refactor remote file handline

nevillelyh:


scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -0,0 +1,320 @@
+/*
```
andrewsmartin: Why Java?

nevillelyh: It's mostly using Java APIs. And this class maybe reused in a pure Java DF pipeline.

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -0,0 +1,320 @@
...
+            shouldDownload = false;
+          } else {
+            LOG.warn("Destination exists with wrong size. {} [{}] -> {} [{}]",
+                src, srcSize, dst, dstSize);
+            Files.delete(dst);
+          }
+        }
+        if (shouldDownload) {
+          copyToLocal(src, dst);
+          LOG.info("Downloaded {} -> {} [{}]", src, dst, srcSize);
```
ravwojdyla: `"Downloaded {} -> {} [{}] bytes"`?

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -0,0 +1,320 @@
...
+        // Remote URI
+        long srcSize = getRemoteSize(src);
+        boolean shouldDownload = true;
+        if (Files.exists(dst)) {
+          long dstSize = Files.size(dst);
+          if (srcSize == dstSize) {
+            LOG.info("URI {} already downloaded", src);
+            shouldDownload = false;
+          } else {
+            LOG.warn("Destination exists with wrong size. {} [{}] -> {} [{}]",
```
ravwojdyla: "[{}] bytes"

ravwojdyla:
Tests are missing for the new logic.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -49,55 +44,43 @@ trait SparkeyUri extends Serializable {
...
+      .exists(e => rfu.remoteExists(new URI(basePath + e)))
 }
 
 private[sparkey] class SparkeyWriter(val uri: SparkeyUri) {
 
   private val localFile = uri match {
-    case u: LocalSparkeyUri => u.toString
-    case u: GcsSparkeyUri => u.localBasePath
+    case u: LocalSparkeyUri => u.basePath
+    case _: RemoteSparkeyUri => Files.createTempDirectory("sparkey").resolve("data").toString
```
andrewsmartin: So the hash prefix is no longer needed right? This should be unique every time it's called / internally uses some kind of hash in the name?

andrewsmartin: Yeah uses random numbers 👍

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
ravwojdyla: `B` sounds good and simple.

PR #491, opened 2017-03-16
Support BQ table desc

ravwojdyla:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -67,19 +67,33 @@ private[types] object TypeProvider {
     schemaToType(c)(schema, annottees, traits, overrides)
   }
 
+  private def getTableDescription(c: blackbox.Context)(tree: Seq[c.universe.Tree])
+  : List[c.universe.Tree] = {
+    import c.universe._
+    tree.head.asInstanceOf[ClassDef].mods.annotations
+      .filter(_.children.head.toString() == "new description")
```
nevillelyh: Is this safe enough? What if there's an annotation like `description2`? Use a `"\>"` regex?

nevillelyh:
👍

Issue #495, opened 2017-03-17
Port to latest cloud-bigtable


nevillelyh:
https://mvnrepository.com/artifact/com.google.cloud.bigtable/bigtable-client-core

It's 0.9.2 in Beam 0.6.0 but 0.9.5.1 available.
https://github.com/apache/beam/blob/release-0.6.0/sdks/java/io/google-cloud-platform/pom.xml#L35

nevillelyh:
@andrewsmartin FYI

andrewsmartin:
Will submit pr to beam. In the meantime I will try overriding in Scio and if that works, that's good enough until beam upgrades.

andrewsmartin:
https://github.com/apache/beam/pull/2265/files

nevillelyh:
I managed to run the upstream integration test after editing the hard coded `tableId` in `BigtableReadIT` and `projectId` + `instanceId` in `BigtableTestOptions`.

`mvn -DbeamTestPipelineOptions='["--project=scio-playground"]' -Dtest=org.apache.beam.sdk.io.gcp.bigtable.BigtableReadIT -pl sdks/java/io/google-cloud-platform test`

And netty seems to be the culprit:
```
-      <version>1.1.33.Fork18</version>
+      <version>1.1.33.Fork19</version>
```

andrewsmartin:
The latest bigtable-client uses a class from `netty-common:4.1.6.Final` that didn't exist in `4.1.3.Final`. Bumped the upstream version: https://github.com/apache/beam/pull/2265/commits/54393e1f707d394b4006f15247a2ff4c39eed23c

PR #497, opened 2017-03-17
Monkey-patch BigtableIO for 0.9.5.1 #495

andrewsmartin:


nevillelyh:
Rename `BigtableIO` to `PatchedBigtableIO` in case any class loading issue?
Also does this work with the test job?

andrewsmartin:
Decided against renaming -> `PatchedBigqueryIO` because this class passes `this` references to other API calls which expect `BigqueryIO`.

nevillelyh:
:+1:

regadas:
👍

PR #503, opened 2017-03-21
TestStream and LeaderBoardTest

nevillelyh:


nevillelyh:
@ravwojdyla @andrewsmartin this is a direct port of Beam stream tests but is not idiomatic ScalaTest.
Get it in first and think about better API later?

PR #505, opened 2017-03-23
Revert "use buffered prefix count in JIterableWrapperSerializer"

ravwojdyla:
This reverts commit 611d375aa73e3cb3eb138c6ed33bfa40d486ab6c.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
@@ -26,46 +26,22 @@ import com.twitter.chill.KSerializer
 
 import scala.collection.JavaConverters._
 
-/**
- * Based on [[org.apache.beam.sdk.coders.IterableLikeCoder]] and
- * [[org.apache.beam.sdk.util.BufferedElementCountingOutputStream]].
- */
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
```diff
@@ -26,46 +26,22 @@ import com.twitter.chill.KSerializer
...
 private class JIterableWrapperSerializer[T] extends KSerializer[Iterable[T]] {
 
-  private val bufferSize = 64 * 1024
-
   override def write(kser: Kryo, out: Output, obj: Iterable[T]): Unit = {
-    val buffer = new Output(bufferSize)
-    var count = 0
     val i = obj.iterator
     while (i.hasNext) {
-      if (buffer.position() >= bufferSize) {
```
ravwojdyla: This does not guarantees that there will be enough space to put bytes from `writeClassAndObject`, which would result in `KryoException` "Buffer overflow".

After thinking about other solutions - like keeping second buffer, trying to estimate size based on previous data etc, I think the original implementation tho not optimal - works and is simple.

ravwojdyla:
Doubling buffer size won't work because the buffer might be partially written too thus "corrupted", also if we keep our own buffer we would need to add option for user to specify the size in case of single objects bigger than buffer (think Iterable of large CMSs).

PR #507, opened 2017-03-24
JIterable Serialize optimization

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
```diff
@@ -19,29 +19,76 @@ package com.spotify.scio.coders
...
+      }
+      val toSer = i.next()
+      var retry = true
+      while (retry) {
+        try {
+          kser.writeClassAndObject(buffer, toSer)
+          retry = false
+          count += 1
+        } catch {
+          case e: KryoException if e.getMessage.startsWith("Buffer overflow") =>
```
nevillelyh: This will happen for every element that overflows if the `Iterable` contains many of them. Double the buffer size (until a max of say 1024K) in that case?

scio-test/src/test/scala/com/spotify/scio/coders/CoderTestUtils.scala
```diff
@@ -36,4 +45,13 @@ object CoderTestUtils {
     result == value
   }
 
+  def testRoundTrip[T: ClassTag](ser: Serializer[T],
+                                 elems: T,
+                                 k: Kryo = KryoSerializer.registered.newKryo(),
+                                 bufferSize: Int = 1024): Unit = {
+    val o = ByteBuffer.allocate(bufferSize).array()
+    ser.write(k, new Output(o), elems)
+    elems === ser.read(k, new Input(o), ScioUtil.classOf[T])
```
nevillelyh: Is `===` equivalent to `a should equal b`? I prefer to spell it out for clarity.
Also this is technically a `SerializerTestUtil` and not for `Coder[T]`. Since it's used in only one class, keep it in `JIterableWrapperSerializerTest` until there's a need to split?

scio-test/src/test/scala/com/spotify/scio/coders/JIterableWrapperSerializerTest.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  "JIterableWrapperSerializer" should "cope with the internal buffer overflow" in {
+    val ser = new JIterableWrapperSerializer[String](128)
+    val input = List.fill(100)("foo")
+    CoderTestUtils.testRoundTrip(ser, input)
+  }
+
+  it should "be able to serialize object larger than the internal buffer" in {
+    val ser = new JIterableWrapperSerializer[String](10)
+    val input = Seq("o" * 10)
```
nevillelyh: `"o" * 10` = 10 bytes? Make it `>> 10` so it's more obvious?

ravwojdyla: Nope - it does not have to be 10 bytes - just something (single object) that will overflow the 10 bytes buffer. I don't see how `>> 10` will make it more obvious, also the description of the test is pretty descriptive I think.

scio-test/src/test/scala/com/spotify/scio/coders/CoderTestUtils.scala
ravwojdyla: 👍

scio-core/src/main/scala/com/spotify/scio/coders/JIterableWrapperSerializer.scala
ravwojdyla: 👍

PR #509, opened 2017-03-28
add some extra DoFns

nevillelyh:


nevillelyh:
@ravwojdyla this is ready, take a look please?

rgruener:
Adding an example of usage in scio-examples would definitely be helpful

nevillelyh:
@rgruener they're in a WIP branch now and i'll add once cleaned up.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
```diff
@@ -0,0 +1,104 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.UUID;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * A {@link DoFn} that manages an external resource.
+ */
+public abstract class DoFnWithResource<InputT, OutputT, ResourceT> extends DoFn<InputT, OutputT> {
```
ravwojdyla: This looks like a rabbit hole and bad pattern - given the properties (fault tolerance) of DoFns. This problem should be solved by State API.

nevillelyh: Not quite the same though, state one is intended  for value data and this is for resource management. Also state doesn't have fine-grained sharing model like what we need for async clients.

ravwojdyla: Yea - but what if people use this to keep state, which afaiu this API it allows, but should not.

nevillelyh: They can use the vanilla `DoFn<I, O>` too since it can just be a member field. This just abstract away the error prone process of managing shared resources. Things in `scio-extra` are considered experimental and use at your own risk. We don't publicize them and usually only those who know what they're doing use them.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/FileDoFn.java
```diff
@@ -0,0 +1,109 @@
...
+  public FileDoFn(SerializableFunction<Path, OutputT> fn, int batchSize, boolean keep) {
+    this.fn = fn;
+    this.batch = Lists.newArrayList();
+    this.batchSize = batchSize;
+    this.keep = keep;
+  }
+
+  @StartBundle
+  public void startBundle(Context c) {
+    this.remoteFileUtil = RemoteFileUtil.create(c.getPipelineOptions());
```
ravwojdyla: Instead of doing this in `@StartBundle` you might want to move it to `@Setup` - in case of streaming bundle might be very short thus this might be expensive. In `@Setup` you won't have access to options and you will probably need to add `RemoteFileUtil.createFromDefaultApplicationCredentials` kinda thing, which is available on VMs.

nevillelyh: 👍

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
ravwojdyla: Yes - they can, and maybe they should, I would disagree this is a common problem. That said maybe instead of allowing any resources here, this API should restrict to only read only resources?

nevillelyh: How would you express a read only resource? Also it doesn't necessarily mean read-only right? An HTTP client can still make requests and cause side effects.

ravwojdyla: HTTP client does not make a change in its own state. So the resource should be stateless.

nevillelyh: How would you express a read only resource then? `ResourceT` can be a mutable `Map` that changes state. An HTTP client may change internal state (pending requests, connection pool etc.) but won't affect element processing.

`createResource()` will create the resource once later to be retrieved via `getResource`, which somewhat expresses the read-only-ness. But other than that we can't really express it via the API or type signature?

scio-extra/src/main/java/com/spotify/scio/extra/transforms/FileDoFn.java
nevillelyh: Actually this might cause problem later when we wanna support different runners. Also assuming input files are not tiny (MBs for audio, GBs for Avro parts), the overhead of creating `RemoteFileUtil` should be negligible?

scio-extra/src/main/java/com/spotify/scio/extra/transforms/BaseAsyncDoFn.java
```diff
@@ -0,0 +1,98 @@
...
+    FutureT future = addCallback(processElement(c.element()), r -> {
+      results.add(r);
+      futures.remove(uuid);
+      return null;
+    }, t -> {
+      errors.add(t);
+      futures.remove(uuid);
+      return null;
+    });
+    futures.put(uuid, future);
```
ravwojdyla: Just checking - is there possible race condition here if [remove](https://github.com/spotify/scio/pull/509/files#diff-090a5f98b929a7979f715501ada577edR80) gets executed before this `put`?

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
```diff
@@ -0,0 +1,144 @@
...
+   * lines.apply(ParDo.of(f1));
+   * lines.apply(ParDo.of(f2));
+   * </code></pre>
+   */
+  public enum ResourceType {
+    /**
+     * One instance of the resource per sub-class.
+     *
+     * All instances of the same class in a remote JVM share a single `ResourceT`, e.g. all clones
+     * of `f1` and `f2`. This is useful for sharing resources globally, e.g. static look up table,
```
ravwojdyla: Would remove `globally` it's misleading - maybe be explicit and say `sharing resources with the same JVM`.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/BaseAsyncDoFn.java
nevillelyh: Possible but shouldn't cause any problem. `remove` will return `null` if key is not found. We wait for all futures in `finishBundle`, including those that were `put` but never removed.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/FileDoFn.java
ravwojdyla: This is with the assumption that files are small, so the user would have to read this code to understand this limitation - how about being explicit instead. About runners afaiu `RemoteFileUtil` currently supports only GCS anyway, it's already pretty specific.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
nevillelyh: Fixed.

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
```diff
@@ -0,0 +1,94 @@
...
+   */
+  implicit class PipeSCollection(val self: SCollection[String]) extends AnyVal {
+
+    /**
+     * Pipe elements through an external command via StdIn & StdOut.
+     * @param command the command to call
+     * @param environment environment variables
+     * @param dir the working directory of the sub-process
+     */
+    def pipe(command: String,
```
ravwojdyla: If not through beam mechanism, maybe download from *FS?

scio-extra/src/test/scala/com/spotify/scio/extra/transforms/DoFnWithResourceTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    c2.getResource shouldNot equal (c3.getResource)
+    c2.getResource shouldNot equal (c4.getResource)
+
+    c3.getResource shouldNot equal (c1.getResource)
+    c3.getResource shouldNot equal (c2.getResource)
+    c3.getResource shouldNot equal (c4.getResource)
+
+    DoFnTester.of(c1).processBundle("a", "b", "c").asScala should equal (Seq("A", "B", "C"))
+  }
+
```
ravwojdyla: -

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
nevillelyh: i wanna defer this for now. We could have a dummy transform (that doesn't process elements) like `p.runOnWorkers("apt-get install ...")` or `p.runOnWorkers("wget ...")` but that doesn't feel super natural.

scio-extra/src/test/scala/com/spotify/scio/extra/transforms/PipeDoFnTest.scala
```diff
@@ -0,0 +1,81 @@
...
+  it should "support environment" in {
+    val tmpDir = Files.createTempDirectory("pipedofn")
+    val file = tmpDir.resolve("tr.sh")
+    GFiles.write("tr $PARG1 $PARG2", file.toFile, Charsets.UTF_8)
+    val env = Map("PARG1" -> "[:lower:]", "PARG2" -> "[:upper:]")
+
+    runWithContext { sc =>
+      val d1 = new PipeDoFn(s"bash $file", env.asJava, null)
+      val d2 = new PipeDoFn(Array("bash", file.toString), env.asJava, null)
+      val p1 =  sc.parallelize(input).applyTransform(ParDo.of(d1))
```
ravwojdyla: double space before `sc`

ravwojdyla: also below

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
ravwojdyla: How would the users install their own executables in this use case?

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
```diff
@@ -0,0 +1,144 @@
...
+     * `f1` and `f2` has its own copy. This is useful for thread local resources, e.g. resources
+     * that are not thread-safe.
+     */
+    PER_CLONE
+  }
+
+  /**
+   * Get resource type.
+   */
+  public abstract ResourceType getResourceType();
```
ravwojdyla: Why not an argument of a CTOR?

nevillelyh: No particular reason but since the class is abstract and user will have to implement some methods anyway. It's either that or implement a ctor in the sub-class and call parent ctor.

ravwojdyla: Wouldn't you usually make such case a CTOR argument?

nevillelyh: :+1:

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
ravwojdyla: Can't think of an easy way to express this.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/BaseAsyncDoFn.java
ravwojdyla: Could you please add it here as a comment, just so future generations won't stumble up this one with wonder in their mind, and fire in their eyes.

scio-extra/src/main/java/com/spotify/scio/extra/transforms/DoFnWithResource.java
nevillelyh: I remembered why. `BaseAsyncDoFn` extends `DoFnWithResources` and has 2 abstract sub-classes `GuavaAsyncDoFn` and `JavaAsyncDoFn`. As an abstract method, only the final concrete class needs to implement it but as a ctor argument I have to implement it in all 3 abstract sub-classes and pass it alone.

ravwojdyla: I see, one downside of the method is that the user might code some weird logic to conditionally return different type - which this API does not prevent. Is this something that this should allow for - if not - honestly would prefer simple ctor argument and copy pass it along in the abstract subclasses - the intention is much more clear.

nevillelyh: Well these are abstract classes to be extended by user so there's no way to prevent bad code else where. When you extend it in a concrete class, auto-complete fill in the 2 method stubs and comments look pretty clear to me:

```scala
class ADoFn extends DoFnWithResource[String, String, TestResource] {
  /**
   * Get resource type.
   */
  override def getResourceType: ResourceType = ???

  /**
   * Create resource.
   *
   * {@link DoFnWithResource#getResourceType()} determines how many times this is called.
   */
  override def createResource(): TestResource = ???
}
```

ravwojdyla: ctor would look something like this (?):

```
class ADoFn extends DoFnWithResource[String, String, TestResource](ResourceType.PER_CLASS) {

  /**
   * Create resource.
   *
   */
  override def createResource(): TestResource = ???
}
```

If you think this is less intuitive - let's keep yours.

nevillelyh: Yeah no prob. I'm not quite happy with all the pieces yet so it'll take a while to get them in shape.

PR #510, opened 2017-03-28
Bump Avro version to 1.8.1

SamSchlegel:
This matches the version upstream in beam. It looks like `org.apache.avro.test._` was removed or isn't generating correctly in 1.8.1, specifically `TestRecord` which is used in `HdfsTapTest`. Want to have the tests run so I can see what else broke, since they OOM on my machine


SamSchlegel:
Will just spin up a quick GCE instance to run the tests.

SamSchlegel:
Looks like https://issues.apache.org/jira/browse/AVRO-1755 removed the avro-idl:test dependency from avro-mapred. Should be good to go now

build.sbt
```diff
@@ -312,6 +312,7 @@ lazy val scioHdfs: Project = Project(
   description := "Scio add-on for HDFS",
   libraryDependencies ++= Seq(
     "org.apache.avro" % "avro-mapred" % avroVersion classifier("hadoop2"),
+    "org.apache.avro" % "avro-ipc" % avroVersion classifier("tests"),
```
nevillelyh: This should probably be a `"test"` scope?

nevillelyh:
Thanks!

PR #511, opened 2017-03-31
Proposing additional SCollection Functions(flatten, asKey, asKeyBy)

richwhitjr:
These are somewhat lifted from the Scalding API.  There are three proposed new api functions in this PR.

1. flatten 
  a. Given a SCollection of Transversable Elements this will flatten it out to a SCollection of those elements.
2. asKey
  a. Given a SCollection of some element this lifts the elements to the Keys and adds a Unit as the values.
3. asKeyBy
  a. Similar to above but you can provide a function that extracts the key.

My only worry for the last two is that they can be confused with the current function keyBy.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -311,6 +311,13 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     this.parDo(Functions.flatMapFn(f))
 
   /**
+    * Return a new SCollection of Transversables by flatterning them.
+    * Similar to collection.flatMap(identity)
+    */
+  def flatten[U: ClassTag](implicit ev: T <:< TraversableOnce[U]): SCollection[U] =
+    this.parDo(Functions.flatMapFn(_.asInstanceOf[TraversableOnce[U]]))
+
+  /**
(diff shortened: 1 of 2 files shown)


nevillelyh:
I prefer to not add syntactic sugar for things that can be easily done with a one-liner (< 20 chars really). Scalding has too many of those and caused some confusion for users.
`SCollection` also has many window/timestamp related methods so its tab-completion list is quite bloated already.
Does that make sense?

- `flatten` = `flatMap(identity)`
- `asKey` = `map((_, Unit))`
- `asKeyBy` = `map((f(_), Unit))`

richwhitjr:
Makes sense.  Do you think flatten would be a good addition though?  flatMap(identity) may be confusing for people not from a functional background.  

I was a little iffy on the other two.

nevillelyh:
TBH a bit iffy on `flatten` too, neither spark nor flink has it, and I'm trying to be more consistent with them.

richwhitjr:
OK I can go ahead and close this.  May be worth seeing if flatten is a request feature in the future.  I tend to prefer the Scalding API to Spark but I have heard many people mention that Scalding is overwhelming for beginners so maybe a simple API is the way to go.

nevillelyh:
Yeah methods with similar name and signature usually cause confusion. The one we had to explain most is `groupBy` vs `groupByKey` :)
It's usually quick & easy to just ask in slack when something is missing.

PR #514, opened 2017-04-04
adding numMaxRecords and maxReadTime to use with DirectRunner

andrisnoko:
DirectRunner throws a NPE if numMaxRecords is not specified.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -49,7 +49,7 @@ import org.apache.beam.sdk.transforms.DoFn.ProcessElement
 import org.apache.beam.sdk.transforms.{Create, DoFn, PTransform}
 import org.apache.beam.sdk.values._
 import org.apache.beam.sdk.{Pipeline, io => gio}
-import org.joda.time.Instant
+import org.joda.time.{Duration, Instant}
 import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


nevillelyh:
LGTM but these 2 settings seem useful to `DirectRunner` only and creates bounded source instead of unbounded? Not sure if it's something we want to open up. Is this to allow users to test with real pubsub locally?

andrisnoko:
@nevillelyh:  yeah this is for testing pubsub locally.

nevillelyh:
I would prefer not to include these since they might make public API and common use cases confusing. Plus this is easily achievable with `sc.customInput()`. Should we add a note in FAQ instead?

andrisnoko:
Yes, `sc.customInput()` should work for such cases, however, I ran into issues with it in scio 0.3

```scala
   sc.customInput(
      PubsubIO.Read
       .topic("projects/data-university/topics/data-university")
      .idLabel("id")
        .timestampLabel("ts")
        .withCoder(StringUtf8Coder.of()).maxNumRecords(50))```

**Comment by andrisnoko** [comment]
Thanks @nevillelyh, the name was missing from `sc.customInput()`. We really don't these changes. I will go ahead and close this PR.

```scala
sc.customInput("ReadFromPubsub",
      PubsubIO.read()
        .topic("projects/data-university/topics/data-university")
        .idLabel("id")
        .timestampLabel("ts")
        .withCoder(StringUtf8Coder.of()).maxNumRecords(50))
```

nevillelyh:
Added a FAQ item here:
https://github.com/spotify/scio/wiki/FAQ#how-do-i-read-pubsub-input-in-a-local-pipeline

vertexclique:
Currently, Since Beam API doesn't have those methods, this example doesn't work anymore.
Any alternative to achieve this?

PR #515, opened 2017-04-04
Make sure run is invoked on JobTest

ravwojdyla:


scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -65,26 +65,32 @@ import scala.util.control.NonFatal
  */
 object JobTest {
 
-  case class Builder(className: String, cmdlineArgs: Array[String],
+  private case class BuilderState(className: String, cmdlineArgs: Array[String],
                      inputs: Map[TestIO[_], Iterable[_]],
                      outputs: Map[TestIO[_], SCollection[_] => Unit],
-                     distCaches: Map[DistCacheIO[_], _]) {
+                     distCaches: Map[DistCacheIO[_], _], wasRunInvoked: Boolean = false)
```
nevillelyh: Why adding this class? To make internal state like `inputs`, `outputs` invisible to users?
Also align parameters?

scio-test/src/main/scala/com/spotify/scio/testing/EngineerProofJobTest.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.scalatest.{BeforeAndAfterEach, Suite}
+import com.spotify.scio.testing.{JobTest => InnerJobTest}
+
+import scala.collection.mutable.ArrayBuffer
+import scala.reflect.ClassTag
+
+/**
+ * Trait which throws if JobTest was constructed but not executed.
+ */
+trait EngineerProofJobTest extends BeforeAndAfterEach { this: Suite =>
```
nevillelyh: Maybe a slightly more PC name? `RunEnforcementJobTest`?

ravwojdyla: 👍

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -148,16 +158,18 @@ object JobTest {
       tearDown()
     }
 
+    override def toString: String = state.toString
```
nevillelyh: This will print all members of `BuilderState` right? Make it more readable like `JobTest[ClassName]`?

nevillelyh:
Also add a test in `JobTestTest`?

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
ravwojdyla: So the way Builder worked here was to create new instances of Builder, thus in the `EngineerProofJobTest` we would have access only to the initial Builder instances not the one that `run` is invoked on. I don't really like this solution much, essentially what I need is `pointer` or `container` that will store the eventual state of the builder.

nevillelyh: Yeah but no way around it. It's hidden from the user so not a big deal.

scio-test/src/main/scala/com/spotify/scio/testing/RunEnforcementJobTest.scala
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Trait that enforces [[JobTest.Builder.run()]] is called.
+ */
+trait RunEnforcementJobTest extends BeforeAndAfterEach { this: Suite =>
+
+  private val tests = ArrayBuffer.empty[InnerJobTest.Builder]
+
+  // scalastyle:off method.name
+  def JobTest[T: ClassTag]: InnerJobTest.Builder = {
```
nevillelyh: There should be a `JobTest(className: String)` version as well?

scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
```diff
@@ -497,7 +497,7 @@ class SCollectionTest extends PipelineSpec {
         r should containInAnyOrder (Seq(1, 2, 3))
       }
     }
-    stdOutMock.message should contain theSameElementsAs Seq("1", "2", "3")
+    stdOutMock.message should contain allElementsOf Seq("1", "2", "3")
```
nevillelyh: Just curious, what's the differences between these 2?

ravwojdyla: `theSameElementsAs` has to contain all and **only** the elements of expected, while `allElementsOf` all the elements of expected.

nevillelyh:
:+1:

PR #523, opened 2017-04-05
Closes #522. Escape scala reserved words

### LINKED ISSUE #522: BQ should escape reserved words in toPrettyString and plugin dumps
Causes spotify/scio-idea-plugin#5

Example:

```
scio> TT3.toPrettyString()
res9: String =
@BigQueryType.toTable
case class TT3(foo: Option[String], bar: Option[Double], type: Option[String], match: Option[Long], class: Option[String], ola: Option[String])

```

ravwojdyla:


nevillelyh:
:+1:

Issue #527, opened 2017-04-07
ScioContext.close() incompatible with DataflowTemplateJob


jobegrabber:
If you run a `DataflowTemplateJob`, i.e. run a pipeline on Dataflow and `setTemplateLocation`, calling `close()` on the `ScioContext`  will throw an `UnsupportedOperationException`: `The result of template creation should not be used.`.

From the stacktrace I infer this is because the `ScioResult` [that gets built](https://github.com/spotify/scio/blob/v0.3.0-beta3/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L341) queries the pipeline's inner state [here](https://github.com/spotify/scio/blob/v0.3.0-beta3/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala#L96) - `DataflowTemplateJob.getState()` will then [throw aforementioned Exception](https://github.com/apache/beam/blob/v0.6.0/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/DataflowTemplateJob.java#L67). 

This was observed on Scio 0.3.0 Beta 2, though I don't see anything that's different to Beta 3 in this regard.

nevillelyh:
Thanks for reporting. Can you paste some minimal snippet of template job usage please?

jobegrabber:
```
def main(cmdlineArgs: Array[String]): Unit = {
    val (sc, args) = ContextAndArgs(cmdlineArgs)
    SetupDataflowJob(sc)

    sc.customInput("ConsumeFromKafka", kafkaReader)
      .withName("DropKeys").map { _.getValue }
      // Some parsing and filtering
      .withName("CreateBigtableMutations").createBigtableMutations
      .saveAsBigtable()

    sc.close()
  }

def SetupDataflowJob(sc: ScioContext) = {
    val opts = sc.optionsAs[DataflowPipelineOptions]
    opts setRunner classOf[DataflowRunner]

    opts setTemplateLocation config("dataflow.template-location")
    // Some other options set (numWorkers, machineType etc.)
    opts setStreaming true
  }
```

Will throw:
```
Dataflow SDK version: 0.6.0
java.lang.UnsupportedOperationException: The result of template creation should not be used.
        at org.apache.beam.runners.dataflow.util.DataflowTemplateJob.getState(DataflowTemplateJob.java:68)
        at com.spotify.scio.ScioResult.state(ScioResult.scala:96)
[...]

PR #530, opened 2017-04-11
Use UNKNOWN state when not supported, eg in DataflowTemplateJob #527.

andrewsmartin:


nevillelyh:
LGTM. @jobegrabber can you confirm that this addresses your issue?

jobegrabber:
Yes, this addresses the issue I had.
Though as mentioned earlier (in #527) I was not yet able to build Scio from source, I'm most certain this commit will resolve it.
I'll report back if not.

nevillelyh:
Try building it in SBT directly?

Issue #531, opened 2017-04-11
Add `toString` to anonymous functions


nevillelyh:
Just an idea. People often post stack trace of `NotSerializableException` of `Functions$$anon$$...` and couldn't easily tell where it's from. Maybe having a custom `.toString` method would help.

PR #532, opened 2017-04-11
Add toString to anonymous functions #531.

andrewsmartin:


nevillelyh:
Threre're a couple more `Function*` utilities in https://github.com/spotify/scio/tree/master/scio-core/src/main/scala/com/spotify/scio/util
Should we cover them also?

andrewsmartin:
Sure.

scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -195,4 +195,14 @@ private[scio] object Functions {
     override def reduceOption(accumulator: Iterable[T]): Option[T] = _mon.sumOption(accumulator)
   }
 
+  /** Helper trait to decorate anonymous functions with meaningful toString. */
+  private trait PrettyFn {
+    val callsite = CallSites.getCurrent
+    override def toString: String = callsite
```
nevillelyh: Make it something like `s"anonymous function $callsite"`?

scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -195,4 +195,14 @@ private[scio] object Functions {
     override def reduceOption(accumulator: Iterable[T]): Option[T] = _mon.sumOption(accumulator)
   }
 
+  /** Helper trait to decorate anonymous functions with meaningful toString. */
```
andrewsmartin: @nevillelyh Small style question, would you prefer these stay here or move to another file `PrettyFn.scala`? I sort of prefer moving them so they're more visible now that they're being used across the package.

nevillelyh: Just this 1 trait right? Don't have a preference but I won't move it :)

andrewsmartin: Well the trait and the 3 lines below as well

nevillelyh:
LGTM merge when tests pass.

PR #538, opened 2017-04-13
Resolve #535 Add PubSub attribute access

### LINKED ISSUE #535: Google PubSub Attribute Access?
Now that Google has the DataFlow 2.0.x (Beam 0.6.y) version available, would it be trivial to enable read/write access to PubSub attributes? As you may know, they were not previously available on the Java API.

Much appreciated if you can help!

Jon

shenoybr:
Provided ability to pass in parseFn for attributes on Pubsub Read
Provided ability to pass in formatFn for attributes on Pubsub Write

@nevillelyh @ravwojdyla 👀

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -34,7 +34,7 @@ import com.spotify.scio.coders.AvroBytesUtil
 import com.spotify.scio.io.{TFRecordOptions, TFRecordSource, Tap}
 import com.spotify.scio.options.ScioOptions
 import com.spotify.scio.testing._
-import com.spotify.scio.util.{CallSites, ScioUtil}
+import com.spotify.scio.util.{CallSites, Functions, ScioUtil}
 import com.spotify.scio.values._
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
@@ -592,7 +592,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
   def pubsubSubscription[T: ClassTag](sub: String,
                                       idLabel: String = null,
                                       timestampLabel: String = null,
-                                      parseFn: SimpleFunction[gio.PubsubIO.PubsubMessage,T] = null)
+                                      parseFn: gio.PubsubIO.PubsubMessage => T = null)
   : SCollection[T] = requireNotClosed {
     if (this.isTest) {
       this.getTestInput(PubsubIO(sub))
@@ -606,7 +606,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
         transform = transform.timestampLabel(timestampLabel)
       }
       if (parseFn != null) {
-        transform = transform.withAttributes(parseFn)
+        transform = transform.withAttributes(Functions.simpleFn(parseFn))
       }
       wrap(this.applyInternal(transform)).setName(sub)
     }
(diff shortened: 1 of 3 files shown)


ravwojdyla:
Should we hide `SimpleFunction` under scala function to keep the API scala friendly? Also no tests?

SamSchlegel:
Could add `Functions.simpleFn`. Since with `SimpleFunction` we'd be instantiating a class instead of creating an anonymous one, would we still want to extend it with `NamedFn`?

scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -150,6 +150,11 @@ private[scio] object Functions {
     override def partitionFor(elem: T, numPartitions: Int): Int = g(elem)
   }
 
+  def simpleFn[T, U](f: T => U): SimpleFunction[T, U] = new SimpleFunction[T, U] {
```
SamSchlegel: Might want to follow suit with `NamedDoFn` and add `private[util] class NamedSimpleFn[T, U] extends SimpleFunction[T, U] with NamedFn` to `NamedFn.scala` and use that instead, so that we get better callsite information.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -614,7 +618,8 @@ class ScioContext private[scio] (val options: PipelineOptions,
    */
   def pubsubTopic[T: ClassTag](topic: String,
                                idLabel: String = null,
-                               timestampLabel: String = null)
+                               timestampLabel: String = null,
+                               parseFn: gio.PubsubIO.PubsubMessage => T = null)
```
nevillelyh: How would this work with types like Avro? In that case `T` is an Avro `SpecificRecord` and may or may not contain fields that can represent the extra attributes? Or the user has to decode manually from `byte[]` to Avro?

Looking at `PubsubMessage` it's just 2 fields, `byte[] message` and `java.util.Map<java.lang.String,java.lang.String> attributes`.

How about variations of these 2 methods that return `[(T, Map[String, String])]` where `T` is the main message and `Map` is the attributes? That way we can still use the inferred coder. And the API feels more idiomatic scala.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1063,12 +1063,18 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * Save this SCollection as a Pub/Sub topic.
    * @group output
    */
-  def saveAsPubsub(topic: String): Future[Tap[T]] = {
+  def saveAsPubsub(topic: String,
```
nevillelyh: Same reasoning with the read counterpart, how about a variation that saves a `SCollection[(T, Map[String, String])]` by encoding `T` as message and `Map` as attributes?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
SamSchlegel: something like `pubsubTopicWithAttributes` ?

SamSchlegel: I was going to say I see a use for both, as I would imagine use cases where the pubsub attributes describe the format of the payload, but in that case `T` can just be `Array<Byte>` and the decoding can happen in a separate step.

PR #544, opened 2017-04-14
Make tap serializable

ravwojdyla:


nevillelyh:
What's the scenario of `Tap` being serialized?

ravwojdyla:
@nevillelyh The use case is to move `Tap` between closures. But his might be useful later on once SDF is out.

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -71,9 +72,10 @@ case class TextTap(path: String) extends Tap[String] {
  * @param schema must be not null if `T` is of type
  *               [[org.apache.avro.generic.GenericRecord GenericRecord]].
  */
-case class AvroTap[T: ClassTag](path: String, schema: Schema = null) extends Tap[T] {
-  override def value: Iterator[T] = FileStorage(path).avroFile(schema)
-  override def open(sc: ScioContext): SCollection[T] = sc.avroFile[T](path, schema)
+case class AvroTap[T: ClassTag](path: String, @transient schema: Schema = null) extends Tap[T] {
```
nevillelyh: `schema` will be null after the tap goes through ser/de right? That breaks the contract. And the user might access it accidentally.

Can't think of a way to fit this in a case class. Make it a boring class, `_schema: Schema` a ctor arg, and `def schema: Schema` to safely get it? Or is there any better way?

ravwojdyla: @nevillelyh how about making it `private`?

nevillelyh: Sure if that works on a case class :)

Issue #543, opened 2017-04-14
Intellij compilation fails on duplicate classes from scio-schemas


ravwojdyla:
This is not coming from scio itself, but to document this issues: if one wants to compile from intellij, to say run tests etc, Intellij fails with duplicate class from scio-schemas. This most likely is something related to Intellij settings, sbt or some of the plugins. Would be nice to figure this one out so that external contributors are not bothered by this issue.

In the meantime, manual workaround is to unmark `main` as `source directory` in `scio-schema/target/scala-*/src-managed` and recompile/rerun, and keep both `compiled_avro` and `compiled_protobuf` as source directories.

nevillelyh:
Could be related:
http://stackoverflow.com/questions/16885489/intellij-compile-failures-is-already-defined-as/16888290

PR #546, opened 2017-04-15
Add *withAttributes Pubsub I/O

### LINKED ISSUE #535: Google PubSub Attribute Access?

samschlegel:
Resolves #535, but by adding new input and output functions, as discussed in #538 

Open to renaming. Needs tests.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -47,7 +47,7 @@ import org.apache.beam.sdk.io.gcp.{bigquery => bqio, datastore => dsio}
 import org.apache.beam.sdk.options._
 import org.apache.beam.sdk.transforms.Combine.CombineFn
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
-import org.apache.beam.sdk.transforms.{Create, DoFn, PTransform, SimpleFunction}
+import org.apache.beam.sdk.transforms.{Create, DoFn, PTransform}
 import org.apache.beam.sdk.util.CoderUtils
 import org.apache.beam.sdk.values._
 import org.apache.beam.sdk.{Pipeline, io => gio}
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -34,7 +34,7 @@ import com.spotify.scio.coders.AvroBytesUtil
 import com.spotify.scio.io.{TFRecordOptions, TFRecordSource, Tap}
 import com.spotify.scio.options.ScioOptions
 import com.spotify.scio.testing._
-import com.spotify.scio.util.{CallSites, Functions, ScioUtil}
+import com.spotify.scio.util._
 import com.spotify.scio.values._
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -643,7 +643,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
       val outputCoder  = pipeline.getCoderRegistry.getScalaCoder[(T, Map[String, String])]
       val parseFn = Functions.simpleFn { msg: PubsubMessage =>
         val element = CoderUtils.decodeFromByteArray(elementCoder, msg.getMessage)
-        val attributes = msg.getAttributeMap.asScala.toMap
+        val attributes = JMapWrapper.of(msg.getAttributeMap)
         (element, attributes)
       }
       val input: gio.PubsubIO.Read[(T, Map[String, String])] =
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -587,6 +587,32 @@ class ScioContext private[scio] (val options: PipelineOptions,
       }
     }
 
+  private def pubsubIn[T: ClassTag](isSubscription: Boolean,
+                                    name: String,
+                                    idLabel: String,
+                                    timestampLabel: String)
+  : SCollection[T] = requireNotClosed {
+    if (this.isTest) {
+      this.getTestInput(PubsubIO(name))
+    } else {
+      val coder = pipeline.getCoderRegistry.getScalaCoder[T]
+      var transform =
+        if (isSubscription) {
+          gio.PubsubIO.read().subscription(name).withCoder(coder)
+        } else {
+          gio.PubsubIO.read().topic(name).withCoder(coder)
+        }
+
+      if (idLabel != null) {
+        transform = transform.idLabel(idLabel)
+      }
+      if (timestampLabel != null) {
+        transform = transform.timestampLabel(timestampLabel)
+      }
+      wrap(this.applyInternal(transform)).setName(name)
+    }
+  }
+
   /**
    * Get an SCollection for a Pub/Sub subscription.
    * @group input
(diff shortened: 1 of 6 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -609,6 +611,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+                                      idLabel: String = null,
+                                      timestampLabel: String = null)
+  : SCollection[(T, Map[String, String])] = requireNotClosed {
+    if (this.isTest) {
+      this.getTestInput(PubsubIO(sub))
+    } else {
+      val elementCoder = pipeline.getCoderRegistry.getScalaCoder[T]
+      val outputCoder  = pipeline.getCoderRegistry.getScalaCoder[(T, Map[String, String])]
+      val parseFn = Functions.simpleFn { msg: PubsubMessage =>
+        val element = CoderUtils.decodeFromByteArray(elementCoder, msg.getMessage)
```
SamSchlegel: This can throw a `CoderException`. It looks like in some places Beam catches this and rethrows as a `RuntimeException`. Not sure what we should do here

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -632,6 +665,36 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+                               idLabel: String = null,
+                               timestampLabel: String = null)
+  : SCollection[(T, Map[String, String])] = requireNotClosed {
+    if (this.isTest) {
+      this.getTestInput(PubsubIO(topic))
+    } else {
+      val elementCoder = pipeline.getCoderRegistry.getScalaCoder[T]
+      val outputCoder  = pipeline.getCoderRegistry.getScalaCoder[(T, Map[String, String])]
+      val parseFn = Functions.simpleFn { msg: PubsubMessage =>
+        val element = CoderUtils.decodeFromByteArray(elementCoder, msg.getMessage)
```
SamSchlegel: Same as above re: exceptions.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1074,6 +1076,29 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+  def saveAsPubsubWithAttributes[V: ClassTag](topic: String)
+                                             (implicit ev: T <:< (V, Map[String, String]))
+  : Future[Tap[V]] = {
+    if (context.isTest) {
+      context.testOut(PubsubIO(topic))(this)
+    } else {
+      val elementCoder = internal.getPipeline.getCoderRegistry.getScalaCoder[V]
+      val tupleCoder = internal.getPipeline.getCoderRegistry.getScalaCoder[T]
+      val formatFn = Functions.simpleFn { tuple: T =>
+        val element = CoderUtils.encodeToByteArray(elementCoder, tuple._1)
```
SamSchlegel: Same as above re: exceptions.

SamSchlegel:
How should I add tests to this? Just reuse `PubsubIO` but pass in `(String, Map[String, String])`, or create a new `TestIO` type?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: It only throws exception for corrupt message right? In that case not much we can do and should be fine to let it escalate.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -609,6 +611,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+                                      timestampLabel: String = null)
+  : SCollection[(T, Map[String, String])] = requireNotClosed {
+    if (this.isTest) {
+      this.getTestInput(PubsubIO(sub))
+    } else {
+      val elementCoder = pipeline.getCoderRegistry.getScalaCoder[T]
+      val outputCoder  = pipeline.getCoderRegistry.getScalaCoder[(T, Map[String, String])]
+      val parseFn = Functions.simpleFn { msg: PubsubMessage =>
+        val element = CoderUtils.decodeFromByteArray(elementCoder, msg.getMessage)
+        val attributes = msg.getAttributeMap.asScala.toMap
```
nevillelyh: You can use a private class that extends `Map[String, String]` to do the lazy wrapping. It's fine since the underlying java Map is never exposed.

https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala#L98

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -609,6 +611,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
   }
 
   /**
+    * Get an SCollection for a Pub/Sub subscription that includes message attributes.
+    * @group input
+    */
+  def pubsubSubscriptionWithAttributes[T: ClassTag](sub: String,
+                                      idLabel: String = null,
```
nevillelyh: Nitpick: align arguments?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -632,6 +665,36 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+  def pubsubTopicWithAttributes[T: ClassTag](topic: String,
+                               idLabel: String = null,
+                               timestampLabel: String = null)
+  : SCollection[(T, Map[String, String])] = requireNotClosed {
+    if (this.isTest) {
+      this.getTestInput(PubsubIO(topic))
+    } else {
+      val elementCoder = pipeline.getCoderRegistry.getScalaCoder[T]
+      val outputCoder  = pipeline.getCoderRegistry.getScalaCoder[(T, Map[String, String])]
+      val parseFn = Functions.simpleFn { msg: PubsubMessage =>
```
nevillelyh: `elementCoder` & `parseFn` are identical in `pubsubSubscriptionWithAttributes` and `pubsubTopicWithAttributes`, put them in a helper method? You can even put almost everything in a helper method and parameterize only the `gio.PubsubIO.read().subscription(topic)` vs `gio.PubsubIO.read().topic(topic)` part.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1074,6 +1076,29 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   }
 
   /**
+    * Save this SCollection as a Pub/Sub topic using the given map as message attributes.
+    * @group output
+    */
+  def saveAsPubsubWithAttributes[V: ClassTag](topic: String)
```
nevillelyh: I guess you want `idLabel` and `timestamp` args here too?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
SamSchlegel: Yep, only for corrupt message.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ravwojdyla: Isn't this a little bit too much for pubsub attributes? I mean if someone is using this methods specifically, they do care about getting attributes. Seems like a unnecessary optimization here?

samschlegel: They do care about getting them, but modifying that map is probably unlikely, so delaying the conversion makes sense.

nevillelyh: Seems pretty common to wrap `j.u.Map` as immutable `Map` instead of the mutable one from `asScala`. We can move those from `SideInput.scala` to a util file and reuse them.

samschlegel: The wrapper in `SideInput` also converts `JIterable` to `Iterable`, so would have to specialize for that anyways, no?

nevillelyh: No that's specific to the `IterableSideInput` case. But wouldn't hurt to pull it into a util file like the regular `Map[K, V]` wrapper.

samschlegel: Since `parseFn` is a `SimpleFunction<T, U>`, which is invariant in U, it would have to leak the wrapper. Is that okay, or is there some way to get around that?

nevillelyh: What do mean by leak? `parseFn` and it's signature doesn't leave this method right? Can you provide a snippet?

samschlegel: The output type of `parseFn` is what's used as the output type of the read transform:
```
public Read<T> withAttributes(SimpleFunction<PubsubMessage, T> parseFn)
```
and thus the element type of the returned SCollection. Would casting the wrapper to an instance of `Map[String, String]` be the way to go?

scio-core/src/main/scala/com/spotify/scio/util/JMapWrapper.scala
```diff
@@ -0,0 +1,39 @@
+package com.spotify.scio.util
+
+import java.lang.{Iterable => JIterable}
+import java.util.{Map => JMap}
+
+import scala.collection.JavaConverters._
+
+private[scio] object JMapWrapper {
+  def ofMultiMap[A, B](self: JMap[A, JIterable[B]]): Map[A, Iterable[B]] =
```
samschlegel: `ofJIterable`? 🤷‍♂️

nevillelyh: Either way don't have a strong opinion. We can keep `ofMultiMap`.

nevillelyh:
LGTM but there's some conflict against master. Can you fix? Thanks.

nevillelyh:
Somehow squash works. Weird.
Anyway thanks for the great work!

jonopwell:
Appreciate the work, guys!

Would you expect it to be immediately included in an upcoming release? If so, when?

nevillelyh:
Don't have any immediate plan and we're still waiting on a lot of upstream changes like Beam release and scala 2.12 fixes. Can you give the snapshot a try first?

jonopwell:
Sure...we can try the snapshot.

smadigan-dexcom:
Hello, I work with jonopwell and tried to build the current source, it fails with 
`[error] /Users/smadigan/git/scio/scio-schemas/src/main/scala/com/spotify/scio/avro/AvroUtils.scala:54: not found: type TestRecord
[error]   def newSpecificRecord(i: Int): TestRecord =
[error]                                  ^
[error] /Users/smadigan/git/scio/scio-schemas/src/main/scala/com/spotify/scio/avro/AvroUtils.scala:55: not found: type TestRecord
[error]     new TestRecord(i, i.toLong, i.toFloat, i.toDouble, true, "hello")
[error]         ^
[error] two errors found
[error] (scio-schemas/compile:compileIncremental) Compilation failed`

I tried working around that error and hit further errors.

nevillelyh:
Are you building in IntelliJ? There's a known IntelliJ issue that can be worked around, see #543
We also publish snapshots on every master commit so you don't have to build yourself.

smadigan-dexcom:
I built from the command line, sbt compile
Where are the snapshots published?  (Forgive my ignorance, please)

samschlegel:
They're published to the Sonatype Snapshots repo which you can add to your build.sbt with:
```sbt
 resolvers += Resolver.sonatypeRepo("snapshots")
```

nevillelyh:
Added this to the wiki FAQ.
https://github.com/spotify/scio/wiki/FAQ#how-do-i-use-the-snapshot-builds

dgouyette:
There is no unit test on saveAsPubsubWithAttributes.
There is one unit test on saveAsPubsub

The signature of saveAsPubsubWithAttributes is difficult to read. An example with a unit test would be a good thing

nevillelyh:
Filed #590, would be great if you already have something and can submit a PR.

Issue #545, opened 2017-04-14
Handle highly uneven outer join case


nevillelyh:
In this case we have a full outer join where LHS is much larger than the RHS. Both sides contain 1-to-1 key-value mapping. LHS contains historical aggregation with all previously seen keys and RHS is an increment with new keys that mostly overlap with LHS. A full outer join will shuffle the entire LHS. We could probably solve it like this:

- Computer intersect of LHS & RHS keys
- Split LHS & RHS into 2 partitions, one with intersect keys and one without
- Full outer join intersect partitions
- Concatenate with non-intersect partitions

nevillelyh:
Fixed in #548

PR #548, opened 2017-04-17
add sparseFullOuterJoin #545

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -353,6 +352,39 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+   * Full outer join for cases when `this` is much larger than `that` which cannot fit in memory,
+   * but contains a mostly overlapping set of keys as `this`. A Bloom Filter of keys in `that` is
+   * used to split `this` into 2 partitions. Only those with keys in the filter go through the
+   * join and the rest are concatenated. This is useful for joining historical aggregates with
+   * incremental updates.
+   * @group join
+   * @param thatNumKeys estimated number of keys in `that`
+   * @param fpProb false positive probability when computing the overlap
+   */
+  def bloomFilterOuterJoin[W: ClassTag](that: SCollection[(K, W)],
```
nevillelyh: Not sure about the name though. Some other candidates are: `unbalanced`, `lookup`, `incremental`. What else?

ravwojdyla: How about `sparseOuterJoin`, would be more consistent with `skewJoin`. `skew` is partitioning to get the hot keys, and `sparse` is partitioning to get the "cold" keys.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -353,6 +352,39 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+                                        thatNumKeys: Int,
```
ravwojdyla: `Int` => `Long`

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -353,6 +352,39 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
     hotJoined ++ chillJoined
   }
 
+  /**
```
ravwojdyla: Does it require to import any of `algebird` (like for `Hash128`), if so might be worth adding in the doc like in the `skewJoin`.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -353,6 +352,39 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+   * Full outer join for cases when `this` is much larger than `that` which cannot fit in memory,
```
ravwojdyla: Might be worth linking to the algebird doc for bloom filters for the `fpProb`.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ravwojdyla: I guess not - cause `optimalWidth` wants Int anyway.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
nevillelyh: No need to. `trait Hash128` has a companion `object Hash128` with common instances and those are in the default lookup scope. Nice little implicit trick.

PR #554, opened 2017-04-18
Fix #553. Use GCP temp location or throw

### LINKED ISSUE #553: NPE in materialize if tempLocation not set
```
scio> :scioOpts --gcpTempLocation=gs://scio-playground/user/rav/df-staging --runner=DataflowRunner --project=scio-playground --zone=us-west1-b
Scio options updated. Use :newScio to get a new Scio context.

scio> :newScio
[run-main-0] WARN com.spotify.scio.VersionUtil$ - Using a SNAPSHOT version of Scio: 0.3.0-SNAPSHOT
Scio context available as 'sc'

scio> sc.parallelize(1 to 10).closeAndCollect
  at com.spotify.scio.values.SCollection$class.materialize(SCollection.scala:831)
  at com.spotify.scio.values.SCollectionImpl.materialize(SCollection.scala:1167)
  at com.spotify.scio.repl.package$ReplSCollection.closeAndCollect(package.scala:31)
  ... 45 elided
```

that is due to `null` coming from `getTempLocation` in:

```
  def materialize: Future[Tap[T]] = {
    val filename = "scio-materialize-" + UUID.randomUUID().toString
    val tmpDir = context.options.getTempLocation
    val path = tmpDir + (if (tmpDir.endsWith("/")) "" else "/") + filename
```

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -827,7 +831,18 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    */
   def materialize: Future[Tap[T]] = {
     val filename = "scio-materialize-" + UUID.randomUUID().toString
-    val tmpDir = context.options.getTempLocation
+    val tmpDir = if (context.options.getTempLocation != null) {
+      context.options.getTempLocation
+    } else {
+      val m = "Specify a temporary location via --tempLocation or PipelineOptions.setTempLocation."
```
nevillelyh: We should use a local directory if temp location is null?

ravwojdyla: For locally runner that is already the case, but for other runners was thinking of using reusing gcp temp location or throwing meaningful error like in this PR.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -827,7 +831,18 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+      scala.util.Try(context.optionsAs[GcpOptions].getGcpTempLocation).toOption match {
+        case Some(l) => {
```
nevillelyh: Minor: `case Success/Failure` so you don't need the `.toOption`?

nevillelyh:
👍

PR #567, opened 2017-04-24
Fix #566. Use build in totalCount and all default params

### LINKED ISSUE #566: Simplify skewedJoin API
Instead of accepting CMS parameters (delta, eps) accept estimated size of the stream, and compute upper bound error internally.

ravwojdyla:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -272,17 +273,19 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
 
     val leftSideKeys = if (sampleFraction < 1.0) {
-      self.sample(withReplacement, sampleFraction).keys
+      self.withName("Sample LHS").sample(withReplacement, sampleFraction).keys
     } else {
       self.keys
     }
 
-    val cms = leftSideKeys.aggregate(keyAggregator)
+    val cms = leftSideKeys.withName("Compute CMS of keys in LHS").aggregate(keyAggregator)
```
nevillelyh: `"Compute CMS of LHS keys"` is more readable?

nevillelyh:
:+1:

PR #568, opened 2017-04-24
Fix serialization of LocalDateTime and LocalDate

kanterov:
See https://github.com/twitter/chill/issues/226. 

I was lazy to extend it for other types from joda, but they likely have the same problem. AFAIK newest Kryo is able to serialize java8 time, but it seems Beam still uses Joda, not sure for how long it would be relevant.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
@@ -1,0 +1,59 @@
+package com.spotify.scio.coders
+
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.esotericsoftware.kryo.{Kryo, Serializer}
+import org.joda.time.{LocalDate, LocalDateTime}
+import org.joda.time.chrono.ISOChronology
+
+class JodaLocalDateTimeSerializer extends Serializer[LocalDateTime] {
+  setImmutable(true)
+
+  def write(kryo: Kryo, output: Output, ldt: LocalDateTime): Unit = {
+    output.writeInt(ldt.getYear, /*optimizePositive=*/ false)
+    output.writeByte(ldt.getMonthOfYear)
+    output.writeByte(ldt.getDayOfMonth)
+    output.writeLong(ldt.getMillisOfDay)
+
+    val chronology = ldt.getChronology
+    if (chronology != null && chronology != ISOChronology.getInstanceUTC) {
+      sys.error(s"Unsupported chronology: $chronology")
+    }
+  }
+
+  def read(kryo: Kryo, input: Input, tpe: Class[LocalDateTime]): LocalDateTime = {
+    val year = input.readInt(/*optimizePositive=*/ false)
+    val month = input.readByte().toInt
+    val day = input.readByte().toInt
+
+    val millis = input.readLong()
+    val hour = (millis / 3600000L).toInt
+    val minute = ((millis % 3600000L) / 60000).toInt
+    val second = ((millis % 60000L) / 1000).toInt
+    val ms = (millis % 1000L).toInt
+
+    new LocalDateTime(year, month, day, hour, minute, second, ms)
+  }
+}
+
+class JodaLocalDateSerializer extends Serializer[LocalDate] {
+  setImmutable(true)
+
+  def write(kryo: Kryo, output: Output, ld: LocalDate): Unit = {
+    output.writeInt(ld.getYear, /*optimizePositive=*/ false)
+    output.writeByte(ld.getMonthOfYear)
+    output.writeByte(ld.getDayOfMonth)
+
+    val chronology = ld.getChronology
+    if (chronology != null && chronology != ISOChronology.getInstanceUTC) {
+      sys.error(s"Unsupported chronology: $chronology")
+    }
+  }
+
+  def read(kryo: Kryo, input: Input, tpe: Class[LocalDate]): LocalDate = {
+    val year = input.readInt(/*optimizePositive=*/ false)
+    val month = input.readByte().toInt
+    val day = input.readByte().toInt
+
+    new LocalDate(year, month, day)
+  }
+}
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.esotericsoftware.kryo.{Kryo, Serializer}
+import org.joda.time.{LocalDate, LocalDateTime}
+import org.joda.time.chrono.ISOChronology
+
+class JodaLocalDateTimeSerializer extends Serializer[LocalDateTime] {
+  setImmutable(true)
+
+  def write(kryo: Kryo, output: Output, ldt: LocalDateTime): Unit = {
+    output.writeInt(ldt.getYear)
```
ravwojdyla: `true` to optimize?

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+    output.writeInt(ldt.getMonthOfYear)
```
ravwojdyla: Write `byte`?

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+    output.writeInt(ldt.getDayOfMonth)
```
ravwojdyla: Write `byte`?

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+    output.writeLong(ldt.getMillisOfDay)
+
+    val chronology = ldt.getChronology
+    if (chronology != null && chronology != ISOChronology.getInstanceUTC) {
+      sys.error(s"Unsupported chronology: $chronology")
+    }
+  }
+
+  def read(kryo: Kryo, input: Input, tpe: Class[LocalDateTime]): LocalDateTime = {
+    val y = input.readInt()
```
ravwojdyla: optimized?

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+    val m = input.readInt()
```
ravwojdyla: `byte`?

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+    val d = input.readInt()
```
ravwojdyla: `byte`?

scio-test/src/test/scala/com/spotify/scio/values/SCollectionPropTest.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import com.fortysevendeg.scalacheck.datetime.joda.ArbitraryJoda._
+import com.spotify.scio.testing.PipelineSpec
+import org.joda.time._
+import org.scalacheck.{Arbitrary, Prop}
+import org.scalatest.prop.Checkers
+
+import scala.reflect.ClassTag
+
+class SCollectionPropTest extends PipelineSpec with Checkers {
```
ravwojdyla: What's the need for pulling `ScioContext`/pipeline for this? This should be part of tests in `com.spotify.scio.coders.*`, and no need to pull `ScioContext`, instead use `CoderTestUtils`'s `testRoundTrip`.

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+
+    new LocalDateTime(
+      y,
+      m,
+      d,
+      (millis / 3600000L).toInt,
+      ((millis % 3600000L) / 60000).toInt,
+      ((millis % 60000L) / 1000).toInt,
+      (millis % 1000L).toInt,
+      null
```
ravwojdyla: Use constructor without `Chronology` or be explicit that you want it to be `ISOChronology.getInstanceUTC`? I would prefer the 2nd to be explicit, also aligned with the `write` logic.

scio-core/src/main/scala/com/spotify/scio/coders/JodaSerializer.scala
```diff
@@ -0,0 +1,68 @@
...
+    )
+  }
+}
+
+class JodaLocalDateSerializer extends Serializer[LocalDate] {
```
ravwojdyla: The same comments as above:
 * use optimized versions of Int
 * use byte for small ints
 * read optimized
 * be explicit about constructor of `LocalDate`

kanterov:
@ravwojdyla thanks for review.

> use optimized versions of Int
> read optimized
didn't do this because year can be negative and tests are failing with optimized, but made it explicit in the code that we don't want optimizing.

> use byte for small ints
> be explicit about constructor of LocalDate
👍

kanterov:
I migrated tests not to depend on `SCollection` and removed dependency to `scalacheck-datetime` library hand-coding arbitrary instance for `LocalDateTime` only.

ravwojdyla:
@kanterov nice. can year be negative tho?

nevillelyh:
LGTM. Test style is a bit more FP than the rest of the code base but not a big deal :)

kanterov:
@ravwojdyla it seems so, joda doesn't' throw any exception, I guess, there could something before christ :)

@nevillelyh I'm not much into coding style there, but I feel it's very valuable to have property-based tests, in fact, in my experience we have discovered bugs in Spark using such tests, for instance [join tests in frameless](https://github.com/typelevel/frameless/blob/master/dataset/src/test/scala/frameless/JoinTests.scala).

kanterov:
@nevillelyh @ravwojdyla thanks for review, I've squashed commits

nevillelyh:
+1 on prop-based testing. Just it's slightly different than the rest of the code base here, which uses `GeneratorDrivenPropertyChecks`. This actually look cleaner. But we can consolidate later :)

PR #569, opened 2017-04-24
Better error messages in case of BQ failure

ravwojdyla:
If user wants to use standard syntax, but the query is invalid, she will see error from legacy syntax, which is misleading, this will simple print both errors. Might be a bit verbose, but at least there is no hidden error.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
@@ -36,11 +36,8 @@ import com.google.common.base.Charsets
 import com.google.common.hash.Hashing
 import com.google.common.io.Files
 import com.spotify.scio.bigquery.types.BigQueryType.HasAnnotation
-import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO
-import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.CreateDisposition._
-import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.WriteDisposition._
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.{CreateDisposition, WriteDisposition}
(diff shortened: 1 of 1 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -531,7 +530,12 @@ class BigQueryClient private (private val projectId: String,
                   "See https://cloud.google.com/bigquery/docs/reference/standard-sql/")
                 logger.warn(s"Legacy query: `$sqlQuery`")
                 true
-              case Failure(f) => throw f
+              case Failure(f) =>
+                logger.error(
+                  s"Tried both standard and legacy syntax, query `$sqlQuery` failed for both!")
+                logger.error("Standard syntax failed due to:", e)
+                logger.error("Legacy syntax failed due to:")
```
nevillelyh: Should there be a `, e` here after "due to:"?

ravwojdyla: Nope - because the throw after will print the exception, otherwise it would be duplicated.

nevillelyh: Will they be printed to the same output stream though? Depends on logger configuration right?

ravwojdyla: Yes it does depend on the configuration. I'm fine with adding it, but it most cases it will result in duplicated stacktrace.

nevillelyh: Or throw a new RuntimeException with e and f as suppressed and a more informative message. A bit too much maybe, up to you :)

ravwojdyla: Thought about that, also about some kind of CompositeException. Btw having an algebra for Exceptions would be nice. Anyway, it's kinda too much for this one problem, so will keep it simple.

PR #574, opened 2017-04-25
Reduce default eps for skewedJoin

ravwojdyla:
For default/large jobs it might be still too much to have 0.0001.
Production pipeline with 4B records was to large for 0.0001, reduce
to 0.001.


ravwojdyla:
My goal here is to have default params working for up to 10B records.

nevillelyh:
LGTM feel free to merge.

Issue #570, opened 2017-04-25
Better JSON IO via Circe?


nevillelyh:
Seems some users are reading/writing JSON text files with Jackson, which is not super user-friendly. We could easily add [Circe](https://github.com/circe/circe)-based JSON IO and improve usability and type-safety.

PR #576, opened 2017-04-26
add JSON IO based on Circe #570

nevillelyh:


nevillelyh:
@ravwojdyla @regadas thoughts?

scio-extra/src/main/scala/com/spotify/scio/extra/json/package.scala
```diff
@@ -0,0 +1,100 @@
...
+  implicit class JsonSCollection[T: ClassTag : Encoder : Decoder](val self: SCollection[T]) {
+    def saveAsJsonFile(path: String, numShards: Int = 0): Future[Tap[T]] = {
+      if (self.context.isTest) {
+        self.context.testOut(JsonIO[T](path))(self)
+        self.saveAsInMemoryTap
+      } else {
+        val encoder = implicitly[Encoder[T]]
+        val decoder = implicitly[Decoder[T]]
+        self
+          .map(_.asJson(encoder).noSpaces)
```
ravwojdyla: Do you need to specify those explicitly, the seem to be in the context, and `asJson` takes implicit encoder?

nevillelyh: Yeah there's some issues, fixing.

scio-extra/src/main/scala/com/spotify/scio/extra/json/package.scala
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Enhanced version of [[SCollection]] with JSON methods.
+   */
+  implicit class JsonSCollection[T: ClassTag : Encoder : Decoder](val self: SCollection[T]) {
+    def saveAsJsonFile(path: String, numShards: Int = 0): Future[Tap[T]] = {
+      if (self.context.isTest) {
+        self.context.testOut(JsonIO[T](path))(self)
+        self.saveAsInMemoryTap
+      } else {
+        val encoder = implicitly[Encoder[T]]
```
ravwojdyla: Do you need those in question below ↓

scio-extra/src/main/scala/com/spotify/scio/extra/json/package.scala
```diff
@@ -0,0 +1,100 @@
...
+        val decoder = implicitly[Decoder[T]]
+        self
+          .map(_.asJson(encoder).noSpaces)
+          .applyInternal(self.textOut(path, ".json", numShards))
+        self.context.makeFuture(TextTap(path + "/part-*").map(decode[T](_)(decoder).right.get))
```
ravwojdyla: Same as question above for explicit decoder ↑

scio-extra/src/main/scala/com/spotify/scio/extra/json/package.scala
nevillelyh: Actually I need these to defeat closure, will add comments once I fix all the ser/de issues :)

ravwojdyla: Mark `self` as `@transient` and make `JsonSCollection` extend `Serializable` and you should be good.

nevillelyh: That works. But it also means `JsonSCollection` gets pulled into the closure. I feel `val encoder = ...` is more explicit w.r.t. what gets pulled in. Thoughts?

ravwojdyla: What's the problem of `JsonSCollection` getting pulled into the closure. Honestly those `vals` are only explicit if you already know what is going on :)

nevillelyh: Nothing. But IMO it's better to be clear what gets into the closure (`Encoder`, `Decoder`) and risks. Those implicits could pull in a lot more and all needs to be serializable.

ravwojdyla: I think the transient/serializable is simpler and more idiomatic solution for the current problem. But it does not matter much cause its not exposed to the user. I will leave it up to you.

nevillelyh: You convinced me :)
Let's do this. We don't have this problem for any other implicit `sc` or `SColl`s but if we do can apply the same pattern.

regadas:
@nevillelyh @ravwojdyla late to the party, but this looks really nice and clean! well done! :)

PR #581, opened 2017-04-26
paritition sparseOuterJoin to prevent BF overflow

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -376,13 +417,38 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+
+  protected def sparseOuterJoinImpl[W: ClassTag](that: SCollection[(K, W)],
+                                               thatNumKeys: Int,
+                                               fpProb: Double)
+                                              (implicit hash: Hash128[K])
+  : SCollection[(K, (Option[V], Option[W]))] = {
     val width = BloomFilter.optimalWidth(thatNumKeys, fpProb).get
     val numHashes = BloomFilter.optimalNumHashes(thatNumKeys, width)
-    val rhsBf = that.keys.aggregate(BloomFilterAggregator[K](numHashes, width)).asSingletonSideInput
+    val rhsBf = {
```
ravwojdyla: You prefer this over the `skewedJoin` solution, which uses `asIterableSideInput` and then when used `if (c(rhsBf).nonEmpty && c(rhsBf).head.maybeContains(e._1)) {`?

nevillelyh: Oh didn't see that one. That's actually better and 1 less transform.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -26,9 +26,50 @@ import com.spotify.scio.util.random.{BernoulliValueSampler, PoissonValueSampler}
...
+      capacity = capacity >> 1
+      width = estimateWidth(capacity, fpProb)
+    }
+    val numBFs = (numEntries / capacity).toInt + 1
+
+    logger.info(s"Optimal Bloom Filter settings for numEntries = $numEntries, fpProb = $fpProb")
+    logger.info(s"BF width = $width, capacity = $capacity, numBFs = $numBFs")
+    val totalBytes = width.toLong * numBFs / 8
+    if (totalBytes > maxTotalBytes) {
+      logger.warn(s"Total BF bytes $totalBytes > 1GB")
```
ravwojdyla: Yeah, now i see how this is gonna be kinda arbitrary warning. Essentially it doesn't really matter from the perspective of this  algo, +/- how well partitioning works. Does this warning actually make sense at the end of the day? It doesn't suggest what to do, nor what is the problem.

nevillelyh: Removed the warning and instead just log total bytes. It's really up to the user's interpretation.

nevillelyh:
@ravwojdyla cleaned up and rebased. Anything else to add?

ravwojdyla:
@nevillelyh LGTM

PR #583, opened 2017-04-28
Open sciocontext in scioresult, no harm

ravwojdyla:
One use case is to pimp ScioResult, and check the state of sciocontext
like `isTest` otherwise would have to pimp both sciocontext and
scioresult.

Also use classOf.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
@@ -43,7 +43,7 @@ import scala.util.{Failure, Success, Try}
 /** Represent a Scio pipeline result. */
 class ScioResult private[scio] (val internal: PipelineResult,
                                 val accumulators: Seq[Accumulator[_]],
-                                private val context: ScioContext) {
+                                val context: ScioContext) {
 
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -43,9 +43,9 @@ import scala.util.{Failure, Success, Try}
 /** Represent a Scio pipeline result. */
 class ScioResult private[scio] (val internal: PipelineResult,
                                 val accumulators: Seq[Accumulator[_]],
-                                private val context: ScioContext) {
+                                val context: ScioContext) {
 
-  private val logger = LoggerFactory.getLogger(this.getClass)
+  private val logger = LoggerFactory.getLogger(classOf[ScioResult])
```
nevillelyh: Any reason not to use `this.getClass` here? Thought that's simpler and more copy-paste-proof?

nevillelyh:
:+1:

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: More scala idiomatic.

PR #584, opened 2017-05-01
make some semigroups monoids

nevillelyh:


andrewsmartin:
What's the reason for this? Just to make it easier to sum with empty `Array[T]`?

nevillelyh:
Yeah I'm not convinced that they are proper monoids since empty array is not the same as zero array. But for breeze it's nice to reduce boilerplate.

nevillelyh:
I'm gonna abort the monoid idea and just generalize Breeze stuff. Will create another one.

Issue #590, opened 2017-05-04
Unit test for saveAsPubsubWithAttributes


nevillelyh:
https://github.com/spotify/scio/pull/546#issuecomment-299128082

Will be great to add unit tests as examples for `ScioContext#pubsub*WithAttributes` and `SCollection#saveAsPubsubWithAttributes`

dgouyette:
I will provide it ASAP

nevillelyh:
Awesome thanks!

dgouyette:
I don't know if that enought. But i didn't see other "unit tests" on PubSub*

nevillelyh:
These are fixed in the beam-2.0 branch. Closing for now.

PR #596, opened 2017-05-05
Unit test for saveAsPubsubWithAttributes #590

dgouyette:


scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -86,6 +87,26 @@ object PubsubJob {
...
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+    sc.pubsubTopic[String](args("input"), null)
+      .map(_ + "X")
+      .map(s => (s, Map("attribute1"-> "value1")))
+      .saveAsPubsubWithAttributes(args("output"))
+    sc.close()
+  }
+}
+object PubsubInWithAttributesJob {
```
nevillelyh: Minor: empty line between objects.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -86,6 +87,26 @@ object PubsubJob {
   }
 }
 
+object PubsubWithAttributesJob {
```
nevillelyh: `PubsubOutWithAttributesJob` so it's symmetric with `PubsubInWithAttributesJob?

dgouyette: Done

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
dgouyette: Done

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -25,6 +25,7 @@ import com.spotify.scio.avro.AvroUtils.{newGenericRecord, newSpecificRecord}
 import com.spotify.scio.avro.{AvroUtils, TestRecord}
 import com.spotify.scio.bigquery._
 import com.spotify.scio.util.MockedPrintStream
+import com.spotify.scio.values.SCollection
```
nevillelyh: Unused import?

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -86,6 +87,27 @@ object PubsubJob {
   }
 }
 
+object PubsubInWithAttributesJob {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+    sc.pubsubSubscriptionWithAttributes(args("input"))
+      .map((v: (String, Map[String, String])) => (v._1 + "X", v._2))
+      .saveAsPubsubWithAttributes(args("output"))
```
nevillelyh: Discard the map and `.saveAsPubsub()` instead?

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -86,6 +87,27 @@ object PubsubJob {
...
+    sc.close()
+  }
+}
+
+object PubsubOutWithAttributesJob {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+    sc.pubsubTopic[String](args("input"), null)
```
nevillelyh: Don't need the `null` argument since there's already a default.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
dgouyette: This is to illustrate the use of the parameters.
I tried ```.map{ case  (value : String, attributes :  Map[String, String]) => (  value + "X", attributes ) }```
But it doesn't compile

nevillelyh: What I mean is
```scala
.map(kv => kv._1 + "X")
.saveAsPubsub()
```
To demonstrate with attributes in but without attributes out, opposite to the other job.

dgouyette:
I wrote unit test but they didn't pass.

**when context.isTest == true** 

I got the following exception : ```org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.ClassCastException: scala.Tuple2 cannot be cast to java.lang.String
[info]   at org.apache.beam.runners.direct.DirectRunner.run(DirectRunner.java:352)
[info]   at org.apache.beam.runners.direct.DirectRunner.run(DirectRunner.java:71)
[info]   at org.apache.beam.sdk.Pipeline.run(Pipeline.java:210)
[info]   at org.apache.beam.sdk.testing.TestPipeline.run(TestPipeline.java:305)
[info]   at com.spotify.scio.ScioContext$$anonfun$close$1.apply(ScioContext.scala:344)
[info]   at com.spotify.scio.ScioContext$$anonfun$close$1.apply(ScioContext.scala:336)
[info]   at com.spotify.scio.ScioContext.requireNotClosed(ScioContext.scala:363)
[info]   at com.spotify.scio.ScioContext.close(ScioContext.scala:336)
[info]   at com.spotify.scio.testing.PubsubOutWithAttributesJob$.main(JobTestTest.scala:106)
[info]   at com.spotify.scio.testing.PubsubOutWithAttributesJob.main(JobTestTest.scala)```

**when context.isTest == false** 

It's seems that the selected attributes coder is KryoAtomicCoder (default when null) 
It should use MapCoder, but it's for java map, not scala map 

Convert scala map to java map doesn't work (

```Error:(36, 34) Cannot prove that (String, java.util.Map[String,String]) <:< (V, Map[String,String]).
Error occurred in an application involving default arguments.
      .saveAsPubsubWithAttributes(topic)```

So it uses the default coder which is KryoAtomicCoder

``` if (coder == null || coder.getClass == classOf[SerializableCoder[T]]) {
        KryoAtomicCoder[T]
      } else {
        coder
      }```

Any idea to fix it ?

nevillelyh:
We're doing a lot of API refactoring to catch up with Beam 2.0.0 release and this PR may not work after that. Let's revisit later once we're on Beam 2.x

dgouyette:
Thanks for my work and your answers

nevillelyh:
Thank you for your understanding. It's probably better to do it properly then than wasting a lot of work right now.

dgouyette:
No problem. I undertand ;-)  But the attributes as implemented were not working. 
I'll test after uout refactoring.  Good luck

PR #598, opened 2017-05-05
support quoted arguments, fix #593

### LINKED ISSUE #593: Argument processing doesn't handle quoted string
For example if I wanted to pass in a sql query as a quoted string the argument processor splits at commas.  So for example passing in the following as a quoted string:

"select a,b,c from foo"

Results in an exception... something like multiple values defined.  Looking at the scio code.  In args.scio the parsing just uses Java's String.split(",") which ignores quoted strings.

Might want to use katan.csv parser to do the quoted string handling.  It must have it, and it is already included in scio.

nevillelyh:


nevillelyh:
@keithmcneill take a look please? Pretty primitive but should cover your case.

keithmcneill:
+1

PR #599, opened 2017-05-05
Initial commit with Elasticsearch Writer

lmayumi:
@nevillelyh @ravwojdyla @shenoybr @andrisnoko @rculbertson @anilmuppalla


build.sbt
```diff
@@ -292,6 +292,24 @@ lazy val scioBigtable: Project = Project(
   scioTest % "it"
 ).configs(IntegrationTest)
 
+lazy val scioElsticsearch: Project = Project(
```
nevillelyh: typo

build.sbt
```diff
@@ -292,6 +292,24 @@ lazy val scioBigtable: Project = Project(
...
+  "scio-elasticsearch",
+  file("scio-elasticsearch")
+).settings(
+  commonSettings ++ itSettings,
+  description := "Scio add-on for Writing to Elasticsearch Cluster",
```
nevillelyh: "for writing to Elasticsearch"

build.sbt
```diff
@@ -292,6 +292,24 @@ lazy val scioBigtable: Project = Project(
...
+  libraryDependencies ++= Seq(
+    "org.apache.beam" % "beam-sdks-java-io-google-cloud-platform" % beamVersion,
+    "com.google.auto.value" % "auto-value" % "1.2",
+    "joda-time" % "joda-time" % jodaTimeVersion,
+    "org.elasticsearch" % "elasticsearch" % "2.1.0",
```
nevillelyh: Do we plan to support more recent versions like 5.x?

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,38 @@
+package com.spotify.scio
+
+import java.time.Duration
+
+import com.spotify.scio.io.Tap
+import com.spotify.scio.values.SCollection
+
+import scala.concurrent.Future
+
+package object elasticsearch {
```
nevillelyh: Add doc here like the other package objects. The simple "import all" instruction should be enough.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,38 @@
...
+  implicit class ElasticsearchSCollection(val self: SCollection[IndexRequestWrapper]) {
```
nevillelyh: `extends AnyVal` to make this a value class.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,38 @@
...
+        ElasticsearchIO.Write.withElasticsearchOptions(elasticsearchOptions))
+    }
+
+    /**
+      * Save this SCollection into Elasticsearch.
+      * Note that the elements must be of a type 'IndexRequestWrapper'
+      * @param elasticsearchOptions provides clusterName and server endpoints
+      * @param flushInterval delayed applied to rate limit writes to Elasticsearch cluster
+      */
+    def saveAsElasticsearch(elasticsearchOptions: ElasticsearchOptions,
```
nevillelyh: You can merge the 2 methods and have a default `flushInterval`?

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,38 @@
...
+                            flushInterval: Duration)
+    : Future[Tap[IndexRequestWrapper]] = {
+      self.saveAsCustomOutput("Write to Elasticsearch",
```
nevillelyh: Name in `saveAsCustomOutput` is also used for `JobTest` IO matching and might cause conflict if a job has multiple ES sinks. Generate one from `ElasticsearchOptions`?

nevillelyh: Also you might want to bypass custom IO API and go one level down. Take a look at JDBC IO. You might also wanna create your own `ElasticsearchIO extends TestIO` for `JobTest` wiring.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Default flushInterval is set to 1 second. But depending on cluster size/loads and latency requirement they may have to increase/decrease this value so overloading makes sense. Added a comment indicating default flushInterval.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

build.sbt
lmayumi: Technically we should be able to use latest client. However this code is tested under version 2.1.0 so not sure about changing it without testing the code against ES cluster. Should we wait and see?

build.sbt
lmayumi: Done

build.sbt
lmayumi: Done

lmayumi:
@nevillelyh In response to following comment:
"Are IndexRequestWrapper and ElasticsearchOptions generated from auto-value or handcrafted? Using them isn't super idiomatic in Scio, should we provide some equivalents case classes ?"

I handcrafted the class and code generation tool could sometime cause issue when used with sbt. Do you prefer using auto-value? Also what is the benefit of providing case class counterpart?

build.sbt
nevillelyh: We don't have to. Just curious since Viet Yen asked about 5.x on Gitter.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
nevillelyh: I mean argument with default value, like `flushInterval: Duration = Duration.standardSeconds(1)`, would do what you want but simpler and clearer?

nevillelyh:
Auto-value should work in `.java` files, even in a SBT project? Exposing these 2 classes in the Scala wrapper make it less idiomatic, e.g. the user needs to use builders instead of the nature case class construction.

For example, `ElasticsearchOptions` has two fields. The more Beam idiomatic way is to have 2 methods, `withClusterName` and `withServers` instead of `withElasticsearchOptions` plus the additional `ElasticsearchOptions` class. This will make it easier to submit upstream in the future.
And then `saveAsElasticsearch` can either take 2 args `clusterName: String, servers: Array[InetSocketAddress]` or a case class `ElasticsearchOptions`.

As for `IndexRequestWrapper`, looks like you're just copying fields to `org.elasticsearch.action.index.IndexRequest`. Any reason not to use `IndexRequest` directly in the API? Because `IndexRequest` is not serializable? What about instead take a `f: T => IndexRequest` and delay the `IndexRequest` instance creation?

lmayumi:
Good point about having 2 methods rather than having an option class. Initially I did that but since both of the fields are required, I thought it might be more intuitive to get them both at once. But I agree that its important to follow the convention. Will remove option and get the fields separately. And will create case class for Scio users. 

Not sure why I didn't think of `f: T => IndexRequest` !!! Very clean and smart approach. As you suggested, wrapper was introduced to overcome lack of serializability for `IndexRequest`.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: As option class is going to be a case class, I can provide a default value as you suggested.

lmayumi:
@nevillelyh Actually not sure how `f: T => IndexRequest` is feasible. If my understanding is correct, you are asking for something like `PTransform<PCollection<SerializableSupplier<IndexRequest>>, PDone>` correct? First we need to provide a coder for this new object, and user will have to provide this in their pipeline step. Do you have an example of this type of use case? BigTableIO actually does wrapper approach.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,228 @@
...
+      return new Bound().withServers(servers);
+    }
+
+    /**
+     * Returns a transform for writing to Elasticsearch cluster by providing
+     * slight delay specified by flushInterval.
+     *
+     * @param flushInterval delay applied to buffer elements. Defaulted to 1 seconds.
+     */
+    public static Bound withFlushInterval(Duration flushInterval) {
```
ravwojdyla: Any reason why you are using `java.time` instead of `joda.time`. Both scio and beam use `joda`, this would also save you on converting java to joda below.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,228 @@
...
+      }
+
+      public Bound withFlushInterval(Duration flushInterval) {
+        return new Bound(clusterName, servers, flushInterval);
+      }
+
+      @Override
+      public PDone expand(final PCollection<IndexRequestWrapper> input) {
+
+        if (clusterName == null) {
```
ravwojdyla: you can probably just use `checkNotNull` null checks.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private static class AssignToShard
+        extends DoFn<IndexRequestWrapper, KV<Long, IndexRequestWrapper>> {
+      private int numWorkers;
+
+      @StartBundle
+      public void startBundle(Context c) throws Exception {
+        numWorkers = c
+            .getPipelineOptions()
+            .as(DataflowPipelineOptions.class)
```
ravwojdyla: this won't be runner agnostic.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,228 @@
...
+        // to 1 or ThreadLocalRandom's nextLong method will throw an exception.
+        if (numWorkers == 0) {
+          numWorkers = 1;
+        }
+      }
+
+      @ProcessElement
+      public void processElement(ProcessContext c) throws Exception {
+        // assign this element to a random shard
+        final long shard = ThreadLocalRandom.current().nextLong(2);
```
ravwojdyla: `nextLong(2)` or `nextLong(numWorkers)`?

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,228 @@
...
+
+        input
+            .apply("Assign To Shard", ParDo.of(new AssignToShard()))
+            .apply("Re-Window to Global Window", Window.<KV<Long, IndexRequestWrapper>>into(new GlobalWindows())
+                       .triggering(Repeatedly.forever(
+                           AfterProcessingTime
+                               .pastFirstElementInPane()
+                               .plusDelayOf(javaToJoda(flushInterval))))
+                       .discardingFiredPanes())
+            .apply(GroupByKey.create())
```
ravwojdyla: What is this GBK exactly for? This `numOfWorkers` heuristic is also kinda shady.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,228 @@
...
+        this.clusterName = clusterName;
+        this.addresses = addresses;
+      }
+      @Override
+      public Client get() {
+        if (CLIENT.get() == null) {
+          synchronized (CLIENT) {
+            if (CLIENT.get() == null) {
+              CLIENT.set(create(clusterName, addresses));
+              return CLIENT.get();
```
ravwojdyla: What is the reason for this `return`?

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,36 @@
...
+import scala.concurrent.Future
+
+/**
+  * Main package for Elasticsearch APIs. Import all.
+  *
+  * {{{
+  * import com.spotify.scio.elasticsearch._
+  * }}}
+  */
+package object elasticsearch extends AnyVal {
```
ravwojdyla: What does `extends AnyVal` on `package object` do?

lmayumi: Mistake, its supposed to extend implicit class.

lmayumi:
@nevillelyh Changed the implementation as you suggested and API looks much cleaner and intuitive. 
Changes done is:

1. Taking a function T => IndexRequest in place of IndexRequestWrapper. 
2. Providing ElasticsearchOption as case class for Scio. Following Beam convention by collecting fields via factory in Beam API. 

Let me know if you have any thoughts on it!

build.sbt
lmayumi: If there is a request for latest version, at least there is an incentive to test it to support it.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @rav We could pass in as a constructor argument. Is there a way to get this value via ScioContext?

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @ravwojdyla Good catch! did this for testing.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: Removed.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,231 @@
+package com.spotify.scio.elasticsearch;
```
andrewsmartin: Missing license header here. (We should set up automated checks for license headers on Java sources too, Scalastyle obviously only checks Scala sources).

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,43 @@
+package com.spotify.scio
```
andrewsmartin: Missing license header

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @ravwojdyla Would going higher in interface make it close to agnostic? The function `getNumWorkers` is defined in `DataflowPipelineWorkerPoolOptions`. All the option class should be inheriting from this class so it might generalize it? Wouldn't all runner require same/similar options?

ravwojdyla: What do you actually need this num of workers for? Related to the GBK questions below.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: Not sure if I understand your suggestion but what I am trying to achieve here is to manually create Iterable[T] via sharding and grouping elements assigned to the shard together. They key here is flushInterval which does not exist in default implementation of writer transform which was our problem. Basically without it, this transform writes too fast without batching and too many requests are sent to ES. `numOfWorkers` and rest are arbitrary, as noted above, just trying to create efficient Iterable[T]. Can you elaborate on your suggestion? Thanks!

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @andrewsmartin Where can I copy it from?

andrewsmartin: @lmayumi all source files have it, here for example: https://github.com/spotify/scio/blob/master/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala

lmayumi: @andrewsmartin Thank you!

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
ravwojdyla: irl - agreed here that bundles in map might be too small, thus need to control the size/time via GBK.

lmayumi: 👍

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
ravwojdyla: @lmayumi ?

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
regadas: @lmayumi @ravwojdyla I'm using a similar solution to this and I did 2 things:
1) moved the windowing out of the `PTransform` my reasoning was that the implementation is not tied to a specific strategy. Maybe we can abstract this a bit more and provide 2 `Write` implementations with/without windowing.

2) `numOfshards` can actually be dynamic i.e: according to the number of workers ... might be handy if auto-scaling is in place. something like `() => Int` called at `startBundle` kind of works for me.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,265 @@
...
+
+      public Bound<T> withServers(InetSocketAddress[] servers) {
+        return new Bound<>(clusterName, servers, flushInterval, function, numOfShard);
+      }
+
+      public Bound<T> withFlushInterval(Duration flushInterval) {
+        return new Bound<>(clusterName, servers, flushInterval, function, numOfShard);
+      }
+
+      public Bound<T> withFunction(SerializableFunction<T, IndexRequest> function) {
```
regadas: @lmayumi btw, Have you tried doing the type conversion outside of this?
```scala
// assuming windowing outside as i said in comment below

input.map(Iterable[T] => Iterable[ActionRequest]).saveAsElasticsearch(..)
```

build.sbt
lmayumi: @regadas Is breaking change reflected in client library as well?

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: 1. We dump everything into the global window, and trigger runs as fast as possible with default trigger. This is identical to what writer transform which is included in Beam does. Except for the delay applied to the trigger which is the whole point of this implementation. If we remove this, there is no point of this implementation. There is always an option to pass in sink to default writer transform. 
2. Initially we did exactly that, however which coupled our implementation to a specific runner. Therefore we decided to delegate this logic to user. If you read the comment on the factory, it is recommended to be number of workers.

```
context.getPipelineOptions()
        .as(DataflowPipelineOptions.class)
        .getNumWorkers()
```

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: We need to provide a custom coder in that case. Its better to abstract away application to to that transformation.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,265 @@
...
+        c.output(KV.of(shard, c.element()));
+      }
+    }
+
+    private static class ElasticsearchWriter<T> extends DoFn<KV<Long, Iterable<T>>, Void> {
+      private final Logger LOG = LoggerFactory.getLogger(ElasticsearchWriter.class);
+      private final ClientSupplier clientSupplier;
+      private final SerializableFunction<T, IndexRequest> function;
+
+      public ElasticsearchWriter(String clusterName,
```
lmayumi: @regadas Following Beam/Dataflow convention of providing factory with individual fields instead of accepting object with all fields included.

build.sbt
regadas: @lmayumi iirc, transport implementation is diff.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
regadas: @lmayumi 👍 i was taking as example `BigTableIO` which exposes `BigtableOptions`

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
regadas: @lmayumi imho, this kind of window in combination with just `IndexRequest` is kinda limited and is only useful if you are just doing indexing (append only)

My idea is more along this lines:

Use `ActionRequest` instead then we start allowing also deletes. Allows more flexible indexing strategies.

With this, things change a little bit depending on what you want. Given a stream of `[D, I, ...]` `D` and `I` might end up in different shards aka different bulk requests. Your use case might allow this however, If I want to make `D and I` in the same bulk request(offers other guarantees) this is not possible. This means that this window/sharding should be provided as optional.

Anyway this is just my thoughts around this.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @regadas ElasicsearchOption is available as a case class from Scio. Take a look at package object.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @ravwojdyla Done

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

lmayumi:
@nevillelyh @ravwojdyla Please take a look at test.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,77 @@
...
+      * @param elasticsearchOptions defines clusterName and cluster endpoints
+      * @param flushInterval delays writes to Elasticsearch cluster to rate limit
+      * @param f transforms arbitrary type T to the object required by Elasticsearch client
+      * @param numOfShard number of parallel writes to be performed.
+      *                   Note: Recommended to be equal to number of workers in your pipeline.
+      */
+    def saveAsElasticsearch(elasticsearchOptions: ElasticsearchOptions,
+                            flushInterval: Duration = Duration.standardSeconds(1),
+                            f: T => IndexRequest,
+                            numOfShard: Long) :Future[Tap[T]] = {
```
ravwojdyla: You could default this to `0` and the do something similar like [here](https://github.com/spotify/scio/blob/master/scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala#L143) to just apply your recommendation from the comment.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,249 @@
...
+     * Returns a transform for writing to Elasticsearch cluster.
+     * Note: Recommended to set this number as number of workers in your pipeline.
+     *
+     * @param numOfShard to construct a batch to bulk write to Elasticsearch.
+     */
+    public static<T> Bound withNumOfShard(Long numOfShard) {
+      return new Bound<>().withNumOfShard(numOfShard);
+    }
+
+    public static class Bound<T> extends PTransform<PCollection<T>, PDone> {
```
ravwojdyla: Add `@Nullable` to nullable fields.

ravwojdyla: Could you override `populateDisplayData`?

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,249 @@
...
+
+      @ProcessElement
+      public void processElement(ProcessContext c) throws Exception {
+        // assign this element to a random shard
+        final long shard = ThreadLocalRandom.current().nextLong(numOfShard);
+        c.output(KV.of(shard, c.element()));
+      }
+    }
+
+    private static class ElasticsearchWriter<T> extends DoFn<KV<Long, Iterable<T>>, Void> {
```
ravwojdyla: Add `populateDisplayData` if it makes sense here.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,265 @@
...
+        // Elasticsearch throws ActionRequestValidationException if bulk request is empty,
+        // so do nothing if number of actions is zero.
+        if (bulkRequestBuilder.numberOfActions() == 0) {
+          LOG.info("ElasticsearchWriter: no requests to send");
+          return;
+        }
+
+        final BulkResponse bulkItemResponse = bulkRequestBuilder.get();
+        if (bulkItemResponse.hasFailures()) {
+          throw new IOException(bulkItemResponse.buildFailureMessage());
```
lmayumi: @regadas Done

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @ravwojdyla Done

build.sbt
lmayumi: @regadas Do you suggest we support 5.X instead of 2.X? @ravwojdyla mentioned that we should sync with version of Elasticsearch that would be available if we were to setup cluster in GCP. Any thoughts?

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,117 @@
...
+/**
+  * Main package for Elasticsearch APIs. Import all.
+  *
+  * {{{
+  * import com.spotify.scio.elasticsearch._
+  * }}}
+  */
+package object elasticsearch {
+
+  case class ElasticsearchIOTest[T](options: ElasticsearchOptions)
```
nevillelyh: Rename this to `ElasticsearchIO` to be consistent with the other IOs? If it conflicts with `ElasticsearchIO.java` move that class to somewhere under `org.apache.beam`?

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,117 @@
...
+    extends TestIO[T](options.toString)
```
nevillelyh: `.toString` here may not work since `servers` is an `Array`.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,117 @@
...
+      } else {
+        self.applyInternal(
+          ElasticsearchIO.Write
+            .withClusterName(elasticsearchOptions.clusterName)
+            .withServers(elasticsearchOptions.servers)
+            .withNumOfShard(numOfShard)
+            .withFunction(new SerializableFunction[T, IndexRequest]() {
+              override def apply(t: T): IndexRequest = f(t)
+            })
+          .withError(new SerializableConsumer[String]() {
```
nevillelyh: Indentation

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.joda.time.Duration
+
+object ElasticsearchJobSpec {
+  val options = new ElasticsearchOptions("clusterName", Array(new InetSocketAddress(8080)))
+  val data = Seq(1, 2, 3)
+  val shard = 2
+  val toIndexRequest = (x: Int) => new IndexRequest()
+  val flushInterval = Duration.standardSeconds(1)
+}
+object ElasticsearchJob {
```
nevillelyh: blank line between objects and classes.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -0,0 +1,117 @@
...
+    /**
+      * Save this SCollection into Elasticsearch.
+      *
+      * @param elasticsearchOptions defines clusterName and cluster endpoints
+      * @param f transforms arbitrary type T to the object required by Elasticsearch client
+      * @param e handles custom error in case of bulk write by Elasticsearch client
+      */
+    def saveAsElasticsearch(elasticsearchOptions: ElasticsearchOptions,
+                            f: T => IndexRequest,
+                            e: String => Unit) :Future[Tap[T]] = {
```
nevillelyh: should `e` have a default value? Also it's not very clear how it should be implemented from just the signature.

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
```diff
@@ -0,0 +1,63 @@
...
+  import ElasticsearchJobSpec._
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, _) = ContextAndArgs(cmdlineArgs)
+    sc.parallelize(data).saveAsElasticsearch(options, flushInterval, toIndexRequest, shard,_=> ())
```
nevillelyh: Space between `shard,_`

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
```diff
@@ -0,0 +1,63 @@
...
+    sc.close()
+  }
+}
+object ElasticsearchDirectRunnerJob {
+  import ElasticsearchJobSpec._
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, _) = ContextAndArgs(cmdlineArgs)
+    sc.parallelize(data).saveAsElasticsearch(options, toIndexRequest,_=> ())
```
nevillelyh: Space between `,_`

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
```diff
@@ -0,0 +1,63 @@
...
+    sc.close()
+  }
+}
+class ElasticsearchTest extends PipelineSpec {
+  import ElasticsearchJobSpec._
+  "ElasticsearchIO" should "work" in {
+    JobTest[ElasticsearchJob.type]
+      .output(ElasticsearchIOTest[Int](options))(_ should containInAnyOrder(data))
```
nevillelyh: Space between `containInAnyOrder(` like in rest of the code base.

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
```diff
@@ -0,0 +1,63 @@
...
+      .run()
+  }
+  "ElasticsearchIO with DirectRunner" should "work" in {
+    JobTest[ElasticsearchDirectRunnerJob.type]
+      .output(ElasticsearchIOTest[Int](options))(_ should containInAnyOrder(data))
```
nevillelyh: Space between `containInAnyOrder(` like in rest of the code base.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @regadas To reiterate your point, you want this IO to be able to write Delete request as well as Index request. However in that case, we cannot batch write them as a part of same bulk request. It is doable if we introduce an interface which identifies among D or I in that way, we can group by shard and type. However, this feels like an over engineering, I say let's put this out there and wait to see if there is a request for this before we over-complicate this impl. 

For side story, this class was worked with Google engineer and they asked us to open source it under Beam. That's why the impl is very close to writer transform.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: @nevillelyh Just to confirm are you suggesting to create a path org.apache.beam under src.main.java?

nevillelyh: Yeah, under `src/main/java`, something like `org.apache.beam.sdk.io.elasticsearch` similar to upstream.

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
lmayumi: Done

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
lmayumi: Done

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
lmayumi: Done

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
lmayumi: Done

scio-elasticsearch/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
lmayumi: Done

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Renamed identifier to more intuitive name: `errorHandle`. Also included default implementation to throw an exception as an example.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: @nevillelyh Overridden toString

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

lmayumi:
@nevillelyh I think I addressed all the comments.

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
regadas: @lmayumi no need to implement such interface coz already exists as `ActionRequest` part of `elasticsearch` api. Every `IndexRequest` `DeleteRequest` `UpdateRequest` etc extends it. BulkRequests allow `[ActionRequest]`. You mention grouping by shard or type ... those should be left to user implementation ... and depending on the use case this `"sharding"` might not work ... so imo offering it as optional is better. 

My use case requires the above, and i think I'm failing to explain/convey my idea so i'll gladly purpose a PR after this one.

build.sbt
regadas: @lmayumi well honestly not sure ... is providing both versions out of question? 😄 looking at some sort of maven stats ... both 2.X and 5.X show a lot of downloads... I'm also a little bit biased cause i have most of the stuff in 2.X

scio-elasticsearch/src/main/java/com/spotify/scio/elasticsearch/ElasticsearchIO.java
lmayumi: @regadas Problem is, they are not serializable so we will have to. And provide a custom coder.

regadas:
@lmayumi made this for your consideration 

https://github.com/regadas/scio/tree/pr/599

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
lmayumi: Done

lmayumi:
@regadas What am I looking for in your fork? If you would like, you can submit a changes after this PR is done. I feel like for general use case what we have is sufficient.

lmayumi:
@nevillelyh @ravwojdyla Can we merge this PR?

regadas:
@lmayumi lets do that 👍

PR #601, opened 2017-05-06
Add Bigtable TableAdmin utility to ensure tables

tchagnon:


tchagnon:
Let me know if you want this somewhere else (ScioContext extension?) or have other style comments.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -0,0 +1,113 @@
...
+import com.google.cloud.bigtable.grpc.BigtableTableAdminGrpcClient
+import org.slf4j.{Logger, LoggerFactory}
+
+import scala.collection.JavaConverters._
+
+/**
+ * Bigtable Table Admin API helper commands.
+ */
+object TableAdmin {
+  val log: Logger = LoggerFactory.getLogger(TableAdmin.getClass)
```
nevillelyh: Make this private?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Ensure that tables and column families exist.
+   * Checks for existence of tables or creates them if they don't exist.
+   * Then calls `ensureColumnFamilies` to ensure that the column families exist as well.
+   *
+   * @param tablesAndColumnFamilies A map of tables and column families.  Keys are table names,
+   *                                Values are a list of column family names.
```
nevillelyh: "values are" or make the comma before a full stop?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -0,0 +1,113 @@
...
+      channel.shutdownNow()
+    }
+  }
+
+  /**
+   * Ensure that column families exist.
+   * Checks for existence of column families and creates them if they don't exist.
+   *
+   * @param tablePath A full table path that the bigtable API expects, in the form of
+   *                  `projects/&lt;projectid&gt;/instances/&lt;instance&gt;/tables/&lt;table&gt;`
```
nevillelyh: Does "&lt;" etc format properly in the generated doc? Or use "$name" instead?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -0,0 +1,113 @@
...
+   * @param columnFamilies A list of column family names.
+   */
+  def ensureColumnFamilies(client: BigtableTableAdminGrpcClient,
```
nevillelyh: Why does`ensureTables` take an options but this takes a client?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
tchagnon: 👍

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
tchagnon: 👍

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
tchagnon: I'll switch to $name. < doesn't format correctly.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
tchagnon: Good point.  This is really just an internal.  I'll make it private.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
tchagnon: Ugh, so $name gets interpreted as a scaladoc macro http://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#macros and there doesn't appear to be an escape sequence.  I'm just going to remove both all symbols and since this fn will be private, it won't end up in the generated docs anyway.

nevillelyh:
Awesome thanks!

PR #602, opened 2017-05-08
Add TableAdmin.ensureTables to BigtableExample

tchagnon:


tchagnon:
I should have added this to the last PR, but just saw that the Wiki points to this example.  Let me know if you have better ideas for the mock setup.

nevillelyh:
Maybe add a convenience method to `BigtableScioContext` like `updateNumberOfBigtableNodes` and bypass this `if (sc.isTest)`? No need to go crazy with mocking and testing.

PR #607, opened 2017-05-10
Add rate limiting DoFn

rgruener:

So this was necessary for calling external services from dataflow to guarantee that I will not overload them. This also is a way to solve controlling the number of write requests to bigtable (placing this before `saveAsBigtable`)

I don't love this since it kind of relies on knowing where the serialization boundaries are so its possible to use this and have a map that you expect to be rate limited not actually be rate limited. Perhaps something better would be to have an optional rate limit on maps and transforms something like: `applyTransform(myParDo, rateLimit=1000)` though that also isnt super clean since it would need to be added to many methods. For now this does work, having it in scio-extra might not be the worst since if you know what youre doing this works fine.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
@@ -23,6 +23,9 @@ import java.nio.file.Path
 
 import com.spotify.scio.util.{Functions, RemoteFileUtil}
 import com.spotify.scio.values.SCollection
+import org.apache.beam.runners.dataflow.DataflowRunner
+import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions
+import org.apache.beam.runners.direct.DirectRunner
 import org.apache.beam.sdk.transforms.ParDo
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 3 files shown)


scio-extra/src/test/scala/com/spotify/scio/extra/transforms/RateLimiterDoFnTest.scala
```diff
@@ -0,0 +1,48 @@
...
+    val timeStart = System.nanoTime()
+    runWithContext { sc =>
+      val p1 = sc.parallelize(input).rateLimitThroughput(1)
+      p1 should containInAnyOrder (input)
+    }
+    val timeElapsed = TimeUnit.NANOSECONDS.toSeconds(System.nanoTime() - timeStart)
+    timeElapsed should be >= input.size.toLong
+  }
+
+  "RateLimiterDoFn" should "work with larger rate" in {
```
nevillelyh: This could be simplified as `it should`.

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
```diff
@@ -112,4 +116,42 @@ package object transforms {
 
   }
 
+  /**
```
nevillelyh: Use JavaDoc indentation. See https://github.com/spotify/scio/wiki/Style-Guide#intellij-idea

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
```diff
@@ -112,4 +116,42 @@ package object transforms {
...
+      *
+      * Note that this only limits the throughput for the current step as well as any steps
+      * running within the same machine (within a serialization boundary). Therefore any maps
+      * and filters surrounding this will be properly rate limited usually until a groupBy, reduce,
+      * or join (or anything that uses those under the hood e.g. distinct, fold) is hit.
+      *
+      * @param maxElementsPerSecond The maxmimum number of elements which should be processed
+      *                             per second
+      */
+    def rateLimitThroughput(maxElementsPerSecond: Double): SCollection[T] = {
```
nevillelyh: Does `withRateLimit` sound more natural here?

rgruener: I think it does, will change

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
rgruener: Sure, though I did copy this from above. It seems that all implicit classes use scalaDoc

nevillelyh:
Actually I'm not confident that this would work depending on DoFn fusion implementation. If this DoFn outputs to a buffer and the next one picks it up, it might defeat the rate limiting and go at full speed again?

rgruener:
Well I can say with confidence that testing this did work, though yeah its a little weird since it could or could not work depending on a lot of things.

rgruener:
The more i think about this, I am worried this is dangerous as in one day it will work and 2 months from now it might just stop. Which could then cause large issues if you relied on this to not bring down a backend service.

rgruener:
Perhaps we could explore adding an optional rate limit to things like `applyTransform` and `saveAsBigtable` explicitly?

nevillelyh:
+1 on adding to bigtable methods directly

rgruener:
Closing this for now, might reopen with new implementation.

regadas:
@rgruener @nevillelyh interesting stuff and i do understand the issues ... imo it might still be useful to provide this as is and find someway of saying ... "use this with caution".

rgruener:
reopening this since it solves a use case and although does not actually make any rate limiting guarantees does work well enough to be useful. It is use at your own risk

rgruener:
Actually need to open a new one since I force pushed :man_shrugging:

PR #609, opened 2017-05-11
Fix #606 do not support accumulator retrieval during template staging.

### LINKED ISSUE #606: Accumulators throw NPE on template staging
```
:scioOpts --templateLocation=<location> --runner=DataflowRunner
:newScio
val foo = sc.sumAccumulator[Long]("foo")
sc.parallelize(1 to 10).accumulateCount(foo)
val r = sc.close()
r.accumulatorTotalValue(foo)
```

will throw NPE:

```
  at org.apache.beam.runners.dataflow.DataflowPipelineJob.fromMetricUpdates(DataflowPipelineJob.java:471)
  at org.apache.beam.runners.dataflow.DataflowPipelineJob.getAggregatorValues(DataflowPipelineJob.java:443)
  at com.spotify.scio.ScioResult$$anonfun$getAggregatorValues$2.apply(ScioResult.scala:177)
  at com.spotify.scio.ScioResult$$anonfun$getAggregatorValues$2.apply(ScioResult.scala:177)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
  at scala.collection.immutable.Set$Set1.foreach(Set.scala:94)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
  at scala.collection.AbstractSet.scala$collection$SetLike$$super$map(Set.scala:47)
  at scala.collection.SetLike$class.map(SetLike.scala:92)
  at scala.collection.AbstractSet.map(Set.scala:47)
  at com.spotify.scio.ScioResult.getAggregatorValues(ScioResult.scala:177)
  at com.spotify.scio.ScioResult.accumulatorTotalValue(ScioResult.scala:101)
  ... 32 elided
```

This should throw more meaningful error/message.

Coming from #604

andrewsmartin:


andrewsmartin:
@ravwojdyla PTAL when you get a chance. I'm actually not sure if this is the best solution since I don't think there's a way to get the accumulator values when actually running the templated job if we prevent them like this in the first place. Maybe it's better to just stub out the impl during template staging so we can still allow the call in the pipeline code?

scio-test/src/test/scala/com/spotify/scio/ScioResultTest.scala
```diff
@@ -30,4 +29,20 @@ class ScioResultTest extends PipelineSpec {
     r.state shouldBe State.DONE
   }
 
+  it should "not support retrieval of accumulators during template creation" in {
+    val sc = ScioContext.forTest()
```
ravwojdyla: Any reason why not `runWithContext`?

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -176,4 +178,8 @@ class ScioResult private[scio] (val internal: PipelineResult,
     aggregators.getOrElse(acc.name, Nil)
       .map(a => internal.getAggregatorValues(a.asInstanceOf[Aggregator[_, T]]))
 
+  private def isTemplateCreationJob: Boolean =
+    Try(context.optionsAs[DataflowPipelineOptions].getTemplateLocation)
```
ravwojdyla: 👍

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -176,4 +178,8 @@ class ScioResult private[scio] (val internal: PipelineResult,
     aggregators.getOrElse(acc.name, Nil)
```
ravwojdyla: So this is the method that actually throws the exception IIRC. How about `require` here? This is also more future proof in case there is another methods that wants to access `internal.getAggregatorValues`? What do you think?

scio-test/src/test/scala/com/spotify/scio/ScioResultTest.scala
andrewsmartin: It's just so that I can use the `maxI` accumulator in the assertion. With `runWithContext` the declaration is locked into the inner scope, but the `ScioResult` is accessed in the outer scope where `maxI` is not defined.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
andrewsmartin: +1 makes sense to lock it down here instead. What do you think of this in general though (see my other comment), is it better to stub out the impl rather than prevent? So that it can still be used when actually running the template job.

scio-test/src/test/scala/com/spotify/scio/ScioResultTest.scala
ravwojdyla: 👍

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
ravwojdyla: Yeah maybe let's do what you suggest stub the aggregators in template staging + print warning maybe, this would allow people to just create templates without code change, while still be informed that it's only the computation DAG that is executed.

andrewsmartin: :+1:

andrewsmartin:
Closing this, will reopen with stub implementation in new PR.

PR #610, opened 2017-05-12
Better error message for materialize in test.

andrewsmartin:


nevillelyh:
I'm poking around `JobTestTest` and `runWithContext` to figure this out.

nevillelyh:
So the `JobTestTest` behavior is expected. `.materialized` should be ignored in a `JobTestTest`. However `runWithContext` doesn't setup IO wiring in `TestManager` and that's why it's complaining.

nevillelyh:
Having a proper fix will submit soon. Closing this.

PR #611, opened 2017-05-12
Add implicit conversions to java functions

rgruener:
@nevillelyh In the meantime this is super useful for dealing with the AsyncDoFns without writing super verbose scala


nevillelyh:
https://github.com/scala/scala-java8-compat
This does the same thing?

rgruener:
Thats actually slightly annoying since its not implicits but it still works.

PR #612, opened 2017-05-12
add check for test IO outside of JobTest

nevillelyh:


nevillelyh:
@andrewsmartin @richwhitjr take a look?

andrewsmartin:
+1 Nice, although is the error message clear enough for `.materialize`? I know it's output under the hood but might not be clear to the user.

nevillelyh:
It's hard to be more specific coz `saveAs*` are all outputs and can't easily tell from which.

andrewsmartin:
Yeah - alright, I think it's fine then

PR #615, opened 2017-05-15
Remove accumulator API

andrewsmartin:
This will be replaced with the Beam metrics API.


andrewsmartin:
Will reopen against correct base branch.

PR #622, opened 2017-05-16
Fix windowing API.

andrewsmartin:


scio-core/src/main/scala/com/spotify/scio/values/WindowedSCollection.scala
```diff
@@ -32,7 +32,7 @@ case class WindowOptions[W <: BoundedWindow](trigger: Trigger = null,
                                              accumulationMode: AccumulationMode = null,
                                              allowedLateness: Duration = null,
                                              closingBehavior: ClosingBehavior = null,
-                                             outputTimeFn: OutputTimeFn[BoundedWindow] = null)
+                                             timestampCombiner: TimestampCombiner = null)
```
andrewsmartin: @nevillelyh With this change can we remove the `[W <: BoundedWindow]` constraint?

nevillelyh: Probably yes.

PR #623, opened 2017-05-16
Bring back junit-interface from 34f0d83

ravwojdyla:


ravwojdyla:
And you can see that now the junit tests are run [here](https://circleci.com/gh/spotify/scio/1013), previous [build](https://circleci.com/gh/spotify/scio/1000).

PR #629, opened 2017-05-18
Replace IOChannelFactory with FileSystems utility.

andrewsmartin:


nevillelyh:
Instead of fixing the java examples one by one, do another code dump and copy all java examples files Beam (v2.0.0 tag) directly?

andrewsmartin:
Good call, I'll leave the examples alone for now and do all at once later.

andrewsmartin:
Removed the Java example change.

PR #634, opened 2017-05-18
Fixes for TFRecord

andrewsmartin:
@nevillelyh Thoughts? Not sure if this is what you had in mind, but it should be a good starting point. I removed the sink entirely (since compression behavior is the same between the internal sink and the one in Beam), and kept the source for now, but fixed it to work with the new API. 

As far as I can tell there isn't an easy way to reuse more of the beam code if we want to keep the auto-detection because Beam's `TFRecordSource` is not exposed (unless we want to resort to package hackery).


nevillelyh:
We can remove `TFRecord.scala` including the source entirely. I doubt anyone's using auto compression detection and it's an easy fix in the user code anyway.

PR #639, opened 2017-05-24
Add checkpoints

ravwojdyla:


yonromai:
Looks really useful! I've played with checkpoints on my own and came up with a very similar API 
* Was considering pimping SCollection to avoid have to pass sc as a parameter
* Also, wondering if we could avoid having to name the checkpoint by using some hash of the function/context, probably over-engineering though 
 * An example would be useful I think, although tests might be enough

yonromai:
oh, well, hopefully it makes it into beam :)

ravwojdyla:
@nevillelyh @andrewsmartin any comments?

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -84,4 +89,32 @@ private[scio] object ScioUtil {
       s"Could not get dataflow metrics of ${getMetrics.getJobId} in ${getMetrics.getProjectId}")
   }
 
+  def addPartSuffix(path: String, ext: String = ""): String =
+    if (path.endsWith("/")) s"${path}part-*$ext" else s"$path/part-*$ext"
+
+  def getTempFile(@transient context: ScioContext, fileOrPath: String = null): String = {
```
andrewsmartin: What does the `@transient` do here? I've only seen it used for fields.

scio-extra/src/main/scala/com/spotify/scio/extra/checkpoint/package.scala
```diff
@@ -0,0 +1,60 @@
...
+     * Checkpoints are useful for debugging one part of a long flow, when you would otherwise have
+     * to run many steps to get to the one you care about.  To enable checkpoints, sprinkle calls to
+     * [[checkpoint[T]()]] throughout your flow, ideally after expensive steps.
+     *
+     * @param fileOrPath filename or fully qualified path to the checkpoint
+     * @param fn result of this arbitrary [[ScioContext]] => [[SCollection[T]] flow is what is
+     *           checkpointed
+     */
+    def checkpoint[T: ClassTag](fileOrPath: String,
+                                fn: ScioContext => SCollection[T]): SCollection[T] = {
```
andrewsmartin: Would it be good to have the `fn` in a separate parameter list? That way blocks could be used.

andrewsmartin:
Does this work for specifying multiple checkpoints?

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -18,23 +18,28 @@
...
+import org.apache.beam.sdk.options.{GcpOptions, PipelineOptions}
 import org.apache.beam.sdk.util.Transport
+import org.slf4j.LoggerFactory
 
 import scala.reflect.ClassTag
+import scala.util.{Failure, Success}
 
 private[scio] object ScioUtil {
 
+  @transient lazy private val log = LoggerFactory.getLogger(this.getClass)
```
nevillelyh: `private lazy val` so it's easy to grep for `lazy val`?
~There's one in `SCollection` as well but we can fix that separately.~ Just realized this PR removed that one.

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -84,4 +89,32 @@ private[scio] object ScioUtil {
...
+    if ((ScioUtil.isLocalUri(uri) && uri.toString.startsWith("/")) || uri.isAbsolute) {
+      fop
+    } else {
+      val filename = fop
+      val tmpDir = if (context.options.getTempLocation != null) {
+        context.options.getTempLocation
+      } else {
+        val m =
+          "Specify a temporary location via --tempLocation or PipelineOptions.setTempLocation."
+        scala.util.Try(context.optionsAs[GcpOptions].getGcpTempLocation) match {
```
nevillelyh: Import `Try` as well? It's commonly used and not need to use full name.

ravwojdyla:
@andrewsmartin 
 * https://issues.apache.org/jira/browse/BEAM-1458
 * should work for multiple checkpoints - added it as part of a job test

ravwojdyla:
@andrewsmartin @nevillelyh updated to code:
  * made it easier to test job with checkpoints
  * update tempLocation before pipeline creation

Any other comments?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/CheckpointExample.scala
```diff
@@ -0,0 +1,39 @@
...
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.ContextAndArgs
+import com.spotify.scio.extra.checkpoint._
+
+object CheckpointExample {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    val words = sc.checkpoint(args("checkpoint") + "-src"){
```
nevillelyh: Add a few comments here explaining what they do? Also space between `) {`

scio-examples/src/main/scala/com/spotify/scio/examples/extra/CheckpointExample.scala
```diff
@@ -0,0 +1,39 @@
...
+      sc.textFile(args("input"))
+        .map(_.trim)
+        .flatMap(_.split("[^a-zA-Z']+").filter(_.nonEmpty))
+      }
```
nevillelyh: `}` should align with `val words` not `sc.textFile`?

scio-extra/src/main/scala/com/spotify/scio/extra/checkpoint/package.scala
```diff
@@ -0,0 +1,82 @@
...
+ * import com.spotify.scio.extra.checkpoint._
+ * }}}
+ */
+package object checkpoint {
+
+  // scalastyle:off method.name
+  /**
+   * For use in testing, see [[CheckpointExampleTest]].
+   */
+  def CheckpointIO[T](fileOrPath: String): ObjectFileIO[T] = ObjectFileIO[T](fileOrPath)
```
nevillelyh: This is cool. I wonder if it's worth changing some other TestIOs to use this trick? IIRC `ProtoIO` also uses `ObjectFileIO` under the hood?

ravwojdyla: Would probably not buy us much given that there are two paths for saveAsPB - one for testing with protobuf objects and other after conversion to avro. In this case checkpoints use object IO for both paths, so had to do this trick.

PR #642, opened 2017-05-30
Add default bucket utils.

andrewsmartin:
Note: logic copied from https://github.com/apache/beam/pull/2803 and https://github.com/apache/beam/pull/2747


andrewsmartin:
These changes are due to https://issues.apache.org/jira/browse/BEAM-1871

andrewsmartin:
@nevillelyh could you please review this one?

PR #646, opened 2017-05-31
Fix NamedTransformTest

andrewsmartin:


andrewsmartin:
@nevillelyh please review this one too. I would have liked to avoid the `var` and just do the assertion as part of the visitor traversal but that will silently pass tests if for some reason there was a problem with the traversal (like passing in wrong collection for example), so it's a bit of a hack to ensure it fails by default. Open to other solutions.

PR #656, opened 2017-06-06
Fix #641. Abort on null.

### LINKED ISSUE #641: NPE on REPL creation/redrawLine
seems like 0dad9f4ca0f69d47026da6e6f65e3cff69402fa3 introduced a bug where `in` might still be `null` at the time of `in.redrawLine()`.

ravwojdyla:


scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
```diff
@@ -240,6 +240,11 @@ class ScioILoop(scioClassLoader: ScioReplClassLoader,
       newScioCmdImpl("sc")
       loadIoCommands()
     }
+    if (in == null) {
+      sys.error("Could not initialize Scio interpreter - abort." +
+        "One possible reason is inconsistency between scala versions, please use exactly the same" +
```
nevillelyh: Minor nitpick "inconsistent Scala versions. Please use the exact same version of Scala as scio-repl"

PR #659, opened 2017-06-07
Improvements around elasticsearch

regadas:
* improved error. Added bulk execution exception with the respective failures.
* usage of `Iterable[ActionRequest]`.
  * Allows `IndexRequest` `DeleteRequest` and `UpdateRequest`
  * Actions happen in the same bulk request
  * Allows indexing `T` in different indices

@nevillelyh @ravwojdyla @andrewsmartin @lmayumi

### DIFF
--- a/scio-elasticsearch/src/main/java/org/apache/beam/sdk/io/elasticsearch/ThrowingConsumer.java
+++ b/scio-elasticsearch/src/main/java/org/apache/beam/sdk/io/elasticsearch/ThrowingConsumer.java
@@ -18,8 +18,9 @@
 package org.apache.beam.sdk.io.elasticsearch;
 
 import java.io.Serializable;
-import java.util.function.Consumer;
 
-public interface SerializableConsumer<T> extends Consumer<T>, Serializable {
+@FunctionalInterface
+public interface ThrowingConsumer<T> extends Serializable {
 
+  <X extends Throwable> void accept(T t) throws X;
 }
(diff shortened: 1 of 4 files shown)


scio-elasticsearch/src/main/java/org/apache/beam/sdk/io/elasticsearch/ThrowingConsumer.java
```diff
@@ -18,8 +18,9 @@
 package org.apache.beam.sdk.io.elasticsearch;
 
 import java.io.Serializable;
-import java.util.function.Consumer;
 
-public interface SerializableConsumer<T> extends Consumer<T>, Serializable {
+@FunctionalInterface
+public interface ThrowingConsumer<T> extends Serializable {
```
lmayumi: @regadas Maybe should we make this more specific by changing above to:
```
public interface ThrowableConsumer<T extends BulkExecutionException> extends Serializable {}
```
For given function, it would be useful to be able to access methods specific to BulkExecutionException? As this function is *only* applied with this specific exception type. In that was you can just throw T as well.

lmayumi:
@regadas I see what you mean by Iterable[ActionRequest] now. This makes sense. Nice improvement 👍 One comment on TrowableConsumer other than that looks good!

scio-elasticsearch/src/main/java/org/apache/beam/sdk/io/elasticsearch/ThrowingConsumer.java
regadas: @lmayumi good point ... this error handler is just for the bulk exception. I'll keep ThrowableConsumer generic enough but simpler ... let me know about the following fixup.

PR #663, opened 2017-06-08
saveAsElasticsearch api style

regadas:
wondering what do you think of this?

@nevillelyh @ravwojdyla @andrewsmartin @lmayumi

### DIFF
--- a/scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
+++ b/scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
@@ -50,15 +50,9 @@ package object elasticsearch {
   case class ElasticsearchOptions(clusterName: String, servers: Seq[InetSocketAddress])
 
   implicit class ElasticsearchSCollection[T](val self: SCollection[T]) extends AnyVal {
-    /**
-     * Save this SCollection into Elasticsearch.
-     *
-     * @param esOptions Elasticsearch options
-     * @param f function to transform arbitrary type T to Elasticsearch [[ActionRequest]]
-     */
-    def saveAsElasticsearch(esOptions: ElasticsearchOptions,
-                            f: T => Iterable[ActionRequest[_]]):Future[Tap[T]] = {
-      val numOfWorkers = self.context.pipeline.getRunner match {
+
+    private def defaultNumOfShards(): Int = {
+      self.context.pipeline.getRunner match {
         case _: DirectRunner => 1
         case _: DataflowRunner =>
           val opts = self.context.optionsAs[DataflowPipelineOptions]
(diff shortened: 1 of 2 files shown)


scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -82,13 +74,13 @@ package object elasticsearch {
...
+     * @param e function to handle error when performing Elasticsearch bulk writes
      */
     def saveAsElasticsearch(esOptions: ElasticsearchOptions,
-                            f: T => Iterable[ActionRequest[_]],
                             flushInterval: Duration = Duration.standardSeconds(1),
-                            numOfShard: Long,
-                            errorHandler: BulkExecutionException => Unit): Future[Tap[T]] = {
+                            numOfShard: Long = defaultNumOfShards())
+                           (f: T => Iterable[ActionRequest[_]],
+                            e: BulkExecutionException => Unit = m => throw m): Future[Tap[T]] = {
```
nevillelyh: Curry `f` and `e` so that one can write `.saveAsElasticsearch(options) { t => ... }`?
Or maybe put `e` in the first group so it's `(esOptions, flushInterval, numOfShards, e)`?
While you're at it, rename `numOfShard` => `numOfShards` so it's consistent with other IOs.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -50,15 +50,9 @@ package object elasticsearch {
...
-     * Save this SCollection into Elasticsearch.
-     *
-     * @param esOptions Elasticsearch options
-     * @param f function to transform arbitrary type T to Elasticsearch [[ActionRequest]]
-     */
-    def saveAsElasticsearch(esOptions: ElasticsearchOptions,
-                            f: T => Iterable[ActionRequest[_]]):Future[Tap[T]] = {
-      val numOfWorkers = self.context.pipeline.getRunner match {
+
+    private def defaultNumOfShards(): Int = {
```
ravwojdyla: ~`()`~

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -82,13 +74,13 @@ package object elasticsearch {
      * @param f function to transform arbitrary type T to Elasticsearch [[ActionRequest]]
      * @param numOfShard number of parallel writes to be performed, recommended setting is the
      *                   number of pipeline workers
-     * @param errorHandler function to handle error when performing Elasticsearch bulk writes
+     * @param e function to handle error when performing Elasticsearch bulk writes
```
ravwojdyla: I actually prefer `errorHandler` or `errorFn`, but `e` in public API is probably a little bit too cryptic.

scio-elasticsearch/src/main/scala/com/spotify/scio/elasticsearch/package.scala
regadas: @nevillelyh yup that's the idea ... moving in it to the first group gives a nice usage also

PR #666, opened 2017-06-12
Fix #624. Backport BEAM-1835

### LINKED ISSUE #624: Backport BEAM-1835 to scio 0.3.x
It's a NPE in DirectRunner, it's solved in Beam 2.x, but until we are on that version would be nice to solve backport given that people keep raising this issue.

ravwojdyla:


build.sbt
```diff
@@ -225,7 +225,8 @@ lazy val scioCore: Project = Project(
     "com.fasterxml.jackson.module" %% "jackson-module-scala" % jacksonScalaModuleVersion,
     "com.google.auto.service" % "auto-service" % autoServiceVersion,
     "me.lyh" %% "protobuf-generic" % protobufGenericVersion,
-    "junit" % "junit" % junitVersion % "provided"
+    "junit" % "junit" % junitVersion % "provided",
```
nevillelyh: Should this be `"test"`?

ravwojdyla: Seems like it would be fine with `test`, but since unrelated to this PR will push in separate commit.

Issue #600, opened 2017-05-05
Update to BEAM-59


nevillelyh:
`IOChannelFactory` is removed and there might be some significant breaking changes in IO components. Need to update our codebase and prepare for that.

nevillelyh:
2.0.0-RC4 is out so we should get started with the migration.

ravwojdyla:
Closing this - it's required/explicit in the beam 2.x / scio 4.0

PR #667, opened 2017-06-12
Disable scio-hdfs and scio-examples, fix checkpoint test.

andrewsmartin:
I'd like to disable the `scio-hdfs` and `scio-examples` projects for now to make it easier to focus on migrating everything else. We might be able to remove `scio-hdfs` completely but we need to address https://github.com/spotify/scio/issues/600 first.


andrewsmartin:
Also this change means there is no more compilation failures so we can start regularly getting test feedback.

PR #676, opened 2017-06-15
TF API

ravwojdyla:


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/package.scala
```diff
@@ -0,0 +1,47 @@
...
+  implicit def makeTensorFlowSCollectionFunctions[T: ClassTag](s: SCollection[T])
+  : TensorFlowSCollectionFunctions[T] = new TensorFlowSCollectionFunctions(s)
+
+  /** Implicit conversion from [[SCollection]] to [[TFRecordSCollectionFunctions]]. */
+  implicit def makeTFRecordSCollectionFunctions[T](s: SCollection[T])
+                                                  (implicit ev: T <:< Array[Byte])
+  : TFRecordSCollectionFunctions[T] = new TFRecordSCollectionFunctions(s)
+
+  /** Implicit conversion from [[ScioContext]] to [[TFRecordSCollectionFunctions]]. */
+  implicit def makeTFScioContextFunctions(s: ScioContext): TFScioConextFunctions =
```
andrewsmartin: Just out of curiosity, what's the benefit of doing this instead of having an implicit class within this package object? Is it just so there's less code here?

ravwojdyla: Yea - less code here, right now those implicits are split into 3 classes/files.

ravwojdyla:
Fyi - this is a first idea, kinda low level, but minimal and functional, opened #678 to make it easier to work with Example protos.

PR #681, opened 2017-06-15
Use metric API helpers in broken examples

andrewsmartin:


andrewsmartin:
This is pending https://github.com/spotify/scio/pull/680, that should get merged first.

PR #692, opened 2017-06-21
Fix #669. Support options file

### LINKED ISSUE #669: Consider allow passing config as file
Instead of running 
`sbt runMain com.foo.Bar --numWorkers=100 --diskType=(I dont even remember) --region=... --project=... --autoscaling_algorithm=...` 
etc., it might be nice to have a config file that can be loaded instead (e.g. `sbt runMain com.foo.Bar --conf=~/scio.conf`)

Not sure how big of an issue it would be, since I can do `sbt runMain com.foo.Bar $(cat ~/scio.txt)` too.

ravwojdyla:
And idea to fix #669. If this is something we accept, will add tests.

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
+++ b/scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
@@ -17,6 +17,7 @@
 
 package com.spotify.scio.options;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.apache.beam.sdk.options.Description;
 import org.apache.beam.sdk.options.PipelineOptions;
 
@@ -41,4 +42,9 @@ public interface ScioOptions extends PipelineOptions {
   @Description("Custom application arguments")
   String getAppArguments();
   void setAppArguments(String arguments);
+
+  @JsonIgnore
+  @Description("Path to newline separated file with command line options")
+  String getOptionsFile();
+  void setOptionsFile(String configFile);
 }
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -120,11 +123,17 @@ object ScioContext {
       cmdlineArgs.partition(arg => optPatterns.exists(_.findFirstIn(arg).isDefined))
 
     val pipelineOpts = PipelineOptionsFactory.fromArgs(optArgs: _*).as(optClass)
-    val args = Args(appArgs)
-    if (appArgs.nonEmpty) {
-      pipelineOpts.as(classOf[ScioOptions]).setAppArguments(args.toString("", ", ", ""))
+    val configFile = pipelineOpts.as(classOf[ScioOptions]).getConfigFile
+    if (configFile != null) {
+      parseArguments(cmdlineArgs.filterNot(_.startsWith("--configFile")) ++
```
nevillelyh: "--configFile=" just to be safe?

ravwojdyla:
@bcleenders would this work for you? Basically:
 * put newline separated options in a config file
 * specify config file via `--configFile`
 * accepts both `--configFile` and extra options, but obviously not duplicates/overrides

Btw the same config file can still be used via command line like `sbt $(paste -s <config-file>)`.

scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
```diff
@@ -41,4 +42,9 @@
   @Description("Custom application arguments")
   String getAppArguments();
   void setAppArguments(String arguments);
+
+  @JsonIgnore
+  @Description("Config file with options")
```
nevillelyh: Make this clearer? Something like "JSON file to be parsed as PipelineOptions instead of command line arguments"?

PR #698, opened 2017-06-23
Add custom ClosureCleaner

andrewsmartin:
The `ClosureCleaner` we've been using seems to be broken in a few cases, especially when it comes to multiple layers of nesting. For example it's possible to have a closure whose enclosing scope references an outer that is not serializable, but such that (by tracing the transitive method calls / field references starting from the inner closure), we don't actually need the reference. So we can null it out. The test case I added to `ClosureTest` is an example of this.

I thought about instead just submitting this upstream but it's a fairly big change and it seems they haven't really touched their version in a couple years. Thoughts? Spark uses this approach too.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,7 @@ import sbtassembly.AssemblyPlugin.autoImport._
 import com.typesafe.sbt.SbtSite.SiteKeys._
 import com.typesafe.sbt.SbtGit.GitKeys.gitRemoteRepo
 
+val asmVersion = "5.2"
 val beamVersion = "2.0.0"
 val algebirdVersion = "0.13.0"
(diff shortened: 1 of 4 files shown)


andrewsmartin:
This also solves the problem we had with using metrics along with checkpoints

ravwojdyla:
Nice. Will need to walk through this one. That said I would still submit upstream - especially if it isn't much work to submit the PR, worst case PR will stay there but at least it will be a sort of "documentation of the bug".

nevillelyh:
BTW should this be back ported to master?

andrewsmartin:
Yeah, the same problem is there we just happened to catch it with the checkpoint test here. I'll cherry pick once this gets merged

PR #700, opened 2017-06-23
Bump bigtable client to 0.9.6.2

andrewsmartin:


nevillelyh:
Should this be ported to master as well?

andrewsmartin:
No, Beam 0.6.0 is still on the client version `0.9.2` which is not compatible with `0.9.6.2`, because of different netty version (among other things).

PR #702, opened 2017-06-26
add Cassandra sinks

nevillelyh:
Had to monkey patch a lot in `scio-cassandra2` due to an old Guava dependency. Otherwise 2 & 3 are sharing Scala code and handling API changes via 2 util classes.

Couldn't get `cassandra-unit` to work since `BulkCqlWriter` uses C* internal API and conflicts with embedded C* when running in the same JVM. Will probably need to fork it as integration test. Thoughts?

### DIFF
--- a/scio-cassandra3/src/main/java/com/spotify/scio/cassandra/CompatUtil.java
+++ b/scio-cassandra3/src/main/java/com/spotify/scio/cassandra/CompatUtil.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.cassandra;
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.CodecRegistry;
+import com.datastax.driver.core.DataType;
+import com.datastax.driver.core.ProtocolVersion;
+import org.apache.cassandra.dht.Murmur3Partitioner;
+import org.apache.cassandra.dht.RandomPartitioner;
+
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+
+/**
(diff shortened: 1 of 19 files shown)


scio-cassandra3/src/main/java/com/spotify/scio/cassandra/CompatUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private static RandomPartitioner randomPartitioner = new RandomPartitioner();
+  private static Murmur3Partitioner murmur3Partitioner = new Murmur3Partitioner();
+
+  public static ProtocolVersion getProtocolVersion(Cluster cluster) {
+    return cluster.getConfiguration().getProtocolOptions().getProtocolVersion();
+  }
+
+  public static <T> ByteBuffer serialize(DataType dataType, T value,
+                                     ProtocolVersion protocolVersion) {
```
andrewsmartin: Indentation

scio-cassandra3/src/main/java/org/apache/cassandra/hadoop/cql3/CqlBulkRecordWriterUtil.java
```diff
@@ -0,0 +1,61 @@
...
+    CqlBulkOutputFormat.setTableSchema(conf, table, tableSchema);
+    CqlBulkOutputFormat.setTableInsertStatement(conf, table, insertStatement);
+
+    // workaround for Hadoop static initialization
+    if (!System.getProperties().containsKey("hadoop.home.dir") &&
+        !System.getenv().containsKey("HADOOP_HOME")) {
+      System.setProperty("hadoop.home.dir", "/");
+    }
+
+    DatabaseDescriptor.clientInitialization();
```
andrewsmartin: What's this for? AFAICT this is the only difference from the same class in `scio-cassandra2`

nevillelyh: And
```scala
CqlBulkOutputFormat.setColumnFamilySchema(conf, table, tableSchema);
CqlBulkOutputFormat.setColumnFamilyInsertStatement(conf, table, insertStatement);
```

vs

```scala
CqlBulkOutputFormat.setTableSchema(conf, table, tableSchema);
CqlBulkOutputFormat.setTableInsertStatement(conf, table, insertStatement);
```

andrewsmartin: 👍 missed that

nevillelyh:
Added integration test but only for `scio-cassandra3` since CircleCI supports Cassandra 3.4 only.

nevillelyh:
Also this won't work very well in streaming mode since it does a `groupByKey` and only start writing at the end after `CqlBulkRecordWriter#close()`.

PR #703, opened 2017-06-27
remove unused coverage steps

nevillelyh:


andrewsmartin:
The point of including it there was to keep the settings the same so they could be cached. If we remove it but then keep it in the test step, it causes a re-application of settings and forces the resolver to do everything again.

nevillelyh:
Do you mean cache for Ivy artifacts? `coverage` is from a plugin and its artifacts are downloaded when SBT starts and shouldn't matter whether we run the step or not? On the other hand having `coverage` on may slow down some tests? May doesn't matter after all since neither `compile` nor `assembly` runs any test.

andrewsmartin:
It's for the sbt resolution cache (not ivy cache). https://github.com/spotify/scio/pull/484#issue-214212473

nevillelyh:
:+1:

PR #709, opened 2017-07-05
Only try to create a table if the CreateDisposition is set accordingly

neurolabs:
Without checking the CreateDisposition, when the table already exists,
an http 409 warning is logged. Also, the code fails when writing to a
specific day partition.

With this check, it is possible to use insertAll for inserting into a
day partition.


nevillelyh:
Great catch, thanks!

PR #711, opened 2017-07-06
Add helper to run jobs with managed ScioContext.

andrewsmartin:
Not sure about the `submitJob` name, open to suggestions. `runWithContext` seems better but we're also using that for tests.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/package.scala
@@ -67,6 +67,19 @@ package object scio {
     """version in .+"([^"]+)"""".r.findFirstMatchIn(line).get.group(1)
   }
 
+  /**
+   * Helper to submit and run a job in a context-manager style, which will automatically create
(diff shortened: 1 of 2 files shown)


nevillelyh:
What's the use case for this?

andrewsmartin:
Forgetting to call `sc.close()` has been a bit of a gotcha sometimes. This is just an optional, alternative style to write jobs without needing to explicitly open and close a context. It's not much code so I feel it's worth having.

nevillelyh:
But won't work with a lot of other things we have like `ScioResult` and metrics. In the end probably adds more confusion than problems it fixes?

andrewsmartin:
The helper returns the `ScioResult` so the metrics and all that should work properly. But like we discussed offline `JobTest` would fail if the context is not closed, so maybe this isn't useful outside of experimental / pre-production work. Up to you whether or not it's worth keeping.

nevillelyh:
Let's close for now. It'll be harder to tell people to switch to this than just not forget `sc.close()` :)

PR #713, opened 2017-07-06
Bump sbt-protoc and compiler plugin.

andrewsmartin:


andrewsmartin:
This version uses protobuf `3.3.1` but we can override if there's a problem.

PR #714, opened 2017-07-06
exclude protobuf-lite globally

nevillelyh:


andrewsmartin:
👍 I wonder, is it worth trying to add another IT to the repl (using DataflowRunner instead of DirectRunner) or is this just an oddity?

nevillelyh:
Yes but a DF job will make the build much longer. We can rethink this later.

PR #717, opened 2017-07-07
simultaneous support for elasticsearch 2 & 5

Alevy01:
Added support for Elasticsearch 5. Split Elasticsearch support into two projects, one for each version.


build.sbt
```diff
@@ -342,16 +346,56 @@ lazy val scioCassandra3: Project = Project(
...
+    "com.twitter" % "jsr166e" % jsr166eVersion,
+    "org.apache.logging.log4j" % "log4j-api" % log4jVersion,
+    "org.apache.logging.log4j" % "log4j-core" % log4jVersion
+  )
+).dependsOn(
+  scioCore,
+  scioTest % "test"
+)
+
+lazy val scioElasticsearch5: Project = Project(
```
josefalcon: you've listed 'scioElasticsearch5' project twice

Alevy01: Fixed

build.sbt
```diff
@@ -342,16 +346,36 @@ lazy val scioCassandra3: Project = Project(
...
+  "scio-elasticsearch5",
+  file("scio-elasticsearch5")
+).settings(
+  commonSettings,
+  description := "Scio add-on for writing to Elasticsearch",
+  libraryDependencies ++= Seq(
```
regadas: imo you only require this deps

```
    "org.elasticsearch.client" % "transport" % elasticsearch5Version
```

scio-elasticsearch5/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,328 @@
...
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.apache.beam.sdk.transforms.windowing.AfterProcessingTime;
+import org.apache.beam.sdk.transforms.windowing.GlobalWindows;
+import org.apache.beam.sdk.transforms.windowing.Repeatedly;
+import org.apache.beam.sdk.transforms.windowing.Window;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PDone;
+import com.twitter.jsr166e.ThreadLocalRandom;
```
regadas: replace with `import java.util.concurrent.ThreadLocalRandom;` ?

scio-elasticsearch5/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,328 @@
...
+          LOG.info("ElasticsearchWriter: no requests to send");
+          return;
+        }
+
+        final BulkRequest bulkRequest = new BulkRequest();
+        Iterables.concat(docWriteRequests)
+            .forEach(x -> {
+              try {
+                bulkRequest.add(x);
+              } catch(InvalidIndexNameException e) {
```
regadas: `InvalidIndexNameException` already a RuntimeException there's no need for special treatment.

```
final BulkRequest bulkRequest = new BulkRequest().add(Iterables.concat(docWriteRequests));
```

build.sbt
nevillelyh: And are the log4j dependency really needed? Can you use `slf4j` instead?

Alevy01: Without the log4j dependencies we get: `org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.NoClassDefFoundError: org/apache/logging/log4j/Logger`

Elasticsearch5 is dependent on the Elasticsearch transport client. Seems like the transport client is coupled with log4j: https://github.com/elastic/elasticsearch/issues/22671

scio-elasticsearch5/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
Alevy01: 👍 removing the twitter dependency from elasticsearch2 and elasticsearch5

build.sbt
nevillelyh: Are both `log4j-api` & `log4j-core` needed then? If it behaves like `slf4j` then without core it'll just fall back to no-op? Users can supply their own impl then.

Alevy01:
We removed both logging implementations. As you suggested the user can provide their own logging implementation. The library behaves as expected and writes to ES with no issue. In our client example our implementation dependency version is `"org.apache.logging.log4j" % "log4j-core" % "2.7"` However we observed this warning given from dataflow `Please use a logger instead of System.out or System.err.` as well as this Error logged from the Elasticsearch client library `ERROR StatusLogger Log4j2 could not find a logging implementation. Please add log4j-core to the classpath. Using SimpleLogger to log to the console`.

After some investigation we have concluded to leave the library as is because it functions as expected. If you have any suggestions let me know. @nevillelyh @regadas @josefalcon

build.sbt
```diff
@@ -342,16 +346,33 @@ lazy val scioCassandra3: Project = Project(
...
+  "scio-elasticsearch5",
+  file("scio-elasticsearch5")
 ).settings(
   commonSettings,
   description := "Scio add-on for writing to Elasticsearch",
   libraryDependencies ++= Seq(
     "com.google.guava" % "guava" % guavaVersion,
     "joda-time" % "joda-time" % jodaTimeVersion,
-    "org.elasticsearch" % "elasticsearch" % elasticsearchVersion
+    "org.elasticsearch" % "elasticsearch" % elasticsearch5Version,
```
regadas: @Alevy01 I think you can drop this one also. The only thing required is the Client.

build.sbt
```diff
@@ -35,7 +35,8 @@ val circeVersion = "0.8.0"
 val commonsIoVersion = "2.5"
 val commonsMath3Version = "3.6.1"
 val csvVersion = "0.2.0"
-val elasticsearchVersion = "2.1.0"
+val elasticsearch2Version = "2.1.0"
+val elasticsearch5Version = "5.5.0"
```
regadas: @Alevy01 +1 on using 5.5.0 was about to suggest the same. While you are at it and if don't mind, we could bump ES 2.1.0 to 2.4.5

build.sbt
regadas: just confirmed that it works with just client.transport dep

build.sbt
Alevy01: done 👍

Alevy01:
Upgrading from elasticsearch 2.1.0 to 2.4.5 deprecates support for elasticsearch 2.3.4 clusters. Reverting back to 2.1.0 @regadas

regadas:
@Alevy01 hmm thats weird ... In ES world clients and clusters should be in same major version and clients can connect to clusters in different minor versions (new funcionality may not be supported). Out of curiosity did you get any error?

regadas:
anyway this is not blocking thing for this PR we can always update it later and user can always override deps as needed 😄

Alevy01:
@regadas when trying to write to an ES 2.3.4 cluster using the 2.4.5 client it was failing with `NoNodeAvailableException`. Reverting back to 2.1.0 allowed the same job to successfully write

build.sbt
```diff
@@ -59,6 +60,8 @@ val shapelessDatatypeVersion = "0.1.4"
 val slf4jVersion = "1.7.25"
 val sparkeyVersion = "2.1.3"
 val tensorFlowVersion = "1.2.0"
+val jsr166eVersion = "1.1.0"
+val log4jVersion = "2.7"
```
regadas: @Alevy01 sorry 😄 this 2 are not used anymore 🎨

Alevy01: fixed

nevillelyh:
@Alevy01 ready to merge?

Alevy01:
@nevillelyh yep

PR #719, opened 2017-07-09
Exemplary fix #718

### LINKED ISSUE #718: BigQuery: Thread leak of one thread per writeTypedRows-call
When inserting into BigQuery via the BigQueryClient, for every call to writeTypedRows, another Thread is created. I haven't exactly figured out why, but my hunch is that a new ExecutorService gets created in BigQueryServicesImpl#insertAll using the passed options (which scio has influence on).

Please find attached a minimal project reproducing the issue. To get it up and running put a serviceaccount keyfile into src/main/resources and edit projectId, datasetId and tableId accordingly in src/main/scala/Main.scala.

It inserts 1000 documents into bigquery using separate calls to writeTypedRows while printing the overall threadcount every second.

neurolabs:
The ExecutorService used by the google bigquery driver needs to be
shutdown explicitly, or else it will keep holding on to its thread pool
with one thread in it.


nevillelyh:
Build failed with ```/home/ubuntu/scio/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala:269: not found: type GcsOptions```, can you please fix?

PR #731, opened 2017-07-18
monkey patch BigQuerySourceBase for BEAM-2532, fix #730

### LINKED ISSUE #730: Workaround BEAM-2532 BigQueryIO performance regression
Might be a blocker for 0.4.0 adoption.
https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQuerySourceBase.java#L174

nevillelyh:


ravwojdyla:
👍

PR #734, opened 2017-07-20
Support for AvroType macros

i-maravic:
Added support for:

@AvroType.fromSchema -
  generates case classes and their companion objects from Avro schema json
@AvroType.fromPath -
  generates case classes and their companion objects from GCS path. Globs are supported.
@AvroType.toSchema -
  generates Avro schema and compantion object from a case class.

A convenience function is provided in ScioContext to read Avro files to generated case classes:

ScioContext.typedAvroFile

A convenince function is provided to SCollection to write Avro files from generated case classes:

SCollection.saveAsTypedAvroFile

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -218,7 +218,8 @@ lazy val root: Project = Project(
   scioRepl,
   scioExamples,
   scioSchemas,
-  scioTensorFlow
+  scioTensorFlow,
+  scioAvro
 )
 
 lazy val scioCore: Project = Project(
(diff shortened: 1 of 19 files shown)


ravwojdyla:
@i-maravic could you give a short explanation of what problem you are trying to solve?

i-maravic:
@ravwojdla my goal is to be able to generate case classes based on Schema in a GCS folder, similar to how BQType does it for BQ tables.

nevillelyh:
Great stuff. LGTM except some minor comments but I can merge and fix later.

i-maravic:
@nevillelyh Let me fix the docs and comments and I (or you) can merge this later today.

PR #735, opened 2017-07-20
Perform full input stream decoding if not a BufferedPrefixInputStream.

andrewsmartin:
Mainly a compatibility fix. Older versions of Scio did not use `BufferedPrefixOutputStream` for serialization, and so recent versions break when reading data produced by an older version because they expect input streams to be prefixed with the length.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
@@ -114,10 +114,15 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
   }
 
   override def decode(inStream: InputStream): T = {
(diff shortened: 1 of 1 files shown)


nevillelyh:
Just realized that just `kryo.get().readClassAndObject(new Input(inStream))` may not be enough since we read an int before the original commit 4d3e9e18c585466b65e72a2df0d2b3fbc6a524f3.

nevillelyh:
Nevermind, the Avro/Proto bytes in question were written with the following code from `AvroBytesUtil` where `isWholeStream == true`, so reading fully is fine.
```java
public static <T> byte[] encodeToByteArray(Coder<T> coder, T value) throws CoderException {
  return encodeToByteArray(coder, value, Coder.Context.OUTER);
}
```

nevillelyh:
Fixed in f4dc6ff8c1d145fab2c6e4bbe919c310d61b4395

PR #738, opened 2017-07-22
protobuf file fixes

nevillelyh:


scio-test/src/test/scala/com/spotify/scio/io/FileFormatTest.scala
```diff
@@ -0,0 +1,74 @@
...
+  // Legacy files generated with 0.3.0-beta1, the last release before the
+  // BufferedPrefix{Input,Output}Stream change.
+
+  private val path = "scio-test/src/test/resources"
+
+  private val objects = (1 to 100).map(x => (x, x.toDouble, x % 2 == 0, "s" + x))
+  private val protobufs = (1 to 100)
+    .map(x => com.google.protobuf.Timestamp.newBuilder.setSeconds(x * 1000).setNanos(x).build)
+
+  // Object file is NOT backwards compatible
```
nevillelyh: In fact I got `KryoException: Unable to find class` error when reading an object file written with 0.3.0-beta1. Need to investigate if it's fixable.

PR #741, opened 2017-07-27
Try to avoid ReflectData in SpecificAvroSerializer

samschlegel:
If you don't pass the schema in `AvroCoder.of`, it falls back to using reflection to rebuild the schema. Since we have `T <: SpecificRecordBase`, it makes sense to try and avoid this if possible. This won't work in the case where `T` hasn't defined `getClassSchema`, but given that the avro compiler includes that all the time afaict, this shouldn't be an issue in practice.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/AvroSerializer.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/AvroSerializer.scala
@@ -58,11 +58,10 @@ private class SpecificAvroSerializer[T <: SpecificRecordBase] extends KSerialize
   private lazy val cache: MMap[Class[T], AvroCoder[T]] = MMap()
 
   private def getCoder(cls: Class[T]): AvroCoder[T] =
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/AvroSerializer.scala
```diff
@@ -56,7 +57,12 @@ private class SpecificAvroSerializer[T <: SpecificRecordBase] extends KSerialize
 
   private lazy val cache: MMap[Class[T], AvroCoder[T]] = MMap()
 
-  private def getCoder(cls: Class[T]): AvroCoder[T] = cache.getOrElseUpdate(cls, AvroCoder.of(cls))
+  private def getCoder(cls: Class[T]): AvroCoder[T] =
+    cache.getOrElseUpdate(cls, {
```
nevillelyh: Don't really need the `{}`?

PR #745, opened 2017-07-28
add Scala 2.12 support

nevillelyh:


scripts/circleci_parallel_run.sh
```diff
@@ -20,8 +20,7 @@ set -e
 case $CIRCLE_NODE_INDEX in
   0) export CI_SCALA_VERSION="2.11.11"
     ;;
-  1)
-    exit 0
+  1) export CI_SCALA_VERSION="2.12.2"
```
ravwojdyla: `2.12.3` or `2.12.2`?

scio-core/src/main/scala/com/spotify/scio/util/ClosureCleaner.scala
```diff
@@ -18,232 +18,19 @@
...
-  /** Clean the closure in place. */
-  def apply[T <: AnyRef](func: T): T = {
+  def apply[T <: AnyRef](obj: T): T = {
     try {
-      SerializableUtils.serializeToByteArray(func.asInstanceOf[Serializable])
+      SerializableUtils.serializeToByteArray(obj.asInstanceOf[Serializable])
     } catch {
       case e: IllegalArgumentException if e.getCause.isInstanceOf[NotSerializableException] =>
-        clean(func)
+        com.twitter.chill.ClosureCleaner(obj)
```
andrewsmartin: Do we need this still? With `-Ydelambdafy:inline` I'm guessing this path is not exercised at all (and if it were, it could break - at best it will do nothing because `$OUTER` is no longer a thing, at worst it will crash).

nevillelyh: We need it for 2.11

andrewsmartin: Oh right

PR #749, opened 2017-07-30
Add warning that using GCS globs might be inefficient

i-maravic:


i-maravic:
@nevillelyh

nevillelyh:
@i-maravic FYI next time please prefix branch names with `$USERNAME/` for better organization.

Issue #747, opened 2017-07-28
ASM Dependency mismatch


Alevy01:
Scio is pulling in two versions of ASM (3.1 & 5.2) 3.1 pulls in ClassVisitor as an interface while 5.2 pulls it in as an abstract class. Reverting https://github.com/spotify/scio/commit/4273e1bf68eba0c790fff81c100a72f913b7c6db fixed this.

```
Exception in thread "main" java.lang.IncompatibleClassChangeError: class com.spotify.scio.util.InnerClosureFinder has interface org.objectweb.asm.ClassVisitor as super class
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at com.spotify.scio.util.TransitiveClosureCleaner.innerClasses(ClosureCleaner.scala:153)
	at com.spotify.scio.util.TransitiveClosureCleaner.getAccessedFields(ClosureCleaner.scala:167)
	at com.spotify.scio.util.TransitiveClosureCleaner.storeAccessedFields(ClosureCleaner.scala:174)
	at com.spotify.scio.util.TransitiveClosureCleaner.cleanOuter(ClosureCleaner.scala:110)
	at com.spotify.scio.util.ClosureCleaner$class.clean(ClosureCleaner.scala:59)
	at com.spotify.scio.util.TransitiveClosureCleaner.clean(ClosureCleaner.scala:104)
	at com.spotify.scio.util.ClosureCleaner$.clean(ClosureCleaner.scala:47)
	at com.spotify.scio.util.ClosureCleaner$.apply(ClosureCleaner.scala:40)
[...]

PR #750, opened 2017-07-31
Use shaded asm library (possible fix for #747)

andrewsmartin:


andrewsmartin:
@alevy01 if you don't mind - would you be able to build + publish this locally (from the andrew/closure branch) and see if this fixes your problem?

PR #759, opened 2017-08-07
Add option to set BigQuery priority

hansencc:
Add BigQuery priority (BATCH vs INTERACTIVE) as a system property.


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -115,13 +115,17 @@ class BigQueryClient private (private val projectId: String,
...
   private def isInteractive =
-    Thread
-      .currentThread()
-      .getStackTrace
-      .exists { e =>
-        e.getClassName.startsWith("scala.tools.nsc.interpreter.") ||
-          e.getClassName.startsWith("org.scalatest.tools.")
-      }
+    BigQueryClient.priority
+      .map(_.equals("INTERACTIVE"))
```
nevillelyh: Should be `_ == "INTERACTIVE"` for idiomatic scala?

PR #767, opened 2017-08-11
coder improvements

nevillelyh:
The first 2 commits are minor and unlikely to make any difference. The last one about slow size estimation could help extreme `groupByKey` cases.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
@@ -73,6 +73,8 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
 
   import KryoAtomicCoder.kryo
 
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
   override def encode(value: T, outStream: OutputStream): Unit = {
     if (value == null) {
       throw new CoderException("cannot encode a null value")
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -134,9 +107,36 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
...
-        observer.update(kryoEncodedElementByteSize(i.next()))
+      while (i.hasNext && !aborted) {
+        val size = kryoEncodedElementByteSize(i.next())
+        observer.update(size)
+        count += 1
+        bytes += size
+        val elapsed = System.currentTimeMillis() - start
+        if (elapsed > abortThreshold) {
+          aborted = true
+          logger.warn(s"Slow size estimation for ${wrapper.underlying.getClass}, " +
```
ravwojdyla: `Aborting size estimation for` ?

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -134,9 +107,36 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
...
+        count += 1
+        bytes += size
+        val elapsed = System.currentTimeMillis() - start
+        if (elapsed > abortThreshold) {
+          aborted = true
+          logger.warn(s"Aborting size estimation for ${wrapper.underlying.getClass}, " +
+            s"elapsed: $elapsed ms, count: $count, bytes: $bytes")
+          wrapper.underlying match {
+            case c: _root_.java.util.Collection[_] =>
+              // interpolate remaining bytes in the collection
```
ravwojdyla: You could just in case also add a `warn` or `debug` message here for getting size of collection and how much it took, just in case it's expensive, also might be good information to have, and is only a single message anyway.

PR #771, opened 2017-08-14
add Parquet Avro read support

nevillelyh:


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -0,0 +1,88 @@
...
+
+      AvroParquetInputFormat.setAvroReadSchema(job, readSchema)
+      if (projection != null) {
+        AvroParquetInputFormat.setRequestedProjection(job, projection)
+      }
+      if (predicate != null) {
+        ParquetInputFormat.setFilterPredicate(job.getConfiguration, predicate)
+      }
+
+      val source = HadoopInputFormatIO.read[JBoolean, T]()
```
ravwojdyla: Why `JBoolean` and not `Void`?

ravwojdyla:
👎 tests are missing.

nevillelyh:
Still a rough draft. Will add tests soon.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
nevillelyh: `null` causes NPE in https://github.com/apache/beam/blob/master/sdks/java/io/hadoop/input-format/src/main/java/org/apache/beam/sdk/io/hadoop/inputformat/HadoopInputFormatIO.java#L714

nevillelyh:
Added tests, docs, etc.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  /** Alias for [[me.lyh.parquet.avro.Predicate]]. */
+  val Predicate = me.lyh.parquet.avro.Predicate
+
+  /** Enhanced version of [[ScioContext]] with Parquet Avro methods. */
+  implicit class ParquetAvroScioContext(val self: ScioContext) extends AnyVal {
+
+    // GenericRecord is not supported since HadoopInputFormatIO cannot infer Coder from type
+    /**
+     * Get an SCollection for a Parquet file as Avro records.
```
ravwojdyla: Not sure where is the best place to put this, but it might be useful to comment/document the limitations of Hadoop*IO, mainly the liquid sharding, thus whatever is gonna be the initial split will stay the same until next fusion.

ravwojdyla: @nevillelyh this?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -0,0 +1,118 @@
...
+      } else {
+        val cls = ScioUtil.classOf[T]
+        val readSchema = cls.getMethod("getClassSchema").invoke(null).asInstanceOf[Schema]
+
+        val job = Job.getInstance()
+
+        // Client id/secret of Google-managed project associated with the Cloud SDK
+        // These are needed since `FileInputFormat.setInputPaths` validates paths locally and
+        // requires the user's GCP credentials.
+        job.getConfiguration.setBoolean("fs.gs.auth.service.account.enable", false)
```
ravwojdyla: This is actually not going to work unless driver runs on a *client* machine.

nevillelyh: Any suggestion? Detect if a service account is available?

ravwojdyla: IRL. we could discover `GOOGLE_APPLICATION_CREDENTIALS` and configure it via https://github.com/GoogleCloudPlatform/bigdata-interop/blob/5f1138979919c9cc38f5dfcc4d7e73ffa7b5e176/util/src/main/java/com/google/cloud/hadoop/util/EntriesCredentialConfiguration.java#L60-L76

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -0,0 +1,127 @@
...
+          .map(_.getValue)
+      }
+    }
+
+    private def setInputPaths(job: Job, path: String): Unit = {
+      // These are needed since `FileInputFormat.setInputPaths` validates paths locally and
+      // requires the user's GCP credentials.
+      sys.env.get("GOOGLE_APPLICATION_CREDENTIALS") match {
+        case Some(json) =>
+          job.getConfiguration.set("fs.gs.auth.service.account.json.keyfile", json)
```
nevillelyh: Works locally with `export GOOGLE_APPLICATION_CREDENTIALS=x.json`. Is this sufficient though?

ravwojdyla:
👍

PR #773, opened 2017-08-15
Fetch bigtable cluster sizes

kiarash-rezahanjani:
@ravwojdyla hi rav, as you suggested here is a PR for fetching cluster sizes. I will add some tests as soon as I find time would be good to know any comments meanwhile. 
I also have a hard time running integration test as I dont have access to integration data sets. Can I have read access to them ?


scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
```diff
@@ -103,4 +108,26 @@ public static void updateNumberOfBigtableNodes(
     }
   }
 
+  /**
+   * Get size of all clusters for specified Bigtable instance.
+   *
+   * @param projectId GCP projectId
+   * @param instanceId Bigtable instanceId
+   *
+   * @return map of cluster name to it's number of nodes
```
ravwojdyla: "it's" -> "its"

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -177,6 +177,18 @@ package object bigtable {
     }
 
     /**
+      * Get size of all clusters for specified Bigtable instance.
+      *
+      * @return map of cluster name to it's number of nodes
+      */
+    def getBigtableClusterSizes(projectId: String,
+                                instanceId: String): Map[String, Int] = if (!self.isTest) {
```
ravwojdyla: Mind moving this to new line?

ravwojdyla:
Also is this a duplicate of 
```
val b = new BigtableClusterUtilities(bigtableOptions)
bt.getClusterNodeCount(<clusterId>, <zoneId>)
```
?

kiarash-rezahanjani:
on whether it is a duplicate of BigtableClusterUtilities(..).getClusterNodeCount(..) Yes and no. 
This one returns sizes of all clusters within a big table instance, and the other size of a single one.
Here is the reasoning and please correct me if I am wrong.
I noticed that current BigtableScioContext.updateNumberOfBigtableNodes method assume there can be multiple clusters per big table instance so I tried to keep it aligned with same assumption and return all cluster sizes, so if one wants to re-size to previous number of nodes can fetch all cluster  sizes.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
kiarash-rezahanjani: you mean the if , as follow ?
 ....Map[String, Int] = {
   if (!self.isTest) ..
}

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
kiarash-rezahanjani: ouch!

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
ravwojdyla: yep - it is more alight with the rest of the code.

ravwojdyla:
@kiarash-rezahanjani 👍  let's add tests and we are good

kiarash-rezahanjani:
@ravwojdy Hi,  I added few new changes, would be great if you could take a look again. I had to add powermock to do some unit testing :| (could be a over kill for such logic less method :) but let me know please if you want me to change things

ravwojdyla:
@kiarash-rezahanjani yea that powermock is largely an overkill, could you instead just add IT test similar to `update number of bigtable nodes` one, it would probably also be an ignored one. We might need to figure out a better way to test those utils for bigtable. Mocking does not actually test much; ignored tests are, basically that "ignored" (at least provide an easy way to manually test if needed), while running those it tests would take time. Bigtable emulator doesn't provide admin services afair. It's all suboptimal. Could you for now add test in IT similar to the update one, and we will create an issue to figure out a better way to test those.

ravwojdyla:
Fyi: https://github.com/spotify/scio/issues/788

ravwojdyla:
Thanks @kiarash-rezahanjani

PR #776, opened 2017-08-15
Add safeFlatMap

### LINKED ISSUE #772: Add unsafeFlatMap
Add `unsafeFlatMap` to handle errors and trade accuracy for latency.

Something along the lines of:

```
val errors = SideOutput[(String, Throwable)]
sc.textFile(args.getOrElse("input", ExampleData.KING_LEAR))
  .flatMap(_.split("[^a-zA-Z']+").filter(_.nonEmpty))
  .unsafeFlatMap(errors)(e => riskyOp(e) )
```

where `unsafeFlatMap` is sth like:

```
  def unsafeFlatMap[T, U](f: T => TraversableOnce[U], errorSideOutput: SideOutput[(T, Throwable)])
  : DoFn[T, U] = new NamedDoFn[T, U] {
    val g = ClosureCleaner(f)  // defeat closure
    @ProcessElement
    private[scio] def processElement(c: DoFn[T, U]#ProcessContext): Unit = {
      val i = Try {
        g(c.element()).toIterator
      } match {
        case Success(r) => r
        case Failure(e) =>
          c.sideOutput[T](errorSideOutput.tupleTag, (c.element(), e))
          Iterator.empty
      }
      while (i.hasNext) c.output(i.next())
    }
  }
```

Possibly add threshold of errors (in a bundle(?)) as a parameter to cancel a job.

ravwojdyla:
fixes #772

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
@@ -21,12 +21,14 @@ import java.io.File
 import java.net.URI
 import java.nio.file.Path
 
-import com.spotify.scio.util.{Functions, RemoteFileUtil}
+import com.spotify.scio.util.{ClosureCleaner, Functions, NamedDoFn, RemoteFileUtil}
 import com.spotify.scio.values.SCollection
 import org.apache.beam.runners.dataflow.DataflowRunner
 import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions
 import org.apache.beam.runners.direct.DirectRunner
-import org.apache.beam.sdk.transforms.ParDo
+import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+import org.apache.beam.sdk.transforms.{DoFn, ParDo}
+import org.apache.beam.sdk.values.{TupleTag, TupleTagList}
 
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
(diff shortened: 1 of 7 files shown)


nevillelyh:
Not sure if we should make it available to all `SCollection`s by default. Might be confusing to new users looking for methods via tab-completion?
Maybe better to put this in a separate package and only available via import?

lmayumi:
👍

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
```diff
@@ -158,4 +161,45 @@ package object transforms {
     }
   }
 
+  /**
+   * Enhanced version of [[SCollection]] with specialized versions of flatMap.
+   */
+  implicit class SpecializedFlatMapSCollection[T: ClassTag](val self: SCollection[T]) {
+
+    /**
+     * Latency optimized flavour of [[self.flatMap]], it returns a new SCollection by first
```
nevillelyh: `flavor` ;)

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
```diff
@@ -158,4 +161,45 @@ package object transforms {
...
+     * @group transform
+     */
+    def safeFlatMap[U: ClassTag](f: T => TraversableOnce[U])
+    : (SCollection[U], SCollection[(T, Throwable)]) = {
+      val (mainTag, errorTag) = (new TupleTag[U], new TupleTag[(T, Throwable)])
+      val doFn = new NamedDoFn[T, U] {
+        val g = ClosureCleaner(f) // defeat closure
+        @ProcessElement
+        private[scio] def processElement(c: DoFn[T, U]#ProcessContext): Unit = {
+          val i = Try {
```
nevillelyh: use `try/catch` instead to avoid extra tiny objects?
I'll normally use `Try` as function return type when `match` `Success`/`Failure` happens in a separate place.

scio-extra/src/main/scala/com/spotify/scio/extra/transforms/package.scala
ravwojdyla: fixed but with a heavy heart ...

PR #785, opened 2017-08-17
Fix #784 lazy batch Elasticsearch requests

### LINKED ISSUE #784: Lazily batch in elastic search sink after group by key
https://github.com/spotify/scio/blob/master/scio-elasticsearch2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java#L223

This iterable to list conversion will blow up with large value group.

regadas:


scio-elasticsearch2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -211,34 +226,49 @@ public void processElement(ProcessContext c) throws Exception {
...
-        if (bulkItemResponse.hasFailures()) {
-          error.accept(new BulkExecutionException(bulkItemResponse));
-        }
+        final Stream<ActionRequest> actionRequests =
+            StreamSupport.stream(values.spliterator(), false)
+                .map(toActionRequests::apply)
+                .flatMap(ar -> StreamSupport.stream(ar.spliterator(), false));
+
+        final Iterable<List<ActionRequest>> chunks =
+            Iterables.partition(actionRequests::iterator, maxBulkRequestSize);
```
nevillelyh: Seems `Iterables.partition` is eager though? https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Iterables.html#partition-java.lang.Iterable-int-

Seems it's easier to just do an older school loop and mutable list buffer?

```java
val buffer = Lists.newArrayList()
while (i.hasNext) {
  buffer.append(i.next)
  if (buffer.size > 3000) {
    sendReq(buffer)
    buffer.clear()
  }
}
if (!buffer.empty) {
  sendReq(buffer)
}

File: scio-elasticsearch2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java [hunk 2]
```diff
@@ -211,34 +226,49 @@ public void processElement(ProcessContext c) throws Exception {
...
-        final List<Iterable<ActionRequest<?>>> actionRequests =
-            StreamSupport.stream(c.element().getValue().spliterator(), false)
-                .map(toActionRequests::apply)
-                .collect(Collectors.toList());
+        final Iterable<T> values = c.element().getValue();
 
         // Elasticsearch throws ActionRequestValidationException if bulk request is empty,
         // so do nothing if number of actions is zero.
-        if (actionRequests.isEmpty()) {
+        if (values.iterator().hasNext()) {
```
marklaw000: this should be if(!values.iterator().hasNext()) ?

regadas: @marklaw000 🤦‍♂️

scio-elasticsearch2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
regadas: @nevillelyh looking at this implementation https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Iterators.java#L620 it's what we want, maybe I missed something?

nevillelyh: That's `Iterators.partition` not `Iterables.partition`?

regadas: @nevillelyh It's the underlying implementation for it https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Iterables.java#L526

marklaw000:
it is working ok for me!
+1

ravwojdyla:
👍

PR #791, opened 2017-08-23
Add AvroType.fromFile annotation and make AvroType imported by default in Scio REPL

i-maravic:
@nevillelyh

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/types/AvroType.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/types/AvroType.scala
@@ -128,7 +128,7 @@ object AvroType {
    * Macro annotation for a file which contains Avro schema.
    *
    * Generate case classes for an Avro schema. File can be either local or remote files.
-   * For example file can be located on Google Cloud Storage:
+   * For example file can be located on Google Cloud Storage (GCS):
    *
    * {{{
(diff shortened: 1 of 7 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/types/AvroType.scala
```diff
@@ -125,6 +125,32 @@ object AvroType {
...
+    *
+    * {{{
+    *   @AvroType.fromFile("sub-project/src/main/avro/schema-file.avsc")
+    *   class MyRecord
+    * }}}
+    *
+    * Also generate a companion object with convenience methods.
+    * @group annotation
+    */
+  class fromFile(schemaFile: String) extends StaticAnnotation {
```
nevillelyh: `fromFile` can be a bit confusing with `fromPath` though, which could also take a single file path. `fromSchemaFile` instead?

scio-avro/src/main/scala/com/spotify/scio/avro/types/AvroType.scala
```diff
@@ -125,6 +125,32 @@ object AvroType {
   }
 
   /**
+    * Macro annotation for a file which contains Avro schema.
```
nevillelyh: Use javadoc comment style please. https://github.com/spotify/scio/wiki/Style-Guide#documentation-style

PR #802, opened 2017-08-26
Generate nested case classes with stable names

i-maravic:


i-maravic:
@nevillelyh

scio-avro/src/main/scala/com/spotify/scio/avro/types/ConverterProvider.scala
```diff
@@ -139,11 +139,11 @@ private[types] object ConverterProvider {
 
     def cast(tree: Tree, tpe: Type): Tree = {
       tpe match {
-        case t if t =:= typeOf[Boolean] => tree
-        case t if t =:= typeOf[Int] => tree
-        case t if t =:= typeOf[Long] => tree
-        case t if t =:= typeOf[Float] => tree
-        case t if t =:= typeOf[Double] => tree
+        case t if t =:= typeOf[Boolean] => q"Boolean.box($tree)"
```
nevillelyh: Are these boxing really necessary? Shouldn't the compiler auto-box?

scio-avro/src/main/scala/com/spotify/scio/avro/types/SchemaUtil.scala
```diff
@@ -45,40 +43,52 @@ object SchemaUtil {
...
+            indent)
+        (nestedClassName, Seq(nested))
       case t => throw new IllegalArgumentException(s"Type: $t not supported")
     }
   }
   // scalastyle:on cyclomatic.complexity
 
-  private def getCaseClass(fields: JList[Field], name: String, indent: Int): String = {
-    val xs = fields.asScala
+  private def getCaseClass(className: String,
```
nevillelyh: You can get class name with `schema.getName()`?
Looks like `getCaseClass` is being used recursively, make an entry point version without `className`?

scio-avro/src/main/scala/com/spotify/scio/avro/types/SchemaUtil.scala
```diff
@@ -29,11 +26,12 @@ import scala.collection.JavaConverters._
 object SchemaUtil {
 
   /** Convert schema to case class definitions. */
-  def toPrettyString(schema: Schema, indent: Int = 0): String =
-    getCaseClass(schema.getFields, schema.getName, indent)
+  def toPrettyString(className: String, schema: Schema, indent: Int = 0): String =
```
nevillelyh: use `schema.getName` as `className` so you can keep the function signature?

i-maravic: If I would do that, than the prettyString wouldn't match the class hierarchy exactly.

scio-avro/src/main/scala/com/spotify/scio/avro/types/SchemaUtil.scala
i-maravic: I can't do that because I need to pass nested class name (ex. `ParentClass$NestedClass`) to getCaseClass function.

scio-avro/src/main/scala/com/spotify/scio/avro/types/ConverterProvider.scala
nevillelyh: Those are probably from here:
```scala
case (fieldName, value) =>
  q"if ($value != null) result.put($fieldName, $value)"
```

We can't get rid of the null check since `$value` can be null if it's not a primitive (String, nested record, etc.) but could never be null if it IS, i.e. `case class Record(a: Int, b: Float)`. So the warning is actually safe to ignore but maybe we can do something smarter there to get rid of the warning without adding boxing.

PR #805, opened 2017-08-28
remove runner check in InMemorySink

nevillelyh:


nevillelyh:
Need to rethink this. Closing for now.

PR #808, opened 2017-08-29
Custom body in BQT

ravwojdyla:


ravwojdyla:
ping @nevillelyh

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -217,22 +223,33 @@ private[types] object TypeProvider {
   private def caseClass(c: blackbox.Context)
                        (name: c.TypeName, fields: Seq[c.Tree], body: Seq[c.Tree]): c.Tree = {
     import c.universe._
-    q"case class $name(..$fields) extends ${p(c, SType)}.HasAnnotation { ..$body }"
+    q"@com.spotify.scio.bigquery.types.BigQueryTag case class $name(..$fields) extends ${p(c, SType)}.HasAnnotation { ..$body }"
```
nevillelyh: `@_root_.` just to be safe?

nevillelyh:
LGTM mostly. Just not sure if the need justifies this support but I guess good to have as long as it doesn't break anything.

PR #816, opened 2017-08-30
Support saving TF Example together with feature spec

ravwojdyla:


ravwojdyla:
This will allow to save features and feature-spec together in scio, but more importantly in featran.

ravwojdyla:
ping @nevillelyh

scio-examples/src/main/scala/com/spotify/scio/examples/extra/TFExampleExample.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    sc.textFile(args.getOrElse("input", ExampleData.KING_LEAR))
+      .flatMap(_.split("[^a-zA-Z']+").filter(_.nonEmpty))
+      .countByValue
+      .map { case (s, c) => WordCountFeatures(s.length.toFloat, c.toFloat) }
+      .map(featuresType.toExample(_))
+      .saveAsTfExampleFile(
+        args("output"),
+        featureSpec,
+        featureSpecPath = args.getOrElse("feature-spec-path", null))
```
ravwojdyla: `orNull`

scio-tensorflow/src/main/scala/com/spotify/scio/values/TFSCollectionFunctions.scala
```diff
@@ -126,6 +132,61 @@ class TensorFlowSCollectionFunctions[T: ClassTag](@transient val self: SCollecti
...
+   *                        `${path}/.feature_spec`
+   *
+   * @group output
+   */
+  def saveAsTfExampleFile(path: String,
+                          featureSpec: SCollection[Seq[String]],
+                          suffix: String = ".tfrecords",
+                          compressionType: CompressionType = CompressionType.NONE,
+                          numShards: Int = 0,
+                          @Nullable featureSpecPath: String = null)
```
nevillelyh: `Nullable` is kinda redundant since default is `null`?

scio-tensorflow/src/main/scala/com/spotify/scio/values/TFSCollectionFunctions.scala
```diff
@@ -126,6 +132,61 @@ class TensorFlowSCollectionFunctions[T: ClassTag](@transient val self: SCollecti
...
+                         (implicit ev: T <:< Example)
+  : (Future[Tap[Example]], Future[Tap[String]]) = {
+    require(featureSpec != null, "Feature spec can't be null")
+    require(path != null, "Path can't be null")
+    val _featureSpecPath =
+      Option(featureSpecPath).getOrElse(path.replaceAll("\\/+$", "") + "/.feature_spec")
```
nevillelyh: Does `.dotfile` play well with GCS or S3?

scio-tensorflow/src/test/scala/com/spotify/scio/io/TFTapTest.scala
```diff
@@ -19,10 +19,26 @@ package com.spotify.scio.io
...
+import org.apache.beam.sdk.Pipeline.PipelineExecutionException
+import org.apache.beam.sdk.io.TFRecordIO.CompressionType
 import org.apache.commons.io.FileUtils
+import org.tensorflow.example._
 
 class TFTapTest extends TapSpec {
 
+  private def getDummyExample = {
+    val features = Seq((1.0F, 2.0F), (5.0F, 3.0F))
+    features.map { kv =>
```
nevillelyh: Use a `TensorFlowType[T]` here as well to simplify the code? We should probably expect the users to do that too right?

scio-tensorflow/src/test/scala/com/spotify/scio/io/TFTapTest.scala
```diff
@@ -40,4 +56,42 @@ class TFTapTest extends TapSpec {
...
   }
 
+  it should "support saveAsTfExampleFile" in {
+    import com.spotify.scio.tensorflow._
+    val examples = getDummyExample
+    import org.apache.beam.sdk.io.TFRecordIO.{CompressionType => CType}
+    for (compressionType <- Seq(CType.NONE, CType.ZLIB, CType.GZIP)) {
+      val dir = tmpDir
+      val sc = ScioContext()
+      val featureSpec = sc.parallelize(Seq(Seq("f1", "f2")))
```
nevillelyh: This `Seq(Seq())` and several other places may be confusing. Anyway to simplify it?
We can't overload `saveAsTfExampleFile` due to the default args. What about making the argument `featureSpec: M` where `[M: FeatureSpecType]`? You can then have 3 instances where `M` is a case class, `Seq[String]` or `SCollection[Seq[String]]`.

By doing this, you can probably also remove the following and pass `WordCountFeatures.type` as `featureSpec` or something like that.
`val featureNameSpec: Seq[String] = FeatureNameFeatureSpec[WordCountFeatures]`

Thoughts? Too complicated?

scio-tensorflow/src/main/scala/com/spotify/scio/values/TFSCollectionFunctions.scala
ravwojdyla: What do you mean? it would be another object; unless you use some kinda of abstraction on top of it, like HadoopFS which would actually assume `.`|`_` is hidden.

nevillelyh: I mean if filenames like `.feature_spec` would cause problem on some file systems. Probably not?

ravwojdyla: Is fine on GCS, HDFS. S3 should be fine too http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

ravwojdyla:
@nevillelyh ptal

scio-examples/src/main/scala/com/spotify/scio/examples/extra/TFExampleExample.scala
```diff
@@ -0,0 +1,66 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.ContextAndArgs
+import com.spotify.scio.examples.common.ExampleData
+import shapeless.datatype.tensorflow.TensorFlowType
```
nevillelyh: duplicate imports with those inside `main`? I'll put all here since it's a short enough example and it's more idiomatic.
I'd also `import com.spotify.scio._` as it's designed for.

scio-tensorflow/src/main/scala/com/spotify/scio/values/TFSCollectionFunctions.scala
```diff
@@ -126,6 +132,112 @@ class TensorFlowSCollectionFunctions[T: ClassTag](@transient val self: SCollecti
...
+   *                        `${path}/_feature_spec`
+   *
+   * @group output
+   */
+  def saveAsTfExampleFile(path: String,
+                          featureSpec: FeatureSpec,
+                          suffix: String = ".tfrecords",
+                          compressionType: CompressionType = CompressionType.NONE,
+                          numShards: Int = 0,
+                          @Nullable featureSpecPath: String = null)
```
nevillelyh: Get rid of `@Nullable`? There's really no reason if it defaults to `null`

scio-examples/src/main/scala/com/spotify/scio/examples/extra/TFExampleExample.scala
```diff
@@ -0,0 +1,66 @@
...
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    sc.textFile(args.getOrElse("input", ExampleData.KING_LEAR))
+      .flatMap(_.split("[^a-zA-Z']+").filter(_.nonEmpty))
+      .countByValue
+      .map { case (s, c) => WordCountFeatures(s.length.toFloat, c.toFloat) }
+      .map(featuresType.toExample(_))
+      .saveAsTfExampleFile(
+        args("output"),
+        FeatureSpec.fromCaseClass[WordCountFeatures],
```
nevillelyh: Curious how does `FeatureSpec` gets imported? It's not in `tensorflow` package object.

nevillelyh: Nevermind I see the alias now.

scio-tensorflow/src/main/scala/com/spotify/scio/values/TFSCollectionFunctions.scala
```diff
@@ -126,6 +132,112 @@ class TensorFlowSCollectionFunctions[T: ClassTag](@transient val self: SCollecti
   }
 }
 
+private[scio] object TFExampleSCollectionFunctions {
+
+  sealed trait FeatureSpec
+  final case class SeqStrFeatureSpec(x: Seq[String]) extends FeatureSpec
```
nevillelyh: make the case classes private?

PR #818, opened 2017-08-31
Use kryo 4.0.1

ravwojdyla:
Until we are based on chill that includes https://github.com/twitter/chill/pull/296

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -49,6 +49,7 @@ val jodaTimeVersion = "2.9.9"
 val junitInterfaceVersion = "0.11"
 val junitVersion = "4.12"
 val kantanCsvVersion = "0.2.1"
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -250,6 +251,7 @@ lazy val scioCore: Project = Project(
     "com.twitter" %% "chill" % chillVersion,
     "com.twitter" %% "chill-algebird" % chillVersion,
     "com.twitter" % "chill-protobuf" % chillVersion,
+    "com.esotericsoftware" % "kryo-shaded" % kryoVersion,
```
nevillelyh: Just curious, "shaded" means it contains shaded dependencies? `Kryo` itself is not shaded?

ravwojdyla: Yes - the dependencies are shaded. Also this is the artifact chill is using. Btw, this is how the content looks like: [gist](https://gist.github.com/ravwojdyla/8ab8be9de3050ce166afd38d0ce020d8).

PR #819, opened 2017-08-31
Separate io directory for IOs

ravwojdyla:


nevillelyh:
Maybe hold on to this for now. Avro, BigQuery, TF, etc. are kind of `IO` but include some other logic, not sure how to group them. Also not sure the stuff in `scio-extra` right now.
Maybe save for later when we decide to split up `scio-extra`?

ravwojdyla:
How about we gradually split this, the ones moved in are obvious IOs. And once we figure out how to split BQ/Avro/TF we can move it around. The `scio-extra`, probably does not make sense to split just now yet, but once it's time, there would probably be another directory `extra` with further submodules?

nevillelyh:
This is outdated. Now I think more about it, doesn't seem to add any benefit and it might actually cause more confusion, e.g. project name and path mismatch. Close for now?

nevillelyh:
Gonna close for now.

PR #822, opened 2017-08-31
Add enhanced version of TableReference

ravwojdyla:
Add `asString` to make it simple to convert TableReference to its String
version acceptable by BigQueryClient's `query` or for formatting.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
@@ -17,7 +17,7 @@
 
 package com.spotify.scio
 
-import com.google.api.services.bigquery.model.{TableRow => GTableRow}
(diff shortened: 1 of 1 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
```diff
@@ -68,6 +68,11 @@ package object bigquery {
    */
   type description = com.spotify.scio.bigquery.types.description
 
+  /** Enhanced version of [[TableReference]]. */
+  implicit class RichTableReference(private val r: TableReference) extends AnyVal {
+    def asString: String = s"${r.getProjectId}:${r.getDatasetId}.${r.getTableId}"
```
nevillelyh: Call it `asTableSpec` to differentiate from `toString`? Also most Google code refers to the string as "TableSpec".

PR #823, opened 2017-08-31
Add query method that accepts TblReference

ravwojdyla:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -242,19 +242,33 @@ class BigQueryClient private (private val projectId: String,
...
       queryJob.table
     }
 
+  /**
+   * Make a query and save results to a destination table.
+   *
+   * A temporary table will be created if `destinationTable` is `null` and a cached table will be
+   * returned instead if one exists.
+   */
+  def query(sqlQuery: String,
```
nevillelyh: I think 2 methods with the same name & default arguments will cause problem. Can you double check?

ravwojdyla:
@nevillelyh yes. alright #822 should be enough for now. closing.

Issue #815, opened 2017-08-30
Issue in TF record coders


ravwojdyla:
There seems to be some kind of issue with TF record coders, there has been a couple of reports of issues with coder discovery, and I myself saw random failures like:

```
[info]   java.lang.IllegalArgumentException: requirement failed: Invalid masked CRC32 of length
[info]   at scala.Predef$.require(Predef.scala:277)
[info]   at com.spotify.scio.io.TFRecordCodec$.read(TFRecordCodec.scala:45)
```

or

```
Caused by: java.util.zip.ZipException: invalid stored block lengths
INFO:     at java.util.zip.InflaterInputStream.read(InflaterInputStream.java:164)
INFO:     at org.apache.commons.compress.compressors.deflate.DeflateCompressorInputStream.read(DeflateCompressorInputStream.java:69)
INFO:     at com.spotify.scio.io.TFRecordCodec$.readFully(TFRecord.scala:221)
INFO:     at com.spotify.scio.io.TFRecordCodec$.read(TFRecord.scala:197)
INFO:     at com.spotify.scio.io.TFRecordReader.readNextRecord(TFRecord.scala:103)
```

nevillelyh:
Is there any reproducible data?

ravwojdyla:
@nevillelyh `com.spotify.scio.io.TFTapTest` sporadically fails:

```
> testOnly com.spotify.scio.io.TFTapTest
[info] Compiling 1 Scala source to /Users/rav/projects/scio/scio-tensorflow/target/scala-2.12/test-classes...
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[info] TFTapTest:
[info] SCollection
[info] - should support saveAsTFRecordFile
[info] - should support saveAsTfExampleFile with case class or Seq feature spec *** FAILED ***
[info]   java.lang.IllegalArgumentException: requirement failed: Invalid masked CRC32 of length
[info]   at scala.Predef$.require(Predef.scala:277)
[info]   at com.spotify.scio.io.TFRecordCodec$.read(TFRecordCodec.scala:45)
[info]   at com.spotify.scio.io.TFFileStorageFunctions$$anon$1.<init>(TFFileStorageFunctions.scala:33)
[...]

PR #825, opened 2017-09-02
Fix kryo performance

kanterov:
Looks like there is performance regression introduced in https://github.com/spotify/scio/pull/478. With this fix, shuffle takes 5-10 times less. Add `scio-jmh` project for micro benchmarks for further improvements.

In one of our pipelines, a shuffle of ~1B records with 16 bytes key takes 4.5 hours instead of 22 hours before.

I propose to handle the case of 400 Mb key separately, cover it with benchmarks, and check if there is a regression in benchmarks for the case of a small key. One of other ways would be to explicitly ask users to specify different coder in case key is huge.

This fix is going to save us approximately 800,000 vCPU hours in an upcoming month, I don't mind backporting if we can ship it as 0.4.0.1.

## Benchmark

```
jmh:run -i 20 -wi 20 -f4 -t1
```

## Before

```
Benchmark                                  Mode  Cnt     Score     Error  Units
KryoAtomicCoderBenchmark.customDecode      avgt   80   286.811 ±   7.392  ns/op
KryoAtomicCoderBenchmark.customEncode      avgt   80   847.654 ±  11.846  ns/op
KryoAtomicCoderBenchmark.customKryoDecode  avgt   80  6184.079 ± 100.342  ns/op
KryoAtomicCoderBenchmark.customKryoEncode  avgt   80  6490.695 ± 196.628  ns/op
KryoAtomicCoderBenchmark.javaDecode        avgt   80  7787.809 ±  58.954  ns/op
KryoAtomicCoderBenchmark.javaEncode        avgt   80  6309.093 ± 105.909  ns/op
KryoAtomicCoderBenchmark.kryoDecode        avgt   80  5851.015 ± 126.595  ns/op
KryoAtomicCoderBenchmark.kryoEncode        avgt   80  6510.682 ± 155.126  ns/op
```

## After

```
Benchmark                                  Mode  Cnt     Score    Error  Units
KryoAtomicCoderBenchmark.customDecode      avgt   80   259.765 ±  1.359  ns/op
KryoAtomicCoderBenchmark.customEncode      avgt   80   813.831 ±  5.095  ns/op
KryoAtomicCoderBenchmark.customKryoDecode  avgt   80   224.362 ±  8.869  ns/op
KryoAtomicCoderBenchmark.customKryoEncode  avgt   80   998.436 ±  8.312  ns/op
KryoAtomicCoderBenchmark.javaDecode        avgt   80  7745.240 ± 54.849  ns/op
KryoAtomicCoderBenchmark.javaEncode        avgt   80  5854.949 ± 51.330  ns/op
KryoAtomicCoderBenchmark.kryoDecode        avgt   80   318.231 ±  2.674  ns/op
KryoAtomicCoderBenchmark.kryoEncode        avgt   80  1055.268 ±  6.411  ns/op
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -191,7 +191,8 @@ lazy val root: Project = Project(
   commonSettings ++ siteSettings ++ noPublishSettings,
   unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject
     -- inProjects(scioCassandra2) -- inProjects(scioElasticsearch2)
-    -- inProjects(scioRepl) -- inProjects(scioSchemas) -- inProjects(scioExamples),
+    -- inProjects(scioRepl) -- inProjects(scioSchemas) -- inProjects(scioExamples)
+    -- inProjects(scioBenchJmh),
   // unidoc handles class paths differently than compile and may give older
   // versions high precedence.
   unidocAllClasspaths in (ScalaUnidoc, unidoc) := {
@@ -218,7 +219,8 @@ lazy val root: Project = Project(
   scioTensorFlow,
   scioSchemas,
   scioExamples,
-  scioRepl
+  scioRepl,
+  scioBenchJmh
 )
 
 lazy val scioCore: Project = Project(
@@ -549,6 +551,23 @@ lazy val scioRepl: Project = Project(
   scioCore,
   scioExtra
 )
+
+lazy val scioBenchJmh: Project = Project(
+  "scio-bench-jmh",
+  file("scio-bench-jmh")
+).settings(
+  commonSettings ++ noPublishSettings,
+  description := "Scio JMH Microbenchmarks",
+  addCompilerPlugin(paradiseDependency),
+  sourceDirectory in Jmh := (sourceDirectory in Test).value,
+  classDirectory in Jmh := (classDirectory in Test).value,
+  dependencyClasspath in Jmh := (dependencyClasspath in Test).value,
+  libraryDependencies ++= Seq(
+    "org.slf4j" % "slf4j-nop" % slf4jVersion
+  )
+).dependsOn(
+  scioCore
+).enablePlugins(JmhPlugin)
 
 // =======================================================================
 // Site settings
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,6 +8,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.2")
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "1.0.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.0")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.27")
 
 libraryDependencies ++= Seq(
   "com.github.os72" % "protoc-jar" % "3.3.0.1"
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
@@ -17,9 +17,9 @@
 
 package com.spotify.scio.coders
 
-import java.io.{ByteArrayInputStream, InputStream, OutputStream}
-import java.nio.ByteBuffer
+import java.io.{InputStream, OutputStream}
 
+import com.esotericsoftware.kryo.io.{InputChunked, OutputChunked}
 import com.google.common.io.{ByteStreams, CountingOutputStream}
 import com.google.common.reflect.ClassPath
 import com.google.protobuf.Message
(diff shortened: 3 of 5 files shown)


kanterov:
According to benchmark, it looks like there is a big room for improvement for encoding

kanterov:
Tests are failing because code assumes that `isWholeStream=true`, but this property has been deprecated, going to check other ways to improve this.

kanterov:
Fixed compatibility, new benchmark

```
Benchmark                                  Mode  Cnt     Score     Error  Units
KryoAtomicCoderBenchmark.customDecode      avgt   80   254.170 ±   2.202  ns/op
KryoAtomicCoderBenchmark.customEncode      avgt   80   789.309 ±  12.876  ns/op
KryoAtomicCoderBenchmark.customKryoDecode  avgt   80   416.311 ±   3.471  ns/op
KryoAtomicCoderBenchmark.customKryoEncode  avgt   80  1145.902 ±   8.498  ns/op
KryoAtomicCoderBenchmark.javaDecode        avgt   80  7769.387 ± 106.154  ns/op
KryoAtomicCoderBenchmark.javaEncode        avgt   80  6197.226 ±  67.831  ns/op
KryoAtomicCoderBenchmark.kryoDecode        avgt   80   545.693 ±   3.605  ns/op
KryoAtomicCoderBenchmark.kryoEncode        avgt   80  1280.221 ±   8.888  ns/op
```

kanterov:
Got more performance by reusing Kryo buffers

```
Benchmark                                  Mode  Cnt     Score    Error  Units
KryoAtomicCoderBenchmark.customKryoDecode  avgt   80   214.340 ±  9.182  ns/op
KryoAtomicCoderBenchmark.customKryoEncode  avgt   80  1009.785 ± 12.134  ns/op
```

nevillelyh:
We reverted the change that causes `IndexOutOfBoundsException` and broke this change. Can you rebase please?

build.sbt
```diff
@@ -561,6 +563,23 @@ lazy val scioRepl: Project = Project(
   scioExtra
 )
 
+lazy val scioJmh: Project = Project(
+  "scio-jmh",
```
ravwojdyla: Don't like the naming of this project, a better one might be `scio-bench[mark]` as it's more important that it's benchmarks not what it uses for benchmarks. Also #830

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -147,18 +157,23 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
...
+private[scio] final case class KryoState(kryo: Kryo, input: InputChunked, output: OutputChunked)
+
 private[scio] object KryoAtomicCoder {
   def apply[T]: Coder[T] = new KryoAtomicCoder[T]
 
-  private val kryo: ThreadLocal[Kryo] = new ThreadLocal[Kryo] {
-    override def initialValue(): Kryo = {
+  val bufferSize = 64 * 1024
+
+  private val kryoState: ThreadLocal[KryoState] = new ThreadLocal[KryoState] {
```
ravwojdyla: This will cause IOOB exceptions dues to kryo state being shared across multiple instances (within the same thread). Beam has an internal heuristic to monitor size of collections and that seems to be using causing IOOB if kryo is shared to serialize and deserialize. This was the reason for #829.

kanterov:
@ravwojdyla Sorry it took me awhile to address review comments because we backported this fix internally, and it became a lower priority for us.

I called project `scio-jmh` because it's an important detail. JMH does code generation to avoid certain JIT optimizations, that affects the whole project and requires specific build setup, that's why such tests should be in a dedicated project, and shouldn't be mixed with other macro benchmarks from #830. As an alternative, we can name it as `scio-bench-jmh` to make it more clear that it does benchmarks.

I still don't fully understand how IOOB problem can happen, but I rebased the code according to master. I guess it shouldn't affect performance a lot because there should be only one instance of coder per executor.

I'm running benchmarks to update numbers.

kanterov:
@ravwojdyla updated numbers, please take a look again

ravwojdyla:
@kanterov:
 * `scio-bench-jmh` +1
 * please fix tests
 * `IOOB` is puzzling, in theory it should not happen (and i have validated that at the time of the exception, none of the threads share kryo instances), might be a bug in kryo, there were already similar bugs reported and fixed in kryo. Anyway, I would prefer having kryo in the class for now, until we figure out where exactly is the problem. Also related to this, have validated that if we set references to false, the bug is gone (but obviously that is not a great option).

build.sbt
```diff
@@ -566,6 +568,23 @@ lazy val scioRepl: Project = Project(
...
+  file("scio-jmh")
+).settings(
+  commonSettings,
+  noPublishSettings,
```
ravwojdyla: To keep it consistent with other project:
 * `commonSettings ++ noPublishSettings`
 * add `description`

project/plugins.sbt
```diff
@@ -8,6 +8,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.2")
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "1.0.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.0")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.24")
```
ravwojdyla: `0.2.27`?

build.sbt
kanterov: Fixed

project/plugins.sbt
kanterov: Fixed

kanterov:
@ravwojdyla @nevillelyh there are `scio-tensorflow` tests failing, but this has nothing related to this PR, because they randomly fail on master too, I was able to reproduce this on my machine after several tries.

kanterov:
Seems like a known issue, https://github.com/spotify/scio/issues/815

kanterov:
@ravwojdyla I tricked the system and tests passed this time

PR #826, opened 2017-09-05
Make runWithLocalOutput public to allow more generic testing

richwhitjr:
I find myself writing the body of this function often in my projects.  The runWithData methods are very useful for more thorough testing logic but sometimes my functions accept arguments other than SCollections.

This PR makes the lower level function public to allow for testing with local results.

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
@@ -127,7 +127,16 @@ trait PipelineTestUtils {
     }
   }
 
-  private def runWithLocalOutput[U](fn: ScioContext => SCollection[U]): Seq[U] = {
+  /**
+   * Test pipeline components with materialized resulting collection.
+   *
+   * The result [[com.spotify.scio.values.SCollection SCollection]] from `fn` is extracted and to be
+   * verified.
+   *
+   * @param fn Scio Job
+   * @return Job results in an in-memory Scala list
+   */
+  def runWithLocalOutput[U](fn: ScioContext => SCollection[U]): Seq[U] = {
     val sc = ScioContext()
     val f = fn(sc).materialize
     sc.close().waitUntilFinish()  // block non-test runner
--- a/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
@@ -128,13 +128,13 @@ trait PipelineTestUtils {
   }
 
   /**
-   * Test pipeline components with materialized resulting collection.
+   *  Test pipeline components with a [[ScioContext]] and materialized resulting collection.
    *
    * The result [[com.spotify.scio.values.SCollection SCollection]] from `fn` is extracted and to be
    * verified.
    *
    * @param fn transform to be tested
-   * @return Scio Result Object and Job results in an in-memory Scala list
+   * @return a tuple containing the [[ScioResult]] and the materialized result of fn as a [[Seq[T]]]
    */
   def runWithLocalOutput[U](fn: ScioContext => SCollection[U]): (ScioResult, Seq[U]) = {
     val sc = ScioContext()
--- a/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
@@ -128,13 +128,13 @@ trait PipelineTestUtils {
   }
 
   /**
-   *  Test pipeline components with a [[ScioContext]] and materialized resulting collection.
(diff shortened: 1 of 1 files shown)


scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -127,7 +127,16 @@ trait PipelineTestUtils {
...
   }
 
-  private def runWithLocalOutput[U](fn: ScioContext => SCollection[U]): Seq[U] = {
+  /**
+   * Test pipeline components with materialized resulting collection.
+   *
+   * The result [[com.spotify.scio.values.SCollection SCollection]] from `fn` is extracted and to be
+   * verified.
+   *
+   * @param fn Scio Job
```
ravwojdyla: `@param fn transform to be tested`

ravwojdyla:
@richwhitjr thanks. what is an example where using `runWithContext` or `runWithData` is not sufficient?

richwhitjr:
Something where one of the args is not a SCollection and the output needs a custom assert.  You can get pretty far with satisfy but it is a bit awkward to work with and doesn't show failed tests well.

Imagine some function like below where you create a matrix per user.  You may want to assert some properties on each matrix aside from direct equality.  Also the rank argument prevents using `runWithData`

```scala
def doMath(input: SCollection[Data], rank: Int): SCollection[Matrix]
```

kanterov:
@ravwojdyla internally we use similar thing a lot in property-based tests

ravwojdyla:
yea `satisfy` should allow you to implement arbitrary `predicate`, that said I see your points. while we are opening this up, maybe it should return `(ScioResult, Seq[U])`, since we already want to allow more generic test cases. what do you think @richwhitjr @kanterov ?

richwhitjr:
Sure, added the ScioResult object on the return also.

kanterov:
Exactly, that's what we use now:
```
  def closeAndMaterialize[A: ClassTag](as: SCollection[A]): (ScioResult, List[A]) = {
```

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -124,14 +124,23 @@ trait PipelineTestUtils {
   : Seq[U] = {
     runWithLocalOutput { sc =>
       fn(sc.parallelize(data1), sc.parallelize(data2), sc.parallelize(data3), sc.parallelize(data4))
-    }
+    }._2
   }
 
-  private def runWithLocalOutput[U](fn: ScioContext => SCollection[U]): Seq[U] = {
+  /**
+   * Test pipeline components with materialized resulting collection.
```
andrewsmartin: For consistency with the other `runWith*` methods - `Test pipeline components with a [[ScioContext]].`

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -124,14 +124,23 @@ trait PipelineTestUtils {
...
+   *
+   * The result [[com.spotify.scio.values.SCollection SCollection]] from `fn` is extracted and to be
+   * verified.
+   *
+   * @param fn transform to be tested
+   * @return Scio Result Object and Job results in an in-memory Scala list
```
andrewsmartin: Is something like this more clear? `@return a tuple containing the [[ScioResult]] and the materialized result of fn as a [[Seq[T]]]`

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -124,14 +124,23 @@ trait PipelineTestUtils {
   : Seq[U] = {
     runWithLocalOutput { sc =>
       fn(sc.parallelize(data1), sc.parallelize(data2), sc.parallelize(data3), sc.parallelize(data4))
-    }
+    }._2
   }
 
-  private def runWithLocalOutput[U](fn: ScioContext => SCollection[U]): Seq[U] = {
+  /**
+   *  Test pipeline components with a [[ScioContext]] and materialized resulting collection.
```
andrewsmartin: (minor formatting thing - there's an extra space before `Test`

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -124,14 +124,23 @@ trait PipelineTestUtils {
...
+   *
+   * The result [[com.spotify.scio.values.SCollection SCollection]] from `fn` is extracted and to be
+   * verified.
+   *
+   * @param fn transform to be tested
+   * @return a tuple containing the [[ScioResult]] and the materialized result of fn as a [[Seq[T]]]
```
andrewsmartin: My bad from the previous comment, this should be `Seq[U]` instead of `Seq[T]`

PR #829, opened 2017-09-07
Revert "move kryo to companion object, remove lazy"

ravwojdyla:
This reverts commit 1747f981eee7d10e3cf177ea52b88482292496df.

This commit is causing: IOOBE in GBK


nevillelyh:
👍 
Is there a reproducible snippet so we can include in the integration test/benchmark suites?

ravwojdyla:
@nevillelyh so far i was only able to reproduce this on a large job, working on scoping this to smaller case, but reverting that commit definitely solves the problem.

PR #838, opened 2017-09-13
Add header and footer to saveAsTextFile

labianchin:
Beam `TextIO.Write` now has configuration for `header` and `footer`: https://beam.apache.org/documentation/sdks/javadoc/2.1.0/org/apache/beam/sdk/io/TextIO.Write.html#withHeader-java.lang.String- . With this feature, users can define csv/tsv headers, as noted here: https://stackoverflow.com/a/44786025/1046584 .

This PR exposes this configuration as part of the `SCollection.saveAsTextFile()` method.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -837,11 +837,15 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       .withCodec(codec)
       .withMetadata(metadata.asJava)
 
-  private[scio] def textOut(path: String, suffix: String, numShards: Int) =
+  private[scio] def textOut(path: String, suffix: String, numShards: Int,
+                            header: Option[String] = None,
+                            footer: Option[String] = None) =
     gio.TextIO.write()
       .to(pathWithShards(path))
       .withSuffix(suffix)
       .withNumShards(numShards)
+      .withHeader(header.orNull)
+      .withFooter(footer.orNull)
 
   /**
    * Save this SCollection as an Avro file.
@@ -1140,7 +1144,10 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @group output
    */
   def saveAsTextFile(path: String,
-                     suffix: String = ".txt", numShards: Int = 0): Future[Tap[String]] = {
+                     suffix: String = ".txt",
+                     numShards: Int = 0,
+                     header: Option[String] = None,
+                     footer: Option[String] = None): Future[Tap[String]] = {
     val s = if (classOf[String] isAssignableFrom this.ct.runtimeClass) {
       this.asInstanceOf[SCollection[String]]
     } else {
@@ -1150,7 +1157,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       context.testOut(TextIO(path))(s)
       s.saveAsInMemoryTap
     } else {
-      s.applyInternal(textOut(path, suffix, numShards))
+      s.applyInternal(textOut(path, suffix, numShards, header, footer))
       context.makeFuture(TextTap(ScioUtil.addPartSuffix(path)))
     }
   }


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1140,7 +1144,9 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @group output
    */
   def saveAsTextFile(path: String,
-                     suffix: String = ".txt", numShards: Int = 0): Future[Tap[String]] = {
+                     suffix: String = ".txt", numShards: Int = 0,
```
ravwojdyla: Could you move `numShards: Int = 0` to next line?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1140,7 +1144,9 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+                     header: Option[String] = None,
```
ravwojdyla: Use `String` and `null` instead of `Option` for both `header` and `footer`.

nevillelyh:
Not sure this really belong here though, for a couple reasons:
- We try not to wrap every Beam feature, esp. the less common ones, and rather prefer to expose ones that cover majority of the use cases
- It can be done with a `saveAsCustomOutput`
- This is not symmetrical with `sc.textFile`. Plus I'm not sure how test would work. `JobTest` bypasses the actual IO so header/footer won't show up.

Maybe better to add an example in `examples/extra`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
labianchin: 👍

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
labianchin: Sure! I generally try to avoid nulls in scala. I am a bit lost here, how can I put `@Nullable` here in scala?

labianchin:
> We try not to wrap every Beam feature, esp. the less common ones, and rather prefer to expose ones that cover majority of the use cases

Makes sense. But it seems that csv/tsv is a recurrent use case. That motivated this PR.

> It can be done with a saveAsCustomOutput

Interesting. I will look into that.

> This is not symmetrical with sc.textFile. Plus I'm not sure how test would work. JobTest bypasses the actual IO so header/footer won't show up.

Yes. Good point! That is really weird to test with `TextIO`. 

> Comments + missing tests.

Not sure here, I struggled to find a place where I could change/add tests.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
ravwojdyla: @labianchin you would do `@Nullable header: String = null`, actually, let's not do it here.

labianchin:
I tried the `saveAsCustomOutput` and I think that is really a better option. Hence I am closing this PR.


Below how I used `saveAsCustomOutput`. Do you think there would be a place for this? Maybe `scio-extra`? Should this live in a separate scala file?

```
  implicit class IterableStringSCollection[T <: Iterable[Any]](val self: SCollection[T]) {

    private[IterableStringSCollection] def pathWithShards(path: String) =
      path.replaceAll("\\/+$", "") + "/part"

    def saveAsTsv[O <: POutput](path: String,
                              separator: String = "\t",
                              suffix: String = ".tsv",
                              header: Option[String] = None,
                              numShards: Int = 1):
    Future[Tap[String]] = {
      val transform = gio.TextIO.write()
        .to(pathWithShards(path))
        .withSuffix(suffix)
        .withNumShards(numShards)
        .withHeader(header.orNull)
      self
        .map(_.mkString(separator))
        .saveAsCustomOutput(path, transform)
    }

  }
```

ravwojdyla:
@labianchin as suggested above, you could add it as an `example in examples/extra` (don't forget to add tests), and you don't need to include the whole `implicit` class, just have an example that uses custom output with `TextIO` transform.

PR #842, opened 2017-09-13
Add flatten and flattenValues to SCollection

azymnis:
I noticed that `SCollection` did not have `flatten` and `flattenValues`. This PR adds them.


nevillelyh:
Thanks!

PR #851, opened 2017-09-19
support metrics assertions in JobTest

nevillelyh:
@azymnis FYI

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
@@ -19,8 +19,11 @@ package com.spotify.scio.testing
 
 import java.lang.reflect.InvocationTargetException
 
+import com.spotify.scio.ScioResult
+import com.spotify.scio.metrics.MetricValue
 import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.{metrics => bm}
 
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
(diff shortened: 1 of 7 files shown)


scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -733,12 +753,61 @@ class JobTestTest extends PipelineSpec {
...
     } should have message msg
   }
 
+  // =======================================================================
+  // Test metrics
+  // =======================================================================
+
+  it should "pass correct metrics test" in {
+    JobTest[MetricsJob.type]
+      .counter(ScioMetrics.counter("counter"))(_.committed shouldBe Some(10))
```
ravwojdyla: `committed` gets repeated over, is there a reason to expose `MetricValue` instead of directly the value from `committed` or maybe even value within the `Option` (if possible)?

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -122,6 +131,40 @@ object JobTest {
     }
 
     /**
+     * Evaluate a [[org.apache.beam.sdk.metrics.Counter Counter]] in the pipeline being tested.
+     * @param c counter to be evaluated
+     * @param f assertions for counter result
+     */
+    def counter(c: bm.Counter)(f: MetricValue[Long] => Unit): Builder = {
```
ravwojdyla: Maybe slightly better names for those arguments since it's user facing?

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -183,6 +186,23 @@ object JobWitDuplicateOutput {
   }
 }
 
+object MetricsJob {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+    val counter = ScioMetrics.counter("counter")
```
ravwojdyla: Move metrics outside of the `main` and into the object, just in case people look at this for the example.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
nevillelyh: Not sure, I guess it's a matter of how much we expose in `JobTest`.
I'm not exposing the `Map[MetricName, V]` or the entire `ScioResult` so it's kinda somewhere in between.

Question is how much is a good balance bewteen flexibility and API elegance?

nevillelyh: I guess for the case of `JobTest` it's almost always a batch job and the job must've finished when evaluating counters, so it's safe to just assert on committed result?

azymnis:
@nevillelyh LGTM

PR #852, opened 2017-09-19
remove direct runner dependency

### LINKED ISSUE #777: Remove DirectRunner dependency from scio-core
Currently, the DirectRunner class is pulled in ScioContext's <init> path through here : 
https://github.com/spotify/scio/blob/613b4a7264f89d0567c418f4c22436182a3098bd/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala#L52 

This makes it impossible to evict beam-runners-direct-java and all its dependencies while using scio 
to make the fat archive lighter, I don't think pulling this class here is necessary in the first place (but of course I might be wrong).

Can I work on a PR to use other means to guess if the current runner is local or remote ? 
Additionnally, some runners like flink have a local running mode where the url is unspecified and the host is '[local]'.

nevillelyh:
@Igosuki FYI

### DIFF
--- a/scio-elasticsearch2/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
+++ b/scio-elasticsearch2/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
@@ -19,8 +19,8 @@ package com.spotify.scio.elasticsearch
 
 import java.net.InetSocketAddress
 
-import com.spotify.scio.ContextAndArgs
-import com.spotify.scio.testing.PipelineSpec
+import com.spotify.scio._
+import com.spotify.scio.testing._
 import org.elasticsearch.action.index.IndexRequest
 import org.joda.time.Duration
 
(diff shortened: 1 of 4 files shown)


scio-elasticsearch2/src/test/scala/com/spotify/scio/elasticsearch/ElasticsearchTest.scala
```diff
@@ -44,29 +44,11 @@ object ElasticsearchJob {
   }
 }
 
-object ElasticsearchDirectRunnerJob {
```
ravwojdyla: Why do these ES jobs need to be removed?

nevillelyh: They look almost identical to the one without `DirectRunner` and I don't think they have anything to do with `DirectRunner` or not. It was just lousy review that they got in the first time ;)

Issue #779, opened 2017-08-16
Remove DataflowRunner dependency in scio-core


nevillelyh:
1. ScioContext.scala - job name and files to stage
1. ScioResult.scala - DF service metrics
1. ScioUtil.scala - DF service metrics
1. ScioILoop.scala - `.as(classOf[DataflowPipelineOptions])`, can probably remove?
1. ScioBenchmark.scala - DF specific benchmark and metrics, cannot remove

nevillelyh:
@ravwojdyla thoughts?
For 2 & 3, we'll have do stop including DF specific metrics, breaking but still doable.
4 should be easily doable as well.
How about 1?

nevillelyh:
Fixed in #882

PR #853, opened 2017-09-19
[WIP] add AsyncScioResult, fix #850

### LINKED ISSUE #850: Allow creating ScioResult independently
We should investigate whether it's possible to create a `ScioResult` independently from project ID and Dataflow job ID. This is so that we can submit a non-blocking job asynchronously and later retrieve state and metrics without a long running process.

nevillelyh:
Actually I'm not sure it's the right approach.
`AsyncScioResult` differs significantly from `BlockingScioResult` in behavior, especially the handling of `def getMetrics` since the `options` in this case is incomplete and not the same as the one used to submit the job.

Maybe better to just split up Beam metrics logic into a trait and add a different concrete class to provide those info alone.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -360,7 +360,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
     _isClosed = true
 
     _preRunFns.foreach(_())
-    val result = new ScioResult(this.pipeline.run(), context)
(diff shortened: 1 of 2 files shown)


nevillelyh:
Reworked a bit so that `getMetrics` and `saveMetrics` are not implemented in `AsyncScioResult`.
Not happy with the names though, `Async` and `Blocking` are misleading.

nevillelyh:
Gonna re-work this in the direction of #779. Closing.

PR #854, opened 2017-09-19
remove DataflowPipelineOptions from ScioILoop #779

nevillelyh:


ravwojdyla:
And this still works fine with dataflow jobs and local?

nevillelyh:
Yes works both for local and remote.

ravwojdyla:
@nevillelyh Sounds good. Yea I'm aware of the it being agnostic, just was wandering why was it there in the first place, and whether it had something to do with Classloader, but if it works fine for both local and remote, then it's all good.

PR #856, opened 2017-09-20
Fix for KryoAtomicClass may not be serializable due to logger

jasonjckn:


scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -102,7 +102,6 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
     }
   }
 
-  private val logger = LoggerFactory.getLogger(this.getClass)
```
nevillelyh: Replace with `import KryoAtomicCoder.logger` so the `logger.*` lines can stay the same?

nevillelyh:
Thanks!

PR #868, opened 2017-10-05
Fix #867. Use classbased thing cause scala/10064

### LINKED ISSUE #867: closeAndCollect in scio-repl hangs on scala 2.12.3
See title, to reproduce:

```
sc.parallelize(1 to 10).map(_ => "foo").closeAndCollect
```

the same thing will work just fine in repl on 2.11.11.

ravwojdyla:


ravwojdyla:
Fix/workaround for https://github.com/spotify/scio/issues/867

PR #876, opened 2017-10-10
BigQuery dynamic destinations

### LINKED ISSUE #73: Support toTableReference in BigQueryIO

regadas:
Closes #73

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigquery.dynamic
+
+import com.google.api.services.bigquery.model.TableSchema
+import org.apache.beam.sdk.coders.Coder
+import org.apache.beam.sdk.io.gcp.bigquery.{
+  DynamicDestinations,
+  TableDestination,
+  TableDestinationCoder
+}
+import org.apache.beam.sdk.util.Transport
+import org.apache.beam.sdk.values.ValueInSingleWindow
+
+private[dynamic] object DynamicDestinationsUtil {
+
+  def constant[T](dst: TableDestination,
+                  schema: TableSchema): DynamicDestinations[T, TableDestination] =
+    tableFn(_ => dst, schema)
+
+  def tableFn[T](fn: ValueInSingleWindow[T] => TableDestination,
+                 schema: TableSchema): DynamicDestinations[T, TableDestination] = {
+    val jsonSchema = Transport.getJsonFactory.toString(schema)
+    new DynamicDestinations[T, TableDestination] {
+      @transient private lazy val tableSchema =
+        Transport.getJsonFactory.fromString(jsonSchema, classOf[TableSchema])
+
+      override def getDestination(element: ValueInSingleWindow[T]): TableDestination =
+        fn(element)
+
+      override def getSchema(destination: TableDestination): TableSchema = tableSchema
+
+      override def getTable(destination: TableDestination): TableDestination = destination
+
+      override def getDestinationCoder: Coder[TableDestination] = TableDestinationCoder.of
+    }
+  }
+
+}
--- a/scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigquery
+
+import com.google.api.services.bigquery.model.TableSchema
+import com.spotify.scio.bigquery.types.BigQueryType.HasAnnotation
+import com.spotify.scio.io.Tap
+import com.spotify.scio.util.Functions
+import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.{CreateDisposition, WriteDisposition}
+import org.apache.beam.sdk.io.gcp.bigquery.{DynamicDestinations, TableDestination}
+import org.apache.beam.sdk.io.gcp.{bigquery => bqio}
+import org.apache.beam.sdk.values.ValueInSingleWindow
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+/**
+ * BigQuery package for dynamic table destinations. Import All.
+ *
+ * {{{
+ * import com.spotify.scio.bigquery.dynamic._
+ * }}}
+ */
+package object dynamic {
+
+  implicit class DynamicBigQuerySCollection[T](val self: SCollection[T]) extends AnyVal {
+
+    /**
+     * Save this SCollection to dynamic BigQuery tables using the table and schema
+     * specified by the [[org.apache.beam.sdk.io.gcp.bigquery.DynamicDestinations]] object.
+     */
+    def saveAsBigQuery(name: String,
+                       destinations: DynamicDestinations[T, _],
+                       formatFn: T => TableRow,
+                       writeDisposition: WriteDisposition,
+                       createDisposition: CreateDisposition): Future[Tap[T]] = {
+      if (self.context.isTest) {
+          throw new NotImplementedError(
+            "BigQuery with dynamic destinations cant be used in a test context")
+      } else {
+        var transform = bqio.BigQueryIO.write()
+          .to(destinations)
+          .withFormatFunction(Functions.serializableFn(formatFn))
+        if (createDisposition != null) {
+          transform = transform.withCreateDisposition(createDisposition)
+        }
+        if (writeDisposition != null) {
+          transform = transform.withWriteDisposition(writeDisposition)
+        }
+        self.applyInternal(transform)
+      }
+
+      Future.failed(
+        new NotImplementedError("BigQuery future with dynamic destinations not implemented"))
+    }
+
+    /**
+     * Save this SCollection to dynamic BigQuery tables using the specified table function.
+     * Note that elements must be of type
+     * [[com.google.api.services.bigquery.model.TableRow TableRow]].
+     */
+    def saveAsBigQuery(name: String,
+                       tableFn: ValueInSingleWindow[T] => TableDestination,
+                       schema: TableSchema,
+                       writeDisposition: WriteDisposition,
+                       createDisposition: CreateDisposition)
+                      (implicit ev: T <:< TableRow): Future[Tap[TableRow]] =
+      saveAsBigQuery(
+        name,
+        DynamicDestinationsUtil.tableFn(tableFn, schema),
+        (t: T) => t.asInstanceOf[TableRow],
+        writeDisposition,
+        createDisposition)
+        .asInstanceOf[Future[Tap[TableRow]]]
+
+
+    /**
+     * Save this SCollection to dynamic BigQuery tables using the specified table function.
+     * Note that element type `T` must be annotated with
+     * [[com.spotify.scio.bigquery.types.BigQueryType BigQueryType]].
+     */
+    def saveAsTypedBigQuery(name: String,
+                            tableFn: ValueInSingleWindow[T] => TableDestination,
+                            writeDisposition: WriteDisposition,
+                            createDisposition: CreateDisposition)
+                           (implicit ct: ClassTag[T], tt: TypeTag[T], ev: T <:< HasAnnotation)
+    : Future[Tap[T]] = {
+      val bqt = BigQueryType[T]
+      val destinations = DynamicDestinationsUtil.tableFn(tableFn, bqt.schema)
+
+      saveAsBigQuery(
+        name,
+        destinations,
+        bqt.toTableRow,
+        writeDisposition,
+        createDisposition)
+    }
+
+  }
+
+}


regadas:
@nevillelyh @andrewsmartin @ravwojdyla what do you think about this?

ravwojdyla:
We should not expose APIs with `TableRow`, it's too easy to shot yourself in the knee. Instead maybe we could add sth to typed API, maybe something similar to:

```
  // is writeDisposition, createDisposition still needed?
  def saveAsTypedBigQuery(writeDisposition: WriteDisposition,
                          createDisposition: CreateDisposition)
                         (destFn: T => TableDestination)
                         (implicit ct: ClassTag[T], tt: TypeTag[T], ev: T <:< HasAnnotation)
  : Future[Tap[T]] = {
    val bqt = BigQueryType[T]
    this.saveAsBigQueryWithDest(
          bqt.schema,
          writeDisposition,
          createDisposition,
          destFn,
          bqt.toTableRow
        )
      .map(_.map(bqt.fromTableRow))
  }

  private def saveAsBigQueryWithDest(schema: TableSchema,
                                     writeDisposition: WriteDisposition,
                                     createDisposition: CreateDisposition,
                                     destFn: T => TableDestination,
                                     formatFn: T => TableRow)
  (implicit ct: ClassTag[T], tt: TypeTag[T], ev: T <:< HasAnnotation)
  = {
      <blah>
     // need to handle table description from T if exists ?
      var transform = bqio.BigQueryIO.write[T]()
        .to((input: ValueInSingleWindow[T]) => destFn(input.getValue))
        .withFormatFunction((input: T) => formatFn(input))
      <blah>
  }
```

Plus a small helper to easily construct `TableDestination`. You could also make all the other BQ save methods go to `saveAsBigQueryWithDest` with some proper handling of arguments etc.

nevillelyh:
I'd also put such non-idiomatic API (since it doesn't `Tap`, `JobTest`, etc.) under a different namespace, maybe `com.spotify.io.dynamic`? Or any better names?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.beam.sdk.coders.Coder
+import org.apache.beam.sdk.io.gcp.bigquery.{
+  DynamicDestinations,
+  TableDestination,
+  TableDestinationCoder
+}
+import org.apache.beam.sdk.util.Transport
+import org.apache.beam.sdk.values.ValueInSingleWindow
+
+private[scio] object DynamicDestinationsUtil {
```
nevillelyh: this can be `private[dynamic]`?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+package object dynamic {
+
+  implicit class DynamicBigQuerySCollection[T](val self: SCollection[T]) extends AnyVal {
+
+    def saveAsBigQuery(name: String,
```
nevillelyh: Instead of `formatFn`, remove that and add a `implicit ev T <:< TableRow`? That's more consistent with vanilla `saveAsBigQuery` right?

nevillelyh:
Also can we keep `SCollection.scala` as is? Adding a almost cyclic dependency on `dynamic` is kinda scary.

regadas:
@nevillelyh yup it is. One thing I was about to propose is:

`scio-core` currently depends on `scio-bigquery` just because we need the bigquery types. Would it be madness to move all `saveAS{Typed}BigQuery` to `scio-bigquery`? that way we could invert the dependency arrow ... which is also inline with some of the other "io" modules

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
regadas: @nevillelyh removing the type constraint in favour of a function ... removes the need for `.toTableRow` and `.fromTableRow` map steps when using `saveAsTypedBigQuery`.

Since this is opt-in and only used internally as underlying implementation i thought it was ok. 

If you still think it's better to remove it let me know.

nevillelyh:
Pretty not making breaking changes in IO area right now. Could be confusing and require a lot of doc updating. Save it for later when we do the IO/Tap refactoring?

By that I mean:
- Keep `SCollection.scala` as is
- Add dynamic implicit one
- Keep `BigQueryIO` as is

That feels like minimal change and just an added pkg in `scio-core`. We can move everything over to `scio-bigquery` later, or not, doesn't matter a lot to us :)

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
```diff
@@ -0,0 +1,52 @@
...
+    new DynamicDestinations[T, TableDestination] {
+      override def getDestination(element: ValueInSingleWindow[T]): TableDestination =
+        fn(element)
+
+      override def getSchema(destination: TableDestination): TableSchema =
+        Transport.getJsonFactory.fromString(jsonSchema, classOf[TableSchema])
+
+      override def getTable(destination: TableDestination): TableDestination = destination
+
+      override def getDestinationCoder: Coder[TableDestination] = TableDestinationCoder.of
```
ravwojdyla: Is there a reason to override this?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
```diff
@@ -0,0 +1,52 @@
...
+
+private[scio] object DynamicDestinationsUtil {
+
+  def constant[T](dst: TableDestination,
+                  schema: TableSchema): DynamicDestinations[T, TableDestination] =
+    tableFn(_ => dst, schema)
+
+  def tableFn[T](fn: ValueInSingleWindow[T] => TableDestination,
+                 schema: TableSchema): DynamicDestinations[T, TableDestination] = {
+    val jsonSchema = Transport.getJsonFactory.toString(schema)
```
ravwojdyla: Shouldn't `jsonSchema` live inside `DynamicDestinations`?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
```diff
@@ -0,0 +1,52 @@
...
+    new DynamicDestinations[T, TableDestination] {
+      override def getDestination(element: ValueInSingleWindow[T]): TableDestination =
+        fn(element)
+
+      override def getSchema(destination: TableDestination): TableSchema =
+        Transport.getJsonFactory.fromString(jsonSchema, classOf[TableSchema])
```
ravwojdyla: Should the result of parsing be memoized? You could use a supplier pattern.

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2016 Spotify AB.
```
ravwojdyla: 2017

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2016 Spotify AB.
```
ravwojdyla: 2017

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
regadas: @ravwojdyla TableSchema is not serialisable

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
ravwojdyla: Missing doc.

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,81 @@
...
+          transform = transform.withWriteDisposition(writeDisposition)
+        }
+        self.applyInternal(transform)
+      }
+
+      Future.failed(
+        new NotImplementedError("BigQuery future with dynamic destinations not implemented"))
+    }
+
+    def saveAsTypedBigQuery(name: String,
```
ravwojdyla: Missing doc.

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,81 @@
...
+                       formatFn: T => TableRow,
+                       writeDisposition: WriteDisposition,
+                       createDisposition: CreateDisposition): Future[Tap[T]] = {
+      if (self.context.isTest) {
+        self.context.testOut(TypedBigQueryIO(name))(self)
+      } else {
+        var transform = bqio.BigQueryIO.write()
+          .to(destinations)
+          .withFormatFunction(Functions.serializableFn(formatFn))
+        if (createDisposition != null) {
```
ravwojdyla: No need for `{}` for a single else-branch-less `if`s

regadas: if find it weird if it's multi-line if but ok 😄

regadas:
@nevillelyh @ravwojdyla I think I addressed most of the stuff

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+package object dynamic {
+
+  implicit class DynamicBigQuerySCollection[T](val self: SCollection[T]) extends AnyVal {
+
+    /**
```
nevillelyh: Use javadoc style as rest of the code base. See how to configure IntelliJ in the style guide.

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.beam.sdk.io.gcp.{bigquery => bqio}
+import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.{CreateDisposition, WriteDisposition}
+import org.apache.beam.sdk.io.gcp.bigquery.{DynamicDestinations, TableDestination}
+import org.apache.beam.sdk.values.ValueInSingleWindow
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+package object dynamic {
```
nevillelyh: Add a brief doc here like the other pkg objs.

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  implicit class DynamicBigQuerySCollection[T](val self: SCollection[T]) extends AnyVal {
+
+    /**
+      * Save this SCollection to multiple BigQuery tables using the table and schema
+      * specified by the [[org.apache.beam.sdk.io.gcp.bigquery.DynamicDestinations]] object.
+      * @group output
+      */
+    def saveAsBigQuery(name: String,
```
nevillelyh: Can we do `implicit ev T <:< TableRow` instead of `formatFn`? Or is there reason not to?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,114 @@
...
+                       destinations: DynamicDestinations[T, _],
+                       formatFn: T => TableRow,
+                       writeDisposition: WriteDisposition,
+                       createDisposition: CreateDisposition): Future[Tap[T]] = {
+      if (self.context.isTest) {
+        self.context.testOut(DynamicBigQueryIO(name))(self)
```
nevillelyh: Not sure if it makes sense to capture this in test, since we can't easily verify the dynamic destination logic.
Get rid of `DynamicBigQueryIO` and throw an error here?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
regadas: @nevillelyh the only reason is to allow more freedom when it comes to build the `DynamicDestion`.

users using a case class with annotated BigQuery can easily access fields to build their `TableDestination`. With that type constraint they need to map it to TableRow and then use `get{Type}` methods to access fields.

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/DynamicDestinationsUtil.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def constant[T](dst: TableDestination,
+                  schema: TableSchema): DynamicDestinations[T, TableDestination] =
+    tableFn(_ => dst, schema)
+
+  def tableFn[T](fn: ValueInSingleWindow[T] => TableDestination,
+                 schema: TableSchema): DynamicDestinations[T, TableDestination] = {
+    val jsonSchema = Transport.getJsonFactory.toString(schema)
+    new DynamicDestinations[T, TableDestination] {
+      private lazy val tableSchema =
```
ravwojdyla: Should this be `@transient`?

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
regadas: @nevillelyh removed

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.beam.sdk.io.gcp.bigquery.{DynamicDestinations, TableDestination}
+import org.apache.beam.sdk.io.gcp.{bigquery => bqio}
+import org.apache.beam.sdk.values.ValueInSingleWindow
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+/**
+ * BigQuery package for multiple table destinations. Import All.
```
ravwojdyla: You use `multiple` multiple times here, while more it's might not be necessary true, it's more that it's dynamic, maybe you should use `dynamic` instead of `multiple`?

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -145,6 +145,8 @@ case class AvroIO[T](path: String) extends TestIO[T](path)
 
 case class BigQueryIO(tableSpecOrQuery: String) extends TestIO[TableRow](tableSpecOrQuery)
 
+case class DynamicBigQueryIO[T](name: String) extends TestIO[T](name)
```
ravwojdyla: Still needed if it's never used?

regadas: yeah forgot to remove it 👍

scio-core/src/main/scala/com/spotify/scio/bigquery/dynamic/package.scala
regadas: makes sense 👍

PR #877, opened 2017-10-11
Fix 874. Dump only the info plugin actually requires.

ravwojdyla:


nevillelyh:
No way to fix the plugin to parse and ignore BQTag?

ravwojdyla:
@nevillelyh that (ignoring BQTag) would be possible, but there is 2nd problem, the original quasiquotes did not capture all possible classes, thus some of BQ dumps were empty (resulting in no autocompletion). Since we need to fix scio anyway, we might as well remove all the obsolete (from the plugin point of view) information.

PR #882, opened 2017-10-12
decouple Dataflow runner, fix 779

nevillelyh:
Todo:
- [x] load `RunnerContext` via reflection
- [x] remove DF runner from ScioContextTest
- [x] ~remove DF runner from ScioContextIT~
- [x] verify that scio-examples and scio-repl work without DF runner
- [x] Support #850, allow creating ScioResult independently

Summary:
- `scio-core` depends on DF runner as `"provided"`
- `scio-test` depends on DF runner as `"it"`, in `ScioContextIT` (can be removed by making a dummy `FakeRemoteRunner`), and `ScioBenchmark` (how do we remove this?)
- `scio-repl` depends on DF runner, so it supports DF out of the box
- Runner specific logic in `ScioContext` are moved to `RunnerContext`, and loaded via reflection
- Runner specific logic in `ScioResult` are moved to `RunnerResult`, and loaded via `ScioResult#as`
- DF specific context and result are under `com.spotify.scio.runners.dataflow` in `scio-core`. Nothing is statically initialized so it should be fine without DF runner dependency.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -21,9 +21,8 @@ package com.spotify.scio
 
 import java.beans.Introspector
 import java.io.File
-import java.net.{URI, URLClassLoader}
+import java.net.URI
 import java.nio.file.Files
-import java.util.jar.{Attributes, JarFile}
 
 import com.google.api.services.bigquery.model.TableReference
 import com.google.datastore.v1.{Entity, Query}
@@ -41,8 +40,6 @@ import com.spotify.scio.values._
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecordBase
-import org.apache.beam.runners.dataflow.DataflowRunner
-import org.apache.beam.runners.dataflow.options.DataflowPipelineWorkerPoolOptions
 import org.apache.beam.sdk.PipelineResult.State
 import org.apache.beam.sdk.extensions.gcp.options.{GcpOptions, GcsOptions}
 import org.apache.beam.sdk.io.gcp.{bigquery => bqio, datastore => dsio, pubsub => psio}
@@ -63,6 +60,44 @@ import scala.io.Source
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
 import scala.util.Try
+
+/** Runner specific context. */
+trait RunnerContext {
+  def prepareOptions(options: PipelineOptions, artifacts: List[String]): Unit
+}
+
+private case object NoOpContext extends RunnerContext {
+  override def prepareOptions(options: PipelineOptions, artifacts: List[String]): Unit = Unit
+}
+
+/** Direct runner specific context. */
+private case object DirectContext extends RunnerContext {
+  override def prepareOptions(options: PipelineOptions, artifacts: List[String]): Unit = Unit
+}
+
+/** Companion object for [[RunnerContext]]. */
+private object RunnerContext {
+  private val mapping = Map(
+    "DirectRunner" -> DirectContext.getClass.getName,
+    "DataflowRunner" -> "com.spotify.scio.runners.dataflow.DataflowContext$")
+    .withDefaultValue(NoOpContext.getClass.getName)
+
+  // FIXME: this is ugly, is there a better way?
+  private def get(options: PipelineOptions): RunnerContext = {
+    val runner = options.getRunner.getName
+    val cls = mapping(runner)
+    try {
+      Class.forName(cls).getField("MODULE$").get(null).asInstanceOf[RunnerContext]
+    } catch {
+      case e: Throwable =>
+        throw new RuntimeException(
+          s"Failed to load runner specific context $cls for $runner", e)
+    }
+  }
+
+  def prepareOptions(options: PipelineOptions, artifacts: List[String]): Unit =
+    get(options).prepareOptions(options, artifacts)
+}
 
 /** Convenience object for creating [[ScioContext]] and [[Args]]. */
 object ContextAndArgs {
@@ -223,8 +258,7 @@ class ScioContext private[scio] (val options: PipelineOptions,
   def pipeline: Pipeline = {
     if (_pipeline == null) {
       // TODO: make sure this works for other PipelineOptions
-      Try(optionsAs[DataflowPipelineWorkerPoolOptions])
-        .foreach(_.setFilesToStage(getFilesToStage(artifacts).asJava))
+      RunnerContext.prepareOptions(options, artifacts)
       _pipeline = if (testId.isEmpty) {
         Pipeline.create(options)
       } else {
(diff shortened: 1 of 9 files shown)


nevillelyh:
TODO:
- [ ] remove DF runner from ScioContextIT
- [ ] remove DF runner from ScioContextTest
- [ ] verify that scio-examples and scio-repl work without DF runner

nevillelyh:
@ravwojdyla take a look please?

- Should we remove the `"it"` dep completely from `scio-test`?
- Could use a hand double check REPL functionality.

nevillelyh:
Couldn't figure out a way to retrieve counter values from the Dataflow API directly. Will punt on this.

nevillelyh:
Decided to keep the DF runner `"it"` dependency in `scio-test` since we also need to IT test DF results.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -62,7 +60,41 @@ import scala.concurrent.{Future, Promise}
...
+private object RunnerContext {
+  private val mapping = Map(
+    "DirectRunner" -> DirectContext.getClass.getName,
+    "DataflowRunner" -> "com.spotify.scio.runners.dataflow.DataflowContext$")
+    .withDefaultValue(NoOpContext.getClass.getName)
+
+  // FIXME: this is ugly, is there a better way?
+  private def get(options: PipelineOptions): RunnerContext =
+    Class
+      .forName(mapping(options.getRunner.getName))
```
ravwojdyla: Have some proper `try` and error massage in case of missing class? For example that maybe some runner dependency is missing etc.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -62,7 +60,41 @@ import scala.concurrent.{Future, Promise}
...
+private case object DirectContext extends RunnerContext {
+  override def prepareOptions(options: PipelineOptions, artifacts: List[String]): Unit = Unit
+}
+
+/** Companion object for [[RunnerContext]]. */
+private object RunnerContext {
+  private val mapping = Map(
+    "DirectRunner" -> DirectContext.getClass.getName,
+    "DataflowRunner" -> "com.spotify.scio.runners.dataflow.DataflowContext$")
+    .withDefaultValue(NoOpContext.getClass.getName)
```
ravwojdyla: Would it be better to fail instead of some default `NoOpContext`?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: It's getting one of the value strings in `mapping` with a fallback `NoOpContext` so this should never fail.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: We could print a warning message, but should have some default behavior for unknown runners. Right now `RunnerContext` only injects Scio-REPL specific logic for `DataflowRunner`.

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -20,52 +20,41 @@ package com.spotify.scio
...
+}
 
 /** Represent a Scio pipeline result. */
-class ScioResult private[scio] (val internal: PipelineResult, val context: ScioContext) {
+abstract class ScioResult private[scio] (val internal: PipelineResult) {
 
-  private val logger = LoggerFactory.getLogger(this.getClass)
+  /** Get a Beam runner specific result. */
+  def as[T <: RunnerResult : ClassTag]: T =
+    ScioUtil.classOf[T].getConstructor(classOf[PipelineResult]).newInstance(internal)
```
ravwojdyla: Should we have some error handling (log message) here - in case of exception from `getCNTR` or `newInstance`?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ravwojdyla: Well if there is unknown runner wouldn't it fail anyway, so would it make sense to just fail asap (?)

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ravwojdyla: Unless classnames change - like `"com.spotify.scio.runners.dataflow.DataflowContext$"`, or that class is for some reason now available.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: Unknown runner should work out of the box for the most case. We only use this to inject files to stage for DF runner. So we shouldn't fail them.

ravwojdyla: @nevillelyh ah i see, that makes sense.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -62,7 +60,41 @@ import scala.concurrent.{Future, Promise}
...
+
+  // FIXME: this is ugly, is there a better way?
+  private def get(options: PipelineOptions): RunnerContext =
```
ravwojdyla: This is not part of any performance critical path right?

scio-core/src/main/scala/com/spotify/scio/metrics/package.scala
```diff
@@ -36,11 +36,9 @@ package object metrics {
    */
   case class Metrics(version: String,
```
ravwojdyla: Btw - since you are changes stuff around here, could you please make all those case classes `final`?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: No it's used in `def prepareOptions` once per context.

scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
```diff
@@ -0,0 +1,124 @@
...
+
+/** Represent a Dataflow runner specific result. */
+class DataflowResult(val internal: DataflowPipelineJob) extends RunnerResult {
+
+  def this(internal: PipelineResult) = this(internal.asInstanceOf[DataflowPipelineJob])
+
+  private val client = DataflowResult.getOptions(internal.getProjectId).getDataflowClient
+
+  /** Get Dataflow [[Job]]. */
+  lazy val getJob: Job = client.projects().jobs()
```
ravwojdyla: I think there should be retries here, a single failure of any of those calls would fail a whole workflow - usually at the very end.

scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
```diff
@@ -0,0 +1,124 @@
...
+
+}
+
+/** Companion object for [[DataflowResult]]. */
+object DataflowResult {
+  /** Create a new [[DataflowResult]] instance. */
+  def apply(projectId: String, jobId: String): DataflowResult = {
+    val options = getOptions(projectId)
+
+    val job = options.getDataflowClient.projects().jobs()
```
ravwojdyla: Duplicated code.

scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
```diff
@@ -0,0 +1,124 @@
...
+      .get(options.getProject, jobId)
+      .setView("JOB_VIEW_ALL")
+      .execute()
+    val transformStepNames: Map[AppliedPTransform[_, _, _], String] =
```
ravwojdyla: Where did you get this from?

nevillelyh: Trial and error, a lot of :D

ravwojdyla: Maybe add a comment here that explains why this exists, and what you are doing here?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: Actually `DataflowResult` can be re-constructed with just project and job ID, async, and independent of submission so maybe keep it simple here and let caller (some centralized service that monitors job completion) handle retry/fallback?

scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
nevillelyh: Actually `DataflowResult` can be re-constructed with just project and job ID, async, and independent of submission so maybe keep it simple here and let caller (some centralized service that monitors job completion) handle retry/fallback?

PR #889, opened 2017-10-18
Fix Option support in flatten, fixes #886

### LINKED ISSUE #886: flatten should support Option
Standard collection's flatten looks like this:
`def flatten[B](implicit asTraversable: A => /*<:<!!!*/ GenTraversableOnce[B])`

And ours look like this:
`def flatten[U: ClassTag](implicit ev: T <:< TraversableOnce[U]): SCollection[U]`

We should fix it so that it doesn't require `TraversableOnce`.

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -308,15 +309,15 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
     this.parDo(Functions.flatMapFn(f))
 
   /**
    * Return a new SCollection[U] by flattening each element of an SCollection[Traversable[U]].
    * @group transform
    */
-  def flatten[U: ClassTag](implicit ev: T <:< TraversableOnce[U]): SCollection[U] =
-    flatMap(_.asInstanceOf[TraversableOnce[U]])
+  def flatten[U: ClassTag](implicit ev: T => GenTraversableOnce[U]): SCollection[U] =
+    flatMap(e => e)
```
nevillelyh: should be `.flatMap(ev)`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -308,15 +309,15 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * this SCollection, and then flattening the results.
    * @group transform
    */
-  def flatMap[U: ClassTag](f: T => TraversableOnce[U]): SCollection[U] =
+  def flatMap[U: ClassTag](f: T => GenTraversableOnce[U]): SCollection[U] =
```
nevillelyh: What's the difference between `TO[U]` and `GTO[U]`? This would break backward compatibility so prefer not to change it.

scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
```diff
@@ -199,6 +199,11 @@ class SCollectionTest extends PipelineSpec {
       val p = sc.parallelize(Seq(Seq("a b", "c d"), Seq("e f", "g h"))).flatten
       p should containInAnyOrder(Seq("a b", "c d", "e f", "g h"))
     }
+
+    runWithContext { sc =>
```
nevillelyh: You can reuse the previous `runWithContext` like other tests and have `val p1 = /*Seq*/` and `val p2 = /*Option*/`.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: @nevillelyh well `TraversableOnce[+A] extends Any with GenTraversableOnce[A]`

probably missing coffee 😄 what do mean by breaking change?

nevillelyh: I mean libs with scio as "provided" dependency may get a `NoSuchMethod` error calling `flatMap(f: T => TO[U])`.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: @nevillelyh 🤦‍♂️ yes!

if we decide to keep `TraversableOnce` than this can be `flatMap(e => ev(e).seq)`

nevillelyh: Do we need `.seq` in this case? It returns a sequential version, I guess to preserve order. But `SCollection` doesn't guarantee order anyway.

PR #895, opened 2017-10-20
Fix #894. Support default value for singleton side input

### LINKED ISSUE #894: Support default values for Singleton SideInput
Currently afaiu, users can't provide default values for SideInputs. They would have to:

```
sc.parallelize(Some(Set(1))).internal.apply(View.asSingleton[Set[Int]]().withDefaultValue(Set.empty))
```

but user can't really use it with `withSideInputs`, also wrap from `SCollection` is only for `PCollection` not `PCollectionView`.

ravwojdyla:


nevillelyh:
Need to think about this. Default may not be enough if the side input content is populated by an external source like file system read. In that case need to make sure that there's a signal event for each window to trigger a reload of external source. Or is there a way for a window to inherit previous one's content?

ravwojdyla:
@nevillelyh well, the default would be used only if there is no side input for that specific window or it can't be projected? Could you give an example where default value would break?

PR #897, opened 2017-10-23
Allow to wrap internal Views as SideInput

ravwojdyla:
Related to #894

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -42,6 +42,43 @@ trait SideInput[T] extends Serializable {
   private[values] val view: PCollectionView[_]
 }
 
+/** Companion object of [[SideInput]]. */
+object SideInput {
+
+  /**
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -42,6 +42,42 @@ trait SideInput[T] extends Serializable {
   private[values] val view: PCollectionView[_]
 }
 
+object SideInput {
```
nevillelyh: I know it's silly but a one liner doc "Companion object of ..." so it shows up nicely in API site?

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -42,6 +42,42 @@ trait SideInput[T] extends Serializable {
...
+
+  /**
+   * Wrap a view of a Singleton as a [[SideInput]]. In most cases you want to use
```
nevillelyh: `singleton`, `multi-map` as they're not types, but `List`, `Iterable`, `Map` as they are proper types?

PR #899, opened 2017-10-24
Honor caching configuration in BigQueryClient

derfloscher:
Disabling caching currently only works for queries running inside a ScioContext. This pull request enables standalone usage of the BigQueryClient without caching of results.


nevillelyh:
Thanks!

PR #901, opened 2017-10-25
make joins artisanal

nevillelyh:


nevillelyh:
Benchmark with artisan join:
```
================================================================================
Benchmark           : Join
Extra arguments     :
State               : DONE
Create time         : 2017-10-25T12:34:03.524821Z
Finish time         : 2017-10-25T12:39:14.293802Z
Elapsed             : 310 seconds
TotalMemoryUsage    : 12902672
TotalPdUsage        : 210004
TotalShuffleUsage   : 0
TotalSsdUsage       : 0
TotalVcpuTime       : 3360
================================================================================
Benchmark           : JoinShuffleService
Extra arguments     : --experiments=shuffle_mode=service
State               : DONE
Create time         : 2017-10-25T12:34:13.055091Z
Finish time         : 2017-10-25T12:38:09.348833Z
Elapsed             : 236 seconds
TotalMemoryUsage    : 9216059
TotalPdUsage        : 15000
TotalShuffleUsage   : 1402.9782047707961
TotalSsdUsage       : 0
TotalVcpuTime       : 2400
================================================================================
Benchmark           : JoinOne
Extra arguments     :
State               : DONE
Create time         : 2017-10-25T12:34:20.450720Z
Finish time         : 2017-10-25T12:39:51.653228Z
Elapsed             : 331 seconds
TotalMemoryUsage    : 14438925
TotalPdUsage        : 235008
TotalShuffleUsage   : 0
TotalSsdUsage       : 0
TotalVcpuTime       : 3760
================================================================================
Benchmark           : JoinOneShuffleService
Extra arguments     : --experiments=shuffle_mode=service
State               : DONE
Create time         : 2017-10-25T12:34:29.414849Z
Finish time         : 2017-10-25T12:39:01.420537Z
Elapsed             : 272 seconds
TotalMemoryUsage    : 10444870
TotalPdUsage        : 17000
TotalShuffleUsage   : 1736.0044514975659
TotalSsdUsage       : 0
TotalVcpuTime       : 2720
```

Without:
```
================================================================================
Benchmark           : Join
Extra arguments     :
State               : DONE
Create time         : 2017-10-25T12:35:41.973046Z
Finish time         : 2017-10-25T12:40:23.674882Z
Elapsed             : 281 seconds
TotalMemoryUsage    : 12288160
TotalPdUsage        : 200002
TotalShuffleUsage   : 0
TotalSsdUsage       : 0
TotalVcpuTime       : 3200
================================================================================
Benchmark           : JoinShuffleService
Extra arguments     : --experiments=shuffle_mode=service
State               : DONE
Create time         : 2017-10-25T12:35:51.096561Z
Finish time         : 2017-10-25T12:39:54.254032Z
Elapsed             : 243 seconds
TotalMemoryUsage    : 9830495
TotalPdUsage        : 16000
TotalShuffleUsage   : 1511.6223033818137
TotalSsdUsage       : 0
TotalVcpuTime       : 2560
================================================================================
Benchmark           : JoinOne
Extra arguments     :
State               : DONE
Create time         : 2017-10-25T12:35:59.388225Z
Finish time         : 2017-10-25T12:41:48.371730Z
Elapsed             : 348 seconds
TotalMemoryUsage    : 15360622
TotalPdUsage        : 250010
TotalShuffleUsage   : 0
TotalSsdUsage       : 0
TotalVcpuTime       : 4000
================================================================================
Benchmark           : JoinOneShuffleService
Extra arguments     : --experiments=shuffle_mode=service
State               : DONE
Create time         : 2017-10-25T12:36:07.869511Z
Finish time         : 2017-10-25T12:41:13.274216Z
Elapsed             : 305 seconds
TotalMemoryUsage    : 13517018
TotalPdUsage        : 22000
TotalShuffleUsage   : 2056.8867823699939
TotalSsdUsage       : 0
TotalVcpuTime       : 3520
```

Not huge but the jobs are probably too small to show significant difference.

nevillelyh:
JMH results, the iterator approach seems to be winning bigly.

```
[info] Benchmark                Mode  Cnt      Score     Error  Units
[info] JoinBenchmark.artisanA   avgt   10     62.385 ±   1.621  ns/op
[info] JoinBenchmark.artisanB   avgt   10     60.965 ±   2.669  ns/op
[info] JoinBenchmark.artisanC   avgt   10   2708.904 ±  62.317  ns/op
[info] JoinBenchmark.artisanD   avgt   10   2104.059 ±  52.542  ns/op
[info] JoinBenchmark.forYieldA  avgt   10    165.752 ±   4.958  ns/op
[info] JoinBenchmark.forYieldB  avgt   10    616.117 ±  12.153  ns/op
[info] JoinBenchmark.forYieldC  avgt   10  13915.338 ± 234.575  ns/op
[info] JoinBenchmark.forYieldD  avgt   10  12352.121 ± 709.163  ns/op
[info] JoinBenchmark.iteratorA  avgt   10     49.629 ±   1.333  ns/op
[info] JoinBenchmark.iteratorB  avgt   10     51.991 ±   3.405  ns/op
[info] JoinBenchmark.iteratorC  avgt   10   1974.178 ±  64.523  ns/op
[info] JoinBenchmark.iteratorD  avgt   10   2113.973 ±  65.832  ns/op
```

nevillelyh:
@kanterov @ravwojdyla take a look? Wanna squeeze this into the next release as well.

kanterov:
@nevillelyh I've found few issues with this code:

1. due to copy-paste `iterator` was actually using `artisan`
2. in order for JVM not to remove computations we have to feed output into `Blackhole`, or return from methods annotated with `@Benchmark`
3. I don't see any significant results in macro-benchmarks, seems that we need to do experiment few times in order to make conclusions. I suspect there would be a high deviation in runtime.

I don't have a good machine to run benchmarks now, but I pushed the code to `gleb/join`, see https://github.com/spotify/scio/commit/639a91647f3444b3ce13f4e3e1860764acaf3bc2

Also, I summoned old-school artisan, I think we should benchmark him too:

```scala
  def artisan2(as: JIterable[Int], bs: JIterable[Int], blackhole: Blackhole): Unit = {
    val ai = as.iterator()

    while (ai.hasNext) {
      val a = ai.next()
      val bi = bs.iterator()

      while (bi.hasNext) {
        blackhole.consume((a, bi.next()))
      }
    }
  }
```

kanterov:
It looks like `iterator` version is slower than `artisan` that is expected because it maintains more sophisticated state.

nevillelyh:
@kanterov 2nd try after fixing 1 & 2. `artisan` is still better and more symmetrical. Should we use that instead?
```
[info] Benchmark                Mode  Cnt      Score     Error  Units
[info] JoinBenchmark.artisanA   avgt   10    129.219 ±  25.141  ns/op
[info] JoinBenchmark.artisanB   avgt   10    114.871 ±   6.662  ns/op
[info] JoinBenchmark.artisanC   avgt   10   8090.208 ± 414.707  ns/op
[info] JoinBenchmark.artisanD   avgt   10   8720.866 ± 377.688  ns/op
[info] JoinBenchmark.forYieldA  avgt   10    223.826 ±  10.835  ns/op
[info] JoinBenchmark.forYieldB  avgt   10    614.489 ±  19.460  ns/op
[info] JoinBenchmark.forYieldC  avgt   10  17163.704 ± 700.357  ns/op
[info] JoinBenchmark.forYieldD  avgt   10  20126.729 ± 807.624  ns/op
[info] JoinBenchmark.iteratorA  avgt   10    171.974 ±   6.787  ns/op
[info] JoinBenchmark.iteratorB  avgt   10    301.541 ±   8.697  ns/op
[info] JoinBenchmark.iteratorC  avgt   10  14887.663 ± 525.564  ns/op
[info] JoinBenchmark.iteratorD  avgt   10  15672.826 ± 577.532  ns/op
```

kanterov:
@neville did you try artisan2?

nevillelyh:
I'm trying a combination of `artisan2` and the `peak` logic in `artisan`. Seems improving slightly over all other previous ones. Running a longer benchmark now.

nevillelyh:
The latest. `muchArtisan` wins all 4. Though the difference is pretty small.
```
[info] Benchmark                   Mode  Cnt      Score      Error  Units
[info] JoinBenchmark.artisanA      avgt   15    138.344 ±    6.689  ns/op
[info] JoinBenchmark.artisanB      avgt   15    123.971 ±    5.232  ns/op
[info] JoinBenchmark.artisanC      avgt   15   9842.301 ±  389.014  ns/op
[info] JoinBenchmark.artisanD      avgt   15  10242.343 ±  668.314  ns/op
[info] JoinBenchmark.iteratorA     avgt   15    205.540 ±    5.738  ns/op
[info] JoinBenchmark.iteratorB     avgt   15    313.193 ±   13.243  ns/op
[info] JoinBenchmark.iteratorC     avgt   15  17093.920 ±  478.775  ns/op
[info] JoinBenchmark.iteratorD     avgt   15  17790.001 ±  939.659  ns/op
[info] JoinBenchmark.muchArtisanA  avgt   15    125.268 ±    7.320  ns/op
[info] JoinBenchmark.muchArtisanB  avgt   15    129.582 ±   10.140  ns/op
[info] JoinBenchmark.muchArtisanC  avgt   15   9339.662 ± 1239.620  ns/op
[info] JoinBenchmark.muchArtisanD  avgt   15   9693.621 ± 2146.684  ns/op
```

nevillelyh:
Fixed everything and switched to plain while loop approach.

kanterov:
Looks good, but I'm concerned about error rate, meaning we don't know for sure which method is better because their 99% confidence intervals overlap a lot. Probably, re-running this benchmarks on a machine without Slack and IntelliJ and more iterations and forks will work :). 

On the other side, it looks reasonable that the most simple approach has won. It would be nice to generalize our benchmarks to test `DoFn` instead. I'm not sure if we want to specialize implementations for `left`, `right`, `outer` join.

What do you think about using [joinlibrary](https://github.com/apache/beam/blob/master/sdks/java/extensions/join-library/src/main/java/org/apache/beam/sdk/extensions/joinlibrary/Join.java) from Beam and converting `KV` to `Tuple2`?

nevillelyh:
Yeah I've been traveling and haven't gotten around to run the benchmarks properly. It was still a tie last time I ran with IntelliJ, Slack, etc. turned off.

I agree the current approach is simplest and should be good enough™, i.e. 2 nested while loops and no specialized left/right/outer. Don't think join library will be any different since it's almost the exact implementation. Should we stick with this, release and see how it works out first?

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def outer[KEY: ClassTag, A: ClassTag, B: ClassTag](name: String,
+                                                     a: SCollection[(KEY, A)],
+                                                     b: SCollection[(KEY, B)])
+  : SCollection[(KEY, (Option[A], Option[B]))] = cogroup(name, a, b)(toOptions, toOptions)
+
+  private def toOptions[A](xs: JIterable[A]): JIterable[Option[A]] =
+    if (xs.iterator().hasNext) {
+      Iterables.transform(xs, new com.google.common.base.Function[A, Option[A]] {
+        override def apply(input: A) = Option(input)
```
ravwojdyla: Should this be `Some(input)`, instead of `Option` specifically if the RHS contains valid `null` that might infer `None` for a valid `null` RHS value.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
```diff
@@ -0,0 +1,94 @@
...
+                                                     b: SCollection[(KEY, B)])
+  : SCollection[(KEY, (Option[A], B))] = cogroup(name, a, b)(toOptions, identity)
+
+  def outer[KEY: ClassTag, A: ClassTag, B: ClassTag](name: String,
+                                                     a: SCollection[(KEY, A)],
+                                                     b: SCollection[(KEY, B)])
+  : SCollection[(KEY, (Option[A], Option[B]))] = cogroup(name, a, b)(toOptions, toOptions)
+
+  private def toOptions[A](xs: JIterable[A]): JIterable[Option[A]] =
+    if (xs.iterator().hasNext) {
```
ravwojdyla: `hasNext` = `isEmpty`?

ravwojdyla: nvm.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
```diff
@@ -0,0 +1,94 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.util
+
+import java.lang.{Iterable => JIterable}
+
+import com.google.common.collect.{AbstractIterator, Iterables}
```
ravwojdyla: Unused import `AbstractIterator`.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
```diff
@@ -0,0 +1,94 @@
...
+  : SCollection[(KEY, (Option[A], B))] = cogroup(name, a, b)(toOptions, identity)
+
+  def outer[KEY: ClassTag, A: ClassTag, B: ClassTag](name: String,
+                                                     a: SCollection[(KEY, A)],
+                                                     b: SCollection[(KEY, B)])
+  : SCollection[(KEY, (Option[A], Option[B]))] = cogroup(name, a, b)(toOptions, toOptions)
+
+  private def toOptions[A](xs: JIterable[A]): JIterable[Option[A]] =
+    if (xs.iterator().hasNext) {
+      Iterables.transform(xs, new com.google.common.base.Function[A, Option[A]] {
```
ravwojdyla: You could probably just use `Iterator`s here, thus use `Iterators.transform()` since you already create a an iterator above. And change `toOption` to return java `Iterator` since it seems like you only need `Iterator` in `cogroup` anyway.

nevillelyh: I actually need `Iterable` since the 2 while loops need to re-iterate over left and right hand side. Also `Iterators.transform` just calls `Iterables.transform` so there's no real performance difference.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
nevillelyh: We should never have null input here since there's no null coder. So `Some` or `Option` doesn't make a huge difference.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
ravwojdyla: You could just create an Iterator inside the `while` loop. It's `Iterables.transform` who calls `Iterators.transform` not the other way around, right?

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
ravwojdyla: One less `if` then :)

nevillelyh: Can't really get rid of the `if` here. Empty list => `List(None)`, non-empty list => `xs.map(Some(_))`. I'll change `Option(input)` to `Some(input)` but that's only for readability.

ravwojdyla: I mean inside the `Option`, there is `if` null check that goes to `Some` or `None`, but since this is always `Some`, should make sense to just use `Some`, but yes - readability.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
nevillelyh: Yes typo got it backwards. `Iterables.transform` calls `Iterators.transform` not the other way around. But with `Iterators.transform` the code is actually less readable.

scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
nevillelyh: 🤣 the invisible `if`

PR #913, opened 2017-11-02
Use ThreadLocalRandom instead of scala Random

ravwojdyla:


nevillelyh:
Is this to address a particular issue?

ravwojdyla:
@nevillelyh no - it's just better to use `ThreadLocalRandom` in a multithreading env.

Issue #916, opened 2017-11-07
hashJoin is using MultiMapSideInput


daikeshi:
`hashJoin` is using `MultiMapSideInput` which may cause poor throughput performance 

https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala#L240

Evaluate using singleton side input, run some performance tests etc.

daikeshi:
https://github.com/spotify/scio/pull/917

PR #917, opened 2017-11-07
Use SingletonSideInput instead of MultiMapSideInput for hashJoin 

daikeshi:
This pr is to fix issue https://github.com/spotify/scio/issues/916

@ravwojdyla @nevillelyh

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -28,6 +28,7 @@ import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.values.{KV, PCollection, PCollectionView}
 import org.slf4j.LoggerFactory
 
+import scala.collection.mutable.{ArrayBuffer, Map => MMap}
 import scala.reflect.ClassTag
 
 private object PairSCollectionFunctions {
@@ -237,9 +238,18 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def hashJoin[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (V, W))] = self.transform { in =>
-    val side = that.asMultiMapSideInput
+    val side = that.combine { case (k, v) =>
+      MMap(k -> ArrayBuffer(v))
+    } { case (combiner, (k, v)) =>
+      combiner.getOrElseUpdate(k, ArrayBuffer.empty[W]) += v
+      combiner
+    } { case (left, right) =>
+        right.foreach { case (k, vs) => left.getOrElseUpdate(k, ArrayBuffer.empty[W]) ++= vs }
+        left
+    }.asSingletonSideInput
+
     in.withSideInputs(side).flatMap[(K, (V, W))] { (kv, s) =>
-      s(side).getOrElse(kv._1, Iterable()).iterator.map(w => (kv._1, (kv._2, w)))
+      s(side).getOrElse(kv._1, ArrayBuffer.empty[W]).map(w => (kv._1, (kv._2, w)))
     }.toSCollection
   }
 
@@ -251,11 +261,20 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def hashLeftJoin[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (V, Option[W]))] = self.transform { in =>
-    val side = that.asMultiMapSideInput
+    val side = that.combine { case (k, v) =>
+      MMap(k -> ArrayBuffer(v))
+    } { case (combiner, (k, v)) =>
+      combiner.getOrElseUpdate(k, ArrayBuffer.empty[W]) += v
+      combiner
+    } { case (left, right) =>
+      right.foreach { case (k, vs) => left.getOrElseUpdate(k, ArrayBuffer.empty[W]) ++= vs }
+      left
+    }.asSingletonSideInput
+
     in.withSideInputs(side).flatMap[(K, (V, Option[W]))] { (kv, s) =>
       val (k, v) = kv
       val m = s(side)
-      if (m.contains(k)) m(k).iterator.map(w => (k, (v, Some(w)))) else Seq((k, (v, None)))
+      if (m.contains(k)) m(k).iterator.map(w => (k, (v, Some(w)))) else Iterator((k, (v, None)))
     }.toSCollection
   }
 


daikeshi:
the downside of this fix is that it will take longer to load the side input.

ravwojdyla:
Could you give share some performance numbers on how this will improve performance for at least some synthetic data?

daikeshi:
Yep, I'll run some jobs on test data

nevillelyh:
Take a look at this: https://github.com/spotify/scio/blob/master/scio-test/src/it/scala/com/spotify/ScioBenchmark.scala
You can run it from SBT directly, with & without your commit in PR. You should also add `--regex=.*HashJoin.*` to run that benchmark only.
Probably wanna increase the data size also.

nevillelyh:
@daikeshi did you get a chance to run the benchmark? Would be nice to compare when side input is close/above cache size.

ravwojdyla:
@daikeshi any update here?

daikeshi:
@ravwojdyla @nevillelyh sorry guys, I was on vacation for past two weeks. I'm going to get this done  this week!

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -237,9 +237,9 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def hashJoin[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (V, W))] = self.transform { in =>
-    val side = that.asMultiMapSideInput
+    val side = that.map(kv => Map(kv._1 -> Seq(kv._2))).sum.asSingletonSideInput
     in.withSideInputs(side).flatMap[(K, (V, W))] { (kv, s) =>
-      s(side).getOrElse(kv._1, Iterable()).toSeq.map(w => (kv._1, (kv._2, w)))
```
nevillelyh: Now I think about it, toSeq may be eager and causing major performer penalty.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -237,9 +237,9 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def hashJoin[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (V, W))] = self.transform { in =>
-    val side = that.asMultiMapSideInput
+    val side = that.map(kv => Map(kv._1 -> Seq(kv._2))).sum.asSingletonSideInput
```
nevillelyh: Even though sum might use Semigroup#sumOption under the hood, I'd make sure if it's comparable with combining into a mutable map.

ravwojdyla: You can use `asSingletonSideInput(Seq.empty[W])`, and you would need `getOrElse` later on.

nevillelyh:
Per slack discussion, there's no need to keep the multi-map side input based implementation since it breaks the RHS in-memory semantic and is probably just as bad as a full join.

OTOH maybe make `hashLeftJoin` a singleton side input too? Should be trivial.

daikeshi:
baseline asMultiMapSideInput (toSeq): 313 seconds
baseline asMultiMapSideInput (iterator): 269 seconds
singleton side input: 215 seconds

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -237,9 +237,9 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
     in.withSideInputs(side).flatMap[(K, (V, W))] { (kv, s) =>
-      s(side).getOrElse(kv._1, Iterable()).iterator.map(w => (kv._1, (kv._2, w)))
+      s(side).getOrElse(kv._1, Seq.empty[W]).map(w => (kv._1, (kv._2, w)))
```
ravwojdyla: using `asSingletonSideInput(Seq.empty[W])` did not work?

daikeshi: hmm, the side input is a map, `getOrElse` is to return an empty seq for key that doesn't exist in the map.

ravwojdyla: @daikeshi my bad - you are right.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -237,9 +238,18 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
...
+      combiner.getOrElseUpdate(k, ArrayBuffer.empty[W]) += v
+      combiner
+    } { case (left, right) =>
+        right.foreach { case (k, vs) => left.getOrElseUpdate(k, ArrayBuffer.empty[W]) ++= vs }
+        left
+    }.asSingletonSideInput
+
     in.withSideInputs(side).flatMap[(K, (V, W))] { (kv, s) =>
-      s(side).getOrElse(kv._1, Iterable()).iterator.map(w => (kv._1, (kv._2, w)))
+      s(side).getOrElse(kv._1, ArrayBuffer.empty[W]).map(w => (kv._1, (kv._2, w)))
```
nevillelyh: Add the `.iterator` back?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -237,9 +238,18 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)])
    */
   def hashJoin[W: ClassTag](that: SCollection[(K, W)])
   : SCollection[(K, (V, W))] = self.transform { in =>
-    val side = that.asMultiMapSideInput
+    val side = that.combine { case (k, v) =>
+      MMap(k -> ArrayBuffer(v))
```
daikeshi: @nevillelyh @ravwojdyla I updated to use combine with mutable.Map and ArrayBuffer, what do you guys think? looks good?

daikeshi:
Benchmark numbers on hashJoin mutable.Map on ArrayBuffer vs Map on Seq:
```
singleton side input (Map + Seq): 215 seconds
singleton side input (mutable.Map + ArrayBuffer): 189 seconds
```

PR #924, opened 2017-11-14
Example for TSV output with headers

labianchin:
As discussed earlier on https://github.com/spotify/scio/pull/838#issuecomment-329491035

> you could add it as an example in examples/extra (don't forget to add tests), and you don't need to include the whole implicit class, just have an example that uses custom output with TextIO transform.

PTAL @ravwojdyla

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/TsvExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/TsvExample.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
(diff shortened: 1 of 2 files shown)


labianchin:
> Maybe worth adding a read example as well?

@nevillelyh yes. I was looking into add a read example. But I found no simple way to read TSV/CSV with headers. In that case I am not sure what I can do for a read example. Still, I left a comment about about that. Let me know if you have other ideas.

ravwojdyla:
@labianchin how about just adding a read example that starts with `textFile` followed by `filter` on a `header`? that is probably the best we can do right now without overly complicating things.

ravwojdyla:
Closing moved to #974

PR #927, opened 2017-11-15
tag and exclude slow tests, fix #909

### LINKED ISSUE #909: DataflowIT fails when scoverage is on
We need to move it to a separate sbt step, maybe also `clean` before `it:test`.

nevillelyh:


build.sbt
```diff
@@ -88,6 +88,13 @@ val commonSettings = Sonatype.sonatypeSettings ++ assemblySettings ++ Seq(
 
   scalastyleSources in Compile ++= (unmanagedSourceDirectories in Test).value,
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v"),
+  testOptions ++= {
+    if (sys.env.contains("SLOW")) {
```
andrewsmartin: This isn't set on circleCI right? Wonder if we should set up separate IT builds (nightly / pre-release)?

nevillelyh: Added a step in release guide to do this manually.

PR #929, opened 2017-11-15
Add PubSub admin helper methods

regadas:


nevillelyh:
`[error] /home/ubuntu/scio/scio-core/src/main/scala/com/spotify/scio/pubsub/PubSubAdmin.scala: Header does not match expected text`

regadas:
@nevillelyh 🤦‍♂️

scio-core/src/main/scala/com/spotify/scio/pubsub/PubSubAdmin.scala
```diff
@@ -0,0 +1,92 @@
...
+import io.grpc.auth.ClientAuthInterceptor
+import io.grpc.netty.{GrpcSslContexts, NegotiationType, NettyChannelBuilder}
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubOptions
+
+import scala.util.Try
+
+object PubSubAdmin {
+
+  private object GrpcClient {
+    @SuppressWarnings(Array("org.wartremover.warts.Null"))
```
nevillelyh: We don't have wartremover in this repo right? Remove this?

nevillelyh:
We can't use the more high level API like this? https://cloud.google.com/pubsub/docs/quickstart-client-libraries#pubsub-client-libraries-java

PR #930, opened 2017-11-16
require extra map in Parquet Avro, fix #928

### LINKED ISSUE #928: parquet read can NPE on chained attributes
When attempting to project parquet records to just certain nested fields (eg `Projection[SomeType](_.getFoo.getBar)` I saw many NPEs apparently from serializing the records while trying to read them. Switching to just projecting the top-level field (`_.getFoo`) got past it.

nevillelyh:
@pdeglopper take a look?

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
@@ -22,7 +22,7 @@ import java.lang.{Boolean => JBoolean}
 import com.spotify.scio.ScioContext
 import com.spotify.scio.io.Tap
 import com.spotify.scio.testing.AvroIO
-import com.spotify.scio.util.ScioUtil
+import com.spotify.scio.util.{ClosureCleaner, ScioUtil}
 import com.spotify.scio.values.SCollection
 import org.apache.avro.Schema
 import org.apache.avro.reflect.ReflectData
@@ -31,6 +31,7 @@ import org.apache.beam.sdk.io.hadoop.inputformat.HadoopInputFormatIO
 import org.apache.beam.sdk.io.{DefaultFilenamePolicy, FileBasedSink, HadoopWriteFiles}
 import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
 import org.apache.beam.sdk.transforms.SimpleFunction
+import org.apache.beam.sdk.values.TypeDescriptor
 import org.apache.hadoop.mapreduce.Job
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat
 import org.apache.parquet.avro.AvroParquetInputFormat
@@ -38,6 +39,7 @@ import org.apache.parquet.filter2.predicate.FilterPredicate
 import org.apache.parquet.hadoop.ParquetInputFormat
 
 import scala.concurrent.Future
+import scala.language.implicitConversions
 import scala.reflect.ClassTag
 
 /**
@@ -59,7 +61,9 @@ package object avro {
   implicit class ParquetAvroScioContext(val self: ScioContext) extends AnyVal {
 
     /**
-     * Get an SCollection for a Parquet file as Avro records.
+     * Get an SCollection for a Parquet file as Avro records. Since Avro records produced by
+     * Parquet column projection may be incomplete and may fail serialization, you must
+     * [[ParquetAvroFile.map map]] the result to extract projected fields from the Avro records.
      *
      * Note that due to limitations of the underlying `HadoopInputFormatIO`,
      * Avro [[org.apache.avro.generic.GenericRecord GenericRecord]] and dynamic work rebalancing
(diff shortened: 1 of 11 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -59,9 +61,11 @@ package object avro {
   implicit class ParquetAvroScioContext(val self: ScioContext) extends AnyVal {
 
     /**
-     * Get an SCollection for a Parquet file as Avro records.
+     * Get an SCollection for a Parquet file as Avro records. Since Avro records produced by
+     * Parquet column projection may be incomplete and may fail serialization, you must
+     * [[ParquetAvroFile.map map]] the result to extract projected fields from the Avro records.
      *
-     * Note that due to limitations of the underlying `HadoopInputFormatIO`,
+     * Note that due to limitations of the underlying HadoopInputFormatIO`,
```
ravwojdyla: ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -122,8 +145,7 @@ package object avro {
    * Enhanced version of [[com.spotify.scio.values.SCollection SCollection]] with Parquet Avro
    * methods.
    */
-  implicit class ParquetAvroSCollection[T : ClassTag]
-  (val self: SCollection[T]) {
+  implicit class ParquetAvroSCollection[T : ClassTag](val self: SCollection[T]) {
```
ravwojdyla: + `extends AnyVal` and move the context bound to the `saveAsParquetAvroFile`?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -71,36 +75,55 @@ package object avro {
...
+      }
+
+      val uCls = ScioUtil.classOf[U]
+      val source = HadoopInputFormatIO.read[JBoolean, U]()
+        .withConfiguration(job.getConfiguration)
+        .withKeyTranslation(new SimpleFunction[Void, JBoolean]() {
+          override def apply(input: Void): JBoolean = true // workaround for NPE
+        })
+        .withValueTranslation(new SimpleFunction[T, U]() {
+          // workaround for incomplete Avro objects
```
ravwojdyla: Could you elaborate here - why is this fixing the problem?

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroTest.scala
```diff
@@ -30,7 +30,7 @@ import org.scalatest._
 object ParquetAvroJob {
   def main(cmdlineArgs: Array[String]): Unit = {
     val (sc, args) = ContextAndArgs(cmdlineArgs)
-    sc.parquetAvroFile[TestRecord](args("input"))
+    sc.parquetAvroFile[TestRecord](args("input")).map(identity)
```
ravwojdyla: What is the purpose of this extra `map`?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
nevillelyh: LOL typo I blame intellij

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
nevillelyh: Now I think about this, it might be better to expose `PCollectionWrapper#ct` so that we can remove the `ClassTag` context bound everywhere. I'll do it in a separate PR.

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroTest.scala
nevillelyh: Because it needs 2 implicit conversions to go from `ParquetAvroFile` to `SCollection` to `ParquetSCollection`. I'll add implicit methods in a companion `object ParquetAvroFile` to make it possible. It may not cover some other cases though, so we won't be able to disguise this completely as a regular `SCollection`.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
ravwojdyla: Are you gonna submit that patch before this PR or after?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
ravwojdyla: Thanks!

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroTest.scala
ravwojdyla: Ah - i see now. What do you thinkg about having it as an argument in `parquetAvroFile` instead of having intermediate type. So it would be:

```
     def parquetAvroFile[T <: SpecificRecordBase : ClassTag, U: ClassTag](path: String,
                                                              projection: Schema = null,
                                                              extraction: (T => U))
```

or curried version:

```
     def parquetAvroFile[T <: SpecificRecordBase : ClassTag, U: ClassTag](path: String,
                                                              projection: Schema = null,
                                                              predicate: FilterPredicate = null)
                                                              (extraction: (T => U))
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
nevillelyh: After, I remember why `ct` was protected now, changing it to public would require changing all subclasses. Better do it in a separate one.

PR #931, opened 2017-11-17
add BigtableDoFn, fix #922

### LINKED ISSUE #922: Abstraction Bigtable side lookup
WIP branch https://github.com/spotify/scio/tree/neville/bigtable

A `DoFn` that abstract away:
- resource management, i.e. sharing a single Bigtable client
- concurrency, i.e. making async requests under the hood in a safe manner compatible with `DoFn`s threading model
- optional  caching

User implements specific querying logic by overriding:
`ListenableFuture<B> processElement(BigtableSession session, A input)`

What's missing:
- Optional caching, maybe a shared Guava `Cache<K, V>`?
- Error handling, right now it's not customizable and there's no retrying logic
- Unit test
- Benchmarks

nevillelyh:


nevillelyh:
@danielnorberg care to take a quick look? Wanna make sure I'm not doing anything silly with the futures.

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
```diff
@@ -0,0 +1,301 @@
...
+        futures.remove(uuid);
+        return output;
+      }
+    });
+
+    // This `put` may happen after `remove` in the callbacks but it's OK since either the result
+    // or the error would've already been pushed to the corresponding queues and we are not losing
+    // data. `waitForFutures` in `finishBundle` blocks until all pending futures, including ones
+    // that may have already completed, and `startBundle` clears everything.
+    futures.put(uuid, f);
```
danielnorberg: Might this leak if this happens after `futures.remove(uuid)` ?

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
```diff
@@ -0,0 +1,301 @@
...
+      }
+    });
+
+    // Handle success
+    ListenableFuture<B> f = Futures.transform(future, new Function<B, B>() {
+      @Nullable
+      @Override
+      public B apply(@Nullable B output) {
+        cacheSupplier.put(instanceId, input, output);
+        results.add(new Result(input, output, c.timestamp(), window));
```
danielnorberg: Could the result be emitted immediately using `c.output()` or is there some benefit to doing that later in `flush` ?

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
```diff
@@ -0,0 +1,301 @@
...
+      @Override
+      public void onFailure(Throwable t) {
+        semaphore.release();
+        errors.add(t);
+        futures.remove(uuid);
+      }
+    });
+
+    // Handle success
+    ListenableFuture<B> f = Futures.transform(future, new Function<B, B>() {
```
danielnorberg: Maybe this transform could be eliminated and the code moved to the callback above? Seeing as it is not actually transforming the input. Or do you need the ordering guarantee?

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
nevillelyh: The buffer behind `c.output()` is not thread safe, so everything has to be done on the main thread in `flush()`. Similar problem to what we had in Storm bolts.

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
nevillelyh: We're doing a blocking `Futures.allAsList(futures.values()).get();` in `finishBundle` to ensure that all pending futures are processed. Without adding the transformed `f` into `futures` the callback may happen after `Futures.allAsList(...).get()` unblocks and may cause some results to be lost in the following `flush`.

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
nevillelyh: Yes, `f` may be already complete before this happens. But we do `Futures.allAsList` in `finishBundle` to ensure all pending ones are handled and `futures.clear()` in `startBundle` to clean up. So there shouldn't be any leak? Is there a better way to handle this?

nevillelyh:
Added tests. @ravwojdyla PTAL. Wanna get it out and get some user feedback, WDYT?

regadas:
@nevillelyh PTAL at this https://gist.github.com/regadas/7371b54721549151f3e1ca31b0dff780. Might be a little bit green still.

I did some simplifications:
* pushed error handling to user.
* pushed cache implementation to user
* use bigtable grpc "throttled" grpc client by default. it offers `ResourceLimiter` underwood and configurable through `BigtableOptions`

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
```diff
@@ -0,0 +1,325 @@
...
+  // same BigtableSession and Cache.
+  private static final ConcurrentMap<UUID, BigtableSession> session = Maps.newConcurrentMap();
+  private static final ConcurrentMap<UUID, Cache> cache = Maps.newConcurrentMap();
+  private final UUID instanceId;
+
+  private final BigtableOptions options;
+  private final CacheSupplier<A, B, ?> cacheSupplier;
+
+  // Data structures for handling async requests
+  private final Semaphore semaphore;
```
regadas: @nevillelyh since we are exposing `BigtableSession` users have access to throttled client. Do you think it's better to have this extra guard?

nevillelyh:
Agree. I'll look into it.

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
nevillelyh: Actually prefer to leave this here. `BigtableSession` has a lot of methods, including 2 data clients and is confusing. This gives us extra guarantee that the `DoFn` won't blow up and be killed by the runner.

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableDoFn.java
algirdas-k: Offtopic: exactly five years later this question and comment helps other developers to achieve their goals. Thank you!

PR #935, opened 2017-11-21
Fix #934 use simple class name in runner context map.

### LINKED ISSUE #934: Dependencies not uploaded to GCS with DataflowRunner in 0.4.4
After decoupling the DF-runner from `scio-core`, dependency jars fail to get uploaded to the staging location in GCS because of a bug in the `RunnerContext` mapping logic. It works when running via `sbt` because classpath entries are uploaded by default, but in other cases it does not work. Will push a fix shortly.

andrewsmartin:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -86,7 +86,7 @@ private object RunnerContext {
 
   // FIXME: this is ugly, is there a better way?
   private def get(options: PipelineOptions): RunnerContext = {
-    val runner = options.getRunner.getName
+    val runner = Class.forName(options.getRunner.getName).getSimpleName
```
nevillelyh: Wait so the original returns `org.apache.beam....DirectRunner` and new one gets `DirectRunner`?
Does `options.getRunner.getSimpleName` work?

andrewsmartin: @nevillelyh correct. And oops, yeah let me fix that.

Issue #902, opened 2017-10-25
TF API should allow access to the input function arguments on output


richwhitjr:
It would be nice if the Tensorflow Predict functions allowed access to the input arguments in the output function.  Sometimes it is helpful to append on a variable from the inputs to the output scores before writing somewhere.

The simplest case would be to pull the T given to the input function to the output function also.

nevillelyh:
Can we close this now?

PR #939, opened 2017-11-23
Propagate input to predict output fn

martinbomio:
This is a very simple way of solving https://github.com/spotify/scio/issues/902.
Tested it with my own workflow and seems to be working.

### DIFF
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
@@ -44,7 +44,7 @@ private class PredictDoFn[T, V](graphBytes: DistCache[Array[Byte]],
                                 fetchOp: Seq[String],
                                 @Nullable config: Array[Byte],
                                 inFn: T => Map[String, Tensor],
-                                outFn: Map[String, Tensor] => V) extends DoFn[T, V] {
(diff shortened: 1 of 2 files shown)


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
```diff
@@ -44,7 +44,7 @@ private class PredictDoFn[T, V](graphBytes: DistCache[Array[Byte]],
                                 fetchOp: Seq[String],
                                 @Nullable config: Array[Byte],
                                 inFn: T => Map[String, Tensor],
-                                outFn: Map[String, Tensor] => V) extends DoFn[T, V] {
+                                outFn: (T, Map[String, Tensor]) => V) extends DoFn[T, V] {
```
martinbomio: Another option is to create a custom type or case class like `PredictionResult(original: T, outputOps: Map[String, Tensor])`

ravwojdyla:
LGTM, still complex API, but we can address that later. @richwhitjr would this help in your case too?

richwhitjr:
Yep, lets me try the API at least in a real word example.

martinbomio:
@richwhitjr I tried it in one of our workflows, but please try it!

ravwojdyla:
@richwhitjr any update here?

richwhitjr:
Don't have a TF model handy right now.  I can pull an old example and see if it works compared to calling the API directly.

richwhitjr:
Yep this works for me on a small local test.  LGTM

PR #942, opened 2017-11-23
Double lock get/cache of side input

ravwojdyla:
Without double locking it is possible multiple instance of DoFns will
try to materialize side input.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -34,8 +34,14 @@ trait SideInput[T] extends Serializable {
   private[values] def get[I, O](context: DoFn[I, O]#ProcessContext): T
   def getCache[I, O](context: DoFn[I, O]#ProcessContext): T = {
     if (cache == null || this.context.ne(context)) {
-      this.context = context
-      cache = get(context)
+      // TODO(rav): reference equality might be too strict, we might want to check pane equality
+      //            instead.
+      context.synchronized {
+        if (cache == null || this.context.ne(context)) {
+          this.context = context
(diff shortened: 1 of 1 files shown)


ravwojdyla:
wot is this shenanigans with those multiple commits.

ravwojdyla:
Now that's better.

nevillelyh:
Did you hit an issue when running the pipeline? What's the symptom like?

ravwojdyla:
@nevillelyh yes - @yonromai was running a pipeline, with a Sparkey side input, that would run out of memory even on large VMs. This is rather significant bug(s), also the reference equality check might be too much here (check of pane might be enough - but need to research it). Need to research this more, but this overall the problem is that for side inputs we might:
 * have a race condition for computing cache value
 * context equality check might be to strict, and cause cache to be overwritten (at least it seems like it does locally) - which in case of expensive Side inputs might be really expensive

Do not merge yet.

nevillelyh:
IIRC Sparkey uses mmap under the hood so multiple instances should share the same OS cache memory as long as they map to the same file?

nevillelyh:
I think #959 is the real culprit. It's probably cache miss on every element since a new `context` is created per element, thus creating many Sparkey reader instances and OOM.
#960 should be a proper fix.

nevillelyh:
@ravwojdyla also I don't think double locking or thread-safety is the issue here, since this `SideInput` instance is pulled in from closure and deserialized once per CPU as part of the `DoFn`.
The only way to share resource among cloned copies is via a singleton concurrent Map + per `SideInput` UUID as key like we did in `BigtableDoFn` but that might be overkill.

1 `SparkeyReader` per CPU is still acceptable as long as they share the underlying mmap.

PR #949, opened 2017-11-29
add flag for kryo.registration, fix #896

### LINKED ISSUE #896: Expose Kryo#setRegistrationRequired
`Kryo#setRegistrationRequired` ensures that every used Kryo coder is pre-registered. It significantly reduces the amount of data shuffled by using class_id lookup instead of fully qualified class names. At the moment, it's possible to enable through custom `KryoRegistrar`:

```scala
@KryoRegistrar
class CustomKryoRegistrar extends IKryoRegistrar {
  def apply(k: Kryo): Unit = {
  }
}
```

It's especially useful in `DirectRunner` that ensures that data in serializable on each pipeline step without doing fusion. We discussed to expose this in a more convenient way because this is considered as a good practice for optimizing big data pipelines.

nevillelyh:


nevillelyh:
@kanterov PTAL?

nevillelyh:
Per slack discussion we want to control this via `PipelineOptions`, so maybe pass that in via `RichCoderRegistry#getScalaCoder`

nevillelyh:
WIP with Kryo options https://github.com/spotify/scio/commit/0fd7532c43cb37113ec7b0099729f8ff24b5385f

nevillelyh:
Proper fix in #955

Issue #952, opened 2017-11-30
hashLeftJoin fails when RHS is empty


devictr:
hashLeftJoin seems to fail when RHS is empty since Scio 0.4.5
Haven't checked if this affects hashJoin yet

```
Caused by: java.util.NoSuchElementException: Empty PCollection accessed as a singleton view.
	at org.apache.beam.sdk.values.PCollectionViews$SingletonViewFn.getDefaultValue(PCollectionViews.java:204)
	at org.apache.beam.sdk.values.PCollectionViews$SingletonViewFn.apply(PCollectionViews.java:229)
	at org.apache.beam.sdk.values.PCollectionViews$SingletonViewFn.apply(PCollectionViews.java:165)
	at org.apache.beam.runners.direct.SideInputContainer$SideInputContainerSideInputReader.get(SideInputContainer.java:246)
	at org.apache.beam.runners.direct.repackaged.runners.core.SimpleDoFnRunner.sideInput(SimpleDoFnRunner.java:206)
	at org.apache.beam.runners.direct.repackaged.runners.core.SimpleDoFnRunner.access$900(SimpleDoFnRunner.java:66)
	at org.apache.beam.runners.direct.repackaged.runners.core.SimpleDoFnRunner$DoFnProcessContext.sideInput(SimpleDoFnRunner.java:408)
	at com.spotify.scio.values.SingletonSideInput.get(SideInput.scala:84)
	at com.spotify.scio.values.SideInput$class.getCache(SideInput.scala:38)
	at com.spotify.scio.values.SingletonSideInput.getCache(SideInput.scala:82)
	at com.spotify.scio.values.SideInputContext.apply(SideInput.scala:114)
	at com.spotify.scio.values.PairSCollectionFunctions$$anonfun$hashLeftJoin$1$$anonfun$apply$10.apply(PairSCollectionFunctions.scala:276)
	at com.spotify.scio.values.PairSCollectionFunctions$$anonfun$hashLeftJoin$1$$anonfun$apply$10.apply(PairSCollectionFunctions.scala:274)
	at com.spotify.scio.util.FunctionsWithSideInput$$anon$2.processElement(FunctionsWithSideInput.scala:51)
```

nevillelyh:
@daikeshi ^
Should use `asSingletonSideInput(defaultValue: T)` and supply an empty map as default.

daikeshi:
Yep, I'm fixing it now

PR #953, opened 2017-11-30
Fix empty RHS for hashLeftJoin

daikeshi:
Fix https://github.com/spotify/scio/issues/952


daikeshi:
@nevillelyh it should be good now!

PR #955, opened 2017-11-30
support custom Kryo options via PipelineOptions, fix #896

### LINKED ISSUE #896: Expose Kryo#setRegistrationRequired

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -208,13 +211,24 @@ private[scio] class KryoAtomicCoder[T] extends AtomicCoder[T] {
...
+      val o = options.as(classOf[ScioOptions])
+      new KryoAtomicCoder[T](KryoOptions(o.getKryoRegistrationRequired))
+    }
+  }
+
+  def apply[T]: Coder[T] = new KryoAtomicCoder[T](KryoOptions())
 
 }
+
+case class KryoOptions(registrationRequired: Boolean = false)
```
kanterov: `private[scio]`? Isn't supposed to be binary compatible

kanterov:
I think it's great that we can configure the behavior of default coders! We can scale this approach to other parameters, like buffer size, because it seems defaults are not always optimal.

I have two suggestions:

1. Remove `KryoAtomicCoder.apply()`, because it's error-prone to create `KryoAtomicCoder` that wouldn't account `PipelineOptions`. It is used in `ScioUtil.getScalaCoder`. We can remove `ScioUtil.getScalaCoder` and refactor usages to use `CoderRegistry` and pass `PipelineOptions`. Otherwise, it looks like there are places where we are going to default to empty `KryoOptions`, looks like it happens in `saveAsObjectFile`.

2. Remove special handling for `null` because it's misleading, values are assumed to be non-nullable, and it looks like we can implement this. In tests, where needed, we can create `new KryoAtomicCoder(KryoOptions())` instead of `KryoAtomicCoder(null)`.

What do you think?

PR #956, opened 2017-12-01
Fix ClosureCleaner so it doesn't fail with ClassNotFound.

andrewsmartin:
On Scala 2.12, `ClosureCleaner` fails with `ClassNotFound` exceptions when given unserializable lambdas. Because of the differences between java7- and java8- compiled lambdas, the Closure Cleaner is not able to do anything in scala 2.12, but in this case it should not fail, and instead let the `NotSerializableException` surface to be actually helpful to users.


andrewsmartin:
@nevillelyh can you PTAL?

scio-core/src/main/scala/com/spotify/scio/util/ClosureCleaner.scala
```diff
@@ -247,3 +245,10 @@ class InnerClosureFinder(output: MSet[Class[_]]) extends ClassVisitor(ASM5) {
       }
     }
 }
+
+object AsmUtil {
```
nevillelyh: Make this private?

PR #963, opened 2017-12-03
Add dataflow runner to scio-examples

regadas:
I noticed that almost all examples reference `--runner=DataflowRunner`. Not sure if we just add the dep or update them to use `DirectRunner`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,10 +15,9 @@
  * under the License.
  */
 
-import sbt._
+import sbt.{Def, _}
 import Keys._
 import sbtassembly.AssemblyPlugin.autoImport._
-import com.typesafe.sbt.SbtSite.SiteKeys._
 import com.typesafe.sbt.SbtGit.GitKeys.gitRemoteRepo
 
 val beamVersion = "2.2.0"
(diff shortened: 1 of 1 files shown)


nevillelyh:
I'm a little conflicted. Kinda want to remove runner deps for the sake of cleanness, but also find myself adding it back often to test a job. I'd prefer if there's a way to add runner deps via some configuration flag, like a maven profile. Any idea how to do so in sbt?

regadas:
@nevillelyh something like this?

```
sbt scio-examples/dataflow:runMain ...

or
set scio-examples/direct:runMain ...
```

nevillelyh:
That'll require custom sbt tasks though? I'm thinking more like `-Drunner=...` to modify dependencies, but that's not perfect either. Open to other ideas :)

regadas:
@nevillelyh as is no need for custom tasks by prefixing the default tasks you are only adding the respective runner dep to the classpath.

nevillelyh:
Cool that's close to what I was thinking.

nevillelyh:
LGTM, how about making direct the default though, so that `sbt runMain WordCount ...` still works out of the box? We should update examples & wiki also once this is in.

regadas:
@nevillelyh sounds good!

build.sbt
```diff
@@ -506,13 +514,15 @@ lazy val scioSchemas: Project = Project(
   compileOrder := CompileOrder.JavaThenScala
 )
 
+
```
nevillelyh: Nitpicking: extra blank line

regadas:
@nevillelyh decided to fallback to the idea of having `-DbeamRunner=` because:
1) using configs and defaults was starting to get complicated.
2) running job requires the `--runner` to be set so it would become weird to have `dataflow:runMain ... --runner=DataflowRunner`

with this new changes you can 

```
// if -DbeamRunners not set default is DirectRunner
sbt -DbeamRunners=DirectRunner,DataflowRunner 
```

Also, there a couple more sub projects that use it. `scioTest` `scioRepl` should we use the same approach?

regadas:
@nevillelyh let me know what you think about this 😄

build.sbt
```diff
@@ -510,9 +532,8 @@ lazy val scioExamples: Project = Project(
   "scio-examples",
   file("scio-examples")
 ).settings(
-  commonSettings ++ noPublishSettings ++ soccoSettings,
+  commonSettings ++ noPublishSettings ++ soccoSettings ++ beamRunnerSettings,
   libraryDependencies ++= Seq(
-    "org.apache.beam" % "beam-runners-direct-java" % beamVersion,
```
nevillelyh: We should either keep this or change `def beamRunnerSettings` to include `DirectRunner` by default?

regadas: @nevillelyh `beamRunnerSettings` includes `DirectRunner` if no `-DbeamRunners` or `BEAM_RUNNERS=` is found. 

https://github.com/spotify/scio/pull/963/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR207

https://github.com/spotify/scio/pull/963/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR209

regadas: @nevillelyh the idea is that ... if we need a runner in some subproject we just include beamRunnerSettings

nevillelyh: LOL missed `.getOrElse(Seq(directRunnerDependency))`. LGTM.

regadas:
@nevillelyh now you can do 
```
sbt:scio-examples> set beamRunners := "DirectRunner,DataflowRunner"
```

PR #965, opened 2017-12-05
support dynamic file destinations, fix #919

### LINKED ISSUE #919: Dynamic file destinations support
Hello i see you guys added support for dynamic destinations for big query. In this commit in beam there is a similar thing for text io 

https://github.com/apache/beam/commit/77ba7a35cdae0b036791cce0682beefeb3fd809b#diff-fbb44cd19552f243afca93d8155c35af

Any plans to support it in an up coming release?

nevillelyh:


nevillelyh:
@LukeDefeo @regadas PTAL

I only implemented text and Avro IO. Still missing protobuf, `TableRow`, object file, etc. Probably not as critical as these 2?

scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
```diff
@@ -0,0 +1,151 @@
...
+      }
+
+      Future.failed(
+        new NotImplementedError("Avro file future with dynamic destinations not implemented"))
+    }
+
+    /**
+     * Save this SCollection as Avro files specified by the destination function.
+     */
+    def saveAsTextFile(path: String, schema: Schema = null,
```
regadas: @nevillelyh should this be `saveAsAvroFile`?

nevillelyh: Duh! 🤯

PR #972, opened 2017-12-12
Support ByteChannelFactory in TextIO

ravwojdyla:
This apart from other stuff enables compression on output of text io.


ravwojdyla:
If good with you, will add test.

nevillelyh:
Not sure if we should expose WBC in API though. Maybe not so straightforward to use for those who don't know the underlying IO module?

nevillelyh:
Either wrap the impls in an enum/case object, or alias the Java type?

ravwojdyla:
@nevillelyh what do you think about this?

ravwojdyla:
Example of usage would be:

```
import com.spotify.scio.io.compression._
sc.parallelize(1 to 100).saveAsTextFile("/tmp/t5", numShards=1, compression=Compression.GZIP)
```

to save a single `GZIP`ed file.

nevillelyh:
LGTM. What do you think about other non user facing classes in io package? Remove/hide to make it easier to import io._?

ravwojdyla:
@nevillelyh i guess out of the scope of this PR, but +1 to hiding everything that users doesn't need to see.

ravwojdyla:
@nevillelyh added tests, also updated to use `Compression` to align with `textFile` method.

PR #977, opened 2017-12-18
Use Compression in scio-tensorflow

ravwojdyla:


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordCodec.scala
```diff
@@ -84,9 +85,8 @@ private object TFRecordCodec {
         } else {
           pushback
         }
-      case CompressionType.NONE => stream
-      case CompressionType.ZLIB => new DeflateCompressorInputStream(stream, deflateParam)
-      case CompressionType.GZIP => new GzipCompressorInputStream(stream)
+      case Compression.UNCOMPRESSED => stream
+      case _ => Channels.newInputStream(compression.readDecompressed(Channels.newChannel(stream)))
```
ravwojdyla: This is the only interesting change, all the rest is just changing types.

PR #979, opened 2017-12-19
Optional arguments validation

stesua:
I noticed that Apache Beam arguments validation was missing in Scio so this is my proposal to improve it.


nevillelyh:
Thanks!

PR #982, opened 2017-12-19
Update new bigquery datetime and timestamp specification and remove "Fixme" 

syodage:
Big query standard-sql and legacy-sql support both datetime and timestamp formats with 'T' and ' ' (space) as time separator. 

Standard-sql: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime-type

Legacy-sql: https://cloud.google.com/bigquery/data-types

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
@@ -176,10 +176,9 @@ package object bigquery {
 
   /** Utility for BigQuery `TIMESTAMP` type. */
   object Timestamp {
-
-    // FIXME: verify that these match BigQuery specification
-    // YYYY-[M]M-[D]D[ [H]H:[M]M:[S]S[.DDDDDD]][time zone]
+    // YYYY-[M]M-[D]D[( |T)[H]H:[M]M:[S]S[.DDDDDD]][time zone]
     private val formatter = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSSSSS ZZZ")
+
     private val parser = new DateTimeFormatterBuilder()
       .append(DateTimeFormat.forPattern("yyyy-MM-dd"))
       .appendOptional(new DateTimeFormatterBuilder()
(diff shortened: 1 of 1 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
```diff
@@ -176,10 +176,9 @@ package object bigquery {
 
   /** Utility for BigQuery `TIMESTAMP` type. */
   object Timestamp {
-
-    // FIXME: verify that these match BigQuery specification
-    // YYYY-[M]M-[D]D[ [H]H:[M]M:[S]S[.DDDDDD]][time zone]
+    // YYYY-[M]M-[D]D[( |T)[H]H:[M]M:[S]S[.DDDDDD]][time zone]
```
nevillelyh: So there could be either a space or `T` between date and hour? It's not captured in the pattern though?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
```diff
@@ -238,8 +237,9 @@ package object bigquery {
 
   /** Utility for BigQuery `DATETIME` type. */
   object DateTime {
-    // YYYY-[M]M-[D]D[ [H]H:[M]M:[S]S[.DDDDDD]]
+    // YYYY-[M]M-[D]D[( |T)[H]H:[M]M:[S]S[.DDDDDD]]
```
nevillelyh: Same Q w.r.t. space or `T`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala
syodage: Yeah, either one is valid
With existing code, when we save to BQ tables it always saves with `T` but when reads it will work with either one.

nevillelyh: Ah gotcha. This is the formatter so we only handle one case. The parser below handles both with 2 `appendOptional`.

nevillelyh:
Gracias!

PR #984, opened 2017-12-22
Add jvmopts

ravwojdyla:


ravwojdyla:
This should help if someone does not have fine tuned sbt opts, and tries to run tests or anything heavier on this codebase.

nevillelyh:
CircleCI is not happy with the 6G Xmx :(

> Your build has exceeded the memory limit of 4G on 2 containers. The results of this build are likely invalid. We have taken a snapshot of the memory usage at the time, which you can find in a build artifact named `memory-usage.txt`. The RSS column in this file shows the amount of memory used by each process, measured in kilobytes.

PR #985, opened 2017-12-25
add TFRecordSpec

yonromai:
Dumps more metadata (along with the TFRecords) to simplify the reading API.
In addition to the feature names, we now dump:
* Each feature's type
* the compression type used
* this data is now stored in a versioned json file (instead of a flat file)

### DIFF
--- a/scio-examples/src/test/scala/com/spotify/scio/examples/extra/TFExampleExampleTest.scala
+++ b/scio-examples/src/test/scala/com/spotify/scio/examples/extra/TFExampleExampleTest.scala
@@ -27,23 +27,25 @@ class TFExampleExampleTest extends PipelineSpec {
   val input = Seq("foo", "bar", "foo")
   val output = Seq(WordCountFeatures(3.0f, 2.0f), WordCountFeatures(3.0f, 1.0f))
     .map(featuresType.toExample(_))
-  val featureNameDesc = Seq("wordLength", "count")
+  val featureNameDesc = Seq("{\"version\":1," +
+    "\"features\":[[\"FloatList\",\"count\"],[\"FloatList\",\"wordLength\"]]," +
+    "\"compression\":\"DEFLATE\"}")
 
   "TFExampleExample" should "work" in {
     JobTest[com.spotify.scio.examples.extra.TFExampleExample.type]
       .args("--input=in", "--output=out")
       .input(TextIO("in"), input)
       .output(TFExampleIO("out"))(_ should containInAnyOrder (output))
-      .output(TextIO("out/_feature_desc"))(_ should containInAnyOrder (featureNameDesc))
+      .output(TextIO("out/_tf_record_spec.json"))(_ should containInAnyOrder (featureNameDesc))
       .run()
   }
 
   it should "work with custom feature desc path" in {
     JobTest[com.spotify.scio.examples.extra.TFExampleExample.type]
-      .args("--input=in", "--output=out", "--feature-desc-path=out/custom_path_features")
+      .args("--input=in", "--output=out", "--feature-desc-path=out/my_tf_record_spec.json")
       .input(TextIO("in"), input)
       .output(TFExampleIO("out"))(_ should containInAnyOrder (output))
-      .output(TextIO("out/custom_path_features"))(_ should containInAnyOrder (featureNameDesc))
+      .output(TextIO("out/my_tf_record_spec.json"))(_ should containInAnyOrder (featureNameDesc))
       .run()
   }
 
--- a/scio-examples/src/test/scala/com/spotify/scio/examples/extra/TFExampleExampleTest.scala
+++ b/scio-examples/src/test/scala/com/spotify/scio/examples/extra/TFExampleExampleTest.scala
@@ -27,9 +27,9 @@ class TFExampleExampleTest extends PipelineSpec {
   val input = Seq("foo", "bar", "foo")
   val output = Seq(WordCountFeatures(3.0f, 2.0f), WordCountFeatures(3.0f, 1.0f))
     .map(featuresType.toExample(_))
-  val featureNameDesc = Seq("{\"version\":1," +
-    "\"features\":[[\"FloatList\",\"count\"],[\"FloatList\",\"wordLength\"]]," +
-    "\"compression\":\"DEFLATE\"}")
+  val featureNameDesc = Seq("""{"version":1,""" +
+    """"features":[["FloatList","count"],["FloatList","wordLength"]],""" +
+    """"compression":"DEFLATE"}""")
 
   "TFExampleExample" should "work" in {
     JobTest[com.spotify.scio.examples.extra.TFExampleExample.type]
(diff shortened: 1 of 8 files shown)


richwhitjr:
Could we dump the length of a list also so we can read it using FixedLength instead of VarLength in TF?  Easier to get back a dense tensor instead of a spare tensor when reading in the data.  No need to reshape or cast.

yonromai:
@richwhitjr thanks for your comment!

@ravwojdyla and I discussed this point quite a bit and couldn't find a 100% satisfactory solution:
It's impossible to know, a priori, if a list feature is Var or Fixed (and what's its size). This leaves us with 2 solutions: 
(1) we default to var length and ask the user to cast (in python) its variable if it's actually a fixed length or (2) we ask the user more explicit info about the features (for example via case class annotations) and we dump this data inside the TFRecordSpec so that reading data in python is seamless.

The main problem with (2) is that we would have to implement an annotation that mirrors the `tf.FixedLenFeature` API, including the shape and the default value. Do you think that's still a better user experience than casting in python?

scio-examples/src/test/scala/com/spotify/scio/examples/extra/TFExampleExampleTest.scala
```diff
@@ -27,23 +27,25 @@ class TFExampleExampleTest extends PipelineSpec {
   val input = Seq("foo", "bar", "foo")
   val output = Seq(WordCountFeatures(3.0f, 2.0f), WordCountFeatures(3.0f, 1.0f))
     .map(featuresType.toExample(_))
-  val featureNameDesc = Seq("wordLength", "count")
+  val featureNameDesc = Seq("{\"version\":1," +
```
ravwojdyla: Use `"""` for readability pls?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/FeatureKind.scala
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.spotify.scio.tensorflow
+
+import com.google.protobuf.ByteString
+import com.spotify.scio.tensorflow.FeatureKind.FeatureKind
+
+import scala.reflect.runtime.universe._
+
+object FeatureKind extends Enumeration {
```
ravwojdyla: must this be `public`?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+
+package com.spotify.scio.tensorflow
+
+import com.spotify.scio.tensorflow.FeatureKind._
+import com.spotify.scio.values.SCollection
+import io.circe.generic.auto._
+import io.circe.syntax._
+import org.apache.beam.sdk.io.Compression
+
+object TFRecordSpec {
```
ravwojdyla: Add docs.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+                                       compression: Compression) extends TFRecordSpec
+
+final case class FeatranTFRecordSpec(x: SCollection[Seq[(FeatureKind, String)]],
+                                     compression: Compression) extends TFRecordSpec
+
+private final case class TFRecordSpecConfig(version: Int,
+                                            features: Seq[(FeatureKind, String)],
+                                            compression: Compression) {
+
+  // Circe struggles with Compression (Java Enum)
```
ravwojdyla: add why this case class helps pls.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
```diff
@@ -75,7 +75,7 @@ private class PredictDoFn[T, V](graphBytes: DistCache[Array[Byte]],
       val outTensors = runner.run()
       try {
         import scala.collection.breakOut
-        c.output(outFn(input, (fetchOp zip outTensors.asScala)(breakOut)))
+        c.output(outFn(input, (fetchOp zip outTensors.asScala) (breakOut)))
```
ravwojdyla: ?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
```diff
@@ -106,25 +106,25 @@ private class PredictDoFn[T, V](graphBytes: DistCache[Array[Byte]],
 }
 
 /**
- * Enhanced version of [[com.spotify.scio.values.SCollection SCollection]] with TensorFlow methods.
- */
+  * Enhanced version of [[com.spotify.scio.values.SCollection SCollection]] with TensorFlow methods.
```
ravwojdyla: Fix indention.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
```diff
@@ -106,25 +106,25 @@ private class PredictDoFn[T, V](graphBytes: DistCache[Array[Byte]],
...
-   * @param config configuration parameters for the session specified as a serialized
-   *               `org.tensorflow.framework.ConfigProto` protocol buffer.
-   * @param inFn translates input elements of T to map of input-operation ->
-   *             [[org.tensorflow.Tensor Tensor]]. This method takes ownership of the
-   *             [[org.tensorflow.Tensor Tensor]]s.
-   * @param outFn translates output of prediction from map of output-operation ->
-   *              [[org.tensorflow.Tensor Tensor]], to elements of V. This method takes ownership
-   *              of the [[org.tensorflow.Tensor Tensor]]s.
-   */
+    * Predict/infer/forward-pass on pre-trained GraphDef.
```
ravwojdyla: Fix indention.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
```diff
@@ -243,11 +192,12 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
 class TFRecordSCollectionFunctions[T <: Array[Byte]](val self: SCollection[T]) {
 
   /**
-   * Save this SCollection as a TensorFlow TFRecord file. Note that elements must be of type
-   * `Array[Byte]`. The recommended record encoding is `org.tensorflow.example.Example` protocol
-   * buffers (which contain `org.tensorflow.example.Features` as a field) serialized as bytes.
-   * @group output
-   */
+    * Save this SCollection as a TensorFlow TFRecord file. Note that elements must be of type
```
ravwojdyla: Fix indentation.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
yonromai: @ravwojdyla do you have a scalastyle scheme for intellij? This formatting (and the docs style below) is from intellij reformat feature, using [Spotify checkstyle config](https://github.com/spotify/spotify-checkstyle-config)

ravwojdyla: @yonromai https://github.com/spotify/scio/wiki/Style-Guide

yonromai:
@ravwojdyla please see updates

yonromai:
Made quite a few changes since this last commit - closing this PR; will post a fresh one soon with squashed commits.

PR #990, opened 2017-12-30
Add support for TFRecordSpec

yonromai:
Dumps more metadata (along with the TFRecords files) to simplify the reading API.
In addition to the feature names, we now dump:

* Each feature's type
* The compression type used
* Tags which provide flexibility for advanced use cases (like multi-spec extraction)
* This data is now stored in a versioned Json file (instead of a flat file)

**Sibling PR: https://github.com/spotify/featran/pull/75**

### DIFF
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/FeatureInfo.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/FeatureInfo.scala
@@ -18,8 +18,14 @@
 package com.spotify.scio.tensorflow
 
 import com.google.protobuf.ByteString
+import com.spotify.scio.tensorflow.FeatureKind.FeatureKind
+import io.circe.{Encoder, Json}
+import org.apache.beam.sdk.io.Compression
 
 import scala.reflect.runtime.universe._
+
+/** Information necessary to extract a given feature in TF. */
+final case class FeatureInfo(name: String, kind: FeatureKind, tags: Map[String, String])
 
 /** Mapping between Scala types and TF types */
 object FeatureKind extends Enumeration {
@@ -42,4 +48,14 @@ object FeatureKind extends Enumeration {
   }
 
   // scalastyle:on cyclomatic.complexity
+}
+
+private object CustomCirceEncoders {
+  implicit val compressionEncoder: Encoder[Compression] = new Encoder[Compression] {
+    override def apply(a: Compression): Json = Json.fromString(a.toString)
+  }
+
+  implicit val featureKindEncoder: Encoder[FeatureKind] = new Encoder[FeatureKind] {
+    override def apply(a: FeatureKind): Json = Json.fromString(a.toString)
+  }
 }
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.tensorflow
+
+import com.spotify.scio.ScioContext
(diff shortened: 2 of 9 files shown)


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+import com.spotify.scio.tensorflow.FeatureKind._
+import com.spotify.scio.values.SCollection
+import io.circe.generic.auto._
+import io.circe.syntax._
+import org.apache.beam.sdk.io.Compression
+
+/** Contains metadata about TFRecords. Useful to read the records later on. */
+trait TFRecordSpec {
+  val LATEST_VERSION: Int = 1
+  def toJson(sc: ScioContext, compression: Compression): SCollection[String]
```
ravwojdyla: Don't seen the reason for `sc: ScioContext, compression: Compression` to be part of `toJson`? Shouldn't both `sc` and `compression` be params for specific impl of `TFRecordSpec`. In case of `compression` if it's common it could be a declared in the trait.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
```diff
@@ -116,14 +116,14 @@ class TensorFlowSCollectionFunctions[T: ClassTag](@transient val self: SCollecti
...
    * @param fetchOps names of [[org.tensorflow.Operation]]s to fetch the results from
-   * @param config configuration parameters for the session specified as a serialized
-   *               `org.tensorflow.framework.ConfigProto` protocol buffer.
-   * @param inFn translates input elements of T to map of input-operation ->
-   *             [[org.tensorflow.Tensor Tensor]]. This method takes ownership of the
-   *             [[org.tensorflow.Tensor Tensor]]s.
-   * @param outFn translates output of prediction from map of output-operation ->
-   *              [[org.tensorflow.Tensor Tensor]], to elements of V. This method takes ownership
-   *              of the [[org.tensorflow.Tensor Tensor]]s.
+   * @param config   configuration parameters for the session specified as a serialized
```
nevillelyh: We don't normally align doc. It doesn't affect generated html and just makes edits harder.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+/** Information necessary to extract a given feature in TF. */
+final case class FeatureInfo(name: String, kind: FeatureKind, tags: Map[String, String])
+
+/** TFRecord metadata that will end up serialized. */
+final case class TFRecordSpecConfig(version: Int,
+                                    features: Seq[FeatureInfo],
+                                    compression: Compression) {
+
+  // Circe struggles with Java Enum, the easiest workaround is to turn all Enums into Strings.
+  case class Str(version: Int,
```
nevillelyh: Probably better with a custom codec and remove this intermediate type? https://circe.github.io/circe/codec.html

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
yonromai: 👍 fixed my intellij for next time

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordSpec.scala
yonromai: That's how I started but turned out that using this case class was less code & more simple. I'll give the custom encoder another try and see how it looks.

yonromai:
@nevillelyh PTAL

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowSCollectionFunctions.scala
nevillelyh: Undo the indent changes here?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/FeatureInfo.scala
```diff
@@ -0,0 +1,61 @@
...
+    case t if t =:= typeOf[Double] => FeatureKind.FloatList
+    case t if t =:= typeOf[String] => FeatureKind.BytesList
+    case t if t =:= typeOf[ByteString] => FeatureKind.BytesList
+    case t if t =:= typeOf[Byte] => FeatureKind.BytesList
+  }
+
+  // scalastyle:on cyclomatic.complexity
+}
+
+object CustomCirceEncoders {
```
nevillelyh: This can be (package) `private`

yonromai: nice catch, done

yonromai:
@nevillelyh fixed

PR #992, opened 2018-01-02
use Avro inputs when reading from BigQueryIO, fix #964

### LINKED ISSUE #964: Use Avro inputs from BigQueryIO
Starting 2.2.0, Beam BigQueryIO now supports reading rows as Avro `GenericRecord` directly. This should be faster and saves us the conversion to and from intermediate `TableRow` format.
https://github.com/apache/beam/blob/v2.2.0-RC4/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L364

nevillelyh:
This is a breaking change and should probably go into 0.5.x

- `BigQueryIO` now requires a type param in `JobTest`. On the plus side, tests with `TableRow` vs macro are cleaner, no more `.map(MyType.toTableRow)` in test fixtures.
- Not sure about the `useAvro: Boolean = false` flag since it's a breaking change anyway. Should be on by default if we test thoroughly and maybe make a few pre-releases?

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -31,7 +31,7 @@ import com.spotify.scio.avro.types.AvroType
 import com.spotify.scio.avro.types.AvroType.HasAvroAnnotation
 import com.spotify.scio.bigquery._
 import com.spotify.scio.bigquery.types.BigQueryType.HasAnnotation
-import com.spotify.scio.coders.AvroBytesUtil
+import com.spotify.scio.coders.{AvroBytesUtil, KryoAtomicCoder, KryoOptions}
 import com.spotify.scio.io.Tap
 import com.spotify.scio.metrics.Metrics
 import com.spotify.scio.options.ScioOptions
@@ -43,10 +43,11 @@ import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecordBase
 import org.apache.beam.sdk.PipelineResult.State
 import org.apache.beam.sdk.extensions.gcp.options.{GcpOptions, GcsOptions}
+import org.apache.beam.sdk.io.gcp.bigquery.SchemaAndRecord
 import org.apache.beam.sdk.io.gcp.{bigquery => bqio, datastore => dsio, pubsub => psio}
 import org.apache.beam.sdk.options._
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
-import org.apache.beam.sdk.transforms.{Create, DoFn, PTransform}
+import org.apache.beam.sdk.transforms.{Create, DoFn, PTransform, SerializableFunction}
 import org.apache.beam.sdk.util.CoderUtils
 import org.apache.beam.sdk.values._
 import org.apache.beam.sdk.{Pipeline, PipelineResult, io => gio}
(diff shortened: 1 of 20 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -542,19 +546,36 @@ class ScioContext private[scio] (val options: PipelineOptions,
     }
   }
 
-  /**
-   * Get an SCollection for a BigQuery table.
-   * @group input
-   */
-  def bigQueryTable(table: TableReference): SCollection[TableRow] = requireNotClosed {
+  private def bqReadTable[T: ClassTag](typedRead: bqio.BigQueryIO.TypedRead[T],
+                                         table: TableReference)
```
andrewsmartin: Indentation

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -542,19 +546,36 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+  def bigQuerySelect(sqlQuery: String,
+                     flattenResults: Boolean = false): SCollection[TableRow] =
+    bqReadQuery(bqio.BigQueryIO.readTableRows(), sqlQuery, flattenResults)
+
+  /**
+   * Get an SCollection for a BigQuery table.
+   * @group input
+   */
+  def bigQueryTable(table: TableReference): SCollection[TableRow] =
+    bqReadTable(bqio.BigQueryIO.readTableRows(), table)
```
andrewsmartin: So the untyped API is still reading as `TableRow`, is this preferred over using Avro here as well? More convenient but less performant?

scio-test/src/it/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  private val options = PipelineOptionsFactory
+    .fromArgs(
+      "--project=data-integration-test",
+      "--tempLocation=gs://data-integration-test-us/temp")
+    .create()
+
+  "TypedBigQuery" should "work with Avro input" in {
+    val sc = ScioContext(options)
+    sc.typedBigQuery[Record](table, useAvro = true) should containInAnyOrder (records)
```
andrewsmartin: This doesn't match the API in `ScioContext` (`typedBigQuery` there has no `useAvro` param), unless I'm missing something?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: For backwards compatibility mostly. Generic Avro API is just as bad as JSON even if it might be slightly faster.

scio-test/src/it/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
nevillelyh: This needs fixing. I originally had an `useAvro` flat but it makes no sense now that we decided to make a breaking change and not do `TableRow => T` read at all.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
andrewsmartin: 👍

PR #1010, opened 2018-01-13
Update JobTest to support new io implementations.

syodage:
- [x] Update `JobTest` and `TestDataManager` classes to support nio implemenations. 
- [x] Update `ScioContext` and `SCollection` to invoke nio test inputs/outputs 
- [x] Add a new word count example to use new `TextIO`
- [x] Write a `JobTest` to test new word count example.


syodage:
@ravwojdyla it seems, `nio` branch is not setup with codacy could you setup `nio` branch too?

PR #1012, opened 2018-01-16
Create AsyncLookupDoFn

rculbertson:
This PR creates a generalized version of BigtableDoFn. Users can use this new
class to do lookups using any type of client, and get the caching and threading
logic for free. BigtableDoFn now inherites from this class.


nevillelyh:
Great idea. Not sure about having it in `scio-extra` (since it has a bunch of other crap) and `scio-bigtable` depending on it though. Maybe time to split up a new sub-module? Suggestion for a name?

DanielDeychakiwsky:
you guys rock!

PR #1015, opened 2018-01-17
Fix typo on cannot find Beam version

labianchin:
Also, can you clarify what is the expected behavior when `val beamVersion = ` is not present? Would all this exceptions turn into warnings?

PTAL @nevillelyh

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
@@ -46,7 +46,7 @@ private[scio] object VersionUtil {
       .filter(_.startsWith("val beamVersion = ")).next()
     """val beamVersion = "([^"]+)"""".r.findFirstMatchIn(line) match {
       case Some(m) => m.group(1)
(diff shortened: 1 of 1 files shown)


nevillelyh:
Good catch! Under what's condition was it not present?

labianchin:
> Under what's condition was it not present?

I mean, nothing forces users to declare `val beamVersion =`.

A couple of scenarios to think about:

- They don't use sbt/build.sbt
- They don't use this identifier, e.g. use `val beamV =` instead.
- They don't have a `val`, i.e.: `libraryDependencies ++= Seq("com.spotify" %% "scio-core" % scioVersion, "org.apache.beam" % "beam-runners-google-cloud-dataflow-java" % "2.2.0")`.


`_.startsWith("val beamVersion = ")` feels a smart way to detect the current `beamVersion`...

nevillelyh:
My commit also includes build.sbt in /src/main/resources so it should always be present?

labianchin:
Ahh ok. Now I got it, it is the `beamVersion` from `scio` `build.sbt`.

PR #1017, opened 2018-01-17
Add ByteStringSerializer

andrewsmartin:
This one is for the failing `BigTableExampleTest` in beam-master.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/ByteStringSerializer.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/ByteStringSerializer.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/ByteStringSerializer.scala
```diff
@@ -0,0 +1,31 @@
...
+package com.spotify.scio.coders
+
+import com.esotericsoftware.kryo.Kryo
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.google.protobuf.ByteString
+import com.twitter.chill.KSerializer
+
+class ByteStringSerializer extends KSerializer[ByteString] {
+  override def read(kryo: Kryo, input: Input, tpe: Class[ByteString]): ByteString =
+    ByteString.copyFromUtf8(input.readString)
```
nevillelyh: You shouldn't use string here. `ByteString` can contain `Array[Byte]`s that are not legit strings, e.g. weird encoding, `\0` in the middle.

nevillelyh: Also what makes `ByteString` not deterministic out of the box? It has fields other than the `Array[Byte]`?

nevillelyh: Hm... so `ByteString` is an abstract class and some impls like `BoundedByteString` has extra stateful fields like offset and length.

andrewsmartin: That makes sense, I'll change this to read/write `Array[Byte]` instead.

nevillelyh:
Cherry picked into the branch.

PR #1018, opened 2018-01-18
Add SparseVectorSerializer.

andrewsmartin:
This is still kind of WIP, just want to get some eyes on this. I think I will add some tests for this one since it's not as trivial.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -434,7 +434,7 @@ lazy val scioExtra: Project = Project(
   "scio-extra",
   file("scio-extra")
 ).settings(
-  commonSettings ++ itSettings,
+  commonSettings ++ itSettings ++ macroSettings,
   description := "Scio extra utilities",
   libraryDependencies ++= Seq(
     "com.spotify" % "annoy" % annoyVersion,
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/breeze/BreezeKryoRegistrar.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/breeze/BreezeKryoRegistrar.scala
@@ -24,7 +24,6 @@ import com.esotericsoftware.kryo.io.{Input, Output}
 import com.spotify.scio.coders.KryoRegistrar
 import com.twitter.chill._
 
-import scala.collection.mutable
 import scala.reflect.ClassTag
 
 @KryoRegistrar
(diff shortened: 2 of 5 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/breeze/BreezeKryoRegistrar.scala
```diff
@@ -0,0 +1,63 @@
...
+    k.register(classOf[SparseVector[_]], new SparseVectorSerializer[Any])
+  }
+}
+
+private class SparseVectorSerializer[T] extends KSerializer[SparseVector[T]] {
+  override def read(kryo: Kryo, input: Input, tpe: Class[SparseVector[T]]): SparseVector[T] = {
+    val zero = kryo.readClassAndObject(input).asInstanceOf[T]
+    val len = input.read()
+    val data = new mutable.ArrayBuffer[T](len)
+    val index = new mutable.ArrayBuffer[Int](len)
```
richwhitjr: Can this just be any Array instead since we know the length at creation time?  Could save some overhead since we don't have to worry about it growing.

andrewsmartin: True, will update to just use plain `Array`.

build.sbt
```diff
@@ -434,7 +434,7 @@ lazy val scioExtra: Project = Project(
   "scio-extra",
   file("scio-extra")
 ).settings(
-  commonSettings ++ itSettings,
+  commonSettings ++ itSettings ++ macroSettings,
```
nevillelyh: Why is this needed?

scio-extra/src/main/scala/com/spotify/scio/extra/breeze/breeze.scala
```diff
@@ -17,8 +17,8 @@
 
 package com.spotify.scio.extra
 
-import breeze.linalg.operators.OpAdd
-import breeze.linalg.support.CanCopy
+import _root_.breeze.linalg.operators.OpAdd
```
nevillelyh: Why do we need to prefix all these imports?

build.sbt
andrewsmartin: For the `@KryoRegistrar` macro.

scio-extra/src/main/scala/com/spotify/scio/extra/breeze/breeze.scala
andrewsmartin: Because I moved the existing Breeze stuff into its own `breeze` package, which collides with the root namespace.

andrewsmartin:
Gonna close this one as it doesn't handle specialized types properly. I'll re-open another one with some test cases when it's ready.

PR #1019, opened 2018-01-19
Add BigTable-specific SCollectionMatchers to scio-test

ClaireMcGinty:
Adds scalatest `Matchers` for BigTable abstractions like row keys, column families, set cell values, and mutation types. This trait extends `SCollectionMatchers`.

I also added a `containValue` function to `SCollectionMatchers` to check for presence of a single item within a collection without making assumptions about the rest of the collection. The rest of the `contains__` matchers required you to specify every item present in the collection which can be hard to read and write, or outside the scope of a specific test.

added unit tests for all modified code

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.testing
+
+import java.lang.{Iterable => JIterable}
+import java.util.{Map => JMap}
+
+import com.google.bigtable.v2.Mutation
+import com.google.bigtable.v2.Mutation.MutationCase
+import com.google.protobuf.ByteString
+import com.spotify.scio.values.SCollection
+import org.scalatest.matchers.{MatchResult, Matcher}
+
+import scala.reflect.ClassTag
+
+/**
+  * Trait with ScalaTest [[org.scalatest.matchers.Matcher Matcher]]s for
+  * [[com.spotify.scio.values.SCollection SCollection]]s specific to BigTable output.
+  */
+trait BigTableMatchers extends SCollectionMatchers {
+  type BTRow = (ByteString, Iterable[Mutation])
+  type BTCollection = SCollection[BTRow]
+
+  // Provide an implicit BT serializer for common cell value type String
+  implicit val stringBTSerializer: String => ByteString = ByteString.copyFromUtf8
+
+  /** Check that the BT collection contains only the given keys, in any order. */
+  def containRowKeys(expectedKeys: String*): Matcher[BTCollection] =
+    new Matcher[BTCollection] {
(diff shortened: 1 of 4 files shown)


scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import com.google.bigtable.v2.Mutation
+import com.google.bigtable.v2.Mutation.MutationCase
+import com.google.protobuf.ByteString
+import com.spotify.scio.values.SCollection
+import org.scalatest.matchers.{MatchResult, Matcher}
+
+import scala.reflect.ClassTag
+
+/**
```
nevillelyh: Minor: We use old javadoc format, i.e. no extra space before `*`.
https://github.com/spotify/scio/wiki/Style-Guide#documentation-style

scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
```diff
@@ -0,0 +1,126 @@
...
+  type BTRow = (ByteString, Iterable[Mutation])
+  type BTCollection = SCollection[BTRow]
+
+  // Provide an implicit BT serializer for common cell value type String
+  implicit val stringBTSerializer: String => ByteString = ByteString.copyFromUtf8
+
+  /** Check that the BT collection contains only the given keys, in any order. */
+  def containRowKeys(expectedKeys: String*): Matcher[BTCollection] =
+    new Matcher[BTCollection] {
+      override def apply(left: BigTableMatchers.this.BTCollection): MatchResult = {
```
nevillelyh: `BigTableMatchers.this.BTCollection` can just be `BTCollection`?

scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
```diff
@@ -0,0 +1,126 @@
...
+        containInAnyOrder(expectedKeys).apply(left.keys.map(_.toStringUtf8))
```
nevillelyh: Are `keys` always UTF-8 Strings though? One could use binary keys too right? Better just make `expectedKeys: ByteString`?

scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+
+  /** Check that the BT collection contains only the given column families, unique, in any order. */
+  def containColumnFamilies(expectedCFs: String*): Matcher[BTCollection] =
+    new Matcher[BTCollection] {
+      override def apply(left: BigTableMatchers.this.BTCollection): MatchResult = {
+        val foundCFs = left.flatMap {
```
nevillelyh: Minor: we usually `.flatMap { case ...\n` for readability.

scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
```diff
@@ -0,0 +1,126 @@
...
+    * @param ser Serializer to convert value type V to ByteString for BT format
+    * @tparam V Class of expected value
+    * @return Whether the collection contains this cell, with no assumptions made about the
+    *         contents of the collection.
+    */
+  def containSetCellValue[V: ClassTag](key: String, cf: String, cq: String, value: V)
+                                    (implicit ser: V => ByteString): Matcher[BTCollection] =
+    new Matcher[BTCollection] {
+      override def apply(left: BTCollection): MatchResult = {
+        val flattenedRows = left.flatMap {
```
nevillelyh: Minor: `.flatMap { case ...`
Also `rowValue { cell => ... }` is fewer `{}` and probably cleaner.

scio-test/src/main/scala/com/spotify/scio/testing/BigTableMatchers.scala
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2016 Spotify AB.
```
nevillelyh: Minor: 2018
You can use a `$today.year` variable in IntelliJ's Copyright template.

ClaireMcGinty:
@nevillelyh addressed all comments! keys are now ByteStrings and i updated all the style suggestions.

PR #1027, opened 2018-01-23
Fix #1026, add helpers in ScioContext to initialize counters.

### LINKED ISSUE #1026: Syntactic sugar to create and initialize counters
maybe `sc.initCounter(name: String, ...): Counter`
The context should memorize these counters and create one artificial parallelize + map step to initialize then.

andrewsmartin:


andrewsmartin:
FYI, as part of this I also changed the default namespace to use the package name at the callsite instead of just `ScioMetrics`. This will go into 0.5.0 so IMO the change is fine, but if anyone feels strongly about keeping it as `ScioMetrics` then I can change it back.

nevillelyh:
Can we:
- Keep the package name? Changing it adds more work for others and not much value.
- Support the other 2 metric types, e.g. distribution and gauge?
- Can we memorize these metrics (probably need 3 collections, 1 each type) and make 1 `parallelize+map` for each type instead of each individual metric? Otherwise the graph will be messy for jobs with lots of metrics.

andrewsmartin:
Sure, I can keep the default name the same.

The problem with distribution and gauge is that initialization doesn't really make sense, whereas for counters a starting value of zero is pretty much always sensible. An initialization for a distribution would mean adding an observation which could mess up things like `min` `max` `mean` etc. Maybe it's ok for `Gauge` because the value just get overwritten, but I held back because semantically it indicates some value was seen, which IMO can be confusing if it was only initialized but never used after. WDYT?

nevillelyh:
Hm... Right we can't handle distribution. Gauge takes `Long` so `NaN` is out of the question too. I guess just counter then.

Can you memorize them in `ScioContext` and use 1 parallelize + map only?

andrewsmartin:
Sure, will update shortly.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/MetricsExample.scala
```diff
@@ -22,6 +22,7 @@
 package com.spotify.scio.examples.extra
 
 import com.spotify.scio._
+import com.spotify.scio.util.CallSites
```
nevillelyh: Unused import? I can fix it's minor.

PR #1032, opened 2018-01-26
add applyKvTransform, fix #1020

### LINKED ISSUE #1020: Handle KvCoder in SCollection#applyTransform
Right know coder inference doesn't work if the transform being applied outputs a `KV<K, V>`. We should handle this case.

nevillelyh:


rculbertson:
👍

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -130,8 +130,23 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * [[SCollection]].
    */
   def applyTransform[U: ClassTag](transform: PTransform[_ >: PCollection[T], PCollection[U]])
```
nevillelyh: I wonder if we should make this a `TypeTag` and get `K` and `V` types directly. But then we might as well fix replace `ClassTag` with `TypeTag` everywhere?

nevillelyh:
We should fix ScioContext#customInput as well.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
nevillelyh: Gave it a try but am going to abandon this idea.
`applyTransform` is used in a lot of places and it'd require type signature change everywhere.

PR #1033, opened 2018-01-26
adding cached resolution to speed up build time

flylo:


nevillelyh:
Did some quick tests:

base line, clean 1st run
sbt -mem 8196 update  144.65s user 10.25s system 238% cpu 1:05.01 total

with cache, clean 1st run
sbt -mem 8196 update  159.38s user 10.89s system 237% cpu 1:11.59 total

with cache, 2nd run
sbt -mem 8196 update  171.24s user 11.35s system 253% cpu 1:11.90 total

The PR build after re-triggering with cache, is 16/11 min for scala 2.12/11, not a huge improvement.
I guess we already cache `~/.sbt`, `resolution-cache` and `target/streams` so this won't make a big difference. I also saw a lot of potential issues with this option. Prefer not to introduce it for now.

PR #1035, opened 2018-01-29
Fix #1034, don't use transform name for internal CoGroup.

### LINKED ISSUE #1034: ArtisanJoins and not have stable unique names
When using `PairSCollectionFuctions` `rightOuterJoin()`

The following error happens:

```
WARN org.apache.beam.sdk.Pipeline - The following transforms do not have stable unique names: rightOuterJoin@{Foo.scala:76}12, fooJoin2
```

Example code:

```scala
      self.context.parallelize(Seq(1,2))
        .keyBy[Int](_.toInt)
        .withName("fooJoin")
        .rightOuterJoin(self.context.parallelize(Seq(2,3)).keyBy[Int](_.toInt))
```

The problem seems to be around:

```
  def rightOuterJoin[W: ClassTag](that: SCollection[(K, W)])
  : SCollection[(K, (Option[V], W))] =
    ArtisanJoin.right(self.tfName, self, that)
```

where the same transform name is reused in another transform.

andrewsmartin:


scio-core/src/main/scala/com/spotify/scio/util/ArtisanJoin.scala
```diff
@@ -38,7 +38,7 @@ private[scio] object ArtisanJoin {
     val keyed = KeyedPCollectionTuple
       .of(tagA, a.toKV.internal)
       .and(tagB, b.toKV.internal)
-      .apply(name, CoGroupByKey.create())
+      .apply(s"CoGroupByKey@$name", CoGroupByKey.create())
```
regadas: 👍 this is a nice format!

PR #1038, opened 2018-01-30
added ser/der support for joda DateTime

LukeDefeo:
Added kyro serializer for joda Datetime class


nevillelyh:
What's the purpose for this?
We have serializers for `LocalTime` `LocalDate` `LocalDateTime` coz those are the ones BQ uses.
If it's for you pipeline code it should be done in a `KryoRegistrator`?

LukeDefeo:
I wanted to make a small contribution and assumed others would be using DateTime to represent dates, if its not a common use case then close it.

nevillelyh:
Cool, just make sure we didn't miss anything.

PR #1040, opened 2018-02-01
make password optional for jdbc connection

ajmarcus:
@syodage making password optional for JDBC connection

### DIFF
--- a/scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
+++ b/scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
@@ -23,6 +23,7 @@ import com.spotify.scio.Implicits._
 import com.spotify.scio.io.Tap
 import com.spotify.scio.testing.TestIO
 import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.io.jdbc.JdbcIO.DataSourceConfiguration
 import org.apache.beam.sdk.io.{jdbc => jio}
 
 import scala.concurrent.Future
@@ -42,12 +43,12 @@ package object jdbc {
    * Options for a JDBC connection.
    *
    * @param username database login username
-   * @param password database login password
+   * @param password database login password if exists
    * @param connectionUrl connection url, i.e "jdbc:mysql://[host]:[port]/db?"
    * @param driverClass subclass of [[java.sql.Driver]]
    */
   case class JdbcConnectionOptions(username: String,
-                                   password: String,
+                                   password: Option[String],
                                    connectionUrl: String,
                                    driverClass: Class[_ <: Driver])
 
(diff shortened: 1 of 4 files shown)


scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
```diff
@@ -84,14 +85,31 @@ package object jdbc {
...
     case JdbcReadOptions(connOpts, query, _, _) => jdbcIoId(connOpts, query)
     case JdbcWriteOptions(connOpts, statement, _) => jdbcIoId(connOpts, statement)
   }
 
+  private def getDataSourceConfig(connOpts: jdbc.JdbcConnectionOptions): DataSourceConfiguration = {
+    val dataSourceConfig = jio.JdbcIO.DataSourceConfiguration
+      .create(connOpts.driverClass.getCanonicalName, connOpts.connectionUrl)
+      .withUsername(connOpts.username)
+    if (connOpts.password.isDefined) {
+      dataSourceConfig.withPassword(connOpts.password.get)
```
ajmarcus: not sure if there is a better way to do this than calling `.get`

syodage: don't see any downside as`.get` is safeguard with `if` check

andrewsmartin: Maybe `connOpts.password.foreach(p => dataSourceConfig.withPassword(p))`

syodage:
Introduce many code style changes, make code style consistent with rest? 
Other than that, LGTM! 👍 
@nevillelyh @ravwojdyla have any comment?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
```diff
@@ -84,14 +85,31 @@ package object jdbc {
     def apply[T](jdbcIoOptions: JdbcIoOptions): JdbcIO[T] = JdbcIO[T](jdbcIoId(jdbcIoOptions))
   }
 
-  private def jdbcIoId(opts: JdbcConnectionOptions, query: String): String =
-    s"${opts.username}:${opts.password}@${opts.connectionUrl}:$query"
+  private def jdbcIoId(opts: JdbcConnectionOptions, query: String): String = {
+    val user = if (opts.password.isDefined) {
+      s"${opts.username}:${opts.password}"
```
andrewsmartin: Should this be `opts.password.get`? You could also do `val user = opts.password.fold(s"${opts.username}")(pass => s"${opts.username}:${opts.password}")` which is more idiomatic Scala.

ajmarcus: yes - i should also add a test for that.

ajmarcus:
tried to undo all the style changes

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
```diff
@@ -23,56 +23,57 @@ import com.spotify.scio.Implicits._
...
 import scala.reflect.ClassTag
 
 /**
- * Main package for JDBC APIs. Import all.
- *
- * {{{
- * import com.spotify.scio.jdbc._
- * }}}
- */
+  * Main package for JDBC APIs. Import all.
```
ravwojdyla: Fix comments indents.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
```diff
@@ -154,4 +170,4 @@ package object jdbc {
     }
   }
 
-}
+}
```
ravwojdyla: Add new line.

ajmarcus:
appreciate all the feedback. looking better?

PR #1042, opened 2018-02-05
Make logging less verbose for AsyncLookDoFn

rculbertson:
No longer log messages for startBundle, finishBundle and setup.


nevillelyh:
Late to the party, just wondering if it's better to use `DEBUG` level instead of removing this?
Similar PR in upstream https://github.com/apache/beam/pull/4557

ravwojdyla:
Was wondering similar - but do we actually need this kind of logs tho - just the info about `start` and `end` of a bundle?

rculbertson:
Would switching to debug level still incur the same costs on systems like Google Cloud Logging? Cost was mentioned as a concern in both tickets.

nevillelyh:
It shouldn't produce any log or incur cost if (I assume) default level is above `DEBUG`.

PR #1045, opened 2018-02-09
Fix #726 ProjectId nonEmpty string check in BigQueryClient

### LINKED ISSUE #726: Throw exception if the Project Id isn't valid passed to BigQueryClient.apply
I wasted some time today because I passed an empty string to the BigQueryClient.apply method. An exception was only thrown when I call BigQueryClient.query and the exception was a NullPointerException. It would be a better developer experience if BigQueryClient.apply threw an exception if the project id wasn't valid.

guidj:
To fail on client creation, rather than later, if the projectId is either null or an empty
string. Fixes #726

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
@@ -74,6 +74,9 @@ private[scio] trait QueryJob {
 class BigQueryClient private (private val projectId: String,
                               _credentials: Credentials = null) { self =>
 
+  require(projectId != null && projectId.nonEmpty, "Invalid projectId. " +
+    "It should be a non-empty string")
+
   def this(projectId: String, secretFile: File) =
     this(
       projectId,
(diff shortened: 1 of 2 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -74,6 +74,9 @@ private[scio] trait QueryJob {
 class BigQueryClient private (private val projectId: String,
                               _credentials: Credentials = null) { self =>
 
+  require(projectId != null && projectId.nonEmpty, "Invalid `projectId`. " +
```
ravwojdyla: Probably no need to \` the `projectId`.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/BigQueryClientTest.scala
```diff
@@ -0,0 +1,46 @@
...
+    assertThrows[IllegalArgumentException] {
+      BigQueryClient("")
+    }
+
+    assertThrows[IllegalArgumentException] {
+      BigQueryClient(null)
+    }
+  }
+
+  "BigQueryClient" should "work with non-empty ProjectId" in {
```
ravwojdyla: Use `it` - it's more idiomatic.

Issue #840, opened 2017-09-13
Add extract/load to BigQueryClient


ravwojdyla:
This could be useful as a pre/post step to a pipeline. Say I want to write data as avro/json to GCS with schema computed at runtime, but I also want my data in BQ: I could have a pipeline that writes avro/json to GCS, and when the pipeline is done just launch BQ load job(s) myself (knowing the URI, schema etc).

guidj:
How would you run tests with this?
Seems I need permissions to run the existing integration tests for scio-bigquery

guidj:
Looking at the code, it seems that some functionality, like `waitForJob` and `logJobStats` is tied to Query type jobs. I'm wondering about the best way to structure BigQueryClient to accomodate other job types.

Here's what I have in mind:

```
sealed trait BQJob {
   //common job fields, like JobReference/Id
}

trait/case class QueryJob extends BQJob
trait/case class LoadJob extends BQJob
trait/case class ExtractJob extends BQJob

//Then, for the common methods, like `waitForJob`

def waitForJobs(jobs: BQJob*) {
   //wait for any type of job
}

def getTable(...)

//where specific functionality is required, use pattern matching on different types
def jobStats(job: BQJob) {
  job match {
    case QueryJob:
      //print query, bytes processed
      //...
  }

  //The functionality specific to any type of job could be encapsulated in an inner object:
  object load {
       def csv(paramsForCsv...)
       def avro(paramsForAvro...)
   }

   object query {
     //....
   }
}

```

Encapsulating functionality would break APIs, so maybe that can be postponed or just done for the new job types

Thoughts?

nevillelyh:
It doesn't have to be integrated into the query job scaffolding right? If we need this for convenience only, it could just be a pair of blocking methods in `BigQueryClient`.

On the other hand, IIRC the BQ job logic is never exposed, so breaking change may be fine?

guidj:
True. I was thinking about `def waitForJobs(jobs: QueryJob*)`, as it's public. But `QueryJob` isn't, so you're right, it's not really exposed

The Job types would be to share methods like waitForJob or logStats, instead of having that implemented by each Job type because most of the logic would be the same (ping the API for job status, sleep). 

I updated the previous comment, the logic in `object query` to clarify what I meant about breaking changes. Instead of calling `runQuery`, `loadCsv` you call `query.run (sync)`, `query.submit (async)`. But as we don't currently expose that logic, maybe best to stick `loadCsv`

[...]

PR #1049, opened 2018-02-15
BigQuery load and extract

guidj:
To allow data to be imported from GCS (csv, json, avro) into BigQuery tables synchronous way, and export table to GCS

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
@@ -17,62 +17,36 @@
 
 package com.spotify.scio.bigquery
 
-import java.io.{File, FileInputStream, IOException, StringReader}
-import java.util.UUID
+import java.io.{File, FileInputStream}
 
-import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.http.javanet.NetHttpTransport
 import com.google.api.client.http.{HttpRequest, HttpRequestInitializer}
-import com.google.api.client.json.JsonObjectParser
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.bigquery.model._
 import com.google.api.services.bigquery.{Bigquery, BigqueryScopes}
 import com.google.auth.Credentials
 import com.google.auth.http.HttpCredentialsAdapter
 import com.google.auth.oauth2.GoogleCredentials
-import com.google.cloud.hadoop.util.{ApiErrorExtractor, ChainingHttpRequestInitializer}
+import com.google.cloud.hadoop.util.ChainingHttpRequestInitializer
 import com.google.common.base.Charsets
 import com.google.common.hash.Hashing
 import com.google.common.io.Files
 import com.spotify.scio.bigquery.types.BigQueryType.HasAnnotation
 import org.apache.beam.sdk.extensions.gcp.options.GcpOptions.DefaultProjectFactory
-import org.apache.beam.sdk.extensions.gcp.options.GcsOptions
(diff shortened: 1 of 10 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -306,67 +336,81 @@ class BigQueryClient private (private val projectId: String,
...
+    bigquery.tables().delete(table.getProjectId, table.getDatasetId, table.getTableId)
+      .execute()
+  }
+
   // =======================================================================
   // Type safe API
   // =======================================================================
 
   /**
-   * Get a typed iterator for a BigQuery SELECT query or table.
```
nevillelyh: Haven't looked thoroughly yet but you should fix scaladoc style in Intellij to avoid changing these:
https://github.com/spotify/scio/wiki/Style-Guide#documentation-style

guidj:
@nevillelyh Any ideas on why the build fails? The following seems to be the issue:

```
java.io.IOException: Error reading credential file from environment variable GOOGLE_APPLICATION_CREDENTIALS, value 'scripts/data-integration-test.json': File does not exist.
```

ravwojdyla:
@guidj it seems like circleci is broken. Even tho we have secrets like `ENCRYPTION_KEY` configured in circleci console, they don't seem to be available on the build machine. Need to contact circleci.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryClient.scala
```diff
@@ -104,159 +78,82 @@ class BigQueryClient private (private val projectId: String,
...
-        Thread
-          .currentThread()
-          .getStackTrace
-          .exists { e =>
-            e.getClassName.startsWith("scala.tools.nsc.interpreter.") ||
-            e.getClassName.startsWith("org.scalatest.tools.")
-          }
-      )
+  /** Services */
+  private[scio] val tables = new TableService(projectId, bigquery, credentials)
```
ravwojdyla: This will cause lazy `credentials` to evaluate thus to create GCP credentials, and this will not work in untrusted PR builds. Also you probably don't want this to be eagerly evaluated anyway.

guidj:
@ravwojdyla I see. Should be fixed now

ravwojdyla:
@guidj could you pls fix all invalid style changes. also separate the style fixes into separate commit pls?

guidj:
Thought I fixed all changes, but I’ll take a look.

guidj:
I removed the style changes

ravwojdyla:
Some **partial** small comments above. But overall - I see this as two issues:
 * #840
 * support all BigQuery operations + refactor of `BigQueryClient`

and frankly this PR does the 1st and partially the 2nd. I would prefer the have this as two separate PRs.

I'm thinking whether it makes sense to have 1st PR (which doesn't break anything) just adds:
 * `exportTable` and `loadTable` to `BigQueryClient`

2nd PR:
 * refactor the whole `BigQueryClient` (breaking change), to have a more consistent API across the client

@guidj what do you think?

guidj:
@ravwojdyla  I can see that. I would have just implemented the two functions, but the client file grew very large and I thought it was best to break functionality into separate modules for maintainability. When you say having a more consistent API across the client, would that mean grouping the existing functionality under common modules like I'm doing with `load` and `export`?

Most styling issues are due to me using IntelliJ to automatically style the code, even though I applied some of the config recommendations. Is there a way to have the preferences in the [style guide](https://github.com/spotify/scio/wiki/Style-Guide) automatically enforced or at least highlighted?

ravwojdyla:
@guidj to your 1st question - yes. 2nd: if you configure the style guide, it should work fine.

Issue #1031, opened 2018-01-25
Workaround for BEAM-3311 BigTableIO and Iterable of KV


nevillelyh:
Now that BEAM-3311 is closed as a won't fix, we might need a local workaround.

Extend BigTableIO to write Iterable of KV
https://issues.apache.org/jira/browse/BEAM-3311

andrisnoko:
@nevillelyh I will take on this issue since I had initially explored it [here](https://github.com/andrisnoko/beam/blob/bulk-write-working/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigtable/BigtableIO.java#L734)

PR #1057, opened 2018-02-28
fixes spotify/scio/issues/1031

andrisnoko:
Copy pasted from [https://github.com/apache/beam/tree/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigtable](bigtable) since they are package private. Added BulkWrite in BigTableIO.

Fix for issue [https://github.com/spotify/scio/issues/1031](1031)

@nevillelyh @ravwojdyla @andrewsmartin @fallonchen


nevillelyh:
Since according to https://issues.apache.org/jira/browse/BEAM-3311, BulkWrite won't be fixed upstream, is there a way to fix it here without copying their code and modifying? Otherwise you'll have to fight a loosing battle of constantly rebasing against upstream and fixing breaking changes.

fallonchen:
@andrisnoko just wanted to nudge you on the question above--is there a way to do this without copying code and modifying?

jto:
@andrisnoko It seems to me that what your PR does is essentially the solution proposed in BEAM-3311.

> Alternatively, you can add a set of steps which will buffer data using a trigger.
PubSubIO -> ... original pipeline ... -> ParDo(Choose a random key in [0, 1000)) -> Window.into(new GlobalWindows()).triggering(Repeatedly.forever(AfterProcessingTime.pastFirstElementInPane().plusDelayOf(Duration.standardSeconds(10)))) -> GBK -> Values -> BigtableIO

If so, I don't see why we'd need to copy code from Beam.

Am I missing something ?

scio-bigtable/src/main/java/com/spotify/scio/bigtable/BigtableBulkWriter.java
```diff
@@ -0,0 +1,195 @@
+package com.spotify.scio.bigtable;
```
nevillelyh: License header

scio-bigtable/src/main/java/org/apache/beam/sdk/io/gcp/bigtable/BigtableServiceHelper.java
```diff
@@ -0,0 +1,10 @@
+package org.apache.beam.sdk.io.gcp.bigtable;
```
nevillelyh: License header

scio-bigtable/src/main/java/org/apache/beam/sdk/io/gcp/bigtable/BigtableServiceHelper.java
```diff
@@ -0,0 +1,10 @@
...
+
+import com.google.cloud.bigtable.config.BigtableOptions;
+
+public class BigtableServiceHelper extends BigtableServiceImpl {
```
nevillelyh: Some comments on why this hack exists would be great.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -127,7 +125,9 @@ package object bigtable {
           tableId)
         self.getTestInput[Row](input)
       } else {
-        var read = BigtableIO.read().withBigtableOptions(bigtableOptions).withTableId(tableId)
```
nevillelyh: No real change here except formatting, undo it?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -259,14 +259,38 @@ package object bigtable {
           bigtableOptions.getProjectId, bigtableOptions.getInstanceId, tableId)
         self.context.testOut(output.asInstanceOf[TestIO[(ByteString, Iterable[T])]])(self)
       } else {
-        val sink = BigtableIO.write().withBigtableOptions(bigtableOptions).withTableId(tableId)
```
nevillelyh: No real change here except formatting, undo it?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -259,14 +259,38 @@ package object bigtable {
...
+          .withBigtableOptions(bigtableOptions)
+          .withTableId(tableId)
         self
           .map(kv => KV.of(kv._1, kv._2.asJava.asInstanceOf[java.lang.Iterable[Mutation]]))
           .applyInternal(sink)
       }
       Future.failed(new NotImplementedError("Bigtable future not implemented"))
     }
 
+    /**
```
nevillelyh: Use JavaDoc style please: https://github.com/spotify/scio/wiki/Style-Guide#documentation-style

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -259,14 +259,38 @@ package object bigtable {
...
+    def saveAsBigtable(tableId: String,
+                       numOfShards: Int = 0,
+                       bigtableOptions: BigtableOptions,
+                       flushInterval: Duration = Duration.standardSeconds(1))
+                      (implicit ev: T <:< Mutation)
+    : Future[Tap[(ByteString, Iterable[Mutation])]] = {
+      if (self.context.isTest) {
+        val output = BigtableBulkOutput(
+          bigtableOptions.getProjectId, bigtableOptions.getInstanceId, tableId)
+        self.context.testOut(output.asInstanceOf[TestIO[(ByteString, Iterable[T])]])(self)
```
nevillelyh: You don't really need the `BigtableBulkOutput` case class right? Since the output type is still `(ByteString, Iterable[T])`, same as `BigtableOutput`. The extra `Iterable` grouping is invisible to users.

build.sbt
```diff
@@ -363,7 +363,14 @@ lazy val scioBigtable: Project = Project(
   description := "Scio add-on for Google Cloud Bigtable",
   libraryDependencies ++= Seq(
     "org.apache.beam" % "beam-sdks-java-io-google-cloud-platform" % beamVersion,
-    "org.scalatest" %% "scalatest" % scalatestVersion % "test"
+    "org.scalatest" %% "scalatest" % scalatestVersion % "test",
+    "org.hamcrest" % "hamcrest-all" % hamcrestVersion % "test",
```
nevillelyh: Remove these dependencies now that we're not copying their test code?

Issue #1056, opened 2018-02-27
Ability to cancel job from ScioResult


nevillelyh:
IIRC `ScioResult#waitUntilDone` only throws an exception when the timeout is up but won't cancel the job. Need to pass it to the underlying job (might be runner specific?).

ClaireMcGinty:
as far as working this functionality into `ScioResult`, what do you think about modifying `waitUntilDone(duration: Duration = Duration.Inf)` to include an optional boolean like `waitUntilDone(duration: Duration = Duration.Inf, doCancelJob: Boolean = false)` ? The other option would be to just add an entirely new method like `def waitOrCancel(duration: Duration = Duration.Inf): ScioResult  = {...}`

nevillelyh:
+1 on optional argument. Although it might break binary compatibility? But that should be OK if we include this in 0.5.0 release. cc @regadas

regadas:
optional arg sounds good at first glance. I think we can fit it in the 0.5.0 release

regadas:
@ClaireMcGinty are you planing on taking this one? 😄

ClaireMcGinty:
@regadas yep!! it looks like it's as simple as calling `internal.cancel()` (at least, this worked with the batch Dataflow runner...). I'll put up a PR after I get things cleaned up/tested.

ClaireMcGinty:
posted https://github.com/spotify/scio/pull/1058

regadas:
PR #1058

PR #1058, opened 2018-02-28
Add job cancellation option to ScioResult#waitUntilDone (#1056)

ClaireMcGinty:
Integration-tested in my own Scio batch pipelines, as well as unit tested. let me know if you want me to scrap the unit test, since including hardcoded timeouts in automated tests can be kind of finicky  ¯\\\_(ツ)_/¯

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
@@ -65,15 +65,15 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
   /** Get metrics of the finished pipeline. */
   def getMetrics: Metrics
 
-  /** Wait until the pipeline finishes. If timeout duration is exceeded and `doCancelJob` is set,
+  /** Wait until the pipeline finishes. If timeout duration is exceeded and `cancelJob` is set,
     * cancel the internal [[PipelineResult]]. */
-  def waitUntilFinish(duration: Duration = Duration.Inf, doCancelJob: Boolean = false):
+  def waitUntilFinish(duration: Duration = Duration.Inf, cancelJob: Boolean = false):
   ScioResult = {
     try {
       Await.ready(finalState, duration)
     } catch {
       case e: TimeoutException =>
-        if (doCancelJob) {
+        if (cancelJob) {
           internal.cancel()
           throw new PipelineExecutionException(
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -72,10 +73,20 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
 
   /**
    * Wait until the pipeline finishes with the State `DONE` (as opposed to `CANCELLED` or
-   * `FAILED`). Throw exception otherwise.
+   * `FAILED`). Throw exception otherwise. If `doCancelJob` is set, cancel the [[PipelineResult]].
    */
-  def waitUntilDone(duration: Duration = Duration.Inf): ScioResult = {
-    waitUntilFinish(duration)
+  def waitUntilDone(duration: Duration = Duration.Inf, doCancelJob: Boolean = false): ScioResult = {
+    try {
```
regadas: what about moving this to `waitUntilFinish`, since they both expect a duration and `waitUntilDone` relies on it. wdyt?

ClaireMcGinty: I can do that! I wasn't 100% sure of the semantic implications of each method tbh

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -64,18 +65,30 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
   /** Get metrics of the finished pipeline. */
   def getMetrics: Metrics
 
-  /** Wait until the pipeline finishes. */
-  def waitUntilFinish(duration: Duration = Duration.Inf): ScioResult = {
-    Await.ready(finalState, duration)
+  /** Wait until the pipeline finishes. If timeout duration is exceeded and `doCancelJob` is set,
+    * cancel the internal [[PipelineResult]]. */
+  def waitUntilFinish(duration: Duration = Duration.Inf, doCancelJob: Boolean = false):
```
regadas: @ClaireMcGinty 🎨 this is a nit: i think I would prefer `cancelJob` instead of `doCancelJob`

ClaireMcGinty:
sure thing - just updated the PR.

Issue #1055, opened 2018-02-27
Add command line option to specify waitUntilDone duration


nevillelyh:
https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L360

ClaireMcGinty:
after taking a look how `ScioContext` is set up, I was envisioning implementing this by adding a getter/setter for `jobTimeout` to `ScioOptions.java` and setting it in [ScioContext#parseArguments](https://github.com/spotify/scio/blob/19fee4716f71827ac4affbd23d753bc074c529b8/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L145). that way `ContextScioResult` has access to it when `waitUntilDone` is called. is that roughly what you had in mind? @nevillelyh

nevillelyh:
Pretty much yeah. You shouldn't have to do anything in `parseArguments` since anything in `ScioOptions` will just be passed on. You can do `.as(classOf[ScioOptions])` in `ContextScioResult` and retrieve the setting.

regadas:
#1062

PR #1062, opened 2018-03-02
Add jobTimeout as a command line option (#1055)

ClaireMcGinty:
I set `jobTimeout` as a string serialized from `scala.concurrent.duration.Duration` type, so it's passed in as "1h", "10min" etc.

also, I updated `ScioContext` so if you do pass in a `jobTimeout` param, you don't have to also override the `--blocking=true` to false - it assumes that's what you want to do. thoughts?

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
+++ b/scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
@@ -39,6 +39,10 @@ public interface ScioOptions extends PipelineOptions, KryoOptions {
   boolean isBlocking();
   void setBlocking(boolean value);
 
+  @Description("Time period in scala.concurrent.duration.Duration style to wait for job completion")
+  String getJobTimeout();
+  void setJobTimeout(String value);
+
   @Description("Custom application arguments")
   String getAppArguments();
   void setAppArguments(String arguments);
--- a/scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
+++ b/scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
@@ -39,9 +39,9 @@ public interface ScioOptions extends PipelineOptions, KryoOptions {
   boolean isBlocking();
   void setBlocking(boolean value);
 
-  @Description("Time period in scala.concurrent.duration.Duration style to wait for job completion")
-  String getJobTimeout();
-  void setJobTimeout(String value);
+  @Description("Time period in scala.concurrent.duration.Duration style to block for job completion")
+  String getBlockFor();
+  void setBlockFor(String value);
 
   @Description("Custom application arguments")
   String getAppArguments();
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -258,6 +258,20 @@ class ScioContext private[scio] (val options: PipelineOptions,
       }
     }
 
+  /** Amount of time to block job for. */
+  private[scio] val awaitDuration: Duration = {
+    val blockFor = optionsAs[ScioOptions].getBlockFor
+    try {
+      Option(blockFor)
+        .map(Duration(_))
+        .getOrElse(Duration.Inf)
+    } catch {
+      case e: NumberFormatException =>
+        throw new IllegalArgumentException(s"blockFor param $blockFor cannot be cast to " +
+          s"type scala.concurrent.duration.Duration")
+    }
+  }
+
   // if in local runner, temp location may be needed, but is not currently required by
   // the runner, which may end up with NPE. If not set but user generate new temp dir
   if (ScioUtil.isLocalRunner(options.getRunner) && options.getTempLocation == null) {
@@ -358,9 +372,8 @@ class ScioContext private[scio] (val options: PipelineOptions,
       TestDataManager.closeTest(testId.get, result)
     }
 
-    if (this.isTest || (this.optionsAs[ScioOptions].isBlocking &&
-      Option(this.optionsAs[ScioOptions].getJobTimeout).isEmpty)) {
-      result.waitUntilDone()  // block local runner for JobTest to work
+    if (this.isTest || (this.optionsAs[ScioOptions].isBlocking && awaitDuration == Duration.Inf)) {
+      result.waitUntilDone()
     } else {
       result
     }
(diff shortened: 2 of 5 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -391,6 +393,19 @@ class ScioContext private[scio] (val options: PipelineOptions,
         context.optionsAs[ApplicationNameOptions].getAppName,
         state.toString,
         getBeamMetrics)
+
+    override def getJobTimeout: Duration = {
+      Option(optionsAs[ScioOptions].getJobTimeout) match {
+        case Some(duration) => try {
+          Duration.apply(duration)
```
ravwojdyla: `Duration(duration)` doesn't work here?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -391,6 +393,19 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+        } catch {
+          case e: NumberFormatException =>
```
ravwojdyla: So this will throw after job was already submitted, right? I wonder if it would be better to do this check before we submit the job?

scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
```diff
@@ -39,6 +39,10 @@
   boolean isBlocking();
   void setBlocking(boolean value);
 
+  @Description("Time period in scala.concurrent.duration.Duration style to wait for job completion")
+  String getJobTimeout();
+  void setJobTimeout(String value);
```
regadas: 🎨 I think this name makes the arg meaning vague. Maybe `blockFor` or even just `await`?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -357,7 +358,8 @@ class ScioContext private[scio] (val options: PipelineOptions,
       TestDataManager.closeTest(testId.get, result)
     }
 
-    if (this.isTest || this.optionsAs[ScioOptions].isBlocking) {
+    if (this.isTest || (this.optionsAs[ScioOptions].isBlocking &&
+      Option(this.optionsAs[ScioOptions].getJobTimeout).isEmpty)) {
```
regadas: I think we should check if the option was set and if so ... we should enforce `result.waitUntilFinish(<duration>)` in case the job itself is not calling it.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
regadas: `Duration(duration)`?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ClaireMcGinty: My only concern there was that there would be some duplicated logic - since the timeout is string-serialized in `ScioOptions`, if we added it as a pre-check when `ScioContext` was instantiated, there's not really a place to store the resulting `Duration` object unless you want to introduce some settable `var`. So we'd be deserializing a `Duration` again after submitting the job.

Ultimately not a huge deal since it's still only two invocations :)

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ClaireMcGinty: good catch, i'll update it

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -391,6 +393,19 @@ class ScioContext private[scio] (val options: PipelineOptions,
         context.optionsAs[ApplicationNameOptions].getAppName,
         state.toString,
         getBeamMetrics)
+
+    override def getJobTimeout: Duration = {
```
regadas: maybe just `awaitDuration`?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ClaireMcGinty: actually, I think if I just declare it once as a private val anytime parseArguments has been called it should work

ravwojdyla: Yeah, I think we can take that cost (two invocations), then have this weird situation where the job is submitted and the driver fails with a silly typo. @regadas what do you think?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -391,6 +393,19 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+      Option(optionsAs[ScioOptions].getJobTimeout) match {
```
regadas: maybe just 
```scala
try {
  Option(option)
      .map(Duration(_))
      .getOrElse(Duration.Inf)
catch {
  case e: NumberFormatException =>
      throw new IllegalArgumentException(s"jobTimeout param $duration cannot be cast to " +
               s"type scala.concurrent.duration.Duration")
}
```

regadas: Also If we could do validation at `ScioOptions` level that would be good?

scio-core/src/main/java/com/spotify/scio/options/ScioOptions.java
ClaireMcGinty: I kind of like `blockFor` since I would expect `await` to be a boolean type

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
ClaireMcGinty: you mean something like

```
    if (Option(this.optionsAs[ScioOptions].getBlockFor).isDefined) {
      result.waitUntilFinish(duration)
    } else if (this.isTest || (this.optionsAs[ScioOptions].isBlocking)) {
      result.waitUntilDone()
    } else {
      result
    }
```
?
the only problem is we don't have the value of `cancelJob` in that scope so we don't know whether to call `result.waitUntilFinish(duration, cancelJob = true)` -- and if we added it I feel like `ScioContext#close()` would become functionally the same as `ScioResult#waitUntilFinish()`. what do you think?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
regadas: @ravwojdyla yeah I think we should try doing an early validation of the arg.

ClaireMcGinty:
@regadas @ravwojdyla Refactored the code in response to your comments--I moved the validation/computation of the block duration into a `private[scio] val awaitDuration: Duration = {...}` that's called once, when `ScioContext` is instantiated. I tested it on my local Scio pipeline projects and now an invalid timeout parameter will crash the job before it's sent to Dataflow. let me know what you think

PR #1066, opened 2018-03-05
Change default cancelJob value in ScioResult#waitUntilDone

ClaireMcGinty:


ravwojdyla:
will merge when green.

PR #1060, opened 2018-03-01
Implemented BigQuery load and export

guidj:
Enables exporting tables as csv, json and avro, as well as loading data
in those formats into BigQuery. #840

PR #1067, opened 2018-03-05
Fix up #1060.

ravwojdyla:
Remove obsolete method, fix style.


ravwojdyla:
will merge when green - wanna get circelci test pass.

Issue #1061, opened 2018-03-02
Additional named step automatically added to the end of the pipeline


inejc:
When `withName("BigQuerySink").saveAsTypedBigQuery(...)` transform is used as the last step of the pipeline, an additional step is automatically added at the end of the it. Inspecting the job through Dataflow UI gives named steps that look something like this:
```... --> BigQuerySink --> saveAsTypedBigQuery@{FileName.scala:82}1```

I asked on [gitter](https://gitter.im/spotify/scio) whether there is a way to name the **last** step as well and was asked to open an issue. I'm interested in doing so because every time the `withName("BigQuerySink").saveAsTypedBigQuery(...)` line is moved, the name of the **last** step changes and I end up having trouble updating the job on the fly.

inejc:
I'm interested in looking into this.

inejc:
Could you direct me to where I should start looking for the problem? I see that names are set to transforms in [PCollectionWrapper](https://github.com/spotify/scio/blob/d6635a626868c7db8926273c7c6c142883180ddb/scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala#L44) by calling `apply` on internal `org.apache.beam.sdk.values.PCollection`, getting the name from the [TransformNameable](https://github.com/spotify/scio/blob/d93b05d28e4a59695c3bb35f67c393458a43c94e/scio-core/src/main/scala/com/spotify/scio/values/TransformNameable.scala) trait.

andrewsmartin:
@inejc That's right...so basically when you call `.withName()`, the `TransformNameProvider` is mutated so that it applies the given name to the **next** transform applied, after which the `CallsiteNameProvider` is used again. The `.saveAsTypedBigQuery` "transform" is actually composed of 2 transforms under the hood - first, a `map` to convert from case class to `TableRow`, then a `BigQueryIO.write` to write out the table rows. So here, `.withName` only sets the name for that first map, and then the default callsite naming scheme is used for the `BigQueryIO.write`. 

The easiest way to fix would be to prepend another `.withName` [here](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L1108), with something like `s"givenTfName$Write"`, where `givenTfName` can be obtained from `this`. Lmk if you have any questions!

inejc:
Thanks for the help @andrewsmartin, opened a PR.

inejc:
[...]

PR #1069, opened 2018-03-08
Prepend withName to the saveAsTypedBigQuery inner transform

inejc:
Solves https://github.com/spotify/scio/issues/1061 and includes:
- [x] implementation
- [ ] tests

Additional `.withName` is prepended to the inner (second) transform of the `saveAsTypedBigQuery` function in order to override the default callsite naming scheme.

The name of the transform is changed so that file name and line number are removed completely. The resulting name looks like this: `saveAsTypedBigQuery$Write`. Let me know if this is inappropriate.

The tests are not included as I wasn't able to find any existing `saveAsTypedBigQuery` tests.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -1105,7 +1105,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       import scala.concurrent.ExecutionContext.Implicits.global
       this
         .map(bqt.toTableRow)
-        .saveAsBigQuery(
+        .withName(s"${this.tfName.replaceFirst("@.*", "")}$$Write").saveAsBigQuery(
           table,
           bqt.schema,
           writeDisposition,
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1105,7 +1105,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       import scala.concurrent.ExecutionContext.Implicits.global
       this
         .map(bqt.toTableRow)
-        .saveAsBigQuery(
+        .withName(s"${this.tfName.replaceFirst("@.*", "")}$$Write").saveAsBigQuery(
```
andrewsmartin: While this works, a job that has more than one `.saveAsTypedBigQuery` step would be a bit problematic because they would both try to use the same transform name. IIRC when this happens a number is appended onto one of the transform names so that they are unique, but this isn't deterministic and so could cause problems with streaming updates.

I think to get around this, it would be good to incorporate the name set before calling `.saveAsTypedBigQuery`, which right now is used only in the `.map` above. Since the name provider is reset back to the callsite scheme after the map, `this.tfName` as used here isn't useful. So instead, before calling `this.map(bqt.toTableRow)` we could have a `val tfName = this.tfName`, and use that on this line. WDYT?

inejc: That's a good point, pushed a new commit.

andrewsmartin:
Thanks @inejc!

inejc:
Thanks for the help @andrewsmartin.

PR #1070, opened 2018-03-08
Flink uses EOFException to detect when it should spill to disk instea…

Igosuki:
…d of writing segments in memory, this catches it if it's the cause of a KryoException and rethrows it

This makes scio run properly with the FlinkRunner for batches where memory is smaller than the data sets (think Parquet, Protobuf, Avro files).

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
@@ -17,8 +17,9 @@
 
 package com.spotify.scio.coders
 
-import java.io.{InputStream, OutputStream}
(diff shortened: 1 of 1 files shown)


Igosuki:
This is a known issue in Beam since 2.0 https://issues.apache.org/jira/browse/BEAM-2831
I can provide a running example to demonstrate that it works, otherwise just take a 200mb parquet file and run a sumByKey on a 1024m heap taskmanager.
The code itself is just a port of the equivalent in the Flink java codebase, not sure if it's idiomatic.
To prevent wrapping exceptions Beam uses the @SystemDoFn annotation but I didn't think I should mess with that one so went with a simple solution.

Igosuki:
That is of course, only relevant if you think scio should run on other runners than Dataflow :)

PR #1074, opened 2018-03-12
Bump dependencies

ravwojdyla:


ravwojdyla:
Alright - need to look into what's broken.

PR #1075, opened 2018-03-12
Custom Type Data Validation in Scio BigQuery

jbigred1:
This will add the ability to intercept the mappings from Scala type to BigQuery type, and vice versa, in order to provide your own validation on READ and WRITE when using the Scio `BigQueryType` annotations in your data pipelines


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -81,7 +83,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Array[Byte]] =>
           val b = q"$tree.asInstanceOf[_root_.java.nio.ByteBuffer]"
           q"_root_.java.util.Arrays.copyOfRange($b.array(), $b.position(), $b.limit())"
-
```
nevillelyh: Nit: keep these empty lines to minimize unrelated style changes? Same for the next few cases of adding/removing blank lines.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
```diff
@@ -49,24 +51,27 @@ private[types] object SchemaProvider {
...
-
-    case t if t =:= typeOf[Instant] => ("TIMESTAMP", Iterable.empty)
-    case t if t =:= typeOf[LocalDate] => ("DATE", Iterable.empty)
-    case t if t =:= typeOf[LocalTime] => ("TIME", Iterable.empty)
-    case t if t =:= typeOf[LocalDateTime] => ("DATETIME", Iterable.empty)
-
-    case t if isCaseClass(t) => ("RECORD", toFields(t))
-    case _ => throw new RuntimeException(s"Unsupported type: $tpe")
+  private def rawType(tpe: Type): (String, Iterable[TableFieldSchema]) = {
+    val provider: ValidationProvider = ValidationProviderFinder.getProvider
```
nevillelyh: Is it possible to minimize the re-indentation change here by moving `val provider` outside of the method?

jbigred1: yeah it is the scope is slightly bigger but not a big deal

jbigred1:
good idea about test cases going to add that in

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -118,8 +119,11 @@ private[types] object TypeProvider {
     import c.universe._
     checkMacroEnclosed(c)
 
+    val provider: ValidationProvider = ValidationProviderFinder.getProvider
+
     // Returns: (raw type, e.g. Int, String, NestedRecord, nested case class definitions)
     def getRawType(tfs: TableFieldSchema): (Tree, Seq[Tree]) = tfs.getType match {
+      case t if provider.shouldOverrideType(tfs) => (provider.getScalaType(c)(tfs), Nil)
```
ankscircle: Should we add a docString here explaining what & why we are doing this?

scio-bigquery-test-pipeline/src/test/scala/com/spotify/scio/bigquery/validation/pipeline/BigQueryValidationTest.scala
```diff
@@ -0,0 +1,46 @@
...
+
+
+package com.spotify.scio.bigquery.validation.pipeline
+
+import com.spotify.scio.bigquery.types.BigQueryType
+import com.spotify.scio.bigquery.validation.Country
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{FlatSpec, Matchers}
+
+class BigQueryValidationTest extends FlatSpec with Matchers with GeneratorDrivenPropertyChecks  {
```
ankscircle: docString here explain what this test does?

build.sbt
```diff
@@ -330,6 +331,29 @@ lazy val scioAvro: Project = Project(
   )
 ).configs(IntegrationTest)
 
+lazy val scioBigQueryTestPipeline: Project = Project(
+  "scio-bigquery-test-pipeline",
+  file("scio-bigquery-test-pipeline")
+).settings(
+  // necessary to properly test
```
idreeskhan: Nit, I would change the comment to why its necessary to properly test (need available at compile time)

scio-bigquery-test-pipeline/src/test/scala/com/spotify/scio/bigquery/validation/pipeline/BigQueryValidationTest.scala
```diff
@@ -0,0 +1,46 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+package com.spotify.scio.bigquery.validation.pipeline
+
+import com.spotify.scio.bigquery.types.BigQueryType
+import com.spotify.scio.bigquery.validation.Country
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
```
idreeskhan: Where is this used?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/validation/ValidationProvider.scala
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.spotify.scio.bigquery.validation
+
+import com.google.api.services.bigquery.model.TableFieldSchema
+
+import scala.reflect.macros.blackbox
+import scala.reflect.runtime.universe._
+
+// Interface for all ValidationProvider implementations
```
idreeskhan: Nit, but I think style guide is `/** */` for doc comments

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -250,7 +255,9 @@ private[types] object ConverterProvider {
     // =======================================================================
 
     def cast(tree: Tree, tpe: Type): Tree = {
+      val provider: ValidationProvider = ValidationProviderFinder.getProvider
       tpe match {
+        case t if provider.shouldOverrideType(c)(t) => q"$tree.toString"
```
idreeskhan: Is this always string? Or should we be getting it from the class instead?

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/SampleValidationProvider.scala
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  def getScalaType(c: blackbox.Context)(tfs: TableFieldSchema): c.Tree = {
+    import c.universe._
+    val typeClassOption: Option[Class[_]] = getBySemanticTypeString(tfs)
+    typeClassOption match {
+      case Some(typeClass) => val packageName = typeClass.getPackage.getName
+        val className = TypeName(typeClass.getSimpleName)
+        tq"${c.parse("_root_." + packageName)}.$className"
+      case None => throw new IllegalArgumentException("Should never be here")
```
idreeskhan: What if the string in description is invalid?

jbigred1: so in that case we shouldn't even enter this method since the `shouldOverrideType` should return false so it's correct to throw an error

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
jbigred1: yeah in this situation we need a primitive type if you look at the `DateTime` examples we're printing it as a string format

we don't strictly need a string but I think it's ok to require it in this case

scio-bigquery-test-pipeline/src/test/scala/com/spotify/scio/bigquery/validation/pipeline/BigQueryValidationTest.scala
jbigred1: good catch don't need this

nevillelyh:
Quick question, why adding a new sub-project?

jbigred1:
yeah so we need the implementation class to compile before we actually use it

so we need `SampleOverrideTypeProvider` to compile before we use it in the annotations otherwise the class won't be available at compile time when we do macro expansion

does that make sense

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -266,6 +274,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Instant] => q"_root_.com.spotify.scio.bigquery.Timestamp($tree)"
         case t if t =:= typeOf[LocalDate] => q"_root_.com.spotify.scio.bigquery.Date($tree)"
         case t if t =:= typeOf[LocalTime] => q"_root_.com.spotify.scio.bigquery.Time($tree)"
+
```
idreeskhan: Nit, but I don't like randomly added lines 😛

jbigred1:
link to wiki page as well https://github.com/spotify/scio/wiki/Override-Type-Provider-Specification

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/Country.scala
```diff
@@ -0,0 +1,44 @@
...
+    if (!isValid(data)) {
+      throw new IllegalArgumentException("Not valid")
+    }
+    new Country(data)
+  }
+
+
+  def isValid(data: String): Boolean = data.length == 2
+
+  def parse(data: String): Country = new Country(data)
```
andrewsmartin: Is this used anywhere? It bypasses the validation logic in `apply` - is that intentional?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/validation/DummyOverrideTypeProvider.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override def getScalaType(c: blackbox.Context)(tfs: TableFieldSchema): c.Tree = null
+
+  override def createInstance(c: blackbox.Context)(tpe: c.Type, tree: c.Tree): c.Tree = null
+
+  override def initializeToTable(c: blackbox.Context)(modifiers: c.universe.Modifiers,
+                                             variableName: c.universe.TermName,
+                                             tpe: c.universe.Tree): Unit = Unit
+
+
```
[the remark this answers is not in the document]
nevillelyh: Also parameter indentation

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/SampleOverrideTypeProvider.scala
```diff
@@ -0,0 +1,103 @@
...
+        }.parse($tree)"
+        instanceOfType
+      case None => throw new IllegalArgumentException("Should never be here")
+    }
+  }
+
+  def initializeToTable(c: blackbox.Context)(modifiers: c.universe.Modifiers,
+                                             variableName: c.universe.TermName,
+                                             tpe: c.universe.Tree): Unit = Unit
+
```
nevillelyh: Nit: extra blank lines.

build.sbt
```diff
@@ -292,6 +292,10 @@ lazy val scioTest: Project = Project(
 ).settings(
   commonSettings ++ itSettings,
   description := "Scio helpers for ScalaTest",
+  // necessary to properly test since we need this value at compile time
+  initialize in Test ~= { _ =>
+    System.setProperty( "OVERRIDE_TYPE_PROVIDER", "com.spotify.scio.bigquery.validation.SampleOverrideTypeProvider" )
```
nevillelyh: There could be only 1 type provider per project right? How does it work if one wants to override multiple types?

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/Country.scala
jbigred1: yeah so this is called in the quasiquotes when we construct a new instance

it shouldn't bypass the validation logic let me move it to the value class instead

build.sbt
jbigred1: currently only 1 type provider per project in the future we could extend it to provide a list of type providers and just take the first one that matches

if you want to override multiple types you have to contain it in the same type provider

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/Country.scala
```diff
@@ -0,0 +1,44 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+package com.spotify.scio.bigquery.validation
+
+class Country(data: String) {
```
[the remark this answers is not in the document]
jbigred1: actually we can't do this since we need to provide validation when we construct

build.sbt
nevillelyh: I guess the easiest would be just to support a comma separated list of class names in the JVM property?

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/Country.scala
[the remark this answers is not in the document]
jbigred1: we have to move the validation into the class but that should be it in order to support this now

build.sbt
jbigred1: yeah we should probably just do that now

but in general having one provider to provide multiple types is the paradigm we're going to use internally

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/Country.scala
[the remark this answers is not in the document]
andrewsmartin: Can the validation still exist in `def apply` of the companion object? And parse would just look like 

```
def parse(data: String): Country = Country(data)
```

Calling the apply method to perform validation before actually hitting the constructor.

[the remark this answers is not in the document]
andrewsmartin: Either way, not a big deal since it's just for a test :)

build.sbt
jbigred1: decided to leave this as is since we can extend in the future to allow for multiple types

scio-test/src/test/scala/com/spotify/scio/bigquery/validation/BigQueryValidationTest.scala
```diff
@@ -0,0 +1,84 @@
...
+    assertThrows[IllegalArgumentException] {
+      val tableRow = TableRow("country" -> "USA", "countryString" -> "USA", "noCountry" -> "USA")
+      val input = CountryInput.fromTableRow(tableRow)
+    }
+  }
+
+  "ValidationProvider" should "no error thrown converting valid data" in {
+    val tableRow = TableRow("country" -> "US", "countryString" -> "USA", "noCountry" -> "USA")
+    CountryInput.fromTableRow(tableRow)
+    //val inputTableRow = CountryInput.toTableRow(inputWithExtra)
```
andrewsmartin: Can we remove this?

scio-test/src/test/scala/com/spotify/scio/bigquery/validation/BigQueryValidationTest.scala
```diff
@@ -0,0 +1,84 @@
...
+
+
```
andrewsmartin: Extra blank line

scio-test/src/test/scala/com/spotify/scio/bigquery/validation/BigQueryValidationTest.scala
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+
```
andrewsmartin: Extra blank line

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/validation/SampleOverrideTypeProvider.scala
andrewsmartin: @jbigred1 here too

bbhoss:
Awesome work, this is a great addition.

Issue #1013, opened 2018-01-17
Make Java IO etc. more usable


nevillelyh:
We should add some implicit converters, probably `com.spotify.scio.JavaConverters._`, to make Java APIs more usable. It should include implicit conversions from common Java types like `SerializableFunction`, `DynamicDestination`. etc.

jpvelez:
I'll take a stab at this.

I'm going to start by looking at AvroIO, BigQueryIO, FileIO; examine the builder methods and list out their argument types. Any non-primitive types, let's figure out how to make them more fluent.

jpvelez:
## AvroIO

### Static Tranform Constructors
To make AvroIO.(Read | ReadAll | Parse | ParseAll | Write | TypedWrite) root transforms, you call one of the following `AvroIO` static constructor methods:

```java
Read<T> read(Class<T> recordClass) 
ReadAll<T> readAll(Class<T> recordClass) 
Read<GenericRecord> readGenericRecords(Schema schema) 
ReadAll<GenericRecord> readAllGenericRecords(Schema schema) 
Read<GenericRecord> readGenericRecords(String schema) 
ReadAll<GenericRecord> readAllGenericRecords(String schema) 
Parse<T> parseGenericRecords(SerializableFunction<GenericRecord, T> parseFn) 
ParseAll<T> parseAllGenericRecords(SerializableFunction<GenericRecord, T> parseFn) 
Write<T> write(Class<T> recordClass) 
Write<GenericRecord> writeGenericRecords(Schema schema) 
TypedWrite<UserT, Void, OutputT> writeCustomType()
TypedWrite<UserT, Void, GenericRecord> writeCustomTypeToGenericRecords()
```
The only input types that may need Scala conversion here are:
- `org.apache.avro.Schema` 
- `org.apache.beam.sdk.transforms.SerializableFunction`

### AvroIO.Read PTransform
Each method use a Builder for each Transform type: 
```java
  public static <T> Read<T> read(Class<T> recordClass) {
    return new AutoValue_AvroIO_Read.Builder<T>()
        .setMatchConfiguration(MatchConfiguration.create(EmptyMatchTreatment.DISALLOW))
        .setRecordClass(recordClass)
        .setSchema(ReflectData.get().getSchema(recordClass))
        .setHintMatchesManyFiles(false)
        .build();
  }
```

[...]

PR #1076, opened 2018-03-12
First Stab at Java Converters

jpvelez:
Implemented Java Converters (#1013) for Beam file abstractions `ResourceId` and `FilenamePolicy`. This isn't finished code, just trying out an approach.

@nevillelyh please let me know if this is the direction you had in mind.


scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2016 Spotify AB.
```
nevillelyh: Wrong year. You can use `$today.year` substitution in IntelliJ's copyright settings.

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  implicit def toResourceId(filenamePrefix: String): ResourceId =
+    FileBasedSink.convertToFileResourceIfPossible(filenamePrefix)
+
+  case class FilenamePolicy(baseFilename: String,
+                            shardTemplate: String = null,
+                            templateSuffix: String = null,
+                            windowedWrites: Boolean = false)
+
+  implicit def toFilenamePolicy(policy: FilenamePolicy): DefaultFilenamePolicy = {
```
nevillelyh: maybe worth having a implicit conversion from `String` to `DefaultFilenamePolicy` as well?

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,43 @@
...
+
+package com.spotify.scio
+
+import org.apache.beam.sdk.io.{DefaultFilenamePolicy, FileBasedSink}
+import org.apache.beam.sdk.io.fs.ResourceId
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
+
+object JavaConverters {
+
+  implicit def toResourceId(filenamePrefix: String): ResourceId =
```
regadas: IMHO we should avoid using implicit type conversion. Using implicit class is a better approach.

regadas: 👍 yes that would nicer

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
jpvelez: Assuming that `String` represents a `baseFilename`, and that `shardTemplate` and `templateSuffix` are null and `windowedWrites` is false?

nevillelyh: Yeah, like `"path".asFileNamePolicy`

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
jpvelez: In other words?

```scala
implicit class RichString (x: String) {
  def asResourceId(): ResourceId = FileBasedSink.convertToFileResourceIfPossible(x)
  def asFilenamePolicy(): DefaultFilenamePolicy = DefaultFilenamePolicy.fromStandardParameters(
      StaticValueProvider.of(x),
      null,
      null,
      false)
}
```

nevillelyh: LGTM

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,46 @@
...
+        StaticValueProvider.of(s.toResourceId()), null, null, false)
+  }
+
+  case class FilenamePolicy(baseFilename: String,
+                            shardTemplate: String = null,
+                            templateSuffix: String = null,
+                            windowedWrites: Boolean = false)
+
+  implicit class RichFilenamePolicy(policy: FilenamePolicy) {
+    def toJava(): DefaultFilenamePolicy = DefaultFilenamePolicy.fromStandardParameters(
```
jpvelez: Wasn't sure whether this would be better as a method on `FilenamePolicy` but went with implicit class pattern for consistency.

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package com.spotify.scio
+
+import org.apache.beam.sdk.io.{DefaultFilenamePolicy, FileBasedSink}
+import org.apache.beam.sdk.io.fs.ResourceId
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
+
+object JavaConverters {
+
+  implicit class RichString(s: String) {
```
ravwojdyla: `extends AnyVal`?

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,50 @@
...
+    def toResourceId(): ResourceId = FileBasedSink.convertToFileResourceIfPossible(s)
+    def toFilenamePolicy: DefaultFilenamePolicy = DefaultFilenamePolicy.fromStandardParameters(
+        StaticValueProvider.of(s.toResourceId()), null, null, false)
+  }
+
+  case class FilenamePolicy(baseFilename: String,
```
ravwojdyla: `final`

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,50 @@
...
+                            shardTemplate: String = null,
+                            templateSuffix: String = null,
+                            windowedWrites: Boolean = false)
+
+  implicit class RichFilenamePolicy(policy: FilenamePolicy) {
```
ravwojdyla: `extends AnyVal`

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,50 @@
...
+    def toJava(): DefaultFilenamePolicy = DefaultFilenamePolicy.fromStandardParameters(
+      StaticValueProvider.of(policy.baseFilename.toResourceId()),
+      policy.shardTemplate,
+      policy.templateSuffix,
+      policy.windowedWrites)
+  }
+
+  implicit class RichAny[T](value: T) {
```
ravwojdyla: `extends AnyVal`

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,50 @@
...
+ * under the License.
+ */
+
+package com.spotify.scio
+
+import org.apache.beam.sdk.io.{DefaultFilenamePolicy, FileBasedSink}
+import org.apache.beam.sdk.io.fs.ResourceId
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
+
+object JavaConverters {
```
ravwojdyla: Haven't thought though much about these - but is there anything we need to be worried regarding serialization and these classes?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/JavaConvertersExample.scala
```diff
@@ -0,0 +1,35 @@
...
+ * under the License.
+ */
+
+package com.spotify.scio.examples.extra
+
+import org.apache.beam.sdk.io.{AvroIO, TextIO}
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
+import com.spotify.scio.JavaConverters._
+
+object JavaConvertersExample {
```
ravwojdyla: Would be nice to have test together with (even experimental) PR, preferably as part of some dummy pipeline.

jpvelez: Absolutely, will do that next.

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
nevillelyh: We shouldn't have to. These conversion happens outside of the transform. SDK Java code shouldn't even see the `Rich*` classes.

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
nevillelyh: In this case probably better as a method on `FilenamePolicy` since it's a class we have control of.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/JavaConvertersExample.scala
jpvelez: Done.

Not that the pipeline test doesn't actually test whether the correct file was written to. In test mode, `.saveAsCustomOutput` never actually applies to the transform. It just compares the SCollection immediately prior to output to the expected collection.

So the dummy pipeline mostly provides an example for users trying to understand the Java Converters. From a testing standpoint, the pipeline test just checks that the code compiles. If we want to makes assertions on the values returned by implicit conversions, we'd need a regular unit test.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/JavaConvertersExample.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.beam.sdk.io.TextIO
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
+
+object JavaConvertersExample {
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    val output = args("output")
+    val t: TextIO.TypedWrite[String, Void] = TextIO.writeCustomType()
+    val transform = args("converter") match {
```
jpvelez: To avoid a bunch of separate example pipelines, thought it would be best to parameterize the Java Conversion applied to the TextIO.typedWrite transform. LMK if you prefer another approach

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
nevillelyh: Also nitpicking: call it `asJava` to be consistent with collection converters?

nevillelyh:
LGMT mostly. Can you address the remaining review comments please?

scio-examples/src/test/scala/com/spotify/scio/examples/extra/JavaConvertersExampleTest
```diff
@@ -0,0 +1,56 @@
...
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.testing.{CustomIO, PipelineSpec, TextIO}
+
+class JavaConvertersExampleTest extends PipelineSpec {
+
+  val output = "gs://bucket/path/to/file"
+  val expected: List[String] = (1 to 10).toList.map(_.toString)
+  val io = CustomIO[String](output)
```
jpvelez: Since the pipeline uses the TextIO transform, it seemed consistent to expect it to return an `SCollection[String]` in the text. Since `.saveAsCustomOutput` doesn't actually apply the transform in testing, though, we have to manually map Int -> String in the pipeline.

scio-examples/src/test/scala/com/spotify/scio/examples/extra/JavaConvertersExampleTest
```diff
@@ -0,0 +1,56 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.testing.{CustomIO, PipelineSpec, TextIO}
+
+class JavaConvertersExampleTest extends PipelineSpec {
```
nevillelyh: File missing `.scala` suffix?

scio-examples/src/test/scala/com/spotify/scio/examples/extra/JavaConvertersExampleTest
```diff
@@ -0,0 +1,56 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.testing.{CustomIO, PipelineSpec, TextIO}
```
nevillelyh: The `testing` package is designed to be imported with a wild card and we should show it in the example here `testing._`

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
jpvelez: Ah, yep. That's what I was aiming at. ;)

jpvelez:
Apologies for broken builds, IntelliJ isn't compiling things correctly atm to harder to catch errors. Should be okay now

jpvelez:
Looks like the build is now breaking for unrelated reasons?

`Inconsistency detected by ld.so: dl-tls.c: 493: _dl_allocate_tls_init: Assertion `listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!
Exited with code 127`

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  implicit class RichString(val s: String) extends AnyVal {
+    def toResourceId(): ResourceId = FileBasedSink.convertToFileResourceIfPossible(s)
```
regadas: nit: this extra `()` can be removed

scio-core/src/main/scala/com/spotify/scio/JavaConverters.scala
```diff
@@ -0,0 +1,48 @@
...
+    def toFilenamePolicy: DefaultFilenamePolicy = DefaultFilenamePolicy.fromStandardParameters(
+        StaticValueProvider.of(s.toResourceId()), null, null, false)
+  }
+
+  final case class FilenamePolicy(baseFilename: String,
+                            shardTemplate: String = null,
+                            templateSuffix: String = null,
+                            windowedWrites: Boolean = false) {
+    def asJava(): DefaultFilenamePolicy = DefaultFilenamePolicy.fromStandardParameters(
```
regadas: nit: same here

regadas:
@jpvelez I'll attach the "nit" changes to this PR

PR #1087, opened 2018-03-27
Update GCS connector from 1.6.1 to 1.6.3.

janlugt:
This changes the GCS batch request endpoint to the new per-api endpoint
(https://github.com/GoogleCloudPlatform/bigdata-interop/blob/branch-1.6.x/gcs/CHANGES.txt),
as the old endpoint is being deprecated
(https://developers.googleblog.com/2018/03/discontinuing-support-for-json-rpc-and.html).


nevillelyh:
Thanks!

PR #1089, opened 2018-03-27
Fix #1088 set schema on the TypedWrite

### LINKED ISSUE #1088: IllegalArgumentException when writing avro using AvroIO.TypedWrite
```
[info]   java.lang.IllegalArgumentException: Unless using DynamicDestinations, .withSchema() is required.
[info]   at org.apache.beam.sdk.repackaged.com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
[info]   at org.apache.beam.sdk.io.AvroIO$TypedWrite.expand(AvroIO.java:1173)
[info]   at org.apache.beam.sdk.io.AvroIO$TypedWrite.expand(AvroIO.java:848)
[info]   at org.apache.beam.sdk.Pipeline.applyInternal(Pipeline.java:537)
[info]   at org.apache.beam.sdk.Pipeline.applyTransform(Pipeline.java:491)
[info]   at org.apache.beam.sdk.values.PCollection.apply(PCollection.java:299)
[info]   at com.spotify.scio.values.PCollectionWrapper.applyInternal(PCollectionWrapper.scala:44)
[info]   at com.spotify.scio.values.PCollectionWrapper.applyInternal$(PCollectionWrapper.scala:42)
[info]   at com.spotify.scio.values.SCollectionImpl.applyInternal(SCollection.scala:1279)
[info]   at com.spotify.scio.values.SCollection.saveAsTypedAvroFile(SCollection.scala:976)
```

andrewsmartin:


scio-avro/src/main/scala/com/spotify/scio/avro/types/AvroType.scala
```diff
@@ -231,7 +231,7 @@ object AvroType {
  *
  * This decouples generated fields and methods from macro expansion to keep core macro free.
  */
-class AvroType[T: TypeTag] {
+class AvroType[T: TypeTag] extends Serializable {
```
nevillelyh: Where is `AvroType` pulled into a lambda?

scio-test/src/it/scala/com/spotify/scio/avro/TypedAvroIT.scala
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+
+package com.spotify.scio.avro
+
+import com.google.common.io.Files
+import com.spotify.scio.{ContextAndArgs, ScioResult}
+import com.spotify.scio.testing.PipelineSpec
+
+class TypedAvroIT extends PipelineSpec {
```
nevillelyh: Don't really need to extend `PipelineSpec` right? Not even using any matchers so `extends FlatSpec` should be enough.

OTOH, is there anything we should assert after running the job?

scio-test/src/it/scala/com/spotify/scio/avro/TypedAvroIT.scala
```diff
@@ -0,0 +1,42 @@
...
+  "Typed Avro IO" should "not throw exception" in {
+    val tempDir = Files.createTempDir()
+    TypedAvroJob.run(tempDir.getAbsolutePath)
+    tempDir.deleteOnExit()
+  }
+}
+
+object TypedAvroJob {
+  def run(tempDir: String): ScioResult = {
```
nevillelyh: For the sake of consistency (and less confusion to those new to the code base), use the standard `def main():Unit = ...`? You can pass in a `--output=$tempDir` arg.

scio-avro/src/main/scala/com/spotify/scio/avro/types/AvroType.scala
andrewsmartin: In the function that converts typed objects to generic record. https://github.com/spotify/scio/pull/1089/files#diff-417755bd55a6dbb7a50a9f58d6a7af4bL973

scio-test/src/it/scala/com/spotify/scio/avro/TypedAvroIT.scala
andrewsmartin: 👍 can change to just stick with `FlatSpec`. I thought about asserting that the file contents are as expected but it seemed a bit out of scope for this test...happy to include something like that if you think it's worthwhile though. I mainly wrote the test to make sure the fix works.

nevillelyh: Maybe use the `noException should be thrownBy` matcher then?
http://www.scalatest.org/user_guide/using_matchers

andrewsmartin: 👍

scio-test/src/it/scala/com/spotify/scio/avro/TypedAvroIT.scala
andrewsmartin: 👍

Issue #1093, opened 2018-03-31
Repl doesn't start on Windows


f-loris:
When using the scio-repl on Windows an error appears while starting:
```
Using Scala version 2.11.12 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_162)

Type in expressions to have them evaluated.
Type :help for more information.

System property 'bigquery.project' not set. BigQueryClient is not available.
Set it with '-Dbigquery.project=<PROJECT-NAME>' command line argument.
<console>:1: error: invalid escape character
val sc: ScioContext = new com.spotify.scio.repl.ReplScioContext(org.apache.beam.sdk.options.PipelineOptionsFactory.fromArgs("").create(), List("C:\Users\xy\AppData\Local\Temp\scio-repl-570657197348397031\scio-repl-session.jar"))
                                                                                                                                                   ^
<console>:1: error: invalid escape character
val sc: ScioContext = new com.spotify.scio.repl.ReplScioContext(org.apache.beam.sdk.options.PipelineOptionsFactory.fromArgs("").create(), List("C:\Users\xy\AppData\Local\Temp\scio-repl-570657197348397031\scio-repl-session.jar"))
                                                                                                                                                              ^
<console>:1: error: invalid escape character
val sc: ScioContext = new com.spotify.scio.repl.ReplScioContext(org.apache.beam.sdk.options.PipelineOptionsFactory.fromArgs("").create(), List("C:\Users\xy\AppData\Local\Temp\scio-repl-570657197348397031\scio-repl-session.jar"))
                                                                                                                                                                   ^
<console>:1: error: invalid escape character
val sc: ScioContext = new com.spotify.scio.repl.ReplScioContext(org.apache.beam.sdk.options.PipelineOptionsFactory.fromArgs("").create(), List("C:\Users\xy\AppData\Local\Temp\scio-repl-570657197348397031\scio-repl-session.jar"))
                                                                                                                                                                         ^
<console>:1: error: invalid escape character
val sc: ScioContext = new com.spotify.scio.repl.ReplScioContext(org.apache.beam.sdk.options.PipelineOptionsFactory.fromArgs("").create(), List("C:\Users\xy\AppData\Local\Temp\scio-repl-570657197348397031\scio-repl-session.jar"))
                                                                                                                                                                               ^
<console>:1: error: invalid escape character
[...]

PR #1094, opened 2018-03-31
#1093 fix scio-repl start on windows

f-loris:
Fix for #1093 to ensure that repl works properly on Windows machines. 
`StringEscapeUtils.escapeJava` is used to ensure that the path to the `nextReplJar` is a valid Java path since `\` is used on Windows as a path separator.

### DIFF
--- a/scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
+++ b/scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
@@ -23,6 +23,7 @@ import org.apache.beam.sdk.options.PipelineOptionsFactory
 import com.spotify.scio.bigquery.BigQueryClient
 import com.spotify.scio.{scalaVersion, scioVersion}
 import org.apache.beam.sdk.extensions.gcp.options.GcpOptions.DefaultProjectFactory
+import org.apache.commons.lang.StringEscapeUtils
 
 import scala.tools.nsc.GenericRunnerSettings
 import scala.tools.nsc.interpreter.{IR, JPrintWriter}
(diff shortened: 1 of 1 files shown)


scio-repl/src/main/scala/com/spotify/scio/repl/ScioILoop.scala
```diff
@@ -89,7 +90,7 @@ class ScioILoop(scioClassLoader: ScioReplClassLoader,
     val sc = if (name.nonEmpty) name else "sc"
     val rsc = "com.spotify.scio.repl.ReplScioContext"
     val opts = optsFromArgs(scioOpts)
-    val nextReplJar = scioClassLoader.getNextReplCodeJarPath
+    val nextReplJar = StringEscapeUtils.escapeJava(scioClassLoader.getNextReplCodeJarPath)
```
regadas: @f-loris thanks for your PR!

instead of introducing a new dependency ... wouldn't the following work as well? 
```scala
val nextReplJar = s"""${scioClassLoader.getNextReplCodeJarPath}"""
```

f-loris: @regadas Thanks for your feedback. Unfortunately this doesn't work, otherwise it would have already worked as the line below adds `nextReplJar` inside a `"""` block. 

The result of `s"""${scioClassLoader.getNextReplCodeJarPath}"""` still looks like this `C:\Users\xy\` but it has to look like `"C:\\Users\\xy\\` since `nextReplJar` get interpreted as such. 

Without introducing a new dependency a simple solution would be replacing `\` with `\\`:

```scala
 val nextReplJar = scioClassLoader.getNextReplCodeJarPath.replace("\\","\\\\")
```

regadas: @f-loris after a closer look i see now that the problem happens when the interpreter eval's the interpolated string. Let's keep `StringEscapeUtils.escapeJava`

Thanks!

PR #1095, opened 2018-04-02
Add unionAll on ScioContext to support empty lists.

### LINKED ISSUE #1092: SCollection#unionAll throws an exception on empty list
When passed an empty `Iterable[SCollection[T]]`, `SCollection#unionAll` throws an exception:

```
Exception in thread "main" java.lang.IllegalArgumentException: must either have a non-empty list of PCollections, or must first call empty(Pipeline)
	at org.apache.beam.sdk.values.PCollectionList.of(PCollectionList.java:97)
	at com.spotify.scio.values.SCollection$.unionAll(SCollection.scala:71)
```

Scio should probably handle this case properly before calling into `PCollectionList`, as the union of an empty list of collections is logically an empty collection.

psobot:
Resolves #1092. @nevillelyh and @andrewsmartin for review. 

`SCollection#unionAll` throws an exception when passed an empty list. To prevent this, this PR adds a `#unionAll`method on `ScioContext` that will return `ScioContext#empty` (i.e.: `sc.parallelize(Seq.empty)`) if the input iterable is empty, and return `SCollection#unionAll` if it is not.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -48,7 +48,7 @@ import org.apache.beam.sdk.io.gcp.{bigquery => bqio, datastore => dsio, pubsub =
 import org.apache.beam.sdk.metrics.Counter
 import org.apache.beam.sdk.options._
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
-import org.apache.beam.sdk.transforms.{Create, DoFn, PTransform, SerializableFunction}
+import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.util.CoderUtils
 import org.apache.beam.sdk.values._
 import org.apache.beam.sdk.{Pipeline, PipelineResult, io => gio}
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -64,7 +64,9 @@ import scala.reflect.runtime.universe._
 /** Convenience functions for creating SCollections. */
 object SCollection {
 
-  /** Create a union of multiple SCollections */
+  /** Create a union of multiple SCollections.
```
nevillelyh: Use JavaDoc indentation please like the rest of the code base. Also you can link API with `[[ScioContext#unionAll]]` or something like that.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -840,6 +840,15 @@ class ScioContext private[scio] (val options: PipelineOptions,
     if (name.length <= maxLength) name else name.substring(0, maxLength - 3) + "..."
   }
 
+  /** Create a union of multiple SCollections. Supports empty lists. */
+  def unionAll[T: ClassTag](scs: Iterable[SCollection[T]]): SCollection[T] = scs match {
+    case Nil => empty()
+    case contents => SCollection.unionAll(contents)
+  }
+
+  /** Form a empty SCollection. */
```
andrewsmartin: Nitpick: `a` -> `an`

psobot: D'oh! Thanks @andrewsmartin 👍

andrewsmartin:
Actually just realized it would be breaking because the impl would have to be moved from `SCollection` to `ScioContext` - I'm fine to break, or else should we keep both and add a deprecation warning for the existing version? @nevillelyh wdyt?

nevillelyh:
Fine to keep both. Empty input case is pretty rare anyway.

andrewsmartin:
@psobot looks like a scalastyle error 
`scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala:69:4: Whitespace at end of line`

PR #1096, opened 2018-04-03
Julient/bigquery module

jto:
**WIP. DO NOT MERGE**

This is a WIP on moving everything BQ related out of core and implementing ScioIOs for BQ.
 
Progress so far:
 - Everything BQ is out of core. Backcompatibility is based on implicit conversions.
 - ScioIOs implemented for read operations
 - Old bqRead* metdhods use nio under the hood
 - ScioIOs writes unimplemented
 - All tests are green.

 I will add a few comments on the code on which feedback is very much appreciated.


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
```diff
@@ -0,0 +1,286 @@
...
+      type ReadP = Unit
+      type WriteP = Nothing // ReadOnly
+
+      def id: String = query
+      def read(sc: ScioContext, params: ReadP): SCollection[T] = {
+        @inline def typedRead(sc: ScioContext) = Reads.avroBigQueryRead[T](sc)
+        Reads.bqReadQuery(sc)(typedRead(sc), query)
+      }
+
+      def write(data: SCollection[T], params: WriteP): Future[Tap[T]] = ???
```
jto: Some IOs like SQL queries are actually read-only. `ScioIO` more or less assume that you can always read and write.
I'm using a little trick here to make `write` impossible to call (you can't pass it an instance of `Nothing`), but splitting `ScioIO` into `Read` and `Write` would probably be a better option. I'd like to limit the scope of this already pretty big PR. I think splitting the traits should be done in another PR.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
```diff
@@ -0,0 +1,286 @@
...
+        Reads.bqReadTable(sc)(typedRead(sc), bqio.BigQueryHelpers.parseTableSpec(id))
+      }
+
+      def write(data: SCollection[T], params: WriteP): Future[Tap[T]] = ???
+      def tap(read: ReadP): Tap[T] = ???
+    }
+
+  private[scio] def dynamic[T <: HasAnnotation : ClassTag : TypeTag](
+    newSource: String
+  ): ScioIO.ReadOnly[T, Unit] = {
```
jto: This method only exists to support `BigQueryScioContext.typedBigQuery` which is why it's private.

build.sbt
```diff
@@ -567,6 +570,7 @@ lazy val scioExamples: Project = Project(
   sources in doc in Compile := List()
 ).dependsOn(
   scioCore,
+  scioBigQuery,
```
jto: dependencies on scio-bigquery now have to be explicit since core is not pulling it anymore. We should document that.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -313,23 +312,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
   private var _pipeline: Pipeline = _
   private var _isClosed: Boolean = false
   private val _promises: MBuffer[(Promise[Tap[_]], Tap[_])] = MBuffer.empty
-  private val _queryJobs: MBuffer[QueryJob] = MBuffer.empty
   private val _preRunFns: MBuffer[() => Unit] = MBuffer.empty
   private val _counters: MBuffer[Counter] = MBuffer.empty
+  private var _onClose: Unit => Unit = identity
+  private val _localInstancesCache: scala.collection.mutable.Map[ClassTag[_], Any] =
```
jto: I'm not super happy with this...

 The bigquery client instance was a local val inside ScioContext which guaranted it to be unique per context. Using implicit conversions has the consequence of instantiating a new client every time `ScioContext` is converted to `BigQueryScioContext`. This thing is used to make sure we still only instantiate one client.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -313,23 +312,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
     new SCollectionImpl[T](p, this)
 
-  // =======================================================================
-  // Miscellaneous
-  // =======================================================================
+  /**
+  * Add callbacks calls when the context is closed.
+  */
+  private[scio] def onClose(f: Unit => Unit): Unit =
+    _onClose = _onClose compose f
```
jto: In the same vein. Used to replace `bigQueryClient.waitForJobs(_queryJobs: _*)`

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -446,12 +457,16 @@ class ScioContext private[scio] (val options: PipelineOptions,
   /**  Whether this is a test context. */
   def isTest: Boolean = testId.isDefined
 
-  private[scio] def testIn: TestInput = TestDataManager.getInput(testId.get)
```
jto: `TestInput` and `TestInputNio` were separated which caused existing tests to fail since they use `.input` but all the bqRead* methods are using nio now.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -843,13 +715,20 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
+    requireNotClosed {
+      if (this.isTest) {
+        this.getTestInputNio(io.id)
+      } else {
+        io.read(this, params)
+      }
     }
-  }
+
+  def read[T: ClassTag](io: ScioIO[T]{ type ReadP = Unit }): SCollection[T] =
```
jto: Pure syntactic sugar. If the IO does not require parameters, you can ignore the second parameters block.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQuerySCollection.scala
```diff
@@ -0,0 +1,218 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio
+package bigquery
```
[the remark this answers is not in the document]
jto: Minor convenience.  Anything in `com.spotify.scio` is automatically in scope.
I'll change it to one line import to be coherent with the rest of the codebase.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/io/taps.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.avro.Schema
+import com.twitter.chill.Externalizer
+import org.apache.avro.generic.GenericRecord
+
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+import scala.language.implicitConversions
+
+
+/** Tap for BigQuery TableRow JSON files on local file system or GCS. */
```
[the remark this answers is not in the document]
jto: Absolutely.

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
```diff
@@ -42,5 +42,36 @@ trait ScioIO[T] {
   def tap(read: ReadP): Tap[T]
```
jto: @nevillelyh Should this method return a `Future[Tap[T]]` instead of simply `Tap[T]` ?
The existing methods (`textFile`, `protobufFile`, `objectFile`, `avroFile` and `typedAvroFile`) from `Taps` are all based on `Taps.mkTap` which returns a `Future`. I'm assuming they return a Future to support polling.

I also noticed that the implementation of `TextIO` is copied from `FileStorage`, `ScioContext` and `SCollection`.  It's probably not great to duplicate that much code. Especially code from `FileStorage` since this class is private and there's no reason not to just delegate to it.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -313,23 +312,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
 
-  private lazy val bigQueryClient: BigQueryClient = {
-    val o = optionsAs[GcpOptions]
-    BigQueryClient(o.getProject, o.getGcpCredential)
+  /**
+  * Get from or put in an object in this context local cache
+  * This method is used un scio-bigquery to only instanciate the bigquery client once
+  * even if there's multiple implicit conversions from [[ScioContext]] to [[BigQueryScioContext]]
+  */
+  private[scio] def cached[T: ClassTag](t: => T): T = {
```
nevillelyh: Where is `cached` used now? Can't seem to find it.

build.sbt
nevillelyh: #1075 might have introduced a conflicting dependency change on `scio-bigquery`. Can you take a look?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
nevillelyh: Where is the new client initiated in `ScioContext` now? `private lazy val bigQueryClient` is removed.

jto: It's initiated in BigQueryIO.scala

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
jto: In BigQueryIO.scala.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
```diff
@@ -0,0 +1,286 @@
...
+  private[scio] def bqReadTable[T: ClassTag](sc: ScioContext)(
+    typedRead: bqio.BigQueryIO.TypedRead[T],
+    table: TableReference)
+  : SCollection[T] = sc.requireNotClosed {
+    val tableSpec: String = bqio.BigQueryHelpers.toTableSpec(table)
+    sc.wrap(sc.applyInternal(typedRead.from(table))).setName(tableSpec)
+  }
+}
+
+ /**
```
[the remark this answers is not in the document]
jto: Done

jto:
@nevillelyh I've rebase on nio

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
syodage: @jto
regarding `Tap[T]`, only write method should return the `Future[Tap[T]]` not this method hence why we have `write(data: SCollection[T], params: WriteP): Future[Tap[T]]`, this method is to create a tap which we can use inside write method. write correct me if I am wrong. 

 The reason behind duplicating code from `FileStorage` is to remove that dependency from new `TextIO` and later delete(or clean all un-necessary util classes) that class.  We can see how much code we need to share with other new `IO` classes and restructure the code accordingly. Following the rule, premature optimization is bad practice :).

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryIOIT.scala
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2016 Spotify AB.
```
syodage: nitpick: 2018?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
```diff
@@ -0,0 +1,414 @@
...
+    tableDescription: String) extends WriteParam
+}
+
+/**
+ * Get an SCollection for a BigQuery table.
+ * @group input
+ */
+final case class TableSpec(tableSpec: String) extends ScioIO[TableRow] {
+  type ReadP = Unit
+  type WriteP = TableSpec.WriteParam
```
syodage: use `TableRef.WriteParam` type? don't see why we will have different write param support for `TableRef` and `TableSpec`. I could be wrong though.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
```diff
@@ -0,0 +1,414 @@
...
+  def tap(read: ReadP): Tap[TableRow] =
+    TableRowJsonTap(path)
+
+  def write(data: SCollection[TableRow], params: WriteP): Future[Tap[TableRow]] =
+    params match {
+      case TableRowJsonFile.Parameters(numShards, compression) =>
+        data
+          .map(e => ScioUtil.jsonFactory.toString(e))
+          .applyInternal(data.textOut(path, ".json", numShards, compression))
+        data.context.makeFuture(TableRowJsonTap(ScioUtil.addPartSuffix(path)))
```
syodage: update tap to be 
```scala
  def tap(read: ReadP): Tap[TableRow] =
    TableRowJsonTap(ScioUtil.addPartSuffix(path))
```
and use it here? 
```scala
data.context.makeFuture(tap())
```

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
```diff
@@ -42,5 +42,37 @@ trait ScioIO[T] {
...
+
+object ScioIO {
+  // scalastyle:off structural.type
+  type ReadOnly[T, R] =
+    ScioIO[T] {
+      type ReadP = R
+      type WriteP = Nothing
+    }
+
+  type Aux[T, R, W] =
```
syodage: Could you explain a bit, what is the purpose of this?

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -70,8 +70,6 @@ object JobTest {
 
   private case class BuilderState(className: String,
                                   cmdlineArgs: Array[String] = Array(),
-                                  inputs: Map[TestIO[_], Iterable[_]] = Map.empty,
```
syodage: I was scratching my palms to do this 👍

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
```diff
@@ -0,0 +1,414 @@
...
+}
+
+object TableRowJsonFile {
+  sealed trait WriteParam
+  final case class Parameters(
+    numShards: Int = 0,
+    compression: Compression = Compression.UNCOMPRESSED) extends WriteParam
+}
+
+object Typed {
```
syodage: wooow, I would like to go for a beer with you @jto  just to talk code in this object :).

syodage:
Great to see that new PRs coming in this direction, Nice work @jto . I left few minor comments, apart from that LGTM!

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
jto: To expose a coherent API to the end user. The parameters for xxxIO should always be found in the xxxIO object.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
jto: Next time I'm in NYC ;)

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
jto: fair enough

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
```diff
@@ -42,5 +42,37 @@ trait ScioIO[T] {
   def tap(read: ReadP): Tap[T]
 
   def id: String
+}
+
+object ScioIO {
+  // scalastyle:off structural.type
+  type ReadOnly[T, R] =
```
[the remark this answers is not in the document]
jto: It's really equivalent. Using type aliases avoid the creation of a new type which I prefer here, but it's really a matter of taste I guess.

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
jto: I use this alias to properly type methods returning ScioIO, which make the compiler keep track of the existential types.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/nio/BigQueryIO.scala
syodage: Yeah, that makes sense 👍

scio-core/src/main/scala/com/spotify/scio/nio/ScioIO.scala
jto: Hum actually @syodage , this can't be correct.
`ScioIO.tap` needs to be passed a ReadP which `ScioIO.write` does not have.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/io/taps.scala
```diff
@@ -0,0 +1,107 @@
...
+import com.spotify.scio.ScioContext
+import com.spotify.scio.values.SCollection
+
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+import scala.language.implicitConversions
+
+
+/** Tap for BigQuery TableRow JSON files on local file system or GCS. */
+case class TableRowJsonTap(path: String) extends Tap[TableRow] {
```
ravwojdyla: these case classes should be final right?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/BigQueryType.scala
```diff
@@ -60,6 +60,14 @@ object BigQueryType {
     def table: String
   }
 
+  /**
+   * Trait for companion objects of case classes generated with table.
+   * Intance of this trait are provided as implicits alowing static discovery.
```
ravwojdyla: `Intance` -> `Instance`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/BigQueryType.scala
```diff
@@ -90,6 +98,14 @@ object BigQueryType {
     def query: String
   }
 
+  /**
+   * Trait for companion objects of case classes generated with query.
+   * Intance of this trait are provided as implicits alowing static discovery.
```
ravwojdyla: `Intance` -> `Instance`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/validation/OverrideTypeProviderFinder.scala
```diff
@@ -19,19 +19,26 @@ package com.spotify.scio.bigquery.validation
...
-    val classInstance = Try(Class.forName(System.getProperty("OVERRIDE_TYPE_PROVIDER", ""))
-      .newInstance()
-      .asInstanceOf[OverrideTypeProvider])
-    classInstance match {
-      case Success(value) => value
-      case Failure(NonFatal(_)) => new DummyOverrideTypeProvider
-    }
+    val default: OverrideTypeProvider = new DummyOverrideTypeProvider
+
+    Option(System.getProperty("OVERRIDE_TYPE_PROVIDER"))
```
ravwojdyla: While you at it, afair we mostly use `sys.props` in other parts of codebase.

scio-core/src/it/scala/com/spotify/scio/testing/util/ItUtils.scala
```diff
@@ -35,11 +34,14 @@ import org.apache.beam.sdk.util.{RetryHttpRequestInitializer, Transport}
 /** Integration test utilities. */
 private[scio] object ItUtils {
 
+  // XXX: Copied from BigQueryClient
+  val PROJECT_KEY: String = "bigquery.project"
```
ravwojdyla: Why?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -314,23 +312,37 @@ class ScioContext private[scio] (val options: PipelineOptions,
...
```
ravwojdyla: if you summon the key in next line you might as well desugar it to implicit argument?

build.sbt
jto: Should be fixed now

scio-core/src/it/scala/com/spotify/scio/testing/util/ItUtils.scala
jto: To avoid a cyclic dependencies between projects.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/validation/OverrideTypeProviderFinder.scala
jto: done

README.md
```diff
@@ -81,6 +81,6 @@ Scio includes the following artifacts:
 
 # License
 
-Copyright 2016 Spotify AB.
+Copyright 2018 Spotify AB.
```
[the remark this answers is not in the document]
jto: Yeah I reverted that commit. You're right it does not make sense to have this in this PR

PR #1105, opened 2018-04-11
Upgrade sparkey to 2.3.0

spkrka:


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -88,6 +88,8 @@ private[sparkey] class SparkeyWriter(val uri: SparkeyUri) {
 
   def close(): Unit = {
     delegate.flush()
+    // TODO: configure how much memory to use for hash table
+    // delegate.setMaxMemory(max memory usage in bytes)
```
regadas: is it ok if we move this comment to an issue? easier to track and increases chances of someone picking it up.

nevillelyh: We can merge now and file ticket to track.

PR #1106, opened 2018-04-11
Add support for setting max memory usage for sparkey objects

spkrka:


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -102,17 +102,27 @@ package object sparkey {
...
      * Write the key-value pairs of this SCollection as a Sparkey file to a specific location.
      *
+     * @param path where to write the sparkey files. Defaults to a temporary location.
+     * @param maxMemoryUsage (optional) how much memory (in bytes) is allowed for writing
+     *                       the index file
      * @return A singleton SCollection containing the [[SparkeyUri]] of the saved files.
      */
-    def asSparkey(basePath: String)(implicit w: SparkeyWritable[K, V])
-    : SCollection[SparkeyUri] = {
+    def asSparkey(path: String = "##tempfile##", maxMemoryUsage: Long = -1)
```
nevillelyh: Is it better to just use `null` here to avoid magic string?

PR #1107, opened 2018-04-11
Add enabled-parameter to SCollection.debug

spkrka:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -577,16 +577,28 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
    */
-  def debug(out: () => PrintStream = () => Console.out, prefix: String = ""): SCollection[T] =
-    this.filter { e =>
-      // scalastyle:off regex
-      out().println(prefix + e)
-      // scalastyle:on regex
-      // filter that never removes
-      true
+  def debug(out: () => PrintStream = () => Console.out, prefix: String = "",
+            enabled: Boolean = true): SCollection[T] = {
```
regadas: 🎨 you could remove the outer `{}`

PR #1115, opened 2018-04-18
Bump plugins

regadas:


project/plugins.sbt
```diff
@@ -2,14 +2,14 @@ addSbtPlugin("com.cavorite" % "sbt-avro-1-8" % "1.1.3")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
 addSbtPlugin("com.eed3si9n" % "sbt-unidoc" % "0.4.1")
 addSbtPlugin("com.github.gseitz" % "sbt-protobuf" % "0.6.3")
-addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.7")
-addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")
+addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.8")
+addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.1")
```
nevillelyh: IIRC one of the plugin upgrades broken console and made it impossible to enter passphrase when releasing. Don't remember which exactly but probably `sbt-pgp`.

project/plugins.sbt
```diff
@@ -2,14 +2,14 @@ addSbtPlugin("com.cavorite" % "sbt-avro-1-8" % "1.1.3")
...
 addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.2")
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "1.0.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.0")
-addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.2.27")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.3")
 
 libraryDependencies ++= Seq(
-  "com.github.os72" % "protoc-jar" % "3.3.0.1"
+  "com.github.os72" % "protoc-jar" % "3.5.1.1"
```
nevillelyh: This may pull in a newer `protoc` which might break the `protobuf-java` dependency we have?
Normally we want it to have the same `protoc` version as our `protobuf-java` dep.

build.sbt
```diff
@@ -279,7 +279,7 @@ lazy val scioCore: Project = Project(
     "com.google.guava" % "guava" % guavaVersion,
     "com.google.protobuf" % "protobuf-java" % protobufVersion,
     "me.lyh" %% "protobuf-generic" % protobufGenericVersion,
-    "org.apache.xbean" % "xbean-asm5-shaded" % asmVersion
+    "org.apache.xbean" % "xbean-asm6-shaded" % asmVersion
```
nevillelyh: I'd wait till @andrewsmartin can sanity check closure cleaner.

regadas: @nevillelyh yeah this requires more changes and i first thought.

project/plugins.sbt
regadas: @nevillelyh good point

build.sbt
regadas: @nevillelyh postponing this update

project/plugins.sbt
regadas: @nevillelyh updated .. also made a PR to `protobuf-generic`

project/plugins.sbt
regadas: @nevillelyh verified that this version works

nevillelyh: Nevermind it was 1.0.1, so I guess it's fixed.
https://github.com/spotify/scio/commit/cf5386339a7cf67a7e3e74e9706b5a97929c5c95#diff-a84f91f20f040218bccd09fed4761fb3

project/plugins.sbt
ravwojdyla: I would be cautious about updating protos, actually would prefer to keep it at the same level as Beam.

regadas:
@ravwojdyla protobuf 3.X are compatible it's ok to use updated version ... however we can definitely keep in sync with beam.

PR #1111, opened 2018-04-13
Julient/avro module

jto:
Based on https://github.com/spotify/scio/pull/1096.
Implement nio for Avro

PR #1117, opened 2018-04-20
Julient/bigtable module

jto:
Based on https://github.com/spotify/scio/pull/1111


scio-avro/src/main/scala/com/spotify/scio/avro/nio/AvroIO.scala
```diff
@@ -90,7 +90,7 @@ case class ObjectFile[T: ClassTag](path: String)
             private[scio] def processElement(c: DoFn[T, GenericRecord]#ProcessContext): Unit =
               c.output(AvroBytesUtil.encode(elemCoder, c.element()))
           })
-          .write(AvroFile(path, AvroBytesUtil.schema))(parameters)
+          .write(AvroFile[GenericRecord](path, AvroBytesUtil.schema))(parameters)
```
nevillelyh: Sorry just got around to look at these PRs. Should this and the BigQuery change go to different commits since they're fixing issues with previous commits?

PR #1119, opened 2018-04-20
Replace thread local with object pool

regadas:
this is based on @bilisie PR. we still need to do some benchmarks on this.


scio-core/src/main/scala/com/spotify/scio/coders/CaseClassSerializer.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.esotericsoftware.kryo.Kryo
+import com.esotericsoftware.kryo.io.Input
+import com.esotericsoftware.kryo.serializers.FieldSerializer
+import com.spotify.scio.util.ScioUtil
+import org.objenesis.instantiator.ObjectInstantiator
+import org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator
+
+import scala.reflect.ClassTag
+
+/**
```
ravwojdyla: Wrong javadoc syntax.

scio-core/src/main/scala/com/spotify/scio/coders/CaseClassSerializer.scala
```diff
@@ -0,0 +1,82 @@
...
+  * @param kryo Kryo instance to be used for underlying serialization
+  * @tparam T Underlying serialized type
+  */
+class CaseClassSerializer[T: ClassTag](kryo: Kryo)
+  extends FieldSerializer[T](kryo, ScioUtil.classOf[T]) {
+
+  private val objectCreator: ObjectInstantiator[T] =
+    new SunReflectionFactoryInstantiator[T](ScioUtil.classOf[T])
+
+  override def create(kryo: Kryo, input: Input, classType: Class[T]): T = {
```
ravwojdyla: No need for `{}`

scio-test/src/test/scala/com/spotify/scio/coders/KryoAtomicCoderTest.scala
```diff
@@ -166,6 +172,8 @@ class RecordAKryoRegistrar extends IKryoRegistrar {
       override def read(kryo: Kryo, input: Input, tpe: Class[RecordA]): RecordA =
         RecordA(input.readString(), input.readInt() + 10)
     })
+  }
+
```
ravwojdyla: ?

scio-test/src/test/scala/com/spotify/scio/coders/KryoAtomicCoderTest.scala
```diff
@@ -60,6 +60,10 @@ class KryoAtomicCoderTest extends PipelineSpec {
     cf should roundTrip (Map("a" -> 1, "b" -> 2, "c" -> 3))
   }
 
+  it should "registered case classes" in {
```
ravwojdyla: `registered case classes` doesn't make much sense.

ravwojdyla: Also how are you validating that this is actually using your serializer? Should that be another test?

scio-core/src/main/scala/com/spotify/scio/coders/CaseClassSerializer.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.esotericsoftware.kryo.serializers.FieldSerializer
+import com.spotify.scio.util.ScioUtil
+import org.objenesis.instantiator.ObjectInstantiator
+import org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator
+
+import scala.reflect.ClassTag
+
+/**
+  * Optimized serializer that can be used for case classes.
+  * Has lower memory and CPU overhead than default serialization.
```
ravwojdyla: Benchmark?

scio-core/src/main/scala/com/spotify/scio/coders/EmptyOnSerializeAtomicReference.scala
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.{
+  IOException,
+  ObjectInputStream,
+  ObjectOutputStream,
+  ObjectStreamException
+}
+import java.util.concurrent.atomic.AtomicReference
+
+private abstract class EmptyOnSerializeAtomicReference[T] extends Serializable {
```
ravwojdyla: Could you pls add comments here why do we need this?

scio-core/src/main/scala/com/spotify/scio/coders/EmptyOnSerializeAtomicReference.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  private var reference = new AtomicReference[T]()
+
+  reference.set(create())
```
ravwojdyla: This could probably by just:

```
private var reference = {
  val ref = new AtomicReference[T]()
  ref.set(create())
  ref
}
```
which binds this logic together.

also does this have to be `var`?

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -75,110 +79,112 @@ private object KryoRegistrarLoader {
...
+            KryoRegistrarLoader.load(k)
+
+            k
+          }
+        }
 
-        k.forClass(new KVSerializer)
-        // TODO:
-        // TimestampedValueCoder
+        new KryoPool.Builder(factory).softReferences().build()
```
ravwojdyla: I wonder if the pool should exists in the object not the class, we must run some benchmarks and get some traces on the objects here before we merge this.

scio-core/src/main/scala/com/spotify/scio/coders/CaseClassSerializer.scala
bilisie: @ravwojdyla I'll attach a benchmark in this PR. I didn't notice the presence of benchmarks around coders.

regadas: @ravwojdyla @bilisie yeah let's do this in a different PR

scio-core/src/main/scala/com/spotify/scio/coders/EmptyOnSerializeAtomicReference.scala
regadas: @ravwojdyla @bilisie I think we can remove this by just using `@transient lazy val`

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
regadas: @ravwojdyla since it's dependent on `KryoOptions` I guess for now it can be under `KryoAtomicCoder` (no strong feelings in moving it to `KryoAtomicCoder` object)

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -79,162 +83,169 @@ private[scio] class KryoAtomicCoder[T](private val options: KryoOptions) extends
 
   import KryoAtomicCoder.logger
 
-  private val kryoState: ThreadLocal[KryoState] =
-    new EmptyOnDeserializationThreadLocal[KryoState] {
-      override def initialValue(): KryoState = {
+  @transient private lazy val kryoPool: KryoPool = {
```
ravwojdyla: Please correct me if I'm wrong but this will possibly create multiple KryoPool for each instance of `KryoAtomicCoder`, if pooling KryoPool is thread safe, why not move this to an object and remove `@transient lazy`?

What's the footprint of this change?

scio-core/src/main/scala/com/spotify/scio/coders/EmptyOnSerializeAtomicReference.scala
ravwojdyla: @regadas @bilisie See comment in the new round below.

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
regadas: @ravwojdyla I was talking to @nevillelyh to clarify a few things and yes I guess this is not ideal. The main reason why we can't move it easily to an object is because we depend on KryoOptions. I guess we need to spend more time on this ...

regadas: @ravwojdyla scratch that ... I think i have something ...

regadas: @ravwojdyla updated wdyt?

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -76,168 +80,175 @@ private object KryoRegistrarLoader {
...
+        KryoRegistrarLoader.load(k)
+
+        k
+      }
+    }
+
+    new KryoPool.Builder(factory).softReferences().build()
+  }
+
+  def kryoPool[R](options: KryoOptions)(f: Kryo => R): R = {
```
ravwojdyla: Nitpick: the name of this method is isn't the best, right? `kryoPool` - what does that mean? Am I creating a KryoPool, am I initializing it, am I getting an instance? Maybe `withKryo`, `borrowKryo` or maybe sth different? I like `withKryo` tho.

ravwojdyla:
Could we also validate that this lowers the number of kryo instances in an actual job?

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
regadas: ok I can change it to `withKryo` no strong feelings

regadas:
I'll be trying to get some job running to see if can get some numbers.

PR #1122, opened 2018-04-23
nio implementation for elasticsearch5

jto:


scio-elasticsearch5/src/main/scala/com/spotify/scio/elasticsearch/package.scala
```diff
@@ -64,25 +59,13 @@ package object elasticsearch {
                             maxBulkRequestSize: Int = 3000,
                             errorFn: BulkExecutionException => Unit = m => throw m)
                            (f: T => Iterable[DocWriteRequest[_]]): Future[Tap[T]] = {
-      if (self.context.isTest) {
```
jto: Is this code missing a `requireNotClosed` @nevillelyh ?

nevillelyh: Probably. Do you need to add it to the refactored code?

PR #1132, opened 2018-04-27
Add MiMa plugin

jto:
This seems useful, especially because of the current work on nio. 
Run it with `mimaReportBinaryIssues`

Interestingly, it's finding compatibility issues between master and 0.5.2
I did not make it run automatically in the CI just yet.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,23 @@ val slf4jVersion = "1.7.25"
 val sparkeyVersion = "2.3.0"
 val tensorFlowVersion = "1.7.0"
 
+lazy val mimaSettings = Seq(
+  mimaPreviousArtifacts :=
+    previousVersion(version.value)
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -65,7 +65,24 @@ val slf4jVersion = "1.7.25"
...
+)
+
+def previousVersion(currentVersion: String): Option[String] = {
+  val Version = """(\d+)\.(\d+)\.(\d+).*""".r
+  val Version(x, y, z) = currentVersion
+  if (z == "0") None
+  else Some(s"$x.$y.${z.toInt - 1}")
+}
+
+val commonSettings = Sonatype.sonatypeSettings ++ assemblySettings ++Seq(
```
nevillelyh: Nit: missing space in `++Seq(`. Otherwise this line shouldn't change?

PR #1143, opened 2018-05-02
Replace kryo thread local with object pool

regadas:


mdvorsky:
Hello!

Could this change be introducing a performance problem? It appears that the time spent encoding/decoding records has increased significantly in Scio versions including this change (I was specifically looking at 0.4.7 vs 0.5.5).

Marian

Issue #1097, opened 2018-04-04
Stricter Scalac options


jto:
Right know we use very fews options of the Scala compiler:

```
-Xmax-classfile-name 100
-target:jvm-1.8
-deprecation
-feature
-unchecked
-Ydelambdafy:inline
```

Scalac support a number of options for linting the code that we could leverage. 
Here's an example of the options I use on a personal project: https://github.com/jto/validation/blob/v3/tagless2/project/scalaOptions.scala.

We may not want to use all those options in Scio but some of them are easy to satisfy and help with the code quality (dead / unused code detection, accidental type or variable shadowing, etc...).

@nevillelyh what do you think ?

nevillelyh:
:+1: if we can document each options like in your repo.
Also do you want to introduce all of them or gradually in small batches? I'm leaning towards a few at a time, together with code fixes, so it's not too overwhelming and we can validate the changes through releases.

jto:
Definitely gradually with a dedicated PR every time. 
We don't necessarily have to use all the options either.

jbx:
Closing because we've added some. @jto feel free to re-open if you think there are some obvious things lacking

PR #1147, opened 2018-05-08
Add basic linting scalac options (see: #1097)

jto:
Add a few linting Scalac options as suggested in https://github.com/spotify/scio/issues/1097.

The changes I made in the source are about:

- Avoiding shadowing types
- Getting rid of automatic type adaptations ie: things like `Some(x, y)` become `Some((x, y))`
- Removing unused imports

Those changes should not have any impact whatsoever. 
There're other warnings but fixing them may have some impact. Mostly because we use a couple of deprecated methods from beam's API a lot in tests. Not sure if it's intended.

They're a bunch of other problems that are detected if we add more linting options. They may be more impactful therefore I'm not activating them for now:

- There're a lot of implicit numeric widening. It usually have no impact but we may want to manually checks that none are accidental.
- A few classes are overriding `private[scio]` defs. We probably want to make those classes final.
- There's class declarations inside `package object` which is discouraged.

### DIFF
--- a/project/ScalacOptions.scala
+++ b/project/ScalacOptions.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import sbt._, Keys._
+
+object Scalac {
+
+  // see: https://tpolecat.github.io/2017/04/25/scalac-flags.html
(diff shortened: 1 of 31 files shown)


project/ScalacOptions.scala
```diff
@@ -0,0 +1,91 @@
...
+
+object Scalac {
+
+  // see: https://tpolecat.github.io/2017/04/25/scalac-flags.html
+  val extraOptions = List(
+    "-encoding", "utf-8", // Specify character encoding used by source files.
+    "-explaintypes", // Explain type errors in more detail.
+    // "-language:existentials", // Existential types (besides wildcard types) can be written and inferred
+    // "-language:experimental.macros", // Allow macro definition (besides implementation and application)
+    // "-language:higherKinds", // Allow higher-kinded types
```
regadas: can we also enable `higherKinds` since we use it in a few places?

jto:
@regadas I did a PR here following up with your comments: https://github.com/spotify/scio/pull/1148.

What do you mean by 
> Add options setting that would give the appropriate options

project/ScalacOptions.scala
jto: We could. Right now we import the features when we need them, but that's something we can change.

regadas:
@jto you got it right in the new PR 👍 I was referring to `commonsOptions` setting but I just named it `options`.

I would just potentially move the scala version check inside `scala212settings` under `commonsOptions` (or `options`)

otherwise 👍

Issue #1146, opened 2018-05-07
Add syntactic sugar for custom parallelism


nevillelyh:
To abstract away the custom `DoFn` with `Semaphore` trick to limit concurrency level of a single transform.

Something like:
```scala
p.mapWithParallelism(2) { /* ... */ }
 .flatMapWithParallelism(2) { /* ... */ }
 .filterWithParallelism(2) { /* ... */ }
```

nevillelyh:
We should also probably use the "resource" trick to make sure different instances on `main` end up with separate `Semaphore`s after being cloned on the workers.
https://github.com/spotify/scio/blob/master/scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java

PR #1151, opened 2018-05-08
Limit Parallellism in DoFns

richwhitjr:
Attempt to implement #1146.  Currently it follow the Google example and does simple checking of the functions.  Looking for suggestions on how to test the parallellism part of the new operations.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/transforms/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/transforms/package.scala
@@ -71,6 +71,78 @@ package object transforms {
 
   }
 
+  implicit class LimitedParallelismDoFns[T: ClassTag](val self: SCollection[T]) {
+    def parallelCollectFn[U](maxDoFns: Int)(pfn: PartialFunction[T, U]): DoFn[T, U] =
+      new ParallelLimitedFn[T, U](maxDoFns) {
+        val isDefined = ClosureCleaner(pfn.isDefinedAt(_)) // defeat closure
+        val g = ClosureCleaner(pfn) // defeat closure
+        def parallelProcessElement(c: DoFn[T, U]#ProcessContext): Unit = {
+          if(isDefined(c.element())){
+            c.output(g(c.element()))
+          }
+        }
+      }
+
+    def parallelFilterFn(maxDoFns: Int)(f: T => Boolean): DoFn[T, T] =
+      new ParallelLimitedFn[T, T](maxDoFns) {
+        val g = ClosureCleaner(f) // defeat closure
+        def parallelProcessElement(c: DoFn[T, T]#ProcessContext): Unit = {
+          if(g(c.element())){
+            c.output(c.element())
+          }
+        }
+      }
+
+    def parallelMapFn[U](maxDoFns: Int)(f: T => U): DoFn[T, U] =
+      new ParallelLimitedFn[T, U](maxDoFns) {
+        val g = ClosureCleaner(f) // defeat closure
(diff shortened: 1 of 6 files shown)


scio-core/src/main/scala/com/spotify/scio/util/ParallelLimitedFn.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+
+/**
+ * Utility class to limit the number of parallel doFns
+ * @param maxDoFns Max number of doFns
+ */
+private[scio] abstract class ParallelLimitedFn[T, U](maxDoFns: Int)
+  extends DoFn[T, U] with NamedFn {
+
+  private val semaphore: Semaphore = new Semaphore(maxDoFns, true)
```
nevillelyh: This instance will be serialized as part of the `ParallelLimitedFn`, shipped to the worker and deserialized once per CPU, which defeats the locking mechanism. We need to use the UUID trick like in the `DoFnWithResource`, i.e. assign a UUID per `DoFn` instance in `main`, so that we can uniquely identify cloned instances on a worker that belong in the same transform. We then use it to lookup from a concurrent map of `UUID -> Semaphore`. That way all cloned instances originated from the same source share the same `Semaphore`.

scio-core/src/main/scala/com/spotify/scio/util/ParallelLimitedFn.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import org.apache.beam.sdk.transforms.DoFn
+import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+
+/**
+ * Utility class to limit the number of parallel doFns
+ * @param maxDoFns Max number of doFns
+ */
+private[scio] abstract class ParallelLimitedFn[T, U](maxDoFns: Int)
+  extends DoFn[T, U] with NamedFn {
```
regadas: @richwhitjr what about making this `DoFn` extends `DoFnWithResource` where the resource is `PER_CLASS`?

scio-core/src/main/scala/com/spotify/scio/transforms/package.scala
```diff
@@ -71,6 +71,78 @@ package object transforms {
 
   }
 
+  implicit class LimitedParallelismDoFns[T: ClassTag](val self: SCollection[T]) {
+    def parallelCollectFn[U](maxDoFns: Int)(pfn: PartialFunction[T, U]): DoFn[T, U] =
```
nevillelyh: These methods should be private

scio-core/src/main/scala/com/spotify/scio/transforms/package.scala
```diff
@@ -71,6 +71,78 @@ package object transforms {
 
   }
 
+  implicit class LimitedParallelismDoFns[T: ClassTag](val self: SCollection[T]) {
```
nevillelyh: Scaladoc with a brief description like the `PipeSCollection` below? It should also `extend AnyVal` and probably be named `LimitedParallelismSCollection`?

nevillelyh: FYI it's available via `self` so you can get rid of `ClassTag` context bound and add `implicit val ct = self.ct` inside the code.

richwhitjr: Ah didn't know that 👏

PR #1158, opened 2018-05-11
Fix annotations preservation

### LINKED ISSUE #1134: If annotating case classes with BigQueryType other static macro annotations aren't preserved
In our investigations for adding a new macro annotation we encountered a Scio bug.  When a user adds a static macro annotation such as `@BigQueryType.toTable` we noticed that any other annotations are removed if they exist below this one.

An example is

```scala
@BigQueryType.toTable
@Test.myMacroAnnotation
case class TestClass
```

This will result in only the first macro being triggered with the second one being ignored.  This is because macros are processed in order.  Since the actual ordering matters and each successive static macro annotation can change the Abstract Syntax Tree if we leave any other static macro annotations from what is produced by a given macro the other ones are never triggered at all.

In our situation this is actually not that big of a deal.  We can require people to put our macro above `@BigQueryType.toTable` and everything still triggers as expected but this seems like a bug in Scio and could lead to user confusion at some point.

Full Example:

```scala
import scala.annotation.StaticAnnotation
  import scala.language.experimental.macros
  import scala.reflect.macros._

  def validateImpl(c: blackbox.Context)(annottees: c.Expr[Any]*): c.Expr[Any] = {
    println("Testing")
    annottees.head
  }
  class bigQueryTypeConstruct extends StaticAnnotation {
    def macroTransform(annottees: Any*): Any = macro validateImpl
  }

  @BigQueryType.toTable
  @bigQueryTypeConstruct
  case class TestNoPrint()

  @bigQueryTypeConstruct
  @BigQueryType.toTable
  case class TestPrints()
```

Note that testing only prints on the second example.

Set `sys.props("bigquery.types.debug") = "true"` in order to see the full case classes generated while macros expand.

oskoi:
Fixes #1134

### DIFF
--- a/scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
+++ b/scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
@@ -15,6 +15,10 @@
  * under the License.
  */
 
+// scalastyle:off file.size.limit
+// scalastyle:off number.of.methods
+// scalastyle:off number.of.types
+
 package com.spotify.scio.avro.types
 
 import com.google.protobuf.ByteString
@@ -775,42 +779,73 @@ class TypeProviderTest extends FlatSpec with Matchers {
     r.test shouldBe 2
   }
 
+  class Annotation1 extends StaticAnnotation
+  class Annotation2 extends StaticAnnotation
 
-  class UserAnnot1 extends StaticAnnotation
-  class UserAnnot2 extends StaticAnnotation
-
-  def containsAllUserAnnotTypes[T: TypeTag]: Assertion =
+  def containsAllAnnotTypes[T: TypeTag]: Assertion =
     typeOf[T]
       .typeSymbol
       .annotations
       .map(_.tree.tpe)
-      .containsSlice(Seq(typeOf[UserAnnot1], typeOf[UserAnnot2])) shouldBe true
+      .containsSlice(Seq(typeOf[Annotation1], typeOf[Annotation2])) shouldBe true
 
-  @UserAnnot1
+  @Annotation1
   @AvroType.fromSchemaFile("scio-avro/src/test/avro/scio-avro-test.avsc")
-  @UserAnnot2
-  class FromResourceWithUserAnnotations
+  @Annotation2
+  class FromResourceWithSurroundingAnnotations
 
-  it should "preserve user defined annotations" in {
-    containsAllUserAnnotTypes[FromResourceWithUserAnnotations]
+  it should "preserve surrounding user defined annotations" in {
+    containsAllAnnotTypes[FromResourceWithSurroundingAnnotations]
   }
 
-  @UserAnnot1
+  @AvroType.fromSchemaFile("scio-avro/src/test/avro/scio-avro-test.avsc")
+  @Annotation1
+  @Annotation2
+  class FromResourceWithSequentialAnnotations
+
+  it should "preserve sequential user defined annotations" in {
+    containsAllAnnotTypes[FromResourceWithSequentialAnnotations]
+  }
+
+  @Annotation1
   @AvroType.fromSchema(
     """{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}""")
-  @UserAnnot2
-  class SchemaWithUserAnnotations
+  @Annotation2
+  class SchemaWithSurroundingAnnotations
 
-  "AvroType.fromSchema" should "preserve user defined annotations" in {
-    containsAllUserAnnotTypes[SchemaWithUserAnnotations]
+  "AvroType.fromSchema" should "preserve surrounding user defined annotations" in {
+    containsAllAnnotTypes[SchemaWithSurroundingAnnotations]
   }
 
-  @UserAnnot1
-  @AvroType.toSchema
-  @UserAnnot2
-  case class RecordWithUserAnnotations(a1: Int)
+  @AvroType.fromSchema(
+    """{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}""")
+  @Annotation1
+  @Annotation2
+  class SchemaWithSequentialAnnotations
 
-  "AvroType.toSchema" should "preserve user defined annotations" in {
-    containsAllUserAnnotTypes[RecordWithUserAnnotations]
+  it should "preserve sequential user defined annotations" in {
+    containsAllAnnotTypes[SchemaWithSequentialAnnotations]
+  }
+
+  @Annotation1
+  @AvroType.toSchema
+  @Annotation2
+  case class RecordWithSurroundingAnnotations(a1: Int)
+
+  "AvroType.toSchema" should "preserve surrounding user defined annotations" in {
+    containsAllAnnotTypes[RecordWithSurroundingAnnotations]
+  }
+
+  @AvroType.toSchema
+  @Annotation1
+  @Annotation2
+  case class RecordWithSequentialAnnotations(a1: Int)
+
+  it should "preserve sequential user defined annotations" in {
+    containsAllAnnotTypes[RecordWithSequentialAnnotations]
   }
 }
+
+// scalastyle:on file.size.limit
+// scalastyle:on number.of.methods
+// scalastyle:on number.of.types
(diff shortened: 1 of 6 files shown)


nevillelyh:
@jbigred1 PTAL

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -87,6 +86,7 @@ private[types] object TypeProvider {
         if (parents.map(_.toString()).toSet != Set("scala.Product", "scala.Serializable")) {
           c.abort(c.enclosingPosition, s"Invalid annotation, don't extend the case class $clazzDef")
         }
+
```
jbigred1: unnecessary new line addition/removal 😄

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -450,4 +453,30 @@ class TypeProviderTest extends FlatSpec with Matchers {
...
+  }
+
+  @UserDefinedAnnotation1
+  @BigQueryType.fromSchema(
+    """{"fields": [ {"mode": "REQUIRED", "name": "f1", "type": "INTEGER"} ]}""")
+  @UserDefinedAnnotation2
+  class ClassWithCustomAnnotations
+
+  "BigQueryType.fromSchema" should "support user defined annotations" in {
+    val annotations = typeOf[ClassWithCustomAnnotations].typeSymbol.annotations.map(_.tree.tpe)
```
jbigred1: you can factor this logic into a separate function that takes a type to test

something like `def containsAllTypes[T]` and then test it using `T` instead

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -450,4 +453,30 @@ class TypeProviderTest extends FlatSpec with Matchers {
...
+    annotations should contain allOf(
+      typeOf[UserDefinedAnnotation1], typeOf[UserDefinedAnnotation2])
+  }
+
+  @UserDefinedAnnotation1
+  @BigQueryType.fromSchema(
+    """{"fields": [ {"mode": "REQUIRED", "name": "f1", "type": "INTEGER"} ]}""")
+  @UserDefinedAnnotation2
+  class ClassWithCustomAnnotations
+
```
jbigred1: do you think it's worthwhile testing with other macro annotations that change the AST or is this sufficient

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -229,9 +230,11 @@ private[types] object TypeProvider {
 
   /** Generate a case class. */
   private def caseClass(c: blackbox.Context)
-                       (name: c.TypeName, fields: Seq[c.Tree], body: Seq[c.Tree]): c.Tree = {
+                       (mods: c.Modifiers, name: c.TypeName, fields: Seq[c.Tree], body: Seq[c.Tree]): c.Tree = {
     import c.universe._
-    q"@_root_.com.spotify.scio.bigquery.types.BigQueryTag case class $name(..$fields) extends ${p(c, SType)}.HasAnnotation { ..$body }"
+    val tagAnnot = q"new _root_.com.spotify.scio.bigquery.types.BigQueryTag"
+    val taggedMods = Modifiers(Flag.CASE, typeNames.EMPTY, tagAnnot :: mods.annotations)
+    q"$taggedMods class $name(..$fields) extends ${p(c, SType)}.HasAnnotation { ..$body }"
```
jbigred1: do we need the `case` modifier or is that what the first argument to this function does

jbigred1:
any idea why the tests are having issues on `2.11.x` versions now

jbigred1:
aside from those small comments it looks good

oskoi:
@jbigred1 I had to reduce the size of Avro TypeProviderTest file a little  because of the limitations (800 lines per file).
All tests passed. Make a review, please)

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
```diff
@@ -21,45 +21,39 @@ import com.google.protobuf.ByteString
...
-      {"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}
-    """)
+    """{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}""")
   class MultiLineStringLiteral
 
   @AvroType.fromSchema(
-    """
-      |{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}
-    """.stripMargin)
+    """{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}""")
```
jbigred1: what's the reason to remove the `stripMargin`

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
```diff
@@ -763,13 +754,45 @@ class TypeProviderTest extends FlatSpec with Matchers {
...
+  @UserAnnot2
+  class SchemaWithUserAnnotations
+
+  "AvroType.fromSchema" should "preserve user defined annotations" in {
+    containsAllUserAnnotTypes[SchemaWithUserAnnotations]
+  }
+
+  @UserAnnot1
+  @AvroType.toSchema
+  @UserAnnot2
```
jbigred1: can you test different orderings as well

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
oskoi: The length of the file went beyond the permissible limits (800 lines per file). I had to go for such tricks. If this is critical, I can return. What do you think?

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
oskoi: The length of the file exceeds the permissible limit. Where it is better to add tests with annotations?

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
jbigred1: @nevillelyh what do you think

that seems like a silly reason to remove stylistic elements and things that we're testing

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
jbigred1: could be time to add another file to specifically test annotations

is it possible to increase the file size for tests @nevillelyh what do you think

nevillelyh: We can just add a pair of `scalastyle:{off,on}` to suppress the limit like this: https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L18

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
nevillelyh: You can also surround this with a pair of `// scalastyle:{off,on} line.size.limit` to suppress the limit.

nevillelyh:
Same, LGTM except the minor comments. Prefer suppressing those style warnings than working around them.

oskoi:
@nevillelyh Oh, thanks for the advice!
@jbigred1 Done. Reverted old changes and added test cases with different annotations ordering

jbigred1:
this looks good

@nevillelyh can you merge it when you have a second

jbigred1:
also thanks for fixing the Avro part as well

PR #1171, opened 2018-05-21
Fix #1170 add BigDecimalSerializer

### LINKED ISSUE #1170: scala BigDecimal serializer?
Scala's `BigDecimal` class has some stateful fields that get serialized with the Kryo-derived serializer, which breaks mutation detection checks in the direct runner. Should we register a custom serializer within Scio?

andrewsmartin:


andrewsmartin:
Added `MutationTest` just to have a common place to check for these kinds of errors, since it's come up a few times.

scio-core/src/main/scala/com/spotify/scio/coders/BigDecimalSerializer.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package com.spotify.scio.coders
+
+import com.esotericsoftware.kryo.Kryo
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.twitter.chill.KSerializer
+
+import java.math.{BigDecimal => JBigDecimal}
+
+class BigDecimalSerializer extends KSerializer[BigDecimal] {
```
regadas: @andrewsmartin can we make this private?

andrewsmartin: yep good call

andrewsmartin:
Inherited the build failure from master, @ravwojdyla are you still looking into that? I'm happy to look too

ravwojdyla:
@andrewsmartin yep - it's fixed now in the master, I went ahead and rebased your branch on the master. sorry for the problem.

Issue #1173, opened 2018-05-21
scio-tensorflow/zoltar dependency conflicts


regadas:
zoltar seems to be using later versions of the google api's which breaks some of beam's functionality.

andrewsmartin:
Hey @regadas I wonder if we can find another solution for this? https://github.com/spotify/scio/commit/e13b3579651c8eec346d71969e25922ccfe13cfa works within Scio, but users trying to upgrade to 0.5.5 and use `scio-tensorflow` would also have to add this override in their project. Does Zoltar need 1.23.0 or can it be downgraded?

PR #1180, opened 2018-05-24
Add google-api-client dep override in scio-tensorflow

regadas:
zoltar brings updated versions of google apis which conflict with beam ones.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -535,6 +535,9 @@ lazy val scioTensorFlow: Project = Project(
 ).settings(
   commonSettings,
   description := "Scio add-on for TensorFlow",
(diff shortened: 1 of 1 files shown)


andrewsmartin:
This version might cause issues with dataflow?https://github.com/GoogleCloudPlatform/DataflowJavaSDK/issues/607

regadas:
@andrewsmartin oh yeah! I was bitten bu this before ... this makes things more interesting ...

andrewsmartin:
@regadas should we close this?

regadas:
@andrewsmartin yeah let's close this one. I'm looking into alternative solution. #1173

Issue #1179, opened 2018-05-24
[NIO] Migrating TestIO to ScioIO


nevillelyh:
- [x] Migrate `PubsubIO`
- [x] Migrate `JsonIO`
- [x] Migrate TensorFlow IOs
- [x] Rename [`JobTest#{inputNio,outputNio}`](https://github.com/spotify/scio/blob/nio/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala#L132) to `input/output`, overwriting the legacy ones.
- [x] Remove [`TestIO[T]`](https://github.com/spotify/scio/blob/nio/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala#L141) and all its sub-classes
- [x] Verify that `requiredNotClosed` and `if (sc.isTest)` usage is localized and consistent
- [x] Merge `JobTestTest`, `TapTest`, etc. from `scio-{avro,bigquery}` back into `scio-test`
- [x] Fix `def materialize`, it should be on `SCollection` not `AvroSCollection`
- [x] Clean up `ReadP/WriteP` `Nothing` vs `Unit` usage
- [x] Integration tests that cover all `nio`s, including `Tap#{open,value}`
- [x] Double check test IO mapping (should be typed instead of relying on `id` only)
- [x] Make sure we use `override` and `final` consistently
- [ ] ~Clean up access control~
- [x] Clean up duplicate taps
- [x] Clean up name spaces
- [x] Clean up imports #1316

`requiredNotClosed` was used to ensure that `ScioContext#close()` is not called before opening a `SCollection` from some input.

Regarding release strategy: we will merge this into a `0.7.0` branch, make a few `alpha` & `beta` releases, before merging it to master and releasing stable `0.7.0`.

nevillelyh:
Gonna leave access control/visibility as is and tweak later.
Everything is considered done for this ticket.

PR #1181, opened 2018-05-25
Update ScioContext and SCollection to use TextIO (ScioIO[T])

syodage:
This PR is focused on removing only old `TextIO` which extends `TestIO` and use new `TextIO` which extends `ScioIO[T]`.  first step for fix https://github.com/spotify/scio/issues/1179

There are two ways we can integrate this changes, either use the same APIs and wrap the abstract read/write nio methods.  (eg: `saveAsTextFile(...)` and `textFile(..)` 
Pros:
 - Will have nice API eg: `sc.textFile(TextIO("in.txt"))`  to read with default configurations and `sc.saveAsTextFile(TextIO("out.txt"))` to write with default configurations

Cons:
- This works ok with TextIO but defeat the one of major requirement of this whole new implementation which is to decouple the IOs from the core. (maybe we can think about a workaround to not to add other IO impl as dependencies but have different APIs for each IO by using implicits) 

or Use abstract methods (eg: `read(...)` and `write(...)`

Pros: 
- Decouple all IO classes from core module (one of main goal of this refactoring) 

Cons:
- API would be bit dusty, as we need to add extra parenthesis read/write with default configurations eg: `sc.read(TextIO("in.txt"))( )` and `sc.write(TextIO("out.txt"))( )` 

Here I have implemented the first approach which uses the existing APIs, once I get the build works with this and based on the suggestions from others will update the code to whatever solutions we agree upon.


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -972,20 +972,10 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
-      this.map(_.toString)
-    }
-
-    if (context.isTest) {
-      context.testOut(textIO)(s)
-      s.saveAsInMemoryTap
-    } else {
-      textIO.write(s, writeP)
-    }
+    write(textIO)(writeP)
```
syodage: This gives me Type inference error, tried couple of ways to resolve that but it seems type information of the higher order type (type T in `ScioIO[T]`) get removed while passing down to the actual write method. 

@jto would be great if you can have a look at this error.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -972,18 +972,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
-      this.map(_.toString)
-    }
-
-    if (context.isTest) {
-      context.testOut(textIO)(s)
-      s.saveAsInMemoryTap
-    } else {
-      textIO.write(s, writeP)
-    }
+   write(textIO)(writeP)
```
syodage: This gives me Type inference error, tried couple of ways to resolve that but it seems type information of the higher order type (type T in ScioIO[T]) get removed while passing down to the actual write method.

@jto would be great if you can have a look at this error.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -594,15 +601,9 @@ class ScioContext private[scio] (val options: PipelineOptions,
    * Get an SCollection for a text file.
    * @group input
    */
-  def textFile(path: String,
-               compression: gio.Compression = gio.Compression.AUTO)
+  def textFile(textIO: TextIO)(implicit readP: textIO.ReadParams = textIO.ReadParams())
```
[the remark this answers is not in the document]
jto: Also I should mention that I think it's ok to let `TextIO` in core. WDYT @nevillelyh ?

jto:
> API would be bit dusty, as we need to add extra parenthesis read/write with default configurations eg: sc.read(TextIO("in.txt"))( ) and sc.write(TextIO("out.txt"))( )

This can very easily be fixed. It's actually already done for read:
```
def read[T: ClassTag](io: ScioIO[T]{ type ReadP = Unit }): SCollection[T] = ...
```

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
jto: Does forcing the type works ? `write[ExpectedType](textIO)(writeP)`

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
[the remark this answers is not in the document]
syodage: Yeah, I was trying to do the same but couldn't make the write method works so decided to make the build works by changing the existing API. Will revert it to original and use `read` and `write` methods inside it.

re: TextIO, I also think it is OK to let the `TextIO` in the core module as a reference implementation of new `ScioIO`, I think we should move all other IO impl to their own modules and decouple those modules from the core module.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
syodage: Tried but this gives me shadowing type compilation error as it shadows the type `T` in `SCollection[T]`

scio-core/src/main/scala/com/spotify/scio/nio/TextIO.scala
```diff
@@ -47,28 +50,18 @@ case class TextIO(path: String) extends ScioIO[String] {
   type WriteP = WriteParams
 
   def read(sc: ScioContext, params: ReadParams): SCollection[String] = sc.requireNotClosed {
-    if (sc.isTest) {
-      sc.getTestInputNio(this.id)
-    } else {
-      sc.wrap(sc.applyInternal(BTextIO.read().from(path)
-        .withCompression(params.compression))).setName(path)
-    }
+    sc.wrap(sc.applyInternal(BTextIO.read().from(path)
```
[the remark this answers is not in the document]
syodage: yeah, that makes the impl more clear and remove the knowledge of testing which I think should handle by the framework instead of IO impl.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
syodage: I think I understood the root cause, working on a fix.

jto:
@syodage what's the status on this ? Still WIP ?

nevillelyh:
Gonna close this since I took over the WIP work and it's now in `neville/nio` branch.

Issue #1064, opened 2018-03-02
Missing (Left|Right|Full)OuterSkewedJoin


joelostlund:
Hello,

Currently scio only implements a regular skewed join. For my project and performance reasons I need to have a LeftOuterSkewedJoin, which SCIO doesn't have. Ideally all needs could easily be solved by a FullOuterSkewedJoin.

https://github.com/spotify/scio/blob/a38ac1139c12837d571219322e343e47346ec13e/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala#L318

The SparseOuterJoin is not performant enough for my situation.

PR #1193, opened 2018-05-31
Add skew join variants

martinbomio:
This PR tries to addresses https://github.com/spotify/scio/issues/1064.

Also move around some of the join hash and skewed variants to its own classes, mostly because the PairSCollection functions class was getting big, but can be moved again.

@nevillelyh @fallonchen


nevillelyh:
It's failing style check right now:
`PairHashSCollectionFunctions.scala:83:0: Whitespace at end of line`

You can run it locally with `sbt scalastyle`.

scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2016 Spotify AB.
```
regadas: @martinbomio 2018 😄

scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
```diff
@@ -0,0 +1,405 @@
+/*
+ * Copyright 2016 Spotify AB.
```
regadas: @martinbomio 2018

scio-test/src/test/scala/com/spotify/scio/values/PairHashSCollectionFunctionsTest.scala
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright 2016 Spotify AB.
```
regadas: @martinbomio 2018

scio-test/src/test/scala/com/spotify/scio/values/PairSkewedSCollectionFunctionsTest.scala
```diff
@@ -0,0 +1,251 @@
+/*
+ * Copyright 2016 Spotify AB.
```
regadas: @martinbomio 2018

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -69,12 +68,10 @@ private object PairSCollectionFunctions {
 
 // scalastyle:off number.of.methods
 /**
- * Extra functions available on SCollections of (key, value) pairs through an implicit conversion.
+ * Extra functions available on SCollections of (key, value) pairs for skew joins
```
nevillelyh: This shouldn't change right? Coz we moved skewed join methods elsewhere.

Issue #1156, opened 2018-05-11
Add method to create BigQuery table


chuwy:
Basically a wrapper for `bigquery.tables().insert()`. Or is there existing convenient method that I've missed?

jto:
@nevillelyh I agree. Should we add a method in ScioContext to give access to the instance of `com.google.api.services.bigquery.Bigquery` that `BigQueryClient` is using ?

chuwy:
Thanks guys! My initial problem was that I couldn't find a short enough snippet to create a table and at first ended up with something bloated to 15 lines of code. Right now though I've found out I can do just following:

```scala
val options = PipelineOptionsFactory.create().as(classOf[BigQueryOptions])
val service = new BigQueryServicesWrapper(options)
service.createTable(tableRef, schema)
```

Which is good enough for me. But I still think exposing `Bigquery` would be really useful as well.

nevillelyh:
Well we monkey pathed `BigQueryServicesWrapper` to be public so that we can use it in Scio but it's not really intended as a public API. We can add a `BigQueryClient#createTable` method but shouldn't try to match Java API's functionality (hard with arg list vs builder).

PR #1195, opened 2018-06-03
Add BigQueryClient#createTable and BigQueryClient#createTypedTable

### LINKED ISSUE #1167: Add support for creating partitioned tables
Now when creating a table using the method `saveAsTypedBigQuery`, it is not possible to specify the field by which to partition the table, the expiration time and the partition type using scio api. I propose to add an annotation that would allow it to be done, or add parameters to the `BigQueryType.toTable`. What do you think about this?

oskoi:
Fixes #1167 , #1156


oskoi:
@regadas In `BigQueryClient`, there are many unsafe methods that throw an exception. Perhaps this problem needs to be solved for all methods. What do you think?

PR #1196, opened 2018-06-04
Add support for reusing side-inputs for hashjoins

spkrka:


scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
```diff
@@ -51,9 +57,15 @@ class PairHashSCollectionFunctions[K, V](val self: SCollection[(K, V)])
     * @group join
     */
   def hashLeftJoin[W: ClassTag](that: SCollection[(K, W)])
-  : SCollection[(K, (V, Option[W]))] = self.transform { in =>
-    val side = combineAsMapSideInput(that)
+   : SCollection[(K, (V, Option[W]))] = hashLeftJoin(that.asArrayBufferMapSideInput
```
regadas: @spkrka syntax error ... missing ')'

scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
```diff
@@ -30,15 +30,21 @@ class PairHashSCollectionFunctions[K, V](val self: SCollection[(K, V)])
   (implicit ctKey: ClassTag[K], ctValue: ClassTag[V]) {
 
    /**
-    * Perform an inner join by replicating `that` to all workers. The right side should be tiny and
-    * fit in memory.
+     * Perform an inner join by replicating `that` to all workers. The right side should be tiny and
```
regadas: @spkrka scio uses javadoc style docs.

scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
```diff
@@ -93,16 +93,15 @@ class PairHashSCollectionFunctions[K, V](val self: SCollection[(K, V)])
     leftHashed.map(x => (x._1, x._2)) ++ rightHashed
   }
 
-  private def combineAsMapSideInput[W: ClassTag](that: SCollection[(K, W)])
-  : SideInput[MMap[K, ArrayBuffer[W]]] = {
-    that.combine { case (k, v) =>
+   private lazy val combineAsMapSideInput: SideInput[MMap[K, ArrayBuffer[V]]] = {
```
spkrka: Is this pointless?

Will ```  implicit def makePairHashSCollectionFunctions[K: ClassTag, V: ClassTag](s: SCollection[(K, V)])
  : PairHashSCollectionFunctions[K, V] =
    new PairHashSCollectionFunctions(s)
``` create a new instance every time anyway?

nevillelyh: Don't think this will actually cache anything. `PairHashSCollectionFunctions` is an implicit class so everytime `that.combineAsMapSideInput` is called the logic below are executed again.

jto: I agree with @nevillelyh. It will get re-evaluated every time.

nevillelyh:
Gonna close this since it won't actually cache anything. Feel free to file an issue if you have a concrete issue affected by this and we'll look into it.

PR #1197, opened 2018-06-06
propagate grpc context in ScioResult

danielnorberg:
Ensure grpc context metadata is available to logging in `org.apache.beam.sdk.PipelineResult.waitUntilFinish()`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,6 +40,7 @@ val elasticsearch5Version = "5.5.0"
 val featranVersion = "0.1.26"
 val gcsConnectorVersion = "1.6.3-hadoop2"
 val gcsVersion = "1.8.0"
+val grpcContextVersion = "1.2.0"
 val guavaVersion = "20.0"
 val hadoopVersion = "2.7.3"
 val hamcrestVersion = "1.3"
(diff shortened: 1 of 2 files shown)


danielnorberg:
@regadas We're working on adding per-task logging in [flo](https://github.com/spotify/flo) by adding the [flo task id](https://github.com/spotify/flo/pull/82) to the [grpc context](https://grpc.io/grpc-java/javadoc/io/grpc/Context.html) and setting up the logging backend to emit [structured json log messages](https://cloud.google.com/kubernetes-engine/docs/how-to/logging#best_practices) decorated with flo task id and styx execution metadata. 

This allows us to construct stackdriver logging filters to display to the user only the logs for a single flo task. The logging backend gets the flo task id from the grpc context, which means that the grpc context must be propagated to all threads that emit log messages. 

In the case of scio there are [valuable log messages](https://github.com/apache/beam/blob/master/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java#L327) [emitted by `PipelineResult.waitUntilFinish()`](https://github.com/apache/beam/blob/master/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/MonitoringUtil.java#L86) that we want tagged with the executing flo task id. As scio calls that method on a [thread pool](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L385) that does not perform automatic grpc context propagation, we need to [explicitly propagate](https://github.com/spotify/scio/pull/1197/files#diff-86dcde70f730bb0f4605f71351fbcb1cR388) it.

build.sbt
```diff
@@ -40,6 +40,7 @@ val elasticsearch5Version = "5.5.0"
 val featranVersion = "0.1.26"
 val gcsConnectorVersion = "1.6.3-hadoop2"
 val gcsVersion = "1.8.0"
+val grpcContextVersion = "1.2.0"
```
jto: Why are you using this version specifically ? It was published on March 2017. The latest version is 1.12.0 and was published on May 2018.

nevillelyh: @danielnorberg ping

PR #1201, opened 2018-06-08
Add timer to scalatest tests

regadas:
I think this can be usefull, wdyt?
```
[info] BigtableDoFn
[info] - should work (1 second, 343 milliseconds)
[info] - should work with cache (77 milliseconds)
[info] - should work with failures (25 milliseconds)
```


regadas:
also under CI i'm looking to remove all slf4j emitted logs.

ravwojdyla:
lgtm

PR #1207, opened 2018-06-13
adding custom pubsubio

amotamed:
This change backports https://issues.apache.org/jira/browse/BEAM-3218 to Scio.  This gets us around the 10 Meg limit.  This is mainly to start the conversation of if we should have our own PubsubIO object since the current one is flawed

### DIFF
--- a/scio-core/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/ScioPubsubIO.java
+++ b/scio-core/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/ScioPubsubIO.java
@@ -1,0 +1,998 @@
+package org.apache.beam.sdk.io.gcp.pubsub;
+
+import static com.google.common.base.Preconditions.checkState;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.Message;
+import java.io.IOException;
+import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


nevillelyh:
Thanks. What's the progress of the upstream PR https://github.com/apache/beam/pull/4275?
Would prefer to wait a bit if that's going to be merged soon. Monkey patch upstream code is not the most sustainable.

amotamed:
@nevillelyh as far as progress: I've contacted google support (opened a ticket) and asked them to put pressure on it.  I upvoted the issue here: https://issues.apache.org/jira/browse/BEAM-3218 I agree monkey patching is an absolute last resort, we could wait a few weeks or a month and if no action has been taken, merge up

amotamed:
@regadas  just a heads up: it looks like we're not getting traction on the beam issue.  We're gzipping on our end to avoid the issue. Should we consider merging if no action has been taken by August 1st?

nevillelyh:
@amotamed so the upstream PR seems stalled. Do you know the original author by any chance? I guess worst case you can always take over the changeset, fix up review comments, and submit as a new PR to unblock this. WDYT?

nevillelyh:
Now that upstream is making progress again (2 separate PRs) I'm leaning towards closing this. Any objection?

amotamed:
yeah, I'll close it.  Hopefully my PR will get merged upstream

PR #1209, opened 2018-06-14
Fix: set kryo options before loading registrars

### LINKED ISSUE #1204: KryoAtomicCoder overrides setRegistrationRequired
It looks like a change in a recent behavior unless we set a flag in pipeline options, 

```
opts.setKryoRegistrationRequired(true)
```

Example code:

```scala
 @KryoRegistrar
 class MyKryoRegistrar extends IKryoRegistrar {
   def apply(k: Kryo): Unit = {
   }
}
```

Expected to throw an exception, but doesn't:

```scala
case class Foo()
val coder = com.spotify.scio.util.ScioUtil.getScalaCoder[Foo]
CoderUtils.encodeToByteArray(coder, Foo())
```

regadas:
Fixes #1204

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
@@ -18,6 +18,8 @@
 package com.spotify.scio.util
 
 import java.lang.{Iterable => JIterable}
+import java.util.concurrent.ConcurrentHashMap
+import java.util.function.Function
 import java.util.{List => JList}
 
 import com.google.common.collect.Lists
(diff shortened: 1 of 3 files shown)


kanterov:
Can you add a test that will catch this? Something similar to what was reported in the original issue.

jto:
LGTM

scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -197,4 +199,12 @@ private[scio] object Functions {
     override def reduceOption(accumulator: Iterable[T]): Option[T] =
       _mon.sumOption(accumulator).orElse(Some(_mon.zero))
   }
+
+  def memoize[T, U](f: T => U): T => U = {
+    lazy val cache = new ConcurrentHashMap[T, U]()
```
nevillelyh: How many `cache` instances will this produce? Will it become a member of `object Functions`?

scio-test/src/test/scala/com/spotify/scio/coders/KryoAtomicCoderTest.scala
```diff
@@ -38,87 +38,85 @@ import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 
 class KryoAtomicCoderTest extends PipelineSpec {
-
```
nevillelyh: Can we undo all the unnecessary style change here?

regadas: yup

scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
regadas: this will produce one `cache` instance per function. I don't believe so.

PR #1216, opened 2018-06-22
Add topicExists function to PubSubAdmin

BenFradet:
We'd need something just like `ensureTopic` but which doesn't create the topic.


BenFradet:
@regadas good point, made the change :+1:

scio-core/src/main/scala/com/spotify/scio/pubsub/PubSubAdmin.scala
```diff
@@ -73,6 +73,18 @@ object PubSubAdmin {
       client.createTopic(topic)
     }
 
+  /**
+   * Check PubSub topic exists.
+   *
+   * @param name Topic path. Needs to follow `projects/projectId/topics/topicName`
+   * @return
+   */
+  def topicExists(pubsubOptions: PubsubOptions, name: String): Try[Topic] =
```
regadas: @BenFradet 🎨 can we rename this to `topic` perhaps? i it suits it better now?

regadas:
Thanks!

Issue #1172, opened 2018-05-21
Clean up Kryo stuff


andrewsmartin:
`KryoAtomicCoder` has blown up quite a bit with the addition of several custom serializers and could use a bit of clean up. 

- [ ] Move `KSerializer` classes into a subpackage, maybe `coders.serializers`
- [ ] Add a `ScioRegistrar` for our custom serializers and load in `KryoAtomicCoder` to get rid of the `k.forSubclass` bloat.

anne-decusatis:
work in progress at https://github.com/spotify/scio/compare/master...anne-decusatis:clean_up_kryo

PR #1220, opened 2018-06-25
Clean up custom serialization in KryoAtomicCoder and in com.spotify.scio.coders package

anne-decusatis:
fixes https://github.com/spotify/scio/issues/1172


scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -85,6 +86,43 @@ private object KryoRegistrarLoader {
 
 }
 
+
+/** serializers we've written in Scio and want to add to Kryo serialization
+  * @see com.spotify.scio.coders.serializers */
+private object ScioKryoRegistrar extends IKryoRegistrar {
```
regadas: @anne-decusatis can we make this a `class`? I prefer to have the same semantic as in the `AlgebebirdRegistrar`.

`new ScioRegistrar().apply(kryo)`

regadas:
Close in favor of #1230

PR #1225, opened 2018-07-02
Make default for `args.getOrElse` call-by-name

TimWSpence:
- Enables usage like `args.getOrElse("arg", args("alternateArg"))`


nevillelyh:
Great thanks!

TimWSpence:
You're welcome! :) Awesome work on Scio, it makes Dataflow actually enjoyable to use!!

PR #1227, opened 2018-07-03
Initial support for tf.metadata (to replace TFRecordSpec)

andrewsmartin:
This adds support for building a [tf.metadata](https://github.com/tensorflow/metadata) schema from a collection of tensorflow examples, replacing our custom TFRecordSpec used to parse saved examples. The schema protobuf allows for storing much more information about the examples which we can gradually add if the need arises. For now however this just saves the feature type and shape for rank 0 or 1 features.

### DIFF
--- a/scio-tensorflow/src/main/protobuf/schema.proto
+++ b/scio-tensorflow/src/main/protobuf/schema.proto
@@ -1,0 +1,335 @@
+// Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+
+
+syntax = "proto2";
+option cc_enable_arenas = true;
+
+package tensorflow.metadata.v0;
+
+option java_package = "org.tensorflow.metadata.v0";
+option java_multiple_files = true;
+
+enum LifecycleStage {
+  UNKNOWN_STAGE = 0;  // Unknown stage.
+  PLANNED = 1;        // Planned feature, may not be created yet.
+  ALPHA = 2;          // Prototype feature, not used in experiments yet.
+  BETA = 3;           // Used in user-facing experiments.
+  PRODUCTION = 4;     // Used in a significant fraction of user traffic.
+  DEPRECATED = 5;     // No longer supported: do not use in new models.
+  DEBUG_ONLY = 6;     // Only exists for debugging purposes.
+};
+
+//
+// Message to represent schema information.
+message Schema {
+  // Features described in this schema.
+  repeated Feature feature = 1;
+
+  // Sparse features described in this schema.
+  repeated SparseFeature sparse_feature = 6;
+
+
+  // declared as top-level features in <feature>.
+  // String domains referenced in the features.
+  repeated StringDomain string_domain = 4;
+
+  // Default environments for each feature.
+  // An environment represents both a type of location (e.g. a server or phone)
+  // and a time (e.g. right before model X is run). In the standard scenario,
+  // 99% of the features should be in the default environments TRAINING,
+  // SERVING, and the LABEL (or labels) AND WEIGHT is only available at TRAINING
+  // (not at serving).
+  // Other possible variations:
+  // 1. There may be TRAINING_MOBILE, SERVING_MOBILE, TRAINING_SERVICE,
+  //    and SERVING_SERVICE.
+  // 2. If one is ensembling three models, where the predictions of the first
+  //    three models are available for the ensemble model, there may be
+  //    TRAINING, SERVING_INITIAL, SERVING_ENSEMBLE.
+  // See FeatureProto::not_in_environment and FeatureProto::in_environment.
+  repeated string default_environment = 5;
+
+}
+
+// Describes schema-level information about a specific feature.
+message Feature {
+
+  // The name of the feature.
+  optional string name = 1;  // required
+
+
+
+
+  // Constraints on the presence of this feature in the examples.
+  optional FeaturePresence presence = 14;
+
+  // The shape of the feature which governs the number of values that appear in
+  // each example.
+  oneof shape_type {
+    // The feature has a fixed shape corresponding to a multi-dimensional
+    // tensor.
+    FixedShape shape = 23;
+    // The feature doesn't have a well defined shape. All we know are limits on
+    // the minimum and maximum number of values.
+    ValueCount value_count = 5;
+  }
+
+  // Type of the feature's values
+  optional FeatureType type = 6;
+
+  // Domain for the values of the feature.
+  oneof domain_info {
+    // Reference to a domain defined at the schema level.
+    string domain = 7;
+    // Inline definitions of domains.
+    IntDomain int_domain = 9;
+    FloatDomain float_domain = 10;
+    StringDomain string_domain = 11;
+    BoolDomain bool_domain = 13;
+  }
+
+  // Constraints on the distribution of the feature values.
+  // Currently only supported for StringDomains.
+  optional DistributionConstraints distribution_constraints = 15;
+
+  // Additional information about the feature for documentation purpose.
+  optional Annotation annotation = 16;
+
+  // Tests comparing the distribution to the associated serving data.
+  optional FeatureComparator skew_comparator = 18;
+
+  // Tests comparing the distribution between two consecutive spans (e.g. days).
+  optional FeatureComparator drift_comparator = 21;
+
+  // List of environments this feature is present in.
+  // Should be disjoint from not_in_environment.
+  // This feature is in environment "foo" if:
+  // ("foo" is in in_environment or default_environments) AND
+  // "foo" is not in not_in_environment.
+  // See Schema::default_environments.
+  repeated string in_environment = 20;
+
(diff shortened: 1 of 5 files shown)


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +270,96 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+   * @group output
+   */
+  def buildAndSaveExampleMetadata(path: String)(implicit ev: T <:< Example): SCollection[Schema] = {
+    val schema = self.buildExampleMetadata(ev)
+    schema.map { s =>
+      val d = FileSystems.matchNewResource(path, false)
+      val chnnl = Channels.newOutputStream(FileSystems.create(d, MimeTypes.BINARY))
+      try {
+        s.writeTo(chnnl)
+      } finally {
```
andrewsmartin: @ravwojdyla We should definitely handle the failure scenario better here. What do you think the expected behavior should be? Maybe we could add retries or something, but if it fails after that we'd kinda want the job to fail right?

andrewsmartin:
BTW - there is no testing for the actual saving of the schema protobuf. Worth adding an IT?

scio-tensorflow/src/main/protobuf/schema.proto
```diff
@@ -0,0 +1,335 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+
+
+syntax = "proto2";
```
regadas: do we know if they package this schemas? I'm not against the copy ... I'm just wondering how we keep update with the schema and how do we check we are not up2date.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +270,96 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+        chnnl.close()
+      }
+    }
+    schema
+  }
+
+  private def examplesToFeatures(examples: SCollection[Example]): SCollection[Feature] = {
```
regadas: 🎨 no need for outer `{}`

regadas:
@andrewsmartin the IT would be nice! is it easy?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +270,96 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+   * @group output
+   */
+  def saveAsTfExampleFileWithMetadata(path: String,
+                                      suffix: String = ".tfrecords",
+                                      compression: Compression = Compression.UNCOMPRESSED,
+                                      numShards: Int = 0)
+                                     (implicit ev: T <:< Example): Future[Tap[Example]] = {
+    import scala.concurrent.ExecutionContext.Implicits.global
+
+    buildAndSaveExampleMetadata(path)(ev)
```
ravwojdyla: Is this suppose to be `path`?

andrewsmartin: Woops

andrewsmartin:
@regadas yep should be straightforward - will add once we decide on how to handle failure scenario.

ravwojdyla:
@regadas PTAL

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +272,114 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+  }
+
+  def saveAsTfExampleFile(path: String,
+                          schema: SCollection[Schema],
+                          schemaFilename: String,
+                          suffix: String,
+                          compression: Compression,
+                          numShards: Int)
+                          (implicit ev: T <:< Example)
+  : (Future[Tap[Example]], Future[Tap[Schema]]) = {
```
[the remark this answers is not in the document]
ravwojdyla: @regadas That would replicate schema for each Example? Why?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +272,114 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+    }
+  }
+
+  /**
+   * Infer a `org.tensorflow.metadata.v0.Schema` from this SCollection of
+   * `org.tensorflow.example.Example`.
+   * @return A singleton `SCollection` containing the schema.
+   */
+  def inferExampleMetadata(implicit ev: T <:< Example): SCollection[Schema] = {
+    val features = examplesToFeatures(self.asInstanceOf[SCollection[Example]])
```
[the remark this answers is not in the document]
ravwojdyla: Fixed.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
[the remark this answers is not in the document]
regadas: @ravwojdyla I guess it makes the return type more concise and closed. You have a combined Future and when it's complete you have (Example, Schema). Replication is not a big deal i think.

scio-tensorflow/src/main/protobuf/schema.proto
yonromai: +1, it's unlikely that they change this schema but if they do :\

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -283,43 +285,56 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+  def saveAsTfExampleFile(path: String,
+                          schema: SCollection[Schema],
+                          schemaFilename: String,
+                          suffix: String,
+                          compression: Compression,
+                          numShards: Int)
+                          (implicit ev: T <:< Example)
+  : (Future[Tap[Example]], Future[Tap[Schema]]) = {
+    val inferedSchema = Option(schema).getOrElse(self.inferExampleMetadata(ev))
+    val schemaPath = path.replaceAll("\\/+$", "") + "/" + schemaFilename
```
regadas: @ravwojdyla can we express this in a better way? can we use string interpolation?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +272,112 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+  private def examplesToFeatures(examples: SCollection[Example]): SCollection[Feature] =
+    examples
+      .flatMap(_.getFeatures.getFeatureMap.asScala)
+      .map { case (name, feature) =>
+        feature.getKindCase match {
+          case KindCase.BYTES_LIST =>
+            ((name, FeatureType.BYTES), feature.getBytesList.getValueCount)
+          case KindCase.FLOAT_LIST =>
+            ((name, FeatureType.FLOAT), feature.getFloatList.getValueCount)
+          case KindCase.INT64_LIST =>
```
yonromai: sure there is no 32bit int?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -263,12 +272,112 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
+          .setName(featureName)
+          .setType(featureType)
+        if (max == min) {
+          // This is a fixed length feature
+          builder.setShape(FixedShape.newBuilder()
+            .addDim(FixedShape.Dim.newBuilder().setSize(max)))
+        }
+        else {
+          // Var length
+          builder.setValueCount(ValueCount.newBuilder().setMin(min).setMax(max))
```
yonromai: Is this how these fields are supposed to be used?
what happens in the case where the max len of a var len field in the training set is N and when doing predictions, a new record comes with length > N? I assume these schema fields are not automatically used, it's up to the users to use it

regadas:
talked with @ravwojdyla IRL let's merge this!

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
ravwojdyla: Yep:

```
  public enum KindCase
      implements com.google.protobuf.Internal.EnumLite {
    BYTES_LIST(1),
    FLOAT_LIST(2),
    INT64_LIST(3),
    KIND_NOT_SET(0);
```

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
ravwojdyla: @yonromai this schema is used only for preprocessing (featran or tf.transform), which themselves would produce another version of schema for inference time, which could or could not include this info - probably not. does that make sense.

PR #1237, opened 2018-07-10
Fix prediction async issue and load model per resource class.

### LINKED ISSUE #1224: Scio-tensorflow PredictDoFn runs very slow because it uses AsyncDoFn
Tensorflow saved model predict function uses JavaAsyncDoFn which creates a new Future for each input record. Since the model is evaluated locally and is probably CPU bound, this leads to very slow processing, and the step gets stuck in dataflow with no progress for a long time. 

Directly using DoFnWithResource (blocking inference) is much faster in some rough tests that I did.

Related issue: https://github.com/spotify/scio/issues/1161

regadas:
Closes #1224

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/JavaAsyncDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/JavaAsyncDoFn.java
@@ -23,6 +23,8 @@ import org.apache.beam.sdk.transforms.DoFn;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A {@link DoFn} that handles asynchronous requests to an external service that returns Java 8
@@ -30,10 +32,13 @@ import java.util.function.Function;
  */
 public abstract class JavaAsyncDoFn<InputT, OutputT, ResourceT>
     extends BaseAsyncDoFn<InputT, OutputT, ResourceT, CompletableFuture<OutputT>> {
+  private Logger logger = LoggerFactory.getLogger(this.getClass());
+
   @SuppressWarnings("unchecked")
   @Override
   protected void waitForFutures(Iterable<CompletableFuture<OutputT>> futures)
       throws InterruptedException, ExecutionException {
+    logger.info(String.format("Futures size: %d", Iterables.size(futures)));
     CompletableFuture.allOf(Iterables.toArray(futures, CompletableFuture.class)).get();
   }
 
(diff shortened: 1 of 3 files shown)


scio-core/src/main/java/com/spotify/scio/transforms/JavaAsyncDoFn.java
```diff
@@ -23,17 +23,22 @@
...
  */
 public abstract class JavaAsyncDoFn<InputT, OutputT, ResourceT>
     extends BaseAsyncDoFn<InputT, OutputT, ResourceT, CompletableFuture<OutputT>> {
+  private Logger logger = LoggerFactory.getLogger(this.getClass());
+
   @SuppressWarnings("unchecked")
   @Override
   protected void waitForFutures(Iterable<CompletableFuture<OutputT>> futures)
       throws InterruptedException, ExecutionException {
+    logger.info(String.format("Futures size: %d", Iterables.size(futures)));
```
nevillelyh: Better use `debug`?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -110,16 +105,28 @@ private[tensorflow] class SavedBundlePredictDoFn[T, V](uri: String,
...
-      .thenApply[V](new Function[TensorFlowModel, V] {
+      .thenApplyAsync[V](new Function[TensorFlowModel, V] {
         override def apply(model: TensorFlowModel): V = f(model.instance().session().runner())
       })
       .toCompletableFuture
 
-  override def createResource(): ModelLoader[TensorFlowModel] = Models.tensorFlow(uri, options)
+  @Teardown
+  def teardown(): Unit = {
+    log.info(s"Closing down predict DoFn $this")
```
nevillelyh: `"Tearing down"` just so it's more precise?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -128,19 +135,30 @@ private[tensorflow] class GraphPredictDoFn[T, V](uri: String,
...
       .get()
-      .thenApply[V](new Function[TensorFlowGraphModel, V] {
+      .thenApplyAsync[V](new Function[TensorFlowGraphModel, V] {
         override def apply(model: TensorFlowGraphModel): V = f(model.instance().runner())
       })
       .toCompletableFuture
+
+  @Teardown
+  def teardown(): Unit = {
+    log.info(s"Closing down predict DoFn $this")
```
nevillelyh: `"Tearing down"`

scio-tensorflow/src/test/scala/com/spotify/scio/tensorflow/TensorflowSpec.scala
```diff
@@ -198,7 +198,8 @@ class TensorflowSpec extends PipelineSpec {
       JobTest[TFGraphJob2Inputs.type]
         .args(s"--graphURI=${graphFile.toUri}", "--output=output")
         .output(TextIO("output")) {
-          _ should containInAnyOrder((1L to 10).map(x => (x, x * 3)).map(_.toString))
+//          _ should containInAnyOrder((1L to 1000000).map(x => (x, x * 3)).map(_.toString))
```
nevillelyh: Commented out for debugging?

PR #1241, opened 2018-07-12
Add support for Distinct with representative values to SCollections

anish749:
Add an api for DistinctWithRepresentativeValueFn which calls Beam's [Distinct.WithRepresentativeValueFn](https://beam.apache.org/documentation/sdks/javadoc/0.4.0/org/apache/beam/sdk/transforms/Distinct.WithRepresentativeValues.html)


nevillelyh:
Thanks. LGTM overall. What about calling it `distinctBy(fn: T => U)`? More concise and should be enough to convey the semantics.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -339,6 +339,22 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   * representative value for each input element.
+   *
+   * @param f The funciton to use to get representative values.
+   * @tparam U The type of representative values used to dedup.
+   * @group transform
+   */
+  def distinctWithRepresentativeValueFn[U](f: T => U)(implicit ctu: ClassTag[U])
+  : SCollection[T] = {
+    this.pApply(Distinct
+      .withRepresentativeValueFn(Functions.serializableFn(f.asInstanceOf[T => U]))
```
regadas: I don't think there's a need for `.asInstanceOf[T => U]` right?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -339,6 +339,22 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+  /**
+   * Returns a new SCollection with distinct elements using given function to obtain a
+   * representative value for each input element.
+   *
+   * @param f The funciton to use to get representative values.
+   * @tparam U The type of representative values used to dedup.
+   * @group transform
+   */
+  def distinctWithRepresentativeValueFn[U](f: T => U)(implicit ctu: ClassTag[U])
+  : SCollection[T] = {
```
regadas: 🎨 remove outer `{}`

scio-examples/src/main/scala/com/spotify/scio/examples/cookbook/DistinctWithRepresentativeValueFnExample.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.google.api.services.bigquery.model.{TableFieldSchema, TableSchema}
+import com.spotify.scio.bigquery._
+import com.spotify.scio.ContextAndArgs
+import com.spotify.scio.examples.common.ExampleData
+
+import scala.collection.JavaConverters._
+
+object DistinctWithRepresentativeValueFnExample {
+
+def main(cmdlineArgs: Array[String]): Unit = {
```
regadas: indentation is off

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -339,6 +339,22 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
 
+  /**
+   * Returns a new SCollection with distinct elements using given function to obtain a
+   * representative value for each input element.
+   *
+   * @param f The funciton to use to get representative values.
+   * @tparam U The type of representative values used to dedup.
+   * @group transform
+   */
+  def distinctWithRepresentativeValueFn[U](f: T => U)(implicit ctu: ClassTag[U])
```
regadas: agree with @nevillelyh this can be called `distinctBy`

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
anish749: You are right, I will remove this.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
anish749: I'll remove this

scio-examples/src/main/scala/com/spotify/scio/examples/cookbook/DistinctWithRepresentativeValueFnExample.scala
anish749: Sorry about this.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
anish749: Yes, we have a few extra functions as part of our codebase, where we have `distinctByKey`. I was confused if we should follow the Beam names, so I kept it as `distinctWithRepresentativeValueFn`, but a simple `distinctBy` makes more sense.
I will also add a `distinctByKey` for `PairSCollectionFunctions`.

anish749:
@nevillelyh, @regadas 
We have a few extra functions as part of our codebase, where we have `distinctByKey`. I was confused if we should follow the Beam names, so I kept it as `distinctWithRepresentativeValueFn`, but a simple `distinctBy` makes more sense.
I also added a `distinctByKey` for `PairSCollectionFunctions`.

Issue #1229, opened 2018-07-05
Make randomSplit accept a random gen


martinbomio:
Would be nice to be able to pass a seed or a (possible seeded) random generator to all variations of `randomSplit` (https://github.com/spotify/scio/blob/72f8a9dcccc1a484f4f6d81678c21ca7cd670c42/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L459)
This will make tests deterministic when using this transform.

jto:
Given that `ThreadLocalRandom` does not support seed (https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadLocalRandom.html#setSeed(long)) , it's not possible in the current implementation. @nevillelyh do you think we can replace `ThreadLocalRandom` by a standard `Random` so that we can set a seed ?

nevillelyh:
I agree. Also unlike spark which has fixed # of partitions and ordered items within each partition, Beam has no such notion since it's runner dependent. So we can't really assign seeds deterministically to each partition and random numbers to each element.

I'm leaning towards closing this as a "won't fix".

martinbomio:
@nevillelyh One alternative is to has the input and split based on the hash if sc.isTest, with the right hash function, this will make the splits deterministic, at least for testing, which was my initial use case.

nevillelyh:
Hashing will break the weight distribution though since the underlying elements may hash unevenly.

PR #1243, opened 2018-07-12
Add support for seeded random split

martinbomio:
Addresses #1229


ravwojdyla:
@martinbomio why do you want to set a seed?

martinbomio:
@ravwojdyla I want to get some determinism while doing random split for testing, but looks like the number of records that get on each split depends also on how data is partitioned across workers

ravwojdyla:
@martinbomio given the nature of the SCollection - the collection is unordered, setting seed for RND won't help to achieve determinism. To support deterministic splits for *tests* we might want to use hash/encoded-value of the element to split - this is obviously suboptimal strategy for "random" split but it might be an option for tests *only*.

PR #1249, opened 2018-07-16
Bump TensorFlow version to 1.9.0

regadas:


build.sbt
```diff
@@ -55,7 +55,7 @@ val kryoVersion = "4.0.2" // explicitly depend on 4.0.1+ due to https://github.c
 val parquetAvroExtraVersion = "0.2.2"
 val parquetVersion = "1.9.0"
 val protobufGenericVersion = "0.2.4"
-val protobufVersion = "3.3.1"
+val protobufVersion = "3.5.1"
```
regadas: Bump version to 3.5.1 is required since `tensorflow-proto` uses `3.5.1` @see [pom](http://repo1.maven.org/maven2/org/tensorflow/proto/1.9.0/proto-1.9.0.pom)

regadas:
I think we should wait for `0.5.6` release.

PR #1250, opened 2018-07-16
Alter saveAsTfExample slightly

ravwojdyla:
* if no schema provided, always infer and save schema
 * if schema provided save given schema


ravwojdyla:
Don't merge - need to make some changed discussed IRL.

ravwojdyla:
@yonromai better?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -306,77 +306,84 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
                              compression = compression)
 
   /**
-   * Save this SCollection of `org.tensorflow.example.Example` as a TensorFlow TFRecord file,
-   * along with a protobuf file representing the inferred `org.tensorflow.metadata.v0.Schema`.
-   * If you need to alter the schema, use [[inferExampleMetadata]] and pass altered schema to
-   * [[saveAsTfExampleFile()]].
```
andrewsmartin: Why remove the doc here? This still saves an inferred schema right?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -306,77 +306,84 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
   }
 
+  /**
+   * Saves this SCollection of `org.tensorflow.example.Example` as a TensorFlow TFRecord file,
+   * along with  `org.tensorflow.metadata.v0.Schema`.
+   * @return
+   */
   def saveAsTfExampleFile(path: String,
-                          schema: SCollection[Schema],
+                          schema: Schema,
```
andrewsmartin: This limits the use case of having Scio infer a Schema and then being able to modify it within the job (since this type change means it requires being materialized now) - is that use case important?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -306,77 +306,84 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
...
-  : (Future[Tap[Example]], Future[Tap[Schema]]) = {
-    val inferedSchema = Option(schema).getOrElse(self.inferExampleMetadata(ev))
+                          (implicit ev: T <:< Example): Future[Tap[Example]] = {
+    require(schemaFilename != null && schemaFilename != "", "schema filename has to be set!")
     val schemaPath = path.replaceAll("\\/+$", "") + "/" + schemaFilename
+    if (schema == null) {
+      // by default if there is no schema provided infer and save schema
+      self.inferExampleMetadata(schemaPath)(ev)
+    } else {
+      // TODO: maybe enforce some schema checks, but at least save the schema along the data
```
andrewsmartin: Should we track this in a separate issue?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
ravwojdyla: This approach follows the TFX - basically what you and @yonromai suggested. In the future the flow would be:
 * schema is inferred once manually
 * schema is altered by user by hand (with some aid)
 * schema might be provided to enforce it - but not change at this point

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
ravwojdyla: @andrewsmartin i was thinking of waiting for Google to release the validation tools and then make a decision whether it even makes sense to do it in scio, but if you wanna create an issue, just in case I'm +1.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -412,6 +419,22 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
       }
 }
 
+private object TFExampleSCollectionFunctions {
+  def saveExampleMetadata(schema: SCollection[Schema], schemaPath: String): Unit = {
```
regadas: 🎨 outer `{}`

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
andrewsmartin: Nope that sounds good to me, let's wait for Google.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
andrewsmartin: Got it 👍 and just to be clear when you say "manual" you just mean calling this without passing in a Schema (so it stays null) to have Scio infer it right?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
regadas: +1 on removing the `TODO` in favor of an issue pointing to this.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
ravwojdyla: @andrewsmartin i guess there would be a separate tool, maybe based on this code maybe not, maybe it would be provided by Google and ppl would just pass the result here - we will see :)

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
yonromai: +1

PR #1264, opened 2018-07-31
fix #1262, AsyncDoFn deadlock

### LINKED ISSUE #1262: AsyncDoFn deadlock
We're using Dataflow to build our pipelines. Recently we had an incident with one of them.
Our Dataflow job stopped consuming messages and had 0 CPU load. 
We asked Google support to assist us and they pointed to a deadlock in `ScalaAsyncDoFn.waitForFutures` method call.
Even that `waitForFutures` deadlock was likely a result of our poorly written `processElement` there are still some concerns about `ScalaAsyncDoFn`.
Google pointed us that `DoFn`'s should be stateless and we should avoid adding threading, blocking, queues or pooling.
And I'm quoting the most interesting part of message in Google Case:

The DoFn should not depend on the state of others elements within the bundle. If the process of one element fails, dataflow may rebundle the elements into a different set, and the DoFn instance may be used on another set. The elements may be re-distributed to other bundles. Per the "No Global Shared State" section of the ParDo API documentation: "programs should not access any mutable static variable state in their DoFn, without understanding that the Java processes for the main program and workers will each have its own independent copy of such state"   
   
The normal recommended pattern of DoFn is as you suspected: "process messages synchronously during processElement invocation".  There can be exceptions, but I am not familiar with them or the rest of your application enough to recommend them at this time.  Here are some more guidelines from our team that I hope are helpful. 

1) There are no guarantees of how many times processElement() will be called between startBundle() and finishBundle(). There is no guarantee the number will stay the same if a bundle is re-tried.
2) Avoid adding threading, blocking, queues or pooling. Let Dataflow handle concurrency
3) Make sure processElement is thread-safe

I have links to internal incident ticket and Google Case but will not post them here.

nevillelyh:
@welvet can you verify if this works?

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
@@ -52,6 +52,7 @@ public abstract class BaseAsyncDoFn<InputT, OutputT, ResourceT, FutureT>
 
   @StartBundle
   public void startBundle() {
+    waitForFutures();
     futures.clear();
     results.clear();
     errors.clear();
@@ -59,6 +60,11 @@ public abstract class BaseAsyncDoFn<InputT, OutputT, ResourceT, FutureT>
 
   @FinishBundle
   public void finishBundle(FinishBundleContext c) {
+    waitForFutures();
+    flush(c);
+  }
+
+  private void waitForFutures() {
     if (!futures.isEmpty()) {
       try {
         waitForFutures(futures.values());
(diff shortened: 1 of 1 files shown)


scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
```diff
@@ -52,13 +52,19 @@ protected abstract FutureT addCallback(FutureT future,
 
   @StartBundle
   public void startBundle() {
+    waitForFutures();
```
nevillelyh: On second thought I'm not sure this would matter. Not waiting for a pending future will not cause deadlock whereas waiting for one that never completes will.

nevillelyh:
Closing as this will not fix the deadlock.

Issue #1183, opened 2018-05-25
Profile and optimize transforms


nevillelyh:
There're a lot of transforms on `SCollection` and `PairSCollectionFunctions`. Some of them may have big performance overhead and we should figure out a way to profile them thoroughly and optimize.

nevillelyh:
WIP branch: https://github.com/spotify/scio/tree/neville/flame
`cd scio-bench && ./init.sh && ./micro-bench.sh` to run. Load the output SVG files in browser to view the graphs.
Pretty sure this is not the best way to profile though since the JVM OOMs very easily even with 32GB heap and there's a fair amount of overhead from Beam runner.

jto:
Would it be possible to write microbench with JMH for those ?

nevillelyh:
Not sure how effective that'll be since it depends on how frequent different runners invoke the various methods in a `DoFn` or its variations.
Also to invoke methods of DoFn, CombineDoFn, etc. without creating a `DirectRunner` (which has pretty big overhead), we'll have to call the underlying methods manually?

nevillelyh:
Hacked together a super simple benchmark for `CombineFn` wrappers:
https://github.com/spotify/scio/blob/neville/bench/scio-jmh/src/test/scala/com/spotify/scio/util/FunctionsBenchmark.scala

```
[info] FunctionsBenchmark.benchAggregate  avgt   10  477628.990 ± 364749.568  ns/op
[info] FunctionsBenchmark.benchCombine    avgt   10  443416.943 ±   8332.099  ns/op
[info] FunctionsBenchmark.benchMonoid     avgt   10  307335.158 ±  17936.943  ns/op
[info] FunctionsBenchmark.benchReduce     avgt   10  482259.889 ±  42456.045  ns/op
[info] FunctionsBenchmark.benchSemigroup  avgt   10  395404.521 ±  36882.540  ns/op
```

Will keep iterating.

nevillelyh:
Looks like some `mergeAccumulators` implementations are doing `map/flatMap+reduce` that could be replaced with a mutable version to speed things up.

`val combined = accumulators.asScala.map(fold).reduce(c)`
`val combined = accumulators.asScala.flatMap(foldOption).reduce(mc)`

jto:
@kanterov may have insights to share on this ?

jbx:
blocked by Algebird release

nevillelyh:
Closing this for now. Opened #1330 for tracking future work.

PR #1268, opened 2018-08-02
reduce copies in CombineFn #1183

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -74,7 +74,7 @@ private[scio] object Functions {
       accumulator._2.asScala.foldLeft(accumulator._1)(s)
 
     override def mergeAccumulators(accumulators: JIterable[(U, JList[T])]): (U, JList[T]) = {
-      val combined = accumulators.asScala.map(fold).reduce(c)
+      val combined = accumulators.asScala.iterator.map(fold).reduce(c)
```
regadas: we can get `AbstractIterator` as well by getting the `iterator` from the java collection, i.e `iterator.asScala`. (which is also the strategy that flink uses). I know that use cases may vary ... but i would probably make: `before converting to scala get the iterator` a rule of thumb.

PR #1272, opened 2018-08-03
Add a warning on update when an older version of a lib is selected

jto:
Looks like this:

```
[info] Set current project to scio-core (in build file:/Users/julient/Documents/Spotify/scio/)
[warn] +- com.google.code.findbugs:jsr305:3.0.1 was selected but 3.0.2 should be used instead
[warn] +- com.google.protobuf:protobuf-java:3.2.0 was selected but 3.4.0 should be used instead
[warn] +- com.google.code.findbugs:jsr305:3.0.1 was selected but 3.0.2 should be used instead
[warn] +- com.google.protobuf:protobuf-java:3.2.0 was selected but 3.4.0 should be used instead
```

(related to https://github.com/spotify/scio/pull/1265)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -182,7 +182,42 @@ val commonSettings = Sonatype.sonatypeSettings ++ assemblySettings ++ Seq(
   },
 
   buildInfoKeys := Seq[BuildInfoKey](scalaVersion, version, "beamVersion" -> beamVersion),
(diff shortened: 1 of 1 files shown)


jto:
Closing the PR since the benefit / cost ratio is not convicing enough

Issue #1164, opened 2018-05-15
Create an example of using writeDynamic


ravwojdyla:
`FileIO#writeDynamic` comes up every now and then, would be good to have an example how to use it. It also seems like it's not very well documented in Beam itself. An example could be to write data per key to GCS with some smartish filenames.

LukeDefeo:
```   
 def saveToFilesDynamic(collection: SCollection[JustDataEvent],
                               basePath: String,
                               shards: Int = 1,
                               compression: Compression = Compression.GZIP,
                               transformName: String = "Save text files"): Future[Tap[JustDataEvent]] =
        {
            val fileNaming = new SerializableFunction[DayPartition, Write.FileNaming]
            {
                override def apply(partition: DayPartition): Write.FileNaming =
                    (window: BoundedWindow, pane: PaneInfo, numShards: Int, shardIndex: Int, compression: Compression) =>
                        s"${Gcs.path(partition)}${window.toString}$shardIndex-of-$numShards.jsonl${compression.getSuggestedSuffix}"
            }

            val transform =
                FileIO
                    .writeDynamic[DayPartition, JustDataEvent]()
                    .to(basePath)
                    .by((input: JustDataEvent) => input.toDayPartition())
                    .via(Contextful.fn[JustDataEvent, String]((element: JustDataEvent) => compact(render(element.toJson()))), TextIO.sink())
                    .withNaming(fileNaming)
                    .withCompression(compression)
                    .withDestinationCoder(new DayPartitionCoder())
                    .withNumShards(shards)

            collection.saveAsCustomOutput(transformName, transform)
        }
    }
```

Here is a rough thing i cooked up. You need an intermediate type which defines a destination. A fn to go from the element type to this destination case class. You can also define exactly how it is named. Since we need to replicate sparks partitioning scheme of year=2018/month=10.. we encapsulate this in a case class and then in the name mapping for each destination we emit this path. Within the partition you can use the windows timings for the file name as i have done.

One issue you will face is that the intermediate type for partitioning, in my case the case class DayPartition will need to have a manually created coder. For some reason scios magic with avro atomic coder doesnt kick in. I dont know enough to comment as to why not, maybe @nevillelyh will be able to say

PR #1278, opened 2018-08-06
Add example of FileIO#writeDynamic (#1164)

ClaireMcGinty:
`TextIO` couldn't support this use case so no unit test right now. I ran the job on Dataflow using the King Lear text and the output was as expected.

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2018 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.ContextAndArgs
+import com.spotify.scio.examples.common.ExampleData
+import org.apache.beam.sdk.coders.StringUtf8Coder
+import org.apache.beam.sdk.io.{FileIO, TextIO}
+import org.apache.beam.sdk.transforms.Contextful
+
+// `sbt runMain "com.spotify.scio.examples.extra.WriteDynamicExample
+// --project=[PROJECT] --runner=DataflowRunner --zone=[ZONE]
+// --input=gs://apache-beam-samples/shakespeare/kinglear.txt
+// --output=[OUTPUT]"
+object WriteDynamicExample {
+  case class LinesPerCharacter(name: String, lines: Long)
+
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    // Group records according to first letter of the character's name
+    // All characters starting with letter A will share an output path, etc
+    val dynamicOutput: FileIO.Write[String, LinesPerCharacter] = FileIO
+      .writeDynamic[String, LinesPerCharacter]()
+      .by((linesPerCharacter: LinesPerCharacter) =>
+        linesPerCharacter.name.charAt(0).toString
+      )
+      .withNaming((characterFirstLetter: String) =>
+        FileIO.Write.defaultNaming(s"characters-starting-with-$characterFirstLetter", ".txt")
+      )
+      .withDestinationCoder(StringUtf8Coder.of())
+      .withNumShards(1) // Since input is small, restrict to one file per bucket
+      .via(
+        Contextful.fn[LinesPerCharacter, String]((lines: LinesPerCharacter) => lines.toString),
+        TextIO.sink() // Serialize LinesPerCharacter records as Strings
+      )
+      .to(args("output"))
+
+    // Compute # of times each character speaks in the King Lear text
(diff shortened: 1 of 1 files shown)


scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio.ContextAndArgs
+import com.spotify.scio.examples.common.ExampleData
+import org.apache.beam.sdk.coders.StringUtf8Coder
+import org.apache.beam.sdk.io.{FileIO, TextIO}
+import org.apache.beam.sdk.transforms.Contextful
+
+// `sbt runMain "com.spotify.scio.examples.extra.WriteDynamicExample
```
nevillelyh: Add "Example" and "Usage" likes like here?
https://github.com/spotify/scio/blob/master/scio-examples/src/main/scala/com/spotify/scio/examples/WordCount.scala#L18

Also these are in the generated socco web pages. So move all of them to above `package`? So that the generated page formats nicely.

http://spotify.github.io/scio/examples/WordCount.scala.html

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+      )
+      .withDestinationCoder(StringUtf8Coder.of())
+      .withNumShards(1) // Since input is small, restrict to one file per bucket
+      .via(
+        Contextful.fn[LinesPerCharacter, String]((lines: LinesPerCharacter) => lines.toString),
+        TextIO.sink() // Serialize LinesPerCharacter records as Strings
+      )
+      .to(args("output"))
+
+    // Compute # of times each character speaks in the King Lear text
```
nevillelyh: Make sure `#` formats properly with socco plugin's markdown?
See https://github.com/spotify/scio/blob/master/scripts/make-site.sh#L5

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+    sc.textFile(args.getOrElse("input", ExampleData.KING_LEAR))
+      .flatMap(line => {
```
nevillelyh: For multi-line lambdas, use `flatMap { x =>\n...` style to be consistent with the other examples?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+        val characterAndFirstLine = line.split("\t").filter(_.nonEmpty)
```
nevillelyh: This logic (`name\tline`) is a little hard to decipher without looking at the text. Add some explanation?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+  case class LinesPerCharacter(name: String, lines: Long)
+
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    // Group records according to first letter of the character's name
+    // All characters starting with letter A will share an output path, etc
+    val dynamicOutput: FileIO.Write[String, LinesPerCharacter] = FileIO
+      .writeDynamic[String, LinesPerCharacter]()
+      .by(Contextful.fn[LinesPerCharacter, String]((linesPerCharacter: LinesPerCharacter) =>
```
nevillelyh: Guess we need the explicit lambda arg type here for scala 2.11 compatibility?

regadas: yup

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+        linesPerCharacter.name.charAt(0).toString.toUpperCase
+      ))
+      .withNaming((characterFirstLetter: String) =>
+        FileIO.Write.defaultNaming(s"characters-starting-with-$characterFirstLetter", ".txt")
+      )
+      .withDestinationCoder(StringUtf8Coder.of())
+      .withNumShards(1) // Since input is small, restrict to one file per bucket
+      .via(
+        Contextful.fn[LinesPerCharacter, String]((lines: LinesPerCharacter) => lines.toString),
```
regadas: same here `new SerializableFunction {...}`

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
```diff
@@ -0,0 +1,67 @@
...
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    // Group records according to first letter of the character's name
+    // All characters starting with letter A will share an output path, etc
+    val dynamicOutput: FileIO.Write[String, LinesPerCharacter] = FileIO
+      .writeDynamic[String, LinesPerCharacter]()
+      .by(Contextful.fn[LinesPerCharacter, String]((linesPerCharacter: LinesPerCharacter) =>
+        linesPerCharacter.name.charAt(0).toString.toUpperCase
+      ))
+      .withNaming((characterFirstLetter: String) =>
```
regadas: same here `new SerializableFunction {...}`

scio-examples/src/main/scala/com/spotify/scio/examples/extra/WriteDynamicExample.scala
regadas: or maybe write it like this

```
        line.split("\t").toList match {
          case name::text::restOfList => Some(name)
          case _ => None
        }
```

PR #1281, opened 2018-08-07
add StatefulExample, fix #448

### LINKED ISSUE #448: Validate/test stateful processing
Validate/test stateful processing. Check if we need/should make it easier in scio api.
https://beam.apache.org/blog/2017/02/13/stateful-processing.html

nevillelyh:


nevillelyh:
Right now `StateSpecs` has `bag,combining,map,set,value` state types. One way of reducing boilerplate, without making the API too complex, is to add specialized versions of each state types, so that the API may look like this:

```scala
data.mapWithState[ValueState[Int]] { case ((k, v), state) =>
  ...
}
```

jto:
Looks good enough for now IMO. I don't think we really have to put too much effort on this unless ou users explicitly ask for it.

In the future, It may possible to unify the APIs for `SideInput` / `SideOutput` and `ValueState` given that their usage are fairly similar.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/StatefulExample.scala
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2016 Spotify AB.
```
regadas: 🎨 header is a bit off 😄

nevillelyh: Is there a way to automate header updates?
Also I need to update my IntelliJ settings...

PR #1287, opened 2018-08-09
add bump_scio.sh helper script

nevillelyh:


nevillelyh:
It's mainly to help other repos we own that depends on scio, like scio-contrib, ratatool, dbeam, etc.
Also `build.sbt` can be parameterized as a command line argument.

nevillelyh:
Updated this to bump versions in downstream repos and submit PRs automatically. A few caveats:

- it still needs to be run manually, most probably as part of the release process
- user running the script needs push permission to all downstream repos
- it requires [hub](https://github.com/github/hub) to be installed

nevillelyh:
@labianchin you're a bash ninja! Thanks!

PR #1290, opened 2018-08-10
Improve benchmark: Force shuffle and wrap elems in case class

regadas:


scio-test/src/it/scala/com/spotify/ScioBenchmark.scala
```diff
@@ -260,16 +260,22 @@ object ScioBenchmark {
   private val K = 1000
   private val numPartitions = 100
 
-  private def randomUUIDs(sc: ScioContext, n: Long): SCollection[String] =
-    sc.parallelize(Seq.fill(numPartitions)(n / numPartitions))
+  final case class Elem[T](elem: T)
+
+  private def randomUUIDs(sc: ScioContext, n: Long): SCollection[Elem[String]] =
+    sc.parallelize((1 to numPartitions / 10).map(_ => Seq.fill(numPartitions)(n / numPartitions)))
```
nevillelyh: Instead of hard coding 10 here, should it be a multiple of `numWorkers` here? https://github.com/spotify/scio/pull/1290/files#diff-65c72767e390d7fa44ad2f4d472b681dR47

scio-test/src/it/scala/com/spotify/ScioBenchmark.scala
```diff
@@ -259,10 +259,16 @@ object ScioBenchmark {
...
   private val M = 1000000
   private val K = 1000
+
   private final case class Elem[T](elem: T)
 
-  private def partitions(n: Long, numPartitions: Int = 100): Iterable[Iterable[Long]] =
-    (1 to numPartitions).map(_ => Seq.fill((n / numPartitions).toInt)(n / numPartitions))
+  private def partitions(
+      n: Long,
+      numPartitions: Int = Runtime.getRuntime
```
regadas: @nevillelyh what do you think of this?

scio-test/src/it/scala/com/spotify/ScioBenchmark.scala
```diff
@@ -258,18 +259,31 @@ object ScioBenchmark {
...
-  private val numPartitions = 100
 
-  private def randomUUIDs(sc: ScioContext, n: Long): SCollection[String] =
-    sc.parallelize(Seq.fill(numPartitions)(n / numPartitions))
+  private final case class Elem[T](elem: T)
+
+  private def partitions(
+      n: Long,
+      numPartitions: Int = Runtime.getRuntime
+        .availableProcessors() * numOfWorkers): Iterable[Iterable[Long]] =
```
nevillelyh: That's the number of processors in `main` not in a worker though.

regadas: yeah you are right 🤦‍♂️

PR #1291, opened 2018-08-10
add FunctionsBenchmark #1183

nevillelyh:


build.sbt
```diff
@@ -682,6 +682,7 @@ lazy val scioJmh: Project = Project(
   classDirectory in Jmh := (classDirectory in Test).value,
   dependencyClasspath in Jmh := (dependencyClasspath in Test).value,
   libraryDependencies ++= Seq(
+    "org.hamcrest" % "hamcrest-all" % hamcrestVersion,
```
regadas: @nevillelyh is it used? or did I missed something?

scio-jmh/src/test/scala/com/spotify/scio/util/FunctionsBenchmark.scala
```diff
@@ -0,0 +1,57 @@
...
+import scala.collection.JavaConverters._
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Thread)
+class FunctionsBenchmark {
+
+  type T = Set[Int]
+
+  val input = Lists.newArrayList((1 to 100).map(Set(_)).asJava)
```
regadas: @nevillelyh I no this is handy but can we replace this with pure scala/java ... one less dep on guava 😄

jto: `new ArrayList((1 to 100).map(Set(_)).asJava)`

build.sbt
nevillelyh: It's needed by `CombineFnTester`.

Issue #890, opened 2017-10-18
Investigate and create example for streaming job with periodically updating side input


nevillelyh:
The side input could be from a GCS file, a BQ query, etc. but updating at some interval. We might need to use a `GenerateSequence` transform + windowing to periodically refresh the content.

We should add this as some syntactic sugar and/or example.

Reference: https://github.com/spotify/scio/issues/865#issuecomment-333980759

nevillelyh:
@regadas maybe something you'll be interested.

nevillelyh:
Played a bit with the idea in https://github.com/spotify/scio/tree/neville/streaming but haven't got very far.

regadas:
@nevillelyh yeah I can take it

nevillelyh:
#895 #897 are somewhat related.

andrisnoko:
@regadas are you still looking into this? If not, I can take it.

regadas:
@andrisnoko yeah you can take it! completely forgot about this one 😛

jbx:
@andrisnoko pokety poke. Still planning to look into this?

bbhoss:
The example is very helpful, but it doesn't show pulling from an external data source like GCS/BQ. Is there some way to trigger fetching the input with this that I am missing? Seems like we'd need to use the BQ/GCS APIs directly in a map.

gregorath:
@bbhoss did you manage to find how to do this with an external data source?

PR #1295, opened 2018-08-13
#890 add example of streaming job with refreshing side input

ClaireMcGinty:
because the input is pub/sub subscription and the side input is randomly generated, I didn't end up writing a test for this. I could re-do it with a bounded input but I think the refreshing side input behavior would still be hard to unit test

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/RefreshingSideInputExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/RefreshingSideInputExample.scala
@@ -22,9 +22,8 @@
 // --project=[PROJECT] --runner=[RUNNER] --zone=[ZONE] --input=[PUBSUB_SUBSCRIPTION]"`
 package com.spotify.scio.examples.extra
 
-import com.spotify.scio.ScioContext
+import com.spotify.scio._
 import com.spotify.scio.values.WindowOptions
-import org.apache.beam.examples.common.ExampleOptions
 import org.apache.beam.sdk.io.GenerateSequence
 import org.apache.beam.sdk.options.StreamingOptions
 import org.apache.beam.sdk.transforms.windowing.Window.ClosingBehavior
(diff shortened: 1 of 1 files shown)


scio-examples/src/main/scala/com/spotify/scio/examples/extra/RefreshingSideInputExample.scala
```diff
@@ -0,0 +1,107 @@
...
+          allowedLateness = Duration.standardSeconds(5))
+      )
+      .map(_ => random.nextInt(100))
+      .asListSideInput
+
+    // Sample pubsub topic modeling lottery tickets as a comma-separated list of numbers.
+    // For example, a message might contain the string "10,7,3"
+    sc.pubsubSubscription[String](args("input"))
+      .flatMap(toLotteryTicket)
+      .timestampBy(_ => Instant.now())
```
nevillelyh: Probably don't need this and can just use timestamp from pubsub.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/RefreshingSideInputExample.scala
```diff
@@ -0,0 +1,107 @@
...
+        duration = Duration.standardSeconds(10),
+        offset = Duration.ZERO,
+        options = WindowOptions(
+          trigger = Repeatedly.forever(AfterPane.elementCountAtLeast(5)),
+          accumulationMode = AccumulationMode.DISCARDING_FIRED_PANES,
+          closingBehavior = ClosingBehavior.FIRE_IF_NON_EMPTY,
+          allowedLateness = Duration.standardSeconds(5))
+      )
+      .map(_ => random.nextInt(100))
+      .asListSideInput
```
nevillelyh: We can use `asSingletonSideInput(default)` if it's guaranteed to have 1 element per window, and fallback to `default` if it's empty, i.e. beginning of the session when window is incomplete?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/RefreshingSideInputExample.scala
```diff
@@ -0,0 +1,107 @@
...
+      .withFixedWindows(
+        duration = Duration.standardSeconds(10),
+        offset = Duration.ZERO,
+        options = WindowOptions(
+          trigger = Repeatedly.forever(AfterPane.elementCountAtLeast(5)),
+          accumulationMode = AccumulationMode.DISCARDING_FIRED_PANES,
+          closingBehavior = ClosingBehavior.FIRE_IF_NON_EMPTY,
+          allowedLateness = Duration.standardSeconds(5))
+      )
+      .map(_ => random.nextInt(100))
```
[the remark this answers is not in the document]
nevillelyh: Hmm... I misunderstood the original code. So the numbers in `Seq[Int]` aren't ordered like typical lottery numbers, but independent, so matching any one of them wins the lottery?

I still think it's clearer to generate 1 event per refresh window (like a pulse), and `.map(_ => Seq[Int])` to mimic a sequence of lottery numbers.

PR #1298, opened 2018-08-14
Add method to read examples directly from TFRecord files

andrewsmartin:


andrewsmartin:
@ClaireMcGinty I don't think they're really used internally, but if you'd like I can add a helper for that too for the sake of completeness.

Issue #1175, opened 2018-05-22
Remove BigDecimalSerializer after next release of chill


andrewsmartin:
Submitted upstream: https://github.com/twitter/chill/pull/312

ClaireMcGinty:
Fixed in [#1301](uhttps://github.com/spotify/scio/pull/1301l)

PR #1301, opened 2018-08-15
#1175 bump chill to 0.9.3, remove patched BigDecimalSerializer

ClaireMcGinty:
the new BigDecimal serializer has [private access in chill-scala](https://github.com/twitter/chill/blob/e774819966ea29c4d2ebf1ac1539fe4a300fd21d/chill-scala/src/main/scala/com/twitter/chill/BigDecimalSerializer.scala) and is accessed via the [AllScalaRegistrar](https://github.com/twitter/chill/blob/b5dc7ee7c89a9245ceb18f10f290d75b809d0381/chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala#L253)


scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -245,6 +244,7 @@ private[scio] object KryoAtomicCoder {
         k.setReferences(options.referenceTracking)
         k.setRegistrationRequired(options.registrationRequired)
 
+        new AllScalaRegistrar()(k)
```
regadas: @ClaireMcGinty I think there's no need to include this. `AllScalaRegistrar` is already registered as part of the `ScalaKryoInstantiator`. (which is the one we use)

PR #1307, opened 2018-08-16
port JsonIO to NIO

nevillelyh:
@jto looks like I have to use `Either[Decoder, T]` for both read & write. The test code is a bit ugly with the extra conversion but not sure how to work around that.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/json/nio/JsonIO.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/json/nio/JsonIO.scala
@@ -18,7 +18,6 @@
 package com.spotify.scio.extra.json.nio
 
 import com.spotify.scio.ScioContext
-import com.spotify.scio.extra.json.DecodeError
 import com.spotify.scio.io.Tap
 import com.spotify.scio.nio.{ScioIO, TextIO}
 import com.spotify.scio.util.ScioUtil
@@ -33,7 +32,7 @@ import scala.reflect.ClassTag
 import scala.util.{Left, Right}
 
 case class JsonIO[T: ClassTag : Encoder : Decoder](path: String)
-  extends ScioIO[Either[DecodeError, T]] {
+  extends ScioIO[T] {
 
   case class WriteParams(suffix: String = ".json",
                          numShards: Int = 0,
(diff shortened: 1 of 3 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/json/nio/JsonIO.scala
```diff
@@ -0,0 +1,93 @@
...
+import io.circe.parser._
+import io.circe.syntax._
+import org.apache.beam.sdk.io.{Compression, FileBasedSink, TextIO => BTextIO}
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.util.{Left, Right}
+
+case class JsonIO[T: ClassTag : Encoder : Decoder](path: String)
+  extends ScioIO[Either[DecodeError, T]] {
```
regadas: would it be worth it to consider other `ADT` instead of plain `Either` to give better semantic? i.e:

```scala
sealed trait JsonSerDe[+T]
final case class Error(error: io.circe.Error, input: String) extends JsonSerDe[Nothing]
final case class Elem[T](e: T) extends JsonSerDe[T]
```

regadas:
I like the 2 "effect" type params approach! 👍 However, could consider the same effect type for reads and writes?

```scala
trait ScioIO[T] {
  ... 
  def read(sc: ScioContext, params: ReadP): SCollection[Either[Throwable, T]]
  def write(data: SCollection[T], params: WriteP): Future[Tap[Either[Throwable, T]]
  def tap(read: ReadP): Tap[T]
  ...
}
```

@jto @nevillelyh @syodage WDYT?

nevillelyh:
Actually for this case we don't need different type params for read/write. I'm using the same `Error[DecodeError, T]`. The ugliness is from mock data in `JobTest` and tap.

When reading we have:
`ScioContext#jsonFile: SCollection[Either[E, T]]` but when testing one has to call `JobTest#input(JsonIO[T])(xs: Seq[Either[E, T]]`. This is technically correct (in case the user wants to test error handling when reading JSONs) but feels cumbersome since most cases we just want to feed `Seq[T]`.

When writing we have:
`SCollection[T]#saveAsJsonFile: Future[Tap[Either[E, T]]`. Mapping from `T => Right[T])` is inside the method. We need `Either` in `Tap[_]` in case there's error reading it back, or opening taps not saved in Scio. Again the ugliness is in `JobTest`, since one has to call `JobTest#output(JsonIO[T]) { _ matchInAnyOrder (expected: Seq[Either[E, T]]) }` since again most cases we just want to check against `Seq[T]`.

nevillelyh:
Or we could get rid of `Either` and hard fail on any errors, since that's the behavior of `Avro/Proto/Object` IOs right now. This will make it a lot simpler.

jto:
@nevillelyh yeah that'd be consistent with the rest of the API

Issue #1311, opened 2018-08-20
saveAsJdbc does not have a way to configure the batch size


ccarpita:
Configuring the batch size for streaming inserts can be a potentially useful strategy for managing deadlock frequency at the expense of insert performance.

Current snippet of code in scio-jdbc:
```scala
var transform = jio.JdbcIO.write[T]()
          .withDataSourceConfiguration(getDataSourceConfig(writeOptions.connectionOptions))
          .withStatement(writeOptions.statement)
```

Proposed feature change:
```scala
var transform = jio.JdbcIO.write[T]()
          .withDataSourceConfiguration(getDataSourceConfig(writeOptions.connectionOptions))
          .withStatement(writeOptions.statement)
          .withBatchSize(writeOptions.batchSize)
```

This would require a batchSize option on writeOptions as well, perhaps `Option[Int]: None`.

syodage:
Fixed with https://github.com/spotify/scio/pull/1314

PR #1314, opened 2018-08-21
Allow users to configure fetch and batch size for jdbc reads and writes

syodage:
Fixes https://github.com/spotify/scio/issues/1311

### DIFF
--- a/scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
+++ b/scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
@@ -65,7 +65,8 @@ package object jdbc {
   case class JdbcReadOptions[T](connectionOptions: JdbcConnectionOptions,
                                 query: String,
                                 statementPreparator: PreparedStatement => Unit = null,
-                                rowMapper: ResultSet => T) extends JdbcIoOptions
+                                rowMapper: ResultSet => T,
+                                fetchSize: Option[Int] = None) extends JdbcIoOptions
 
   /**
    * Options for writing to a JDBC source.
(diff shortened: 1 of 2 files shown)


syodage:
@nevillelyh Done!

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
```diff
@@ -65,7 +65,8 @@ package object jdbc {
   case class JdbcReadOptions[T](connectionOptions: JdbcConnectionOptions,
                                 query: String,
                                 statementPreparator: PreparedStatement => Unit = null,
-                                rowMapper: ResultSet => T) extends JdbcIoOptions
+                                rowMapper: ResultSet => T,
+                                fetchSize: Option[Int] = None) extends JdbcIoOptions
```
nevillelyh: Instead of requiring the user to wrap `Int` in `Some` here, what about use `Int` directly with a default of `-1`, which skips setting the fetch size?

syodage: +1, yes agree.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/package.scala
```diff
@@ -28,25 +28,28 @@ import scala.concurrent.Future
...
+  * {{{
+  * import com.spotify.scio.jdbc._
+  * }}}
+  */
 package object jdbc {
 
   type JdbcIO[T] = jdbc.nio.JdbcIO[T]
   val JdbcIO = jdbc.nio.JdbcIO
 
+  private[jdbc] val USE_BEAM_DEFAULT_BATCH_SIZE = -1L
```
nevillelyh: I wouldn't go crazy as de-magic-fy these since `-1` is a pretty common pattern but that's just me ;)

syodage: thought it is more informative what actually -1 means here.

Issue #1303, opened 2018-08-15
BigQueryType.fromTableRow throws NPE without explanation for nullable fields without an Option


keithmcneill:
When converting from a TableRow to a case class and if the TableRow is missing a field that isn't an Option type, fromTableRow will throw an NPE without any explanation. The Following shows the problem:

```
class MissingFieldTest extends PipelineSpec {

  case class Artist(artist: String, artist2: String)

  "FromJSon should work" should "work" in {
    val goodJson =
      """
        |{
        |    "artist" : "spotify:artist",
        |    "artist2" : "spotify:artist2"
        |}
      """.stripMargin

    val badJson =
      """
        |{
        |    "artist" : "spotify:artist"
        |}
      """.stripMargin

    val goodtr = new JacksonFactory().fromString(goodJson, classOf[TableRow])
    val goodartist: Artist = BigQueryType.fromTableRow[Artist](goodtr)

    val badtr = new JacksonFactory().fromString(badJson, classOf[TableRow])
    val artist: Artist = BigQueryType.fromTableRow[Artist](badtr)
  }
}
```

Gives an exception like:
```
java.lang.NullPointerException
    at com.spotify.MissingVectorTest$$anonfun$1$$anonfun$3.apply(MissingFieldTest.scala:36)
    at com.spotify.MissingVectorTest$$anonfun$1$$anonfun$3.apply(MissingFieldTest.scala:36)
    at com.spotify.MissingVectorTest$$anonfun$1.apply$mcV$sp(MissingFieldTest.scala:36)
    at com.spotify.MissingVectorTest$$anonfun$1.apply(MissingFieldTest.scala:16)
    at com.spotify.MissingVectorTest$$anonfun$1.apply(MissingFieldTest.scala:16)
    ....
```

An error message beyond an NPE would be nice, since it is kinda hard to figure out what is going on.

keithmcneill:
To avoid a lot if !null checks for each and every field in each and every row that passes through the code you could wrap the serialization of all the fields in 1 big try/catch(NPE) and only run through the null checks if the NPE fires.  

Kinda seems like pre-mature optimization, but that is a lot of fields within a lot of rows over time.

nevillelyh:
But then we won't be able to distinguish NPEs caused by null fields or other genuine bugs in the code. Also stacktrace won't pinpoint exact location of the generated code so we won't know which field caused the NPE, kinda defeats the purpose.

We can try adding the check & message at the leaf level and micro-benchmark first.

PR #1319, opened 2018-08-21
Fix tablerow npe to have more info

anne-decusatis:
fixes https://github.com/spotify/scio/issues/1303

still needs some tests and stuff :) thanks for pairing with me @nevillelyh


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -210,7 +210,14 @@ private[types] object ConverterProvider {
       val name = symbol.name.toString
       val tpe = symbol.asMethod.returnType
 
-      val tree = q"$fn.get($name)"
+
```
nevillelyh: Actually `tree` can be null if the type is `Option`, handled in `def option()` above. We should move this null check into the `else` branch, before `cast`.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -210,7 +210,14 @@ private[types] object ConverterProvider {
       val name = symbol.name.toString
       val tpe = symbol.asMethod.returnType
 
-      val tree = q"$fn.get($name)"
+      
+      val tree = q"""{
+            val v = $fn.get($name)
+            if(v == null) {
+                throw new NullPointerException("null field in non-option field " + $name)
```
nevillelyh: Change the message to `null value in REQUIRED field $name`?

scio-jmh/src/test/scala/com/spotify/scio/jmh/BigQueryBenchmark.scala
```diff
@@ -0,0 +1,48 @@
+/*
```
nevillelyh: Actually we shouldn't check this file in since it doesn't benchmark both versions and show the difference. As long as we can use it to verify that the change doesn't degrade performance.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
nevillelyh: We need to check it in `else if` also since `REPEATED` fields aren't supposed to be null.

nevillelyh:
As for testing, there's a `ConverterProviderSpec` but that one uses scalacheck to verify roundtrips and is not suitable for this particular case. We can add a `ConverterProviderTest` similar to `TypeProviderTest` and verify that it handles `NULLABLE/REQUIRED/REPEATED` correctly with or without null value.

scio-jmh/src/test/scala/com/spotify/scio/jmh/BigQueryBenchmark.scala
```diff
@@ -0,0 +1,48 @@
...
+import com.google.common.collect.AbstractIterator
+import com.google.common.collect.Lists
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.collection.JavaConverters._
+
+import com.spotify.scio.bigquery.types._
+
+// scalastyle:off number.of.methods
```
regadas: in case this file gets in, this can be removed.

scio-jmh/src/test/scala/com/spotify/scio/jmh/BigQueryBenchmark.scala
```diff
@@ -0,0 +1,48 @@
...
+@State(Scope.Thread)
+class BigQueryBenchmark {
+  import BigQueryBenchmark.Simple
+  
+  val s = Simple.toTableRow(Simple("test"))
+  
+  @Benchmark def bq(): Simple = Simple.fromTableRow(s)
+
+}
+// scalastyle:on number.of.methods
```
regadas: remove as well

nevillelyh:
Closing this and use #1332 instead.

Issue #1326, opened 2018-08-27
KryoState per thread is not Garbage Collected leading to memory leaks


anish749:
The `KryoStateMap` in `KryoAtomicCoder` keeps a reference of the `KryoState` for every thread and instance. For streaming jobs, when new threads get created and older threads are killed, the `KryoState` is not garbage collected. This leads to the memory consumption increasing and a GC thrashing occurs. When running on Dataflow, it restarts the JVM, and continues the job. This restart leads to an increase in system lag / data watermark lag. In one of the jobs I have noticed this lag to be as high as 13/14mins. 

I am not sure how to change the ExecutorService to allow ThreadPooling and not have the old threads die out.

I tried to use a `ThreadLocal[HashMap]` to store the KryoState which seems to solve the problem and allow the objects to be garbage collected once the Thread itself dies. The memory usage is constant which confirms garbage collections.

I will create a PR showing the fix. (Edit - #1327 )

I can share heap dumps if required for debugging.

PR #1327, opened 2018-08-27
Use thread local hash maps for storing Kryo State

anish749:
Use ThreadLocal[Map] for storing `KryoState` for to allow garbage collection when a thread is killed.

Fixes https://github.com/spotify/scio/issues/1326


anish749:
Stackdriver Agent Memory usage on a streaming job (orange is free, blue is used) :
screen shot 2018-08-27 at 18 09 19

HeapDump from memory analyser:
screen shot 2018-08-27 at 18 05 44

PR #1328, opened 2018-08-28
DirectRunner with single thread on testing

labianchin:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -180,8 +180,11 @@ object ScioContext {
 
   /** Create a new [[ScioContext]] instance for testing. */
   def forTest(): ScioContext = {
+    // when running unit tests force 1 thread per job,
+    // otherwise each JobTest spawns as many threads as CPU cores available
+    // this is intended to make tests faster
     val opts = PipelineOptionsFactory
-      .fromArgs("--appName=" + TestUtil.newTestId())
+      .fromArgs("--targetParallelism=1", "--appName=" + TestUtil.newTestId())
```
nevillelyh: Some concurrency bugs won't be revealed with single thread though. OTOH it will make those tests less flaky.
Should we make this configurable? As an env var, `sys.props` or optional argument to `forTest` & `JobTest`.

labianchin:
By looking at builds https://circleci.com/gh/spotify/scio it seems this is making builds faster. Before moving forward with this I have some questions:
- Is this really worth or it can cause more confusions? I see the benefit of decreased build times improvement development productivity.
- Is there a more scientific approach to validate that this really decreases build times? Open for ideas.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
labianchin: > Should we make this configurable? As an env var, sys.props or optional argument to forTest & JobTest.

Good idea. I will think about that.

labianchin: So, I realized there are at least 3 places that we might need this logic. `ScioContext`, `JobTest` and [Flo's `ScioOperator`](https://github.com/spotify/flo/blob/master/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala#L64). I am trying to think if there is a way to simplify and/or avoid duplication.

Also, what do we want as default?

- If we choose `1`, this repo and all scio users will easily get faster and less flaky builds. But they might not notice concurrency bugs.
- If we make this optional, benefits are opt in, we need to instruct users on how to use it.

nevillelyh: Alternatively we can configure scalatest to run tests sequentially instead of parallel?
Not sure how much users really benefit from this given that most of our repos are smallish and we build on our own Jenkins/VMs which are more customizable. It'll probably benefit Scio more given the amount of tests and tiny CircleCI containers, but for that we can opt in.
@regadas thoughts?

flylo: If it were an optional arg in the context that would be nice. You should be able to run unit tests in a single-threaded environment so that you can get in-process step-through capabilities on methods that take SCollections as args. Not sure why you'd need single-threadedness when you're running in any context other than a debugger though.

labianchin: > If it were an optional arg in the context that would be nice. You should be able to run unit tests in a single-threaded environment so that you can get in-process step-through capabilities on methods that take SCollections as args.

Yes. I can look into that.

> Alternatively we can configure scalatest to run tests sequentially instead of parallel?
> Not sure why you'd need single-threadedness when you're running in any context other than a debugger though.

Scalatest already runs tests in parallel. I think this is good, because on unit tests that seems the right place to have parallelism. Scio tests generally do not need to run in parallel, as the data being tests is generally small (maybe except property based testing?). So I am in favor of parallel scalatest with single threaded scio test. We could also tests and compare to check what is best (I probably won't have time to help on that).

jto:
@labianchin What's the status on this. I personally don't have any string opinion other than as @nevillelyh pointed out the fact that it **could** hide concurrency issues. wdyt @regadas ?

nevillelyh:
We talked IRL and since it might be cumbersome to pass in options all the way through `JobTest`, what about a system property to set parallelism for all tests, and make it an opt-in?

labianchin:
> what about a system property to set parallelism for all tests, and make it an opt-in?

Yes. I can look into that. (I am also fine if somebody looks into it).

regadas:
@labianchin did you had time to have a look at this 😄. As an heads up, we are thinking in rethinking `JobTest` and if you don't have any more work on this we probably can factor in this change. wdyt?

labianchin:
Sorry I let this stale. I will close this for now, as I am no longer actively working on this.

> As an heads up, we are thinking in rethinking JobTest and if you don't have any more work on this we probably can factor in this change. wdyt?

Yes that would be great!

PR #1332, opened 2018-08-28
fix #1303, meaningful NPE message for nulls in BigQuery macro

### LINKED ISSUE #1303: BigQueryType.fromTableRow throws NPE without explanation for nullable fields without an Option

nevillelyh:


nevillelyh:
Did a micro benchmark

```scala
@BenchmarkMode(Array(Mode.AverageTime))
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
class BQBenchmark {

  val tr = TableRow("a" -> "hello")

  import BQBenchmark._
  @Benchmark def fromTableRow = Record.fromTableRow(tr)
}

object BQBenchmark {
  @BigQueryType.toTable
  case class Record(a: String)
}
```

Before the change:
```
[info] Benchmark                 Mode  Cnt   Score   Error  Units
[info] BQBenchmark.fromTableRow  avgt   30  84.408 ± 2.075  ns/op
```

After the change:
```
[info] Benchmark                 Mode  Cnt   Score    Error  Units
[info] BQBenchmark.fromTableRow  avgt   30  82.580 ± 0.936  ns/op
```

Don't think it'll affect performance much.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/ConverterProviderTest.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.scalatest._
+
+class ConverterProviderTest extends FlatSpec with Matchers {
+
+  import ConverterProviderTest._
+
+  // scalastyle:off no.whitespace.before.left.bracket
+  "ConverterProvider" should "throw NPE with meaningful message for null in REQUIRED field" in {
+    the [NullPointerException] thrownBy {
+      Required.fromTableRow(TableRow())
```
anne-decusatis: I didn't realize you could construct table rows this way in tests! Thanks for the example :)

nevillelyh: It's a syntactic suger in `com.spotify.scio.bigquery`.
https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/package.scala#L102

PR #1350, opened 2018-09-11
Small tweaks around core io and taps

regadas:


regadas:
@nevillelyh yup that's great idea! 👍

PR #1356, opened 2018-09-11
Make PCollection names always unique

### LINKED ISSUE #1323: After upgrading from scio 0.4.7 to 0.5.7, PCollections "both have unique_name" error
After upgrading from Scio 0.4.7 to 0.5.7, several dataflow jobs are unable to start due to an error of the form 

```INFO: Exception in thread "main" java.lang.IllegalArgumentException: pipeline: PCollections [GCS Path]/part-*1 and [GCS Path]/part-* both have unique_name "[GCS Path]/part-*"```

### PR DESCRIPTION
**Pull request by jto**
fix #1323

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -134,7 +134,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   def setCoder(coder: Coder[T]): SCollection[T] = context.wrap(internal.setCoder(coder))
 
   /** Assign a name to this SCollection. */
-  def setName(name: String): SCollection[T] = context.wrap(internal.setName(name))
+  def setName(name: String): SCollection[T] = context.wrap(internal.setName(context.mkUnique(name)))
 
   /**
    * Apply a [[org.apache.beam.sdk.transforms.PTransform PTransform]] and wrap the output in an
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -133,9 +133,6 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   /** Assign a Coder to this SCollection. */
   def setCoder(coder: Coder[T]): SCollection[T] = context.wrap(internal.setCoder(coder))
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: scio-test/src/it/scala/com/spotify/scio/ScioContextIT.scala [hunk 1]
```diff
@@ -54,4 +54,16 @@ class ScioContextIT extends FlatSpec with Matchers {
...
+    val options = PipelineOptionsFactory.create()
+    options.setRunner(classOf[DataflowRunner])
+    options.as(classOf[GcpOptions]).setProject(ItUtils.project)
+    val sc = ScioContext(options)
+
+    val s1 = sc.empty[(String, Int)]()
+    val s2 = sc.empty[(String, Double)]()
+    s1.join(s2)
+    sc.close()
+  }
```
nevillelyh: You should assert on something here?

jto: Not necessary. Beam throws an exception when the name is not unique which makes the test fail.

jto:
@regadas hummm that's a good question. Is there a non-painful way to check that ?

jto:
I ended up completely removing setName. All the tests are passing, and there's no name conflict when creating empty `SCollections`.  I couldn't find anything in Google's doc about `setName` on `PCollection`, nor a call to it in examples. Also searched through ghe for any pipeline using it, could only find 1. 
Given all that, it seems safe to get rid of it.

scio-test/src/it/scala/com/spotify/scio/ScioContextIT.scala
nevillelyh: There's a `noException should be thrownBy` matcher in scalatest, that could make this more explicit.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -133,9 +133,6 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   /** Assign a Coder to this SCollection. */
   def setCoder(coder: Coder[T]): SCollection[T] = context.wrap(internal.setCoder(coder))
 
-  /** Assign a name to this SCollection. */
-  def setName(name: String): SCollection[T] = context.wrap(internal.setName(context.mkUnique(name)))
```
regadas: 👍 😄

PR #1357, opened 2018-09-11
update dynamic file sink

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -960,8 +960,8 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @group output
    */
   def saveAsTextFile(path: String,
-                     suffix: String = ".txt",
                      numShards: Int = 0,
+                     suffix: String = ".txt",
```
nevillelyh: This is technically breaking but it'll be a breaking change anyway.
It's to make the API consistent with other `saveAs*` methods.

scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
```diff
@@ -123,27 +97,27 @@ package object dynamic {
...
-      saveAsTextFile(destinations, fileDestination.windowedWrites, fileDestination.numShards)
-    }
-
   }
 
+
+  private def writeDynamic[A](dest: FileDestinations,
+                              suffix: String,
+                              destinationFn: A => String): FileIO.Write[String, A] = {
+    FileIO.writeDynamic[String, A]()
```
regadas: @nevillelyh should we add a compression option?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: @nevillelyh I'm ok with that!

nevillelyh:
Think I need to rename these 2 `saveAs*` methods since they have default args and don't play well with the ones on `SCollection` with the same name.

PR #1359, opened 2018-09-11
fix #1025, cache schema in Avro SchemaProvider

### LINKED ISSUE #1025: SchemaProvider#schemaOf should cache
This may affect both Avro and BigQuery macros.
Right now `schemaOf` is in the generated converter's code which means they're called for every record being converted.

nevillelyh:


scio-avro/src/main/scala/com/spotify/scio/avro/types/SchemaProvider.scala
```diff
@@ -29,14 +29,17 @@ import scala.reflect.runtime.universe._
 
 private[types] object SchemaProvider {
 
+  private val m: scala.collection.concurrent.Map[Type, Schema] =
```
regadas: 🎨 `private[this]`

PR #1362, opened 2018-09-12
check and warn/fail chained joins

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -100,6 +100,8 @@ object SCollection {
   : PairSkewedSCollectionFunctions[K, V] =
     new PairSkewedSCollectionFunctions(s)
 
+  private[scio] case class State(postCoGroup: Boolean = false)
```
regadas: 🎨 `final`

nevillelyh: It's already private so no need to be so strict? OTOH we still don't have total consistency when it comes to details like this. Should we have one?

PR #1373, opened 2018-09-18
Tweak AvroCoders/JavaCoders scope

regadas:


nevillelyh:
I'm refactoring this and removing some code, so this won't be needed. Close?

PR #1386, opened 2018-09-19
Sparse left and right outer joins

anish749:
A version of sparse left and right outer joins, taken from the original sparse outer join implementation

In some of our pipeline we have a bunch of right outer joins where the RHS of the join is around 5% of the LHS, and but the RHS still doesn't fit in memory. As of now we extract the keys as a Side Input out of RHS which barely fits in memory and then we filter LHS using the keys, before doing the right outer join. Using a BF and sparse joins would help us scale well, which we would need to very soon, but we would need a right outer join variant. So I thought of adding the right and left versions.

 Also,
Can we have sparse multi joins?
The use case is like a small but not tiny dataset to be left joined with multiple very large right sides.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -325,8 +325,8 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
    * @param fpProb false positive probability when computing the overlap
    */
   def sparseRightOuterJoin[W: Coder](that: SCollection[(K, W)],
-                                    thatNumKeys: Long,
-                                    fpProb: Double = 0.01)(
+                                     thatNumKeys: Long,
+                                     fpProb: Double = 0.01)(
     implicit hash: Hash128[K],
     koder: Coder[K],
     voder: Coder[V]): SCollection[(K, (Option[V], W))] =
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -278,47 +278,97 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
-          lhsOverlap
-        } else {
-          lhsUnique
-        }
+    voder: Coder[V]): SCollection[(K, (V, Option[W]))] =
+    SCollection.unionAll(
+      splitSelfUsing(that, thatNumKeys, fpProb).map {
+        case (lhsUnique, lhsOverlap, rhs) =>
+          val unique =
+            lhsUnique.map(kv => (kv._1, (kv._2, Option.empty[W])))(Coder.gen[(K, (V, Option[W]))])
```
anish749: I was getting problems with diverging implicit expansions here and had to explicitly add `(Coder.gen[(K, (V, Option[W]))])`
Can this be done in a better way?

jto: probably just a case of Scalac being an idiot. I'll have a look at it.

jto: Fixed it in https://github.com/spotify/scio/pull/1389 @anish749

regadas: @anish749 can you rebase this? 😄

PR #1387, opened 2018-09-19
Make PTransform names unique

### LINKED ISSUE #1355: Fix `The following transforms do not have stable unique names` warnings
All the messages like `[pool-271-thread-11-ScalaTest-running-SCollectionTest] WARN org.apache.beam.sdk.Pipeline - The following transforms do not have stable unique names: apply@{OutcomeOf.scala:85}1`

jto:
fix #1355


scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
```diff
@@ -59,7 +59,7 @@ private[scio] object CallSites {
     val n = nameCache.merge(name, 1, new BiFunction[Int, Int, Int] {
       override def apply(t: Int, u: Int): Int = t + u
     })
-    if (n == 1) name else name + 1
+    if (n == 1) name else name + n
```
regadas: 🥇

Issue #1354, opened 2018-09-11
Sparse & hash versions of intersectByKey


nevillelyh:
Right now `intersectByKey` is implemented with a `cogroup` and could be inefficient with the RHS (collection of keys) is small or LHS (key-value pairs) is huge. We could implement a Bloom filter backed version similar to `sparseOuterJoin`.
https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala#L257

nevillelyh:
For feature parity we could also have a `hash` based implementation, i.e. backed by a `Set[K]` side input.

PR #1393, opened 2018-09-20
Add sparseIntersectByKey/hashIntersectByKey (#1354)

ClaireMcGinty:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -488,6 +488,55 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+                                         thatNumKeys: Int,
+                                         fpProb: Double = 0.1)(
+    implicit koder: Coder[K],
+    voder: Coder[V],
+    hash: Hash128[K]): SCollection[(K, V)] = {
+    val width = BloomFilter.optimalWidth(thatNumKeys, fpProb).get
+    val numHashes = BloomFilter.optimalNumHashes(thatNumKeys, width)
+    val rhsBf = that
+      .aggregate(BloomFilterAggregator[K](numHashes, width))
+      .asIterableSideInput
```
nevillelyh: Any reason not using singleton side input here?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -488,6 +488,55 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+
+    self
+      .withSideInputs(rhsBf)
+      .filter { case (e, c) => c(rhsBf).headOption.exists(_.maybeContains(e._1)) }
+      .toSCollection
+      .cogroup(that.map((_, ())))
```
nevillelyh: We don't need the `cogroup+flatMap` anymore since it's covered by `filter`?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -488,6 +488,55 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+      val thatParts = that.partition(n, _.hashCode() % n)
+      val joined = (thisParts zip thatParts).map {
+        case (lhs, rhs) =>
+          lhs.sparseIntersectByKeyImpl(rhs, bfSettings.capacity, fpProb)
+      }
+      SCollection.unionAll(joined)
+    }
+  }
+
+  protected def sparseIntersectByKeyImpl(that: SCollection[K],
```
nevillelyh: This can be `private`?

scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
```diff
@@ -112,6 +112,25 @@ class PairHashSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+   * Return an SCollection with the pairs from `this` whose keys are in `that`
+   * given `that` is small enough to fit in memory.
+   * @group per key
+   */
+  def hashIntersectByKey(that: SCollection[K])(implicit koder: Coder[K],
+                                               voder: Coder[V]): SCollection[(K, V)] = {
+    val side = combineAsMapSideInput(that.map((_, ())))
+    self
+      .withSideInputs(side)
+      .flatMap {
```
nevillelyh: This can be a `filter`?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ClaireMcGinty: as a singleton, in the case of an empty `that` it'll fail with `Empty PCollection accessed as a singleton view` when we try to access `rhsBf` through side input context. I based it off of the SparseJoin implementations, which use an `IterableSideInput` in the same way. wdyt?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ClaireMcGinty: i guess it depends on if we expect this function to be approximate or exact -- since the bloom filter can have false positives

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
nevillelyh: Actually there's a `def asSingletonSideInput(defaultValue: T): SideInput[T]` with default now maybe use that for both?

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
nevillelyh: Maybe add a `exact: Boolean` option that defaults to false? If the user joins the result with another data set it defeats the purpose of `cogroup+flatMap`.

scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
ClaireMcGinty: good catch! thanks.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ClaireMcGinty: sounds good! added a unit test for that case in SparseOuterJoins too. Also added an `exact` boolean to the sparse intersection.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
ClaireMcGinty: weirdly when i make it private it's inaccessible from

```
      val joined = (thisParts zip thatParts).map {
        case (lhs, rhs) =>
          lhs.sparseIntersectByKeyImpl(rhs, bfSettings.capacity, computeExact, fpProb)
```

nevillelyh: You're right probably because it's an implicit class.

PR #1397, opened 2018-09-20
use singleton side input instead of iterable side input for sparse joins

anish749:
IterableSideInput performance isn't optimal because it get materialized across all bundles (and threads?) separately, which leads to a significant amount of wall time being spent on reading the SideInput in the consuming DoFn. This also results in the bytes read from side input to be very high. A singleton here would mean the materialization to happen once, and hence the time spent reading would be near 0. The bytes read from side input would also be significantly lower - once per bundle. (more than 3000 times lower in some pipelines I tested)

For streaming joins, there might be a case where the BF would be empty with no elements in the side input, and hence we would need to use the zero value to avoid NoSuchElementException.


anish749:
I would try to refactor this a bit further as well, to support a sparseLookUp function.

Something like:
`pairSColl.sparseLookUp(that1: SCollection[(K, A)], that2: SCollection[(K, B)], thisNumKeys: Long, fpProb: Double): SCollection[(K, (V, Iterable[A], Iterable[B]))]`

The idea here is this being opposite of `hashLookUp`. `that1` and `that2` would be considerably larger and `this` is also large enough to not fit in memory. And have a function for `sparseLookUp(hugeCollection1)` as well.

The left side here ends up being smaller than the right side. Is this way of naming a good idea?

PR #1404, opened 2018-09-24
Register sys.props

regadas:
To list all `SysProps`:
```scala
SysProps.properties: List[SysProps]
```

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/types/TypeProvider.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/types/TypeProvider.scala
@@ -390,7 +390,7 @@ private[types] object TypeProvider {
       Paths
         .get(CoreSysProps.TmpDir.value)
         .resolve(CoreSysProps.User.value)
-        .resolve("avro-classes")
+        .resolve("bigquery-classes")
     }
   }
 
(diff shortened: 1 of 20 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/types/TypeProvider.scala
```diff
@@ -380,18 +382,15 @@ private[types] object TypeProvider {
...
   private def getBQClassCacheDir: Path = {
     // TODO: add this as key/value settings with default etc
-    if (sys.props("bigquery.class.cache.directory") != null) {
-      Paths.get(sys.props("bigquery.class.cache.directory"))
-    } else {
+    AvroSysProps.Debug.valueOption[String].map(Paths.get(_)).getOrElse {
       Paths
-        .get(sys.props("java.io.tmpdir"))
-        .resolve(sys.props("user.name"))
-        .resolve("bigquery-classes")
```
regadas: @nevillelyh any reason why this was using the same folder? I guess we can put them on a different one right?
/cc @jto @ClaireMcGinty

nevillelyh: Just laziness, so that `scio-idea-plugin` works with both Avro & BQ macro out of the box (it only looks in `bigquery-classes`).

regadas: @nevillelyh should we revert the name back for now?

jto: I would rather revert to the old name just to be sure we don't break anyone's intellij. We'll have enough support to do without blowing up IDEs...

PR #1405, opened 2018-09-25
Apply scalafix rules

jto:
- remove procedure syntax
   https://scalacenter.github.io/scalafix/docs/rules/ProcedureSyntax.html
- Make implicit classes members private:
   https://scalacenter.github.io/scalafix/docs/rules/LeakingImplicitClassVal.html


regadas:
@jto now you need to reformat things again 😄

PR #1407, opened 2018-09-25
Method to parse only expected options

mikaelgoldmann:
I've found something like this useful when fixing a bug in
a project using scio. Basically, it protects you from issues
resulting from typos in options, or accidentally running an
older version that silently ignores an option.

When that came up, we solved it with a wrapper, but it would have
been nice if support for failing on unknown arguments was built
into scio.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -229,6 +229,16 @@ object ScioContext {
     }
   }
 
+  def parseOptions[T <: PipelineOptions: ClassTag](args: Array[String],
+                                                   withValidation: Boolean = false): T = {
+    val (opts, extra) = parseArguments[T](args)
+    val extraArgs = extra.asMap.keys
+    if (extraArgs.nonEmpty) {
+      throw new IllegalArgumentException("Unsupported arguments: " + extraArgs.mkString(", "))
+    }
+    opts
+  }
+
   import scala.language.implicitConversions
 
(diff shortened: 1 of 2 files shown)


mikaelgoldmann:
Hi @regadas, I was hoping to get some feedback like that.

I think that does make sense and I'll look at that, but I'd still like to have a version of `ScioContext.parseArgs` that does complains if there are unknown arguments (or is that a bad idea for some reason I don't understand?). A non-breaking way to do that could possibly be to introduce a named boolean parameter where the default value matches current behavior. Would something like that make sense, and would it be preferable to adding the new method proposed here?

jto:
H @mikaelgoldmann,

Could you talk a bit more about your use case ?

Normally when creating a Scio job one uses:
```scala
``` 
or 

```scala
val (sc, args) = ContextAndArgs.typed[Options](cmdlineArgs)
```
but never `ScioContext.parseArgs` directly.

In the first case (`ContextAndArgs(cmdlineArgs)`), you can't restrict the argument list because that'd break the `args("agrName")` calls.

In the later case (`ContextAndArgs.typed`) you can but not by relying on `BeamOptions`. You need to compute the union of the arguments from Beam's `PipelineOptions` AND the arguments defined in your own `Options` case class and fail when something not included in that list is passed to the pipeline.

The best solution would be to leave `ContextAndArgs.apply` as it is but to make `ContextAndArgs.typed` strict by default. It would also promote stricter validation of arguments since you'd get it for free when you use `ContextAndArgs.typed`.

mikaelgoldmann:
Makes sense. I'll fix this so it only affects `ContextAndArgs.typed[T]`

PR #1417, opened 2018-10-01
Set timestamp attribute in JobTest for PubSubIO (fix #1403)

### LINKED ISSUE #1403: pubsubSubscriptionWithAttributes should set timestamp in JobTest
Find this on Stack Overflow
https://stackoverflow.com/questions/49069194/scio-jobtest-pubsubio-pubsubsubscriptionwithattributes-timestampattribute-and

Basically when using `pubsubSubscriptionWithAttributes` with `timestampAttribute` argument, `JobTest` should do a `timestampBy` using the corresponding field in `Map[String, String]` attributes.

We can probably add a condition here, and set timestamp if `ScioContext#isTest` and `timestampAttribute != null`

https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L572

ClaireMcGinty:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -574,7 +574,13 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
     idAttribute: String,
     timestampAttribute: String): SCollection[(T, Map[String, String])] = {
     val io = PubsubIO.withAttributes[T](name, idAttribute, timestampAttribute)
-    this.read(io)(PubsubIO.ReadParam(isSubscription))
+    val read = this.read(io)(PubsubIO.ReadParam(isSubscription))
+
+    if (this.isTest && timestampAttribute != null) {
```
regadas: @ClaireMcGinty i was wondering if we should to this inside `PubsubIOWithAttributes` because now since we have multiple ways of reading the behavior is not consistent, i.e:

```
sc.read(PubsubIO.withAttributes(name, idAttribute, timestampAttribute))(ReadParam(false)))
```

regadas: talked IRL with @ClaireMcGinty and this will require some more thinking.

ClaireMcGinty: will merge this, then refactor our test IO setup in a different PR.

PR #1418, opened 2018-10-01
Fix verifyNondeterministic exception in coders

kanterov:
Now exception message includes offending fields and their coders.
Otherwise, it isn't clear what is the root cause.

Change RecordCoder to encode records instead of sequences

# TODO
- [x] fix scala 2.11
- [x] implement disjunction checks

### DIFF
--- a/scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -198,7 +198,7 @@ private class RecordCoder[T](typeName: String,
 
     problems match {
       case (_, e) :: _ =>
-        val reasons = problems.map { case (reason, e) => reason }
+        val reasons = problems.map { case (reason, _) => reason }
         throw new NonDeterministicException(this, reasons.asJava, e)
       case Nil =>
     }
--- a/scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -25,7 +25,7 @@ import org.apache.beam.sdk.util.common.ElementByteSizeObserver
 import org.apache.beam.sdk.values.KV
 
 import scala.annotation.implicitNotFound
-import scala.collection.JavaConverters
+import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 
 @implicitNotFound(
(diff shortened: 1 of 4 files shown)


scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -100,28 +110,32 @@ private object WrappedBCoder {
 // Coder used internally specifically for Magnolia derived coders.
 // It's technically possible to define Product coders only in terms of `Coder.transform`
 // This is just faster
-private class RecordCoder[T: ClassTag](cs: Array[(String, BCoder[T])])
```
kanterov: ClassTag wasn't doing anything because `T` was always `Any` ¯\_(ツ)_/¯

scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -45,15 +47,23 @@ sealed trait Coder[T] extends Serializable
...
 final case class Transform[A, B] private (c: Coder[A], f: BCoder[A] => Coder[B]) extends Coder[B]
-final case class Disjunction[T, Id] private (idCoder: Coder[Id],
+final case class Disjunction[T, Id] private (typeName: String,
+                                             idCoder: Coder[Id],
                                              id: T => Id,
                                              coder: Map[Id, Coder[T]])
     extends Coder[T]
-final case class Record[T] private (cs: Array[(String, Coder[T])]) extends Coder[Array[T]]
+
+final case class Record[T] private (typeName: String,
```
kanterov: `typeName: String` doesn't look good, but don't know of a better serializable type, don't want to add dependency to `magnolia.TypeName`

scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -45,15 +47,23 @@ sealed trait Coder[T] extends Serializable
 final case class Beam[T] private (beam: BCoder[T]) extends Coder[T]
```
kanterov: `private` doesn't do anything there because it hides only `new`, but not `apply`

kanterov:
While doing this found a minor problem in beam: https://github.com/apache/beam/pull/6526

kanterov:
@regadas @jto  PTAL

scio-coders/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -130,18 +144,42 @@ private class RecordCoder[T: ClassTag](cs: Array[(String, BCoder[T])])
...
+          case e: NonDeterministicException =>
+            val reason = s"field $label is using non-deterministic $c"
+            List(reason -> e)
+        }
+    }
+
+    problems match {
+      case (_, e) :: _ =>
+        val reasons = problems.map { case (reason, e) => reason }
+        throw new NonDeterministicException(this, JavaConverters.seqAsJavaList(reasons), e)
```
regadas: here `.asJava` is fine. In 2.12 it's the same thing as this.
```
  implicit def seqAsJavaListConverter[A](b : Seq[A]): AsJava[ju.List[A]] =
    new AsJava(seqAsJavaList(b))
```

kanterov:
@regadas thanks for the review, I added support for adjunctions as well

kanterov:
@regadas are we waiting for @jto, or you will merge this PR, or should I rebase it?

PR #1421, opened 2018-10-02
Make typed and default args parsing logic more test friendly

regadas:
# TODO
- [x] add typed parser tests

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -145,8 +145,8 @@ object ContextAndArgs {
         }
 
       CaseApp.detailedParseWithHelp[T](customArgs) match {
-        case Left(message) =>
-          Failure(new Exception(message.message))
+        case Left(error) =>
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -96,70 +97,109 @@ private object RunnerContext {
...
-    val Reg = "^-{1,2}(.+)$".r
-    val (customArgs, remainingArgs) =
-      args.partition {
-        case Reg(a) =>
-          val name = a.takeWhile(_ != '=')
-          supportedCustomArgs.contains(name)
-        case _ => true
+
+      CaseApp.detailedParseWithHelp[T](customArgs) match {
+        case Left(message) =>
```
ClaireMcGinty: maybe rename to `case Left(error) => `Failure(new Exception(error.message))` so you don't have to type `message.message` ?

jto:
@regadas is it still a WIP or can we merge this ?

regadas:
@jto merging.

PR #1423, opened 2018-10-02
Remove unnecessary final, rename vals and narrow access

regadas:


scio-coders-macros/src/main/scala/com/spotify/scio/coders/CoderMacros.scala
```diff
@@ -21,15 +21,14 @@ import scala.reflect.macros._
 
 private[coders] object CoderMacros {
 
-  var verbose = true
-  val reported: scala.collection.mutable.Set[(String, String)] =
+  private[this] var verbose = true
+  private[this] val Reported: scala.collection.mutable.Set[(String, String)] =
```
jto: Nitpicking: Why did you uppercase this ?

scio-coders-macros/src/main/scala/com/spotify/scio/coders/CoderMacros.scala
```diff
@@ -21,15 +21,14 @@ import scala.reflect.macros._
...
     scala.collection.mutable.Set.empty
 
-  private[this] val SETTING_SHOW_WARN = "show-coder-fallback"
-  private[this] val DEFAULT_SHOW_WARN = true
-  private[this] val KVS = "show-coder-fallback=(true|false)".r
+  private[this] val ShowWarnDefault = true
+  private[this] val ShowWarnSettingRegex = "show-coder-fallback=(true|false)".r
 
-  val blacklistedTypes = List("org.apache.beam.sdk.values.Row")
+  private[this] val BlacklistedTypes = List("org.apache.beam.sdk.values.Row")
```
jto: Same here. Why uppercase ?

scio-coders-macros/src/main/scala/com/spotify/scio/coders/CoderMacros.scala
regadas: Ah! nice catch i'll revert this since it's mutable

scio-coders-macros/src/main/scala/com/spotify/scio/coders/CoderMacros.scala
regadas: changed because it's a constant, so it falls in the naming convention.

PR #1427, opened 2018-10-04
Add more tests on coders

jto:
Adding tests for the coder that had no test coverage (according to codecov).


jto:
I found a bug in Beam in the process: https://issues.apache.org/jira/browse/BEAM-5645

PR #1430, opened 2018-10-04
Upgrade to beam 2.7

jto:


scio-bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryServicesWrapper.java
```diff
@@ -58,7 +58,14 @@ public long insertAll(TableReference ref, List<TableRow> rowList)
                 PaneInfo.NO_FIRING))
         .collect(Collectors.toList());
     return bqServices.getDatasetService(bqOptions)
-        .insertAll(ref, rows, null, InsertRetryPolicy.alwaysRetry(), null);
+        .insertAll(ref,
+          rows,
+          null,
+          InsertRetryPolicy.alwaysRetry(),
+          null,
```
regadas: should this be an empty list?

regadas: only `insertIdList` is marked with nullable

```
    <T> long insertAll(
        TableReference ref,
        List<ValueInSingleWindow<TableRow>> rowList,
        @Nullable List<String> insertIdList,
        InsertRetryPolicy retryPolicy,
        List<ValueInSingleWindow<T>> failedInserts,
        ErrorContainer<T> errorContainer,
        boolean skipInvalidRows,
        boolean ignoreUnknownValues)
```

jto: I did not change this one. Maybe ?

Issue #936, opened 2017-11-21
PubsubIO batch size should be configurable


nevillelyh:
This might hit service payload size limit. We should fix it upstream and maybe expose the setting.

https://github.com/apache/beam/blob/release-2.1.0/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java

nevillelyh:
https://issues.apache.org/jira/browse/BEAM-3234

ThatRfernand:
https://issues.apache.org/jira/browse/BEAM-3234 is a duplicate of another issue, a PR for which addresses this concern.

nevillelyh:
The new upstream issue is https://issues.apache.org/jira/browse/BEAM-2660

nevillelyh:
https://issues.apache.org/jira/browse/BEAM-3218
https://github.com/apache/beam/pull/4275
https://github.com/spotify/scio/pull/1207

All address the same problem with a different approach - by checking against the max publish byte size.

nevillelyh:
Don't think the upstream PRs will get merged and released in time for 0.5.7 though. Move to 0.6.0 instead?

nevillelyh:
Upstream https://github.com/apache/beam/pull/3619 was recently merged and should be available in the next Beam release.

nevillelyh:
From the mailing list:
> Our release calendar indicates that the process for the 2.7.0 Beam release should start on September 7.

Which means that we won't have the upstream fix until then. Moving this to 0.7.0

PR #1433, opened 2018-10-05
Add PubSubIO batch size write params

regadas:
Fixes: #936

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
@@ -36,7 +36,7 @@ import scala.reflect.ClassTag
 
 sealed trait PubsubIO[T] extends ScioIO[T] {
   override type ReadP = PubsubIO.ReadParam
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
```diff
@@ -45,6 +45,9 @@ sealed trait PubsubIO[T] extends ScioIO[T] {
 object PubsubIO {
   final case class ReadParam(isSubscription: Boolean)
 
+  final case class WriteParam(maxBatchSize: Option[Int] = None,
+                              maxBatchBytesSize: Option[Int] = None)
+
   def apply[T: ClassTag: Coder](name: String,
```
regadas: @jto @ClaireMcGinty it seems that `idAttribute` and `timestampAttribute` are good candidates to be part of `ReadParam` and `WriteParam` types WDYT?

jto: yes indeed

PR #1435, opened 2018-10-08
Automated migration to v0.7.0 using scalafix

jto:
Add missing imports in files and rewrite `sys.prop` calls.

### DIFF
--- a/scalafix/build.sbt
+++ b/scalafix/build.sbt
@@ -1,0 +1,54 @@
+lazy val V = _root_.scalafix.sbt.BuildInfo
+inThisBuild(
+  List(
+    organization := "com.spotify",
+    homepage := Some(url("https://github.com/spotify/scio")),
+    licenses := List("Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")),
+    developers := List(
+      Developer(
+        "jto",
+        "Julien Tournay",
+        "julient@spotify.com",
+        url("https://example.com")
+      )
+    ),
+    scalaVersion := V.scala212,
+    addCompilerPlugin(scalafixSemanticdb),
+    scalacOptions ++= List(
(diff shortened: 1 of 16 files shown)


scalafix/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.1.1
```
regadas: 🎨 is it ok if we use same version as in scio?

jto: Probably yes. I used the g8 template scalafix provide but I should be able to change that.

scalafix/build.sbt
```diff
@@ -0,0 +1,54 @@
+lazy val V = _root_.scalafix.sbt.BuildInfo
```
regadas: can we make this part of scio build instead of different project?

jto:
FYI: I tested it on a pretty big pipeline at Spotify and it gets the compilations errors down from 329 to 6. Fixing those errors takes less than 5 min.

scalafix/build.sbt
jto: Probably not. The build for scalafix rules is pretty specific. I followed their tutorial and creating a separate project seems to be the recommended method

jto:
I need to add a migration rule for methods that use a parameter and a potentially explicit`Ordering` like `SCollection.quantilesApprox` because

```scala
scoll.quantilesApprox(numQuantiles)(Ordering.by(_.smth))
```

is now

```scala
scoll.quantilesApprox(numQuantiles, Ordering.by(_.smth))
```

jto:
@regadas I think we can merge this and improve it later if necessary

regadas:
@jto Yup let's merge this, We will need to pimp a little bit the build info, just in case we decide to publish these rules.

PR #1440, opened 2018-10-09
Add coder for SpecificFixed

kanterov:


kanterov:
@regadas memoization is used to get over serialization issues

regadas:
@kanterov I see `Schema` is not serializable. So in this case I think something along the lines would work as well:

```scala
private final class SpecificFixedCoder[A <: SpecificFixed](cls: Class[A]) extends AtomicCoder[A] {
  private lazy val schema = SpecificData.get().getSchema(cls)
  private lazy val size = schema.getFixedSize

...
}
```

kanterov:
@regadas thanks for the suggestion! fixed the code, I'm waiting for you or @jto to merge it because you own this repo

PR #1441, opened 2018-10-09
Add coder for CharSequence

kanterov:


kanterov:
the test is broken because `new Utf8("foo")` is not equal to `new String("foo")`

kanterov:
it was a bad idea to add such coder because it isn't lawful due to the absence of equality

kanterov:
We should probably add such coder to prevent fallback and make sure it throws in `verifyDeterministic`

regadas:
hm this was interesting .. from `CharSequence` javadoc

```
This interface does not refine the general contracts of the equals and hashCode methods. The result of comparing two objects that implement CharSequence is therefore, in general, undefined. Each object may be implemented by a different class, and there is no guarantee that each class will be capable of testing its instances for equality with those of the other. It is therefore inappropriate to use arbitrary CharSequence instances as elements in a set or as keys in a map.
```

PR #1443, opened 2018-10-10
exclude grpc-okhttp

honnix:
> For Android client, use grpc-okhttp instead of grpc-netty-shaded and grpc-protobuf-lite instead of grpc-protobuf...

As documented by https://github.com/grpc/grpc-java

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -351,7 +351,9 @@ lazy val scioCore: Project = Project(
       "com.google.protobuf" % "protobuf-java" % protobufVersion,
       "me.lyh" %% "protobuf-generic" % protobufGenericVersion,
       "org.apache.xbean" % "xbean-asm6-shaded" % asmVersion,
-      "io.grpc" % "grpc-all" % grpcVersion exclude ("io.opencensus", "opencensus-api"),
+      ("io.grpc" % "grpc-all" % grpcVersion)
+        .exclude("io.opencensus", "opencensus-api")
+        .exclude("io.grpc", "grpc-okhttp"),
       "com.github.alexarchambault" %% "case-app" % caseappVersion,
       "org.scalatest" %% "scalatest" % scalatestVersion % "test"
     )


honnix:
@regadas PTAL

honnix:
Why is codecov so mad about me?

honnix:
Probably my fork.

honnix:
@jto wdyt?

jto:
LGTM. @regadas wdyt ?

honnix:
No merge permission. Please merge it whenever suitable. Thanks.

honnix:
This is introduced by grpc-all and has caused trouble where instead of netty okhttp was chosen as transportation layer. I think in general grpc-all is not recommended to depend on.

regadas:
@honnix Apache Beam as dependencies that declare this as dependency i.e `org.apache.beam:beam-sdks-java-io-google-cloud-platform:2.7.0` so unless this problem is really affecting a Job execution, I don't think we should do it in `scio`. If you have a failing execution could you please share a repro example?

regadas:
@honnix yeah seems like is a 👎 from me. I also did a little digging and seems like if you bring the wrong grpc version to the mix bad things can happen https://github.com/grpc/grpc-java/issues/4423

honnix:
Did a bit more investigation.

The error we had was:
```
Exception in thread "main" java.lang.NoSuchMethodError: io.grpc.okhttp.OkHttpChannelProvider.isAndroid()Z
        at io.grpc.okhttp.OkHttpChannelProvider.priority(OkHttpChannelProvider.java:51)
        at io.grpc.ManagedChannelProvider$1.getPriority(ManagedChannelProvider.java:49)
        at io.grpc.ManagedChannelProvider$1.getPriority(ManagedChannelProvider.java:41)
        ...
```

Reading the source code, it basically means when trying to determine priority of which `ManagedChannelProvider` to choose (okhttp has higher priority than netty in case of Android), `io.grpc.okhttp.OkHttpChannelProvider.isAndroid()Z` doesn't exist. I believe that is due to somehow pulling in lower version of `grpc-okhttp` into classpath. This is very likely not because of the dependency introduced by Scio, but rather user's own dependency.

However that being said, it still doesn't make sense to include `grpc-okhttp`, not even in beam code.

honnix:
This pom doesn't make sense to me: http://central.maven.org/maven2/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.7.0/beam-sdks-java-io-google-cloud-platform-2.7.0.pom

It depends on both `grpc-netty` explicity and `grpc-all`. :/

honnix:
I agree that this issue is not the root cause of our problem. So I'm fine with closing it.

kanterov:
@honnix I've seen this stacktrace before, the issue is solved by consistently pinning each gRPC dependency to `1.13.1`:
```scala
  val grpcVersion = "1.13.1"
  val grpcNetty = "io.grpc" % "grpc-netty" % grpcVersion
  val grpcNettyCore = "io.grpc" % "grpc-core" % grpcVersion
  val grpcAuth = "io.grpc" % "grpc-auth" % grpcVersion
  val grpcAll = "io.grpc" % "grpc-all" % grpcVersion
  val grpcOkHttp = "io.grpc" % "grpc-okhttp" % grpcVersion

dependencyOverrides ++= Seq(
      grpcNetty,
      grpcNettyCore,
      grpcAll,
      grpcAuth,
      grpcOkHttp
    )
```

regadas:
Closing this. @kanterov confirmed what I mentioned earlier.

honnix:
Thanks.
I think I will still send an issue to beam according to https://github.com/grpc/grpc-java/pull/1924

honnix:
https://issues.apache.org/jira/browse/BEAM-5715

PR #1444, opened 2018-10-10
Work around buggy handling of file URI scheme

ulzha:
@fabriziodemaria encountered the problem while using Flo test mode, which tried to saveMetrics to a local file.

@jto PTAL

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
@@ -110,10 +110,16 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
   def state: State = Try(internal.getState).getOrElse(State.UNKNOWN)
 
   /** Save metrics of the finished pipeline to a file. */
-  def saveMetrics(filename: String): Unit = {
+  def saveMetrics(fileUri: String): Unit = {
     require(isCompleted, "Pipeline has to be finished to save metrics.")
     val mapper = ScioUtil.getScalaJsonMapper
-    val resourceId = FileSystems.matchNewResource(filename, false)
+
+    /**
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -110,10 +110,16 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
...
+  def saveMetrics(fileUri: String): Unit = {
     require(isCompleted, "Pipeline has to be finished to save metrics.")
     val mapper = ScioUtil.getScalaJsonMapper
-    val resourceId = FileSystems.matchNewResource(filename, false)
+
+    /**
+     * Workaround for BEAM-5677. If we pass URIs with "file" scheme through,
+     * LocalFileSystem.matchNewResource mishandles them.
+     */
+    val fileUriOrPath = if (fileUri.startsWith("file:")) fileUri.substring(5) else fileUri
```
jto:
```
scala> val fileUri = "file://path/to/yolo.txt"
fileUri: String = file://path/to/yolo.txt

scala> val fileUriOrPath = if (fileUri.startsWith("file:")) fileUri.substring(5) else fileUri
fileUriOrPath: String = //path/to/yolo.txt
```

Is that the expected behaviour ?

ulzha: `file://path/to/yolo.txt` would be a reference to a remote file `/to/yolo.txt` on a host named `path`. I indeed shouldn't break remote files handling. Will fix.

File URIs referring to local files would come in the form `file:///path/to/yolo.txt` (three slashes, not two), and `///path/to/yolo.txt` is fine, same thing as `/path/to/yolo.txt`.

regadas:
There are hasn't been any change on this for a long time now and this hasn't been reported.

Issue #1369, opened 2018-09-17
Create BQ tables for data-integration-test in IT


nevillelyh:
Right now they are created manually. We should run it in `beforeAll` in the same test. Need to figure out compilation order though.

ClaireMcGinty:
some scope expansion related to bigquery IT:

- [ standardize on EU location for BQ reads + temp tables ]
- [ port bigquery IO tests from scio-core to scio-bigquery? ]
- [ test fixture to copy over `bigquery-public-data` dataset (which is in US datacenter) to `data-integration` project in EU ]

ClaireMcGinty:
it's easier to just override retention policy directly in BQ so our test fixtures never expire. closing this ticket for now.

PR #1445, opened 2018-10-10
Standardize on EU location for IT test data in scio-bigquery

ClaireMcGinty:
part of https://github.com/spotify/scio/issues/1369


scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryClientIT.scala
```diff
@@ -39,13 +39,11 @@ class BigQueryClientIT extends FlatSpec with Matchers {
 
   "extractLocation" should "work with legacy syntax" in {
     val query = "SELECT word FROM [data-integration-test:samples_%s.shakespeare]"
-    bq.extractLocation(query.format("us")) shouldBe Some("US")
```
ClaireMcGinty: i removed the US test cases from `extractLocation` -- i would expect that any regressions would be caught by the remaining EU test case. what do you think?

regadas: yup 👍

PR #1453, opened 2018-10-15
use grpc-netty-shaded

danielnorberg:
`grpc-netty-shaded` is recommended: https://github.com/grpc/grpc-java#download


danielnorberg:
I guess we'd also need to upgrade to `grpc >= 1.9.0`

danielnorberg:
Yeah, I realized this will probably not be feasible until beam changes.

PR #1462, opened 2018-10-16
Fix KryoRegistrar scope widenning

regadas:


scio-macros/src/main/scala/com/spotify/scio/coders/KryoRegistrar.scala
```diff
@@ -37,14 +37,14 @@ private object KryoRegistrarMacro {
     val traitT = tq"_root_.com.spotify.scio.coders.AnnotatedKryoRegistrar"
 
     annottees.map(_.tree) match {
-      case List(q"class $name extends ..$parents { ..$body }") =>
+      case List(q"$mod class $name extends ..$parents { ..$body }") =>
```
nevillelyh: Where is `mod` defined? Can't find in this file.

regadas: In this case `$mod` will be extracted as result of the `unapply`.

nevillelyh: Oh lol misread it. Make sense.

PR #1463, opened 2018-10-16
register default file systems in Scio test context (fix #1455)

### LINKED ISSUE #1455: Pipeline test will fail if one uses beam FileSystems and GCS bucket.
According to https://github.com/spotify/scio/wiki/FAQ#how-do-i-access-various-files-outside-of-a-sciocontext one should use `FileSystems` in order to access files outside of ScioContext.

This works in `scio-repl`:

```
scio> import org.apache.beam.sdk.io.FileSystems
import org.apache.beam.sdk.io.FileSystems

scio> val fs = FileSystems.matchNewResource("gs://some-bucket/file", false)
fs: org.apache.beam.sdk.io.fs.ResourceId = gs://some-bucket/file

scio>
```

this will also work in a normal pipeline run, but it will not work in a any local test run (using `PipelineSpec`) to test the pipeline.

If one uses `FileSystems.matchNewResource("gs://some-bucket/file", false)` in a pipeline, during the tests it will throw: 

```
java.lang.IllegalArgumentException: No filesystem found for scheme gs
	at org.apache.beam.sdk.io.FileSystems.getFileSystemInternal(FileSystems.java:459)
	at org.apache.beam.sdk.io.FileSystems.matchNewResource(FileSystems.java:529)
        ...
```

Quick fix:

One needs to use `FileSystems.setDefaultPipelineOptions(PipelineOptionsFactory.create)` before trying to reach a resource with `FileSystems`.

Side note: method `matchNewResource` from `FileSystems` takes two parameters, whereas the linked Scio documentation has an example with only one.

ClaireMcGinty:
doc for `FileSystems.setDefaultPipelineOptions` says:
```
This is expected only to be used by runners after {@code Pipeline.run}, or in tests.
```

Beam manually registers these in their unit testing framework: [TestPipeline.java](https://github.com/apache/beam/blob/4a92d28967a1c3fbe4719d8599bdde7b9ff2e31f/sdks/java/core/src/main/java/org/apache/beam/sdk/testing/TestPipeline.java#L463) so it seems reasonable to also register it in ours?

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -565,7 +565,7 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
   // =======================================================================
 
   /**  Whether this is a test context. */
-  lazy val isTest: Boolean = testId.isDefined
+  def isTest: Boolean = testId.isDefined
 
   // =======================================================================
   // Read operations
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -36,6 +36,7 @@ import com.spotify.scio.values._
 import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import org.apache.beam.sdk.PipelineResult.State
 import org.apache.beam.sdk.extensions.gcp.options.GcsOptions
+import org.apache.beam.sdk.io.FileSystems
 import org.apache.beam.sdk.metrics.Counter
 import org.apache.beam.sdk.options._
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -560,7 +565,7 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
   // =======================================================================
 
   /**  Whether this is a test context. */
-  def isTest: Boolean = testId.isDefined
+  lazy val isTest: Boolean = testId.isDefined
```
regadas: @ClaireMcGinty I think it's ok to keep it as `def` there's no point in having it as `lazy val`

PR #1479, opened 2018-10-22
pass in region to DataflowResult.apply()

danielnorberg:
Unless the region is explicitly set, we get the default region from the pipeline options when looking up the dataflow job and will get a 404 unless the job was running in that region.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
@@ -91,7 +91,7 @@ object DataflowResult {
   }
 
   def apply(job: Job): DataflowResult = {
-    val options = getOptions(job.getProjectId)
+    val options = getOptions(job.getProjectId, job.getLocation)
     apply(options, job)
   }
 
--- a/scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
@@ -41,7 +41,7 @@ class DataflowResult(val internal: DataflowPipelineJob) extends RunnerResult {
     this(internal.asInstanceOf[DataflowPipelineJob])
 
   private val client =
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
```diff
@@ -109,10 +109,10 @@ object DataflowResult {
     new DataflowResult(internal)
   }
 
-  private def getOptions(projectId: String): DataflowPipelineOptions = {
-    val options =
-      PipelineOptionsFactory.create().as(classOf[DataflowPipelineOptions])
+  private def getOptions(projectId: String, region: String): DataflowPipelineOptions = {
+    val options = PipelineOptionsFactory.create().as(classOf[DataflowPipelineOptions])
     options.setProject(projectId)
+    options.setRegion(region)
```
regadas: @danielnorberg that option is still marked as `@Experimental` given that, I kinda tend to prefer the behavior as is ... you specifically set the region.

@nevillelyh WDYT?

danielnorberg: I’m fine with whatever as long as it is possible to pass in the region to `DataflowResult.apply()` and get results for jobs that are outside `us-central1`

danielnorberg: @regadas would you prefer this https://github.com/spotify/scio/pull/1480 ?

danielnorberg: Note that unless we set the region on the `DataflowPipelineOptions`, calling `getJob` on the `DataflowResult` will not work because it uses the region of the `DataflowPipelineOptions` when querying for the dataflow job.

regadas: @danielnorberg yeah I had a second look at this and it seems this is the best approach, plus they use this option all over the place in `apache-beam` `DataflowRunner` and `DataflowClient`.

danielnorberg:
Thanks!

PR #1480, opened 2018-10-23
pass in region to DataflowResult.apply() (alt 2)

danielnorberg:
Unless the region is explicitly set, we get the default region from the pipeline options when looking up the dataflow job and will get a 404 unless the job was running in that region.

Alternative to https://github.com/spotify/scio/pull/1479

**Note**: Calling `getJob` on the returned `DataflowResult` will not work as it uses the region of the  `DataflowPipelineOptions` when querying the dataflow api for the job.


regadas:
Closing in favor of #1479

PR #1482, opened 2018-10-23
#1477: Get rid of Future.failed in IOs

### LINKED ISSUE #1477: Refactor `ScioIO` instances to get rid of the `Future.failed` in writes and `Tap`
The idea being to return a successful `Future[Tap[Nothing]]` instead.

jto:
Fixes #1477

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
@@ -24,6 +24,33 @@ import com.spotify.scio.values.SCollection
 
 import scala.concurrent.Future
 
+sealed trait TapT[A] extends Serializable {
+  type T
+  def saveForTest(data: SCollection[A])(implicit c: Coder[A]): Future[Tap[T]]
+}
+
+object TapT {
+  // scalastyle:off structural.type
+  type Aux[A, T0] = TapT[A] { type T = T0 }
+  // scalastyle:on structural.type
+}
+
+final class EmptyTapOf[A] private extends TapT[A] {
+  type T = Nothing
+  def saveForTest(data: SCollection[A])(implicit c: Coder[A]): Future[Tap[T]] =
+    Future.successful(EmptyTap)
+}
+
+object EmptyTapOf { def apply[A]: TapT.Aux[A, Nothing] = new EmptyTapOf[A] }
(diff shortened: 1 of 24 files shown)


scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
```diff
@@ -24,6 +24,33 @@ import com.spotify.scio.values.SCollection
...
+  def saveForTest(data: SCollection[A])(implicit c: Coder[A]): Future[Tap[T]]
+}
+
+object TapT {
+  // scalastyle:off structural.type
+  type Aux[A, T0] = TapT[A] { type T = T0 }
+  // scalastyle:on structural.type
+}
+
+final class TapOfNothing[A] private extends TapT[A] {
```
regadas: @jto should we name this `EmptyTapOf[A]`

scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
```diff
@@ -24,6 +24,33 @@ import com.spotify.scio.values.SCollection
 
 import scala.concurrent.Future
 
+sealed trait TapT[A] {
```
regadas: @jto I think we could move this to `Tap.scala`?

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -60,6 +60,11 @@ trait Tap[T] extends Serializable { self =>
 
 }
 
+final case object EmptyTap extends Tap[Nothing] {
```
regadas: @jto 🎨 i think `final` here is redundant, since it's a top level object

scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
jto: No. The only use of the class is to encode the fact the write and tap can either return a `Tap[A]` or a `Tap[Nothing]`. It's actually not strictly related to `Tap` but I can't find a better name...

regadas: yeah on a second thought I agree with you! 👍

PR #1483, opened 2018-10-23
Feature/sub exists

BenFradet:
This is the sister PR to #1216.

I also removed the `pubsubOptions` param from `newChannel` since it doesn't seem to be used.

I'm keen on writing unit tests for `PubSubAdmin` next if possible.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/pubsub/PubSubAdmin.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/pubsub/PubSubAdmin.scala
@@ -37,7 +37,7 @@ import scala.util.Try
 object PubSubAdmin {
 
   private object GrpcClient {
-    private def newChannel(pubsubOptions: PubsubOptions): ManagedChannel = {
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/pubsub/PubSubAdmin.scala
```diff
@@ -19,7 +19,7 @@ package com.spotify.scio.pubsub
 
 import com.google.pubsub.v1.PublisherGrpc.PublisherBlockingStub
 import com.google.pubsub.v1.SubscriberGrpc.SubscriberBlockingStub
-import com.google.pubsub.v1.{GetTopicRequest, PublisherGrpc, SubscriberGrpc, Subscription, Topic}
+import com.google.pubsub.v1._
```
regadas: Hi @BenFradet can you keep the expanded version here?

BenFradet: sure :+1:

PR #1484, opened 2018-10-24
Fix mixed content on examples page by serving js from https

lazyval:
before: 
screenshot 2018-10-24 at 17 47 16
after:
screenshot 2018-10-24 at 17 47 40

With that said, I've only tested on Chrome (69.0.3497.100) and Safari (v12 14606.1.36.1.9)

### DIFF
--- a/project/SoccoIndex.scala
+++ b/project/SoccoIndex.scala
@@ -33,7 +33,7 @@ object SoccoIndex {
     """.stripMargin
   val footer =
     """  </xmp>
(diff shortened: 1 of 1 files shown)


lazyval:
wondering if codecov calculations are off, because changing single character in such a place should not suddenly decrease coverage by 10% 

> Merging #1484 into master will decrease coverage by 10.37%.

regadas:
@lazyval yeah, I actually have a revision of codeconv config in my todo list. Just didn't had time to get it done yet 😄

PR #1491, opened 2018-10-29
add Spanner package

ClaireMcGinty:
Currently supports writes from `Mutation`s and reads into `Struct`s. I have the spanner type macros saved for sometime further down the line.

notes:
- Spanner integration test is kind of slow, takes 37 seconds
- SpannerIO's `testId` doesn't support a unit having 2 spanner reads/writes to the same DB since it's just keyed by the `SpannerConfig` params, which don't include table name. For writes the table name is keyed into each `Mutation` object and reads give you a choice of specifying a table name or a literal query
- Haven't implemented Taps yet

### DIFF
--- a/scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
+++ b/scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
@@ -40,7 +40,7 @@ object SpannerIOIT {
   private val adminClient = Spanner.getAdminClient(projectId)
   private val dbClient = Spanner.getDatabaseClient(config)
 
-  private val options = PipelineOptionsFactory.fromArgs(s"--project=$projectId").create()
+  private val options = PipelineOptionsFactory.create()
 }
 
 class SpannerIOIT extends FlatSpec with Matchers with BeforeAndAfterAll {
@@ -107,7 +107,8 @@ class SpannerIOIT extends FlatSpec with Matchers with BeforeAndAfterAll {
       sc,
       SpannerRead.ReadParam(
         readMethod = SpannerRead.FromTable(table, Seq("Key", "Value")),
-        withTransaction = true
+        withTransaction = true,
+        withBatching = true
       )
     )
 
@@ -143,6 +144,7 @@ class SpannerIOIT extends FlatSpec with Matchers with BeforeAndAfterAll {
       sc,
       SpannerRead.ReadParam(
         readMethod = SpannerRead.FromQuery(s"SELECT Key, Value FROM $table"),
+        withTransaction = true,
         withBatching = true
       )
     )
--- a/scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
+++ b/scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
@@ -26,21 +26,40 @@ import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{Await, ExecutionContext}
-import scala.concurrent.duration._
+import scala.concurrent.duration.Duration
 import scala.util.Random
 
 object SpannerIOIT {
-  private val tablePrefix = "test_table"
   private val projectId = "data-integration-test"
+  private val options = PipelineOptionsFactory.create()
   private val config: SpannerConfig = SpannerConfig.create()
     .withProjectId(projectId)
     .withDatabaseId(s"io_it_${Random.nextInt}")
     .withInstanceId("spanner-it")
 
-  private val adminClient = Spanner.getAdminClient(projectId)
-  private val dbClient = Spanner.getDatabaseClient(config)
+  private val adminClient = Spanner.adminClient(projectId)
+  private val dbClient = Spanner.databaseClient(config)
 
-  private val options = PipelineOptionsFactory.create()
+  private final case class FakeSpannerData(asMutations: Seq[Mutation], asStructs: Seq[Struct])
+  private def fakeData(tableName: String) = FakeSpannerData(
+    Seq(
+      Mutation
+        .newInsertBuilder(tableName)
+        .set("Key").to(1L).set("Value").to("foo")
+        .build(),
+      Mutation
+        .newInsertBuilder(tableName)
+        .set("Key").to(2L).set("Value").to("bar")
+        .build()
+    ),
+    Seq(
+      Struct.newBuilder()
+        .set("Key").to(1L).set("Value").to("foo")
+        .build(),
+      Struct.newBuilder()
+        .set("Key").to(2L).set("Value").to("bar")
+        .build())
+  )
 }
 
 class SpannerIOIT extends FlatSpec with Matchers with BeforeAndAfterAll {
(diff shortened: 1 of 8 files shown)


scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
```diff
@@ -0,0 +1,168 @@
...
+class SpannerIOIT extends FlatSpec with Matchers with BeforeAndAfterAll {
+  import SpannerIOIT._
+  implicit val ec: ExecutionContext = ExecutionContext.global
+
+  override def beforeAll(): Unit = {
+    val create = adminClient.createDatabase(
+      config.getInstanceId.get(),
+      config.getDatabaseId.get(),
+      List(
+        s"CREATE TABLE ${tablePrefix}_1 (\n Key INT64, \n Value STRING(MAX) \n) PRIMARY KEY (Key)",
```
nevillelyh: nit: do we really need the `\n`s here? Kinda makes it hard to read.

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
```diff
@@ -0,0 +1,168 @@
...
+  private val dbClient = Spanner.getDatabaseClient(config)
+
+  private val options = PipelineOptionsFactory.create()
+}
+
+class SpannerIOIT extends FlatSpec with Matchers with BeforeAndAfterAll {
+  import SpannerIOIT._
+  implicit val ec: ExecutionContext = ExecutionContext.global
+
+  override def beforeAll(): Unit = {
```
nevillelyh: Also make a helper function for table creation and move the logic to each individual tests? We can also change `_[123]` to something more meaningful in each test to make it more readable.

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
```diff
@@ -0,0 +1,168 @@
...
+      Struct.newBuilder()
+        .set("Key").to(3L)
+        .set("Value").to("foo")
+        .build(),
+      Struct.newBuilder()
+        .set("Key").to(4L)
+        .set("Value").to("bar")
+        .build())
+
+    val awaited = Await.result(result, 10.seconds)
```
nevillelyh: What's the reason for these timeouts?

scio-spanner/src/main/scala/com/spotify/scio/spanner/SpannerIO.scala
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    sc.applyInternal(transform)
+  }
+
+  override protected def write(data: SCollection[Struct], params: WriteP): Future[Tap[Nothing]] =
+    throw new IllegalStateException("SpannerRead is read-only")
+
+  override def tap(params: ReadP): Tap[Nothing] =
+    throw new NotImplementedError("SpannerRead tap is not implemented")
```
nevillelyh: Should be `EmptyTap`?

scio-spanner/src/test/scala/com/spotify/scio/spanner/SpannerIOTest.scala
```diff
@@ -0,0 +1,104 @@
...
+
+object SpannerIOTest {
+  val spannerConfig: SpannerConfig = SpannerConfig
+    .create()
+    .withProjectId("someProject")
+    .withDatabaseId("someDatabase")
+    .withInstanceId("someInstance")
+}
+
+class SpannerIOTest extends PipelineSpec with Matchers {
```
nevillelyh: We have a `ScioIOSpec` so you don't have to handcraft test jobs ;)
https://github.com/spotify/scio/blob/master/scio-test/src/main/scala/com/spotify/scio/testing/ScioIOSpec.scala

Example:
https://github.com/spotify/scio/blob/master/scio-bigtable/src/test/scala/com/spotify/scio/bigtable/BigtableIOTest.scala

scio-spanner/src/main/scala/com/spotify/scio/spanner/SpannerIO.scala
```diff
@@ -0,0 +1,120 @@
...
+}
+
+object SpannerRead {
+  object ReadParam {
+    private[spanner] val DefaultWithTransaction = false
+    private[spanner] val DefaultWithBatching = true
+  }
+
+  sealed trait ReadMethod
+  case class FromTable(tableName: String, columns: Seq[String]) extends ReadMethod
```
regadas: @ClaireMcGinty

```suggestion
  final case class FromTable(tableName: String, columns: Seq[String]) extends ReadMethod
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/SpannerIO.scala
```diff
@@ -0,0 +1,120 @@
...
+  case class FromQuery(query: String) extends ReadMethod
```
regadas: @ClaireMcGinty 

```suggestion
  final case class FromQuery(query: String) extends ReadMethod
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/SpannerIO.scala
```diff
@@ -0,0 +1,120 @@
...
+  )
+}
+
+final case class SpannerWrite(config: SpannerConfig) extends SpannerIO[Mutation] {
+  override type ReadP = Nothing
+  override type WriteP = SpannerWrite.WriteParam
+
+  override protected def write(data: SCollection[Mutation],
+                               params: WriteP): Future[Tap[Nothing]] = {
+    var transform = BSpannerIO.write().withSpannerConfig(config)
```
regadas: @ClaireMcGinty can you do it like this?

```suggestion
    val transform = BSpannerIO
      .write()
      .withSpannerConfig(config)
      .withBatchSizeBytes(params.batchSizeBytes)
      .withFailureMode(params.failureMode)
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/client/Spanner.scala
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+
+package com.spotify.scio.spanner.client
+
+import com.google.cloud.spanner._
+import org.apache.beam.sdk.io.gcp.spanner.SpannerConfig
+
+object Spanner {
+  lazy val defaultInstance: Spanner = {
```
regadas: @ClaireMcGinty 

```suggestion
  private lazy val defaultInstance: Spanner = {
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/client/Spanner.scala
```diff
@@ -0,0 +1,40 @@
...
+    SpannerOptions.newBuilder().build().getService
+  }
+
+  def getDatabaseClient(config: SpannerConfig,
```
regadas:
```suggestion
  def databaseClient(config: SpannerConfig,
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/client/Spanner.scala
```diff
@@ -0,0 +1,40 @@
...
+                        instance: Spanner = defaultInstance): DatabaseClient = {
+    instance.getDatabaseClient(
+      DatabaseId.of(
+        config.getProjectId.get(),
+        config.getInstanceId.get(),
+        config.getDatabaseId.get()
+      ))
+  }
+
+  def getAdminClient(project: String, instance: Spanner = defaultInstance): DatabaseAdminClient =
```
regadas:
```suggestion
  def adminClient(project: String, instance: Spanner = defaultInstance): DatabaseAdminClient =
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/package.scala
```diff
@@ -0,0 +1,78 @@
...
+import com.spotify.scio.io.Tap
+import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.io.gcp.spanner.SpannerConfig
+import org.apache.beam.sdk.io.gcp.spanner.SpannerIO.FailureMode
+
+import scala.concurrent.Future
+
+package object spanner {
+
+  implicit class SpannerScioContext(@transient val self: ScioContext) extends Serializable {
```
regadas:
```suggestion
  implicit class SpannerScioContext(private @transient val self: ScioContext) extends Serializable {
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/package.scala
```diff
@@ -0,0 +1,78 @@
...
+        SpannerRead.FromQuery(query),
+        withTransaction,
+        withBatching
+      )
+
+      self.read(SpannerRead(spannerConfig))(params)
+    }
+  }
+
+  implicit class SpannerSCollection(@transient val self: SCollection[Mutation])
```
regadas:
```suggestion
  implicit class SpannerSCollection(@transient private val self: SCollection[Mutation])
```

scio-spanner/src/main/scala/com/spotify/scio/spanner/SpannerIO.scala
ClaireMcGinty: I chose the `NotImplementedError` because I thought it was something that actually should be implemented in the future for the SpannerRead op. what do you think?

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
ClaireMcGinty: I thought it made since to batch all the Spanner CREATE statements for the sake of simplicity/performance -- I refactored it a bit to provide a Spanner table context per test that handles all the other setup logic

scio-spanner/src/main/scala/com/spotify/scio/spanner/SpannerIO.scala
nevillelyh: Talked IRL. Database source/sinks are not immutable/atomic so reading a tap is not guaranteed to return the same data in the original `SCollection`.

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
nevillelyh: :+1: no strong preference. Either way works.

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  private class ReadableSpannerTable(val tableName: String) {
+    val writeData: FakeSpannerData = fakeData(tableName)
+
+    lazy val readOperationResults: List[Struct] = {
+      val txn = dbClient
+        .readOnlyTransaction()
+        .read(tableName, KeySet.all(), Seq("Key", "Value").asJava)
+
+      List(
```
regadas: @ClaireMcGinty will this brake if we change `FakeSpannerData`?

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
```diff
@@ -0,0 +1,148 @@
...
+    private val read = SpannerRead(config).readWithContext(
+      sc,
+      SpannerRead.ReadParam(
+        readMethod = SpannerRead.FromTable(tableName, Seq("Key", "Value")),
+        withTransaction = true,
+        withBatching = true
+      )
+    ).materialize.map(_.value.toList)
+
+    sc.close()
```
regadas: @ClaireMcGinty 🎨  we could potentially re-use somethings here ```
sc
  .close()
  .waitUntilDone()

read.map(_ should contain theSameElementsAs spannerRows.asStructs)
```

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
ClaireMcGinty: updated it to iterate for the # of elements in fakeData

scio-spanner/src/it/scala/com/spotify/scio/spanner/SpannerIOIT.scala
ClaireMcGinty: done!

PR #1494, opened 2018-10-31
Avoid using kryo coder for TF Schema and Feature

regadas:


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -259,7 +261,8 @@ class TFExampleSCollectionFunctions[T <: Example](val self: SCollection[T]) {
   }
 
   // scalastyle:off method.length
-  private def examplesToFeatures(examples: SCollection[Example]): SCollection[Feature] = {
+  private def examplesToFeatures(examples: SCollection[Example])(
+    implicit coder: Coder[Feature]): SCollection[Feature] = {
```
nevillelyh: This can just be an `implicit val` in the method?

regadas: i think in this case it's ok to have it like that, it's a private method; it will use whatever instance in scope, which in this case will be defined in the public functions

PR #1503, opened 2018-11-06
feature(QueryOps): Allow setting of query dispositions

ccarpita:
This change enables the following settings for Scio BigQuery wrapper when running queries:

* WriteDisposition
* CreateDisposition

For instance, this would enable users to run queries that truncate tables, append to tables,
or fail if the table does not already exist.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
@@ -21,6 +21,7 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.services.bigquery.model._
 import com.spotify.scio.bigquery.client.BigQuery.Client
 import com.spotify.scio.bigquery.{BigQueryUtil, TableRow}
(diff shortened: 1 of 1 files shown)


ccarpita:
@regadas Thank you very much for the quick feedback-- I made the suggested edits, particularly:
1) Ensuring QueryJobConfig privacy
2) Ensuring there cannot be any recursive setting for legacy

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
```diff
@@ -81,129 +82,159 @@ private[client] final class QueryOps(client: Client, tableService: TableOps, job
...
+      val location = extractLocation(query.sql).getOrElse(BigQueryConfig.location)
       tableService.prepareStagingDataset(location)
-      val isLegacy = isLegacySql(sqlQuery, flattenResults)
-      if (isLegacy) {
-        Logger.info(s"Executing legacy query: `$sqlQuery`")
-      } else {
-        Logger.info(s"Executing SQL query: `$sqlQuery`")
-      }
-      val tryRun = run(sqlQuery, destinationTable, flattenResults, isLegacy, dryRun = false)
+      val isLegacy = isLegacySql(query.sql)
```
regadas: @ccarpita This is not used anymore.

PR #1504, opened 2018-11-07
writeRows current behavior is actually append

jonnyjonjon:
making default and throwing exception for other writeDispositions

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQuery.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQuery.scala
@@ -30,7 +30,7 @@ import com.google.auth.oauth2.GoogleCredentials
 import com.google.cloud.hadoop.util.ChainingHttpRequestInitializer
 import com.spotify.scio.bigquery.client.BigQuery.Client
 import com.spotify.scio.bigquery.types.BigQueryType.HasAnnotation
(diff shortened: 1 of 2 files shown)


jto:
I'm not sure I understand what the PR is meant to do. The current behaviour in Scio is consistent with beam's bigquery client (see: https://beam.apache.org/releases/javadoc/2.8.0/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.Write.WriteDisposition.html#WRITE_EMPTY). If for some reason it does not work, I think the bug should be reported and fixed upstream.

jonnyjonjon:
Thanks for the quick reply. Yes, I understand what WRITE_EMPTY is supposed to do. You will notice, though, that in TableOps.scala, Scio drops the writeDisposition parameter entirely in the writeRows method I edited - it does not even pass it to beam.

Under the hood, it's using a beam streaming insert method that only appends (which actually caused a rather nasty bug for us).

The user doesn't get alerted that their disposition parameter is being ignored, nor is there any indication about how the method actually behaves.

This PR isn't meant to change the behavior, just alert a user that WRITE_TRUNCATE will not overwrite, and WRITE_EMPTY will not throw an error if the destination is not empty.

regadas:
Issue #1511

jonnyjonjon:
Great! Thanks for creating the issue. I understand that you'd rather have write dispositions functional, but wouldn't this exception help prevent unexpected behavior for users in the meantime?

regadas:
Closing via #1541

PR #1541, opened 2018-11-19
Fix #1151: add support for more WriteDisposition's in bigquery writeRows

regadas:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala
```diff
@@ -158,40 +145,55 @@ private[client] final class TableOps(client: Client) {
...
+    if (createDisposition == CreateDisposition.CREATE_IF_NEEDED) {
+      service.createTable(table)
+    }
+
+    writeDisposition match {
+      case WriteDisposition.WRITE_TRUNCATE =>
+        delete(tableReference)
+        service.createTable(table)
+      case WriteDisposition.WRITE_EMPTY =>
+        require(service.isTableEmpty(tableReference))
```
regadas: I don't particularly like this, but this `writeRows` method is expected to throw 😞

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala
```diff
@@ -158,40 +145,55 @@ private[client] final class TableOps(client: Client) {
...
+                createDisposition: CreateDisposition): Unit = withBigQueryService { service =>
+    val table = new Table().setTableReference(tableReference).setSchema(schema)
+    if (createDisposition == CreateDisposition.CREATE_IF_NEEDED) {
+      service.createTable(table)
+    }
+
+    writeDisposition match {
+      case WriteDisposition.WRITE_TRUNCATE =>
+        delete(tableReference)
+        service.createTable(table)
```
jonnyjonjon: Wait, isn't a deletion followed by a creation non-atomic? A user might get the intermediate state with the missing table when querying.

regadas: Correct it's not an atomic operation. Documentation states that:

> The replacement may occur in multiple steps - for instance by first removing the existing table, then creating a replacement, then filling it in. This is not an atomic operation, and external programs may see the table in any of these intermediate steps.

Issue #1551, opened 2018-11-23
flatten on SCollection[Iterable[Int]] does not work on 0.7.0-beta1


pishen:
The following code compiles without error in Scio 0.6.1
```scala
sc.parallelize(Seq(1,2,3,4,5)).top(3).flatten
```
But when using Scio 0.7.0-beta1, I met the following error:
```
[error] Main.scala:20:43: ambiguous implicit values:
[error]  both method jBitSetCoder in trait JavaCoders of type => com.spotify.scio.coders.Coder[java.util.BitSet]
[error]  and value jShortCoder in trait JavaCoders of type => com.spotify.scio.coders.Coder[Short]
[error]  match expected type com.spotify.scio.coders.Coder[U]
[error]     sc.parallelize(Seq(1,2,3,4,5)).top(3).flatten
[error]                                           ^
[error] one error found
```

regadas:
Hi @pishen try this:

```
sc.parallelize(Seq(1,2,3,4,5)).top(3).flatten[Int]
```

mpasa:
Flatmap works as expected:

```scala
sc.parallelize(Seq(1,2,3,4,5)).top(3).flatMap(identity)
```

So it seems to be a type inference problem with `T => TraversableOnce[U]` and `Coder[U]`. A workaround is to use an implicit parameter for the coder instead of a context bound, but I'm not sure why that works. We can discuss it in this PR #1552.

jto:
Fixed by #1552

PR #1552, opened 2018-11-24
#1551: Fix flatten type inference changing the coder context bound to an implicit parameter

mpasa:
This PR discusses a possible workaround for #1551

The original problematic code:

```scala
sc.parallelize(Seq(1,2,3,4,5)).top(3).flatten
```

It fails to find a correct implicit in this line:

```scala
def flatten[U: Coder](implicit ev: T => TraversableOnce[U]): SCollection[U]
```

Using a context bound for the coder, the Scala compiler tries to use a `jBitSetCoder` or a `jShortCoder`, which is ambiguous. If we move the context bound to a regular implicit parameter, `flatten` works as expected.

```scala
def flatten[U](implicit ev: T => TraversableOnce[U], coder: Coder[U]): SCollection[U]
```

I'm not 100% about why it works. Using an implicit parameter works whatever the order is. Maybe someone with better knowledge of the compiler can give us a hint.

I haven't included new tests in the PR, but this executes correctly.

```scala
"SCollection" should "support flatten() of iterables" in {
    runWithContext { sc =>
      val p1 = sc.parallelize(Seq(1, 2, 3, 4, 5)).top(3).flatten
      p1 should containInAnyOrder(Seq(3, 4, 5))
    }
}
```

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -388,7 +388,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * Return a new `SCollection[U]` by flattening each element of an `SCollection[Traversable[U]]`.
    * @group transform
    */
-  def flatten[U: Coder](implicit ev: T => TraversableOnce[U]): SCollection[U] =
+  def flatten[U](implicit ev: T => TraversableOnce[U], coder: Coder[U]): SCollection[U] =
     flatMap(ev)
 
   /**
--- a/scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/values/SCollectionTest.scala
@@ -252,10 +252,10 @@ class SCollectionTest extends PipelineSpec {
     runWithContext { sc =>
       val p1 = sc
         .parallelize(Seq(Seq("a b", "c d"), Seq("e f", "g h")))
-        .flatten[String]
+        .flatten
       p1 should containInAnyOrder(Seq("a b", "c d", "e f", "g h"))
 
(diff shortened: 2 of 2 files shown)


regadas:
Hi, thanks for the PR! I found the same  thing regarding the same issue #1551, however because it's but weird I went down the rabbit hole.

given something like this:

```scala
sc.parallelize(Seq(Seq(1, 2, 3, 4, 5))).flatten
```

it will be expanded to something like:

```scala
sc.parallelize[Seq[Int]](
scala.collection.Seq.apply[Seq[Int]](scala.collection.Seq.apply[Int](1, 2, 3, 4, 5)))(coders.this.Coder.seqCoder[Int](coders.this.Coder.intCoder))
.flatten[Nothing]()
```

The reason why I think this happens is due to implicit resolution order. It tries to to resolve `U: Coder` before it resolves `implicit ev: T => TraversableOnce[U]` leading to an implicit divergence, which we also get:

```
diverging implicit expansion for type com.spotify.scio.coders.Coder[K]
```

This leads to un unresolved type `U` hence the `flatten[Nothing]`

Doing the following also shows the same issue:

```
def flatten[U](implicit coder: Coder[U], ev: T => TraversableOnce[U]): SCollection[U]
```

This was filled as an issue some time ago https://issues.scala-lang.org/browse/SI-7205

Given this, I think moving the context bound to a parameter is ok.

mpasa:
Nice. I thought I had tested it with the coder as first implicit parameters and it had worked. Whatever, I've updated the tests to remove the explicit types from the tests. Sorry for the force pushes, I forgot how to write commit messages in English :smiley:

jto:
> The reason why I think this happens is due to implicit resolution order. It tries to to resolve U: Coder before it resolves implicit ev: T => TraversableOnce[U] leading to an implicit divergence, which we also get:

Yeah I'm pretty sure that's what is happening too.

jto:
Thanks for the PR @mpasa 👍

PR #1556, opened 2018-11-26
Deprecate tf schema logic

brianmartin:
Existing schema logic is known to be problematic. We are deprecating in favor of using `tf.data_validation`, an upstream project from Google.

Any existing users who can't immediately switch can use the deprecated functions until deletion in 0.8.x.

cc @regadas @yonromai


scio-tensorflow/src/main/protobuf/schema.proto
```diff
@@ -1,3 +1,5 @@
+// TODO(brianm): Delete when deprecated schema logic goes away
```
regadas: @brianmartin instead of having all of this TODO spread across the codebase I think it's better to have an issue that points to what needs to be deprecated. It's easier to track and not forget.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFScioContextFunctions.scala
```diff
@@ -52,6 +52,7 @@ class TFScioContextFunctions(val self: ScioContext) extends AnyVal {
    * remotely stored `org.tensorflow.metadata.v0.Schema` object available in a DistCache.
    * @group input
    */
+  @deprecated("Schema inference should be done using TensorFlow Data Validation", "Scio 0.7.0")
```
regadas: @brianmartin I would be a bit more specific here and actually mention that this method will be removed.

scio-tensorflow/src/main/protobuf/schema.proto
brianmartin: 👍 #1557

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFScioContextFunctions.scala
brianmartin: 👍 done

PR #1558, opened 2018-11-27
Fix macro generated class directory

regadas:


scio-avro/src/main/scala/com/spotify/scio/avro/AvroSysProps.scala
```diff
@@ -26,6 +26,6 @@ object AvroSysProps {
 
   val DisableDump = SysProp("avro.plugin.disable.dump", "disable class dump")
 
-  val CacheDirectory = SysProp("avro.class.cache.directory", "class cache directory")
+  val CacheDirectory = SysProp("bigquery.class.cache.directory", "class cache directory")
```
regadas: `scio-idea-plugin` right now only supports this sys prop ... in a next release of the plugin we can add support for a `macro.class.cache.directory`

regadas: https://github.com/spotify/scio-idea-plugin/pull/27

PR #1563, opened 2018-12-03
Julient/protobuf backcompat

### LINKED ISSUE #1561: Make sure protobuf files writeen with Scio v0.6 can be read by 0.7
Scio 0.6 gets a protobuf coder using [`val elemCoder = this.getCoder[T]`](https://github.com/spotify/scio/blob/v0.6.1/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L927) which should return an instance of [`ProtobufSerializer` from Chill](https://github.com/spotify/scio/blob/v0.6.1/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala#L106).

`ProtobufSerializer` writes an `Int` and then writes the actual protobuf object [(code)](https://github.com/twitter/chill/blob/develop/chill-protobuf/src/main/java/com/twitter/chill/protobuf/ProtobufSerializer.java#L64-L65).

This ticket goal is to adress 2 issues: 

1. We need to make sure that files written by `saveAsProtobufFile` in Scio 0.6 can be read by Scio 0.7 (which uses [`ProtoCoder`](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/coders/instances/ProtobufCoders.scala#L35)).
1. `ProtobufIO` is [pulling the protobuf `Coder` from the implicit context](https://github.com/spotify/scio/blob/master/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala#L91) which looks like a terrible idea to me. I think we should hardcode `ProtoCoder` instead (or at least have it as a non-implicit parameter with `ProtoCoder` as default)

wdyt about the last point @regadas @ClaireMcGinty ?

jto:
fixes #1561.


scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -88,20 +88,25 @@ object ObjectFileIO {
...
-final case class ProtobufIO[T: ClassTag: Coder](path: String)(implicit ev: T <:< Message)
+final case class ProtobufIO[T: ClassTag](path: String)(implicit ev: T <:< Message)
     extends ScioIO[T] {
   override type ReadP = Unit
   override type WriteP = ProtobufIO.WriteParam
   override final val tapT = TapOf[T]
 
+  private val protoCoder =
+    Coder
+      .protoMessageCoder[Message](classTag[T].asInstanceOf[ClassTag[Message]])
```
jto: The fact the we use an evidence of `T <:< Message` instead of a proper type constraint makes that awkward. I think we should rework that but it's out of the scope of this PR.

PR #1566, opened 2018-12-03
Make tf instances final and self private

regadas:


jto:
👍 to @ClaireMcGinty's comment

regadas:
well `@transient` is only needed if the instance is required to be serialized. I agree that some of them could be `AnyVal`, to be inline with some of other `implicit class`; the other will still be `Serializable`.

regadas:
Actually with a small trick I think we can avoid serialization of on of the implicit class.

PR #1572, opened 2018-12-06
Add elasticsearch 6

regadas:


scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,355 @@
...
+                .map(toDocWriteRequests::apply)
+                .flatMap(ar -> StreamSupport.stream(ar.spliterator(), false));
+
+        final Iterable<List<DocWriteRequest>> chunks =
+            Iterables.partition(docWriteRequests::iterator, maxBulkRequestSize);
+        chunks.forEach(chunk -> {
+          try {
+            final BulkRequest bulkRequest = new BulkRequest().add(chunk.toArray(
+                new DocWriteRequest[0]));
+            final BulkResponse bulkItemResponse = clientSupplier.get().bulk(bulkRequest).get();
```
ClaireMcGinty: depending on how this worker is threaded, would it make more sense to use java Future callbacks instead of blocking for every request?

scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,355 @@
...
+        final Iterable<T> values = c.element().getValue();
+
+        // Elasticsearch throws ActionRequestValidationException if bulk request is empty,
+        // so do nothing if number of actions is zero.
+        if (!values.iterator().hasNext()) {
+          LOG.info("ElasticsearchWriter: no requests to send");
+          return;
+        }
+
+        final Stream<DocWriteRequest> docWriteRequests =
```
ClaireMcGinty: At any point do we want to support `IndexRequest`s/`DeleteRequest`s?

regadas: That's support by `DocWriteRequest`

scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
regadas: Threaded model here is managed by beam and in this case since it's a sink it's kinda "ok" to block, since each list of bulk requests happen in a different thread and we don't have a proper rate-limit/retry police in place.

PR #1575, opened 2018-12-11
Remove unused dependences from scio-parquet

ClaireMcGinty:
Scala steward put up a PR for autovalue, but it seems we aren't actually using that anywhere in scio-parquet. Likewise with `com.google.cloud.bigdataoss.gcs-connector`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,7 +28,6 @@ val annoy4sVersion = "0.7.0"
 val annoyVersion = "0.2.5"
 val asmVersion = "4.9"
 val autoServiceVersion = "1.0-rc4"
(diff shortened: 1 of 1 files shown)


ClaireMcGinty:
cc @regadas if you're aware of any reason these deps are needed

ClaireMcGinty:
@regadas I need to look at that code more, because I saw some configuration params that seemed to originate from `gcs-connector` but no direct usage of the library itself. anyway, put it back into the build file :)

regadas:
correct there's no direct usage only through the hadoop libs which use the connector configuration provided by that config to use the correct connector.

PR #1584, opened 2018-12-18
Map big decimals to BQ numeric data type

lazyval:
Few important points:
- unlike the other types, BD <=> NUMERIC is lossy conversion. Scale is a number of digits after the dot. [NUMERIC has scale =9, precision=38](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric-type), so there is no way to fit BigDecimal with higher scale to NUMERIC data type;
- Scala's BigDecimal is not equal to java one, so for java you would have to access underlying value
- Scala's BigDecimal keeps MathContext as a field.  There is no way to store this in NUMERIC, so math context is also lost and might be different after conversion.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
@@ -78,7 +78,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Float]      => q"$tree.asInstanceOf[Double].toFloat"
         case t if t =:= typeOf[Double]     => q"$tree.asInstanceOf[Double]"
         case t if t =:= typeOf[String]     => q"$tree.toString"
-        case t if t =:= typeOf[BigDecimal] => bdecimal(tree)
+        case t if t =:= typeOf[BigDecimal] => q"BigDecimal($tree.toString)"
 
         case t if t =:= typeOf[ByteString] =>
           val b = q"$tree.asInstanceOf[_root_.java.nio.ByteBuffer]"
@@ -105,11 +105,6 @@ private[types] object ConverterProvider {
           """
         case _ => c.abort(c.enclosingPosition, s"Unsupported type: $tpe")
       }
-    }
-
-    def bdecimal(tree: Tree): Tree = {
-      val mc = q"new java.math.MathContext(38)"
-      q"BigDecimal($tree.toString, $mc)"
     }
 
     def option(tree: Tree, tpe: Type): Tree =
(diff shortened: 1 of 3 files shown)


scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/BigDecimalConversionTest.scala
```diff
@@ -0,0 +1,54 @@
...
+
+package com.spotify.scio.bigquery.types
+
+import com.spotify.scio.bigquery.BigQueryUtil
+import org.scalatest.{FlatSpec, Matchers}
+
+class BigDecimalConversionTest extends FlatSpec with Matchers {
+  case class BigBoy(x: BigDecimal)
+
+  "big decimal values" should "roundtrip if they fit NUMERIC scale" in {
```
regadas: @lazyval we can do property based testing for this. Maybe we can add the test under `ConverterProviderSpec`.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/BigDecimalConversionTest.scala
```diff
@@ -0,0 +1,54 @@
...
+    // at least that's what seen with `bq load` for json data
+    val original = "4288.1111111119"
+    val rounded = "4288.111111112"
+    val before = BigBoy(BigDecimal(original))
+    val after = BigQueryType.fromTableRow[BigBoy](BigQueryType.toTableRow[BigBoy](before))
+
+    after should equal(BigBoy(BigDecimal(rounded)))
+  }
+
+  "bigdecimal" should "be represented as NUMERIC in schema" in {
```
regadas: @lazyval can we add this under `SchemaProviderTest`? we do similar stuff there

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/BigDecimalConversionTest.scala
lazyval: Yeah, I saw it. But it's key idea is that conversions are lossless, whereas bigdecimal is lossy. I can't take an arbitrary BD and expect it to roundtrip just fine: some will not. Limiting the scope of `Arbitrary[BigDecimal]` sounds like a cheating.

regadas: yeah, I think it's ok to just add a custom `Gen` that would generate numbers up to the same scale/precision as BQ `Numeric` and still keep an extra test that verifies the round up.

regadas:
@lazyval Hi! are you close to wrap this up? 😄  we would like to release pretty soon and include this.

lazyval:
@regadas hey! It will take me some time, I'd say I can do this on Monday

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -181,7 +176,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Float]      => q"$s.toFloat"
         case t if t =:= typeOf[Double]     => q"$s.toDouble"
         case t if t =:= typeOf[String]     => q"$s"
-        case t if t =:= typeOf[BigDecimal] => bdecimal(tree)
+        case t if t =:= typeOf[BigDecimal] => q"BigDecimal($tree.toString)"
```
regadas: I was looking into this and the precision here is  `34 >= p <= 38`. so I think it's better and always use the same precision. `38`

PR #1589, opened 2018-12-19
Bump guava to 25.1-jre

regadas:
This version seems to be the most compatible one with current beam version.

This version is already being pulled by `bigdataoss:util:1.9.0` and `bigdataoss:gcsio:1.9.0`


scio-bigtable/src/it/scala/BigtableBenchmark.scala
```diff
@@ -117,7 +118,8 @@ object BigtableBenchmark {
             assert(value == expected)
             value.toStringUtf8
           }
-        }
+        },
+        MoreExecutors.directExecutor()
```
regadas: in higher versions of guava executor is a required param.

PR #1609, opened 2019-01-09
Improve AvroType.toSchema annotation error if a case class is not provided

mpasa:
Currently, if you do something like

```scala
@AvroType.toSchema
class AvroFoo
```

you get a very generic compiler error because `toSchema` expects a case class to generate the Avro schema.

```
[error] Macros.scala:4:4: Invalid annotation List(class AvroFoo extends scala.AnyRef {
[error]   def <init>() = {
[error]     super.<init>();
[error]     ()
[error]   }
[error] })
[error]   @AvroType.toSchema
[error]    ^
```
This could happen if you're new to Scio and you're not used to the API. The error you get is not very descriptive. This PR tries to improve that message to get something like:

```
[error] Macros.scala:4:4: Invalid annotation
[error] 
[error] @AvroType.toSchema annotation requires a case class, which fields will be translated to an Avro schema.
[error] 
[error] Example: case class Foo(fieldA: String, fieldB: Int)
[error] 
[error] Tree:
[error] 
[error] List(class AvroFoo extends scala.AnyRef {
[error]   def <init>() = {
[error]     super.<init>();
[error]     ()
[error]   }
[error] })
[error]           
[error]   @AvroType.toSchema
```
The message can be changed, but I thought that it could be good to have better error messages while generating code, especially for people who are new to Scio.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
@@ -173,7 +173,17 @@ private[types] object TypeProvider {
         """,
          caseClassTree,
          cName.toString())
-      case t => c.abort(c.enclosingPosition, s"Invalid annotation $t")
+      case t =>
+        val error =
+          s"""Invalid annotation:
+             |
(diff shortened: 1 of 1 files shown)


mpasa:
Great idea! Let me update the PR and include some other macros too!

mpasa:
Done, added to both `AvroType.toSchema` and `BigQueryType.toTable`

Issue #1567, opened 2018-12-04
Support type safe arguments in `JobTest`


jto:
Given that scio now supports [type safe arguments](https://github.com/spotify/scio/blob/master/scio-examples/src/main/scala/com/spotify/scio/examples/MinimalWordCountTypedArguments.scala#L36-L41), we could also support them in `JobTest`.

So instead of writing:

```scala
JobTest[com.spotify.scio.examples.WordCount.type]
      .args("--input=in.txt", "--output=out.txt")
// ...
```

one could use:

```scala
JobTest[com.spotify.scio.examples.WordCount.type]
      .args(JobTest.Arguments(
          input = "in.txt",
          output = "out.txt"
      ))
// ...
```

nevillelyh:
@jto same question, still worth pursuing?

PR #1623, opened 2019-01-23
Derive args from case class in JobTest

regadas:
This is a proposal for #1567

### DIFF
--- a/scio-macros/src/main/scala/com/spotify/scio/testing/ArgGenerator.scala
+++ b/scio-macros/src/main/scala/com/spotify/scio/testing/ArgGenerator.scala
@@ -1,0 +1,126 @@
+package com.spotify.scio.testing
+import caseapp.core.util.CaseUtil
+
+sealed trait Arg {
+  type Type
+
+  def name: Option[String]
+
+  def value: Type
+}
+
+final case class IntArg(name: Option[String], value: Int) extends Arg {
+  override type Type = Int
+}
+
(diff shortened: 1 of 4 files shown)


jto:
This looks unnecessarily complex to me. 

We should be able to just pass the `MinimalWordCountTypedArguments.Arguments("in.txt", "out.txt")` since that's what the pipeline expect to get instead of going  `Arguments -> String -> Arguments`. 

wdyt ?

regadas:
@jto 😄 hm, if I'm looking at this right, I don't think we can get way without the conversion to `String` (at least without too many changes). `scio` jobs don't expect that much setup: 

```scala
object Job {
  def main(args: Array[String])
}
```

And because of this `JobTest` does the following:

https://github.com/spotify/scio/blob/3ca4405c612b84b32be91c16e0961b3808855536/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala#L218-L234

Bottom line is, to allow passing a case class without conversion we need to consider rethinking how we setup the job.

scio-macros/src/main/scala/com/spotify/scio/testing/ArgGenerator.scala
```diff
@@ -0,0 +1,182 @@
+/*
+ * Copyright 2016 Spotify AB.
```
ClaireMcGinty: outdated copyright year?

regadas:
yup, let's park this for now!

PR #1626, opened 2019-01-23
Retry bulk chunk requests to Elasticsearch 2, 5, and 6

soundofjw:
Implements retries when saving bulk documents to Elasticsearch.
This is useful for transient connection problems and other failures which are temporary in nature.

The current behavior will cause the entire "Save to Elasticsearch" step to fail and retry, which creates redundant saves to the cluster. There is no existing method to enable a single failed chunk to retry.

This behavior is implemented with two new options, withMaxRetries and withRetryPause, which can be passed in the configuration as "maxRetries" and "retryPause" respectively.

For completeness, this also patches a typo in the name of the "Write to Elasticesarch" transform ;)

---------

TODO:

- [ ] Implement `FluentBackoff`. See comments.

---------

Here we see the logs for the re-attempts in action:

image

### DIFF
--- a/scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
+++ b/scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
@@ -122,8 +122,29 @@ public class ElasticsearchIO {
       return new Bound<>().withMaxBulkRequestSize(maxBulkRequestSize);
     }
 
+    /**
+     * Returns a transform for writing to Elasticsearch cluster.
+     *
+     * @param maxRetries Maximum number of retries to attempt for saving any single chunk of bulk
+     * requests to the Elasticsearch cluster.
+     */
+    public static <T> Bound withMaxRetries(int maxRetries) {
+      return new Bound<>().withMaxRetries(maxRetries);
+    }
+
+    /**
+     * Returns a transform for writing to Elasticsearch cluster.
+     *
+     * @param retryPause Number of seconds to wait between successive retry attempts.
+     */
+    public static <T> Bound withRetryPause(int retryPause) {
+      return new Bound<>().withRetryPause(retryPause);
+    }
+
     public static class Bound<T> extends PTransform<PCollection<T>, PDone> {
       private static final int CHUNK_SIZE = 3000;
+      private static final int DEFAULT_RETRIES = 3;
+      private static final int DEFAULT_RETRY_PAUSE = 5;
 
       private final String clusterName;
       private final InetSocketAddress[] servers;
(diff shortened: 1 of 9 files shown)


soundofjw:
@regadas I have not prior knowledge of this - I think it's a great idea.
I'll add this to the PR proper the coming week. Can you put a "WIP" label on this sucker?

soundofjw:
@regadas FluentBackoff changes in - I did not go through the trouble of adding every FluentBackoff argument. Let me know if you think this is desired - my experience suggests it may be OK to have the defaults.

scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -259,11 +323,50 @@ public void processElement(ProcessContext c) throws Exception {
             Iterables.partition(actionRequests::iterator, maxBulkRequestSize);
 
         chunks.forEach(chunk -> {
+          Exception exception;
+
+          final BackOff backoff = backoffConfig.backoff();
+
+          do {
```
regadas: @soundofjw wdyt of remove this `do { ... } while(true)` in favor of a 
```java
while (BackOffUtils.next(Sleeper.DEFAULT, backoff)) {
 ...
}
```

removes the need for all of this 

```java
try {
  final long sleepTime = backoff.nextBackOffMillis();
  if (sleepTime == BackOff.STOP) {
    break;
  }
  Thread.sleep(sleepTime);
} catch (InterruptedException | IOException e1) {
  LOG.error("Interrupt during backoff", e1);
  break;
}
```

soundofjw: Your proposal will introduce a sleep at the beginning of every attempt:

```java
  public static boolean next(Sleeper sleeper, BackOff backOff)
      throws InterruptedException, IOException {
    long backOffTime = backOff.nextBackOffMillis();
    if (backOffTime == BackOff.STOP) {
      return false;
    }
    sleeper.sleep(backOffTime);
    return true;
  }```

File: scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java [hunk 2]
```diff
@@ -259,11 +323,50 @@ public void processElement(ProcessContext c) throws Exception {
...
+            try {
+              final BulkRequest bulkRequest = new BulkRequest().add(chunk).refresh(false);
```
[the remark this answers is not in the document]
soundofjw: Yes, though it should be noted that when we are indexing with a refresh interval of `-1`, we're getting refreshes anyway (after a certain point, probably hitting the maximum amount of docs that can hang in the bulk queue) - it's really hard to know why this is happening. But to your point, changing this had no effect.

scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -227,18 +284,25 @@ public void processElement(ProcessContext c) throws Exception {
...
+          int maxBulkRequestSize,
+          SerializableFunction<T, Iterable<ActionRequest<?>>> toActionRequests,
+          ThrowingConsumer<BulkExecutionException> error, int maxRetries, int retryPause) {
         this.maxBulkRequestSize = maxBulkRequestSize;
         this.clientSupplier = new ClientSupplier(clusterName, servers);
         this.toActionRequests = toActionRequests;
         this.error = error;
+
+        this.backoffConfig = FluentBackoff.DEFAULT
+            .withMaxRetries(maxRetries)
```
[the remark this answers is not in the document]
soundofjw: If the param were `maxAttempts`, this would make sense - but when we say "3 retries," we are effectively saying 4 attempts. I'd resolve to keep this as is.

scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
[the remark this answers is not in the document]
regadas: Yeah refresh at indexing time can happen do to several things ... document sizes, nr documents, queues... I guess at some point we can offer support, but let's not do it in the same PR.

scio-elasticsearch/es5/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -264,12 +340,53 @@ public void processElement(ProcessContext c) throws Exception {
...
+
         chunks.forEach(chunk -> {
+          Exception exception;
+
+          final BackOff backoff = backoffConfig.backoff();
+
+          do {
+            try {
+              final BulkRequest bulkRequest = new BulkRequest().add(chunk)
+                  .setRefreshPolicy(WriteRequest.RefreshPolicy.NONE);
```
regadas: same here ... this policy is the default one.

scio-elasticsearch/es2/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
regadas: Yeah sorry for poor explanation, yeah correct... what I was trying to say is that I think we can refactor this block by decoupling the logic a little bit and at the same time trying to leverage that util for the retry logic ... I think we can get a better semantic and readability.

jto:
@regadas what's the status of this ?

regadas:
Closed via #1826

PR #1635, opened 2019-01-28
Add support for reusable-side inputs for filter/join purposes

spkrka:


jto:
Hi @spkrka,

Thanks for your PR. A few remark though.

1. Could you provide some context on why this PR is needed ? I've read the code and I have to admit that it's non-obvious.
1. When you submit a PR, please make sure that scalastyle is happy. You can see that circleci failed because of that.
1. Please do not rename existing method like you did with `combineAsMapSideInput` if you didn't make any change in their body.
1. Do not reformat existing code. It makes it harder to review the PR since we have to figure out what is an actual change and what's just cosmetics. I'd also like to point out that we always format the code with scalafmt so you should not have to do anything beside compiling to get it right
1. Please also do make sure that the code actually compiles before you submit the PR. Your code currently has 8 compilation errors.

spkrka:
Thanks for the feedback! Sorry about submitting non-working code, I was hoping to get feedback on the feature itself before finalizing the code.

I created a new PR with a better commit message, so I'll close this one. See https://github.com/spotify/scio/pull/1643

1. Should be answered by the commit message in the next PR.
2. Next PR seems to pass the build/scalastyle.
3. Done in next PR.
4. I've tried to avoid that in the next PR, but it seems the build process itself reformats the code.
5. Done in next PR.

PR #1642, opened 2019-01-30
add Coder matchers to scio-test

### LINKED ISSUE #1631: Create helper methods to test coders
We already have them!
Extract the checkCoder, checkSer etc methods from coders.scala into a utility in scio-test.

ClaireMcGinty:
fix #1631

I kept the coder serializable checks as plain Assertions, not Matchers, since it's such a simple use case although if we want they could become some kind of `coderIs(...)`-style Matcher too.


scio-test/src/test/scala/com/spotify/scio/testing/CoderMatchersTest.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import scala.collection.JavaConverters._
+import com.spotify.scio.coders.Coder
+import com.spotify.scio.testing.CoderMatchers._
+import org.apache.beam.sdk.coders.{AtomicCoder, StringUtf8Coder}
+import org.scalatest.{FlatSpec, Matchers}
+import org.scalatest.exceptions.TestFailedException
+
+class CoderMatchersTest extends FlatSpec with Matchers {
+  class Foo(val id: String) {
```
regadas: why not just use a case class?

scio-test/src/test/scala/com/spotify/scio/testing/CoderMatchersTest.scala
```diff
@@ -0,0 +1,86 @@
...
+  private def incorrectCoder: Coder[Foo] =
+    Coder.beam(new AtomicCoder[Foo] {
+      override def encode(value: Foo, outStream: OutputStream): Unit =
+        StringUtf8Coder.of().encode(value.id, outStream)
+      override def decode(inStream: InputStream): Foo =
+        new Foo(StringUtf8Coder.of().decode(inStream) + "wrongBytes")
+    })
+
+  "CoderMatchers" should "support roundtrip" in {
+    "foo" coderShould roundtrip()
```
regadas: 🎨 Foo("something") since you will use it for testing a wrong coder?

scio-test/src/test/scala/com/spotify/scio/testing/CoderMatchersTest.scala
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright 2016 Spotify AB.
```
regadas: 2019

regadas:
@ClaireMcGinty seems like `JodaCoders` tests were duplicated?

scio-test/src/main/scala/com/spotify/scio/testing/CoderMatchers.scala
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.beam.sdk.coders.{Coder => BCoder}
+import org.apache.beam.sdk.options.{PipelineOptions, PipelineOptionsFactory}
+import org.apache.beam.sdk.util.{CoderUtils, SerializableUtils}
+import org.scalactic.Equality
+import org.scalatest.Assertion
+import org.scalatest.Matchers._
+
+import scala.reflect.ClassTag
+
+object CoderMatchers {
```
regadas: Also, since these are assertions maybe `CoderAssertions`?

scio-test/src/test/scala/com/spotify/scio/testing/CoderMatchersTest.scala
ClaireMcGinty: that....is a very good point. fixed!

PR #1643, opened 2019-01-31
Add support for reusing SideInputs.

spkrka:
This is useful for avoiding recreating side inputs when you want to join multiple things
against the same side input.

Example:
There is a workflow using very big inputs, but you want to filter it to run for a small subset of data.
To avoid big shuffles you could hash-filter all big inputs against the small subset first.

Instead of doing:

    val sarsCollection = sarInput.keyBy(_.getUserId.toString)
    val userdata = userdataInput.keyBy(_.getUserId.toString).hashJoin(sarsCollection)
    val pipelinekeys = keyInput.keyBy(_.getUserId.toString).hashJoin(sarsCollection)

which would create two separate but identical side inputs, you could do this:

    val sarsSide = sarInput.map(_.getUserId.toString).toSideSet
    val userdata = userdataInput.keyBy(_.getUserId.toString).hashIntersectByKey(sarsSide)
    val pipelinekeys = keyInput.keyBy(_.getUserId.toString).hashIntersectByKey(sarsSide)

which only creates a single side input.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideSet.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideSet.scala
@@ -16,6 +16,4 @@
  */
 package com.spotify.scio.values
 
-import scala.collection.mutable.{ArrayBuffer, Map => MMap}
-
-case class SideSet[K](side: SideInput[MMap[K, ArrayBuffer[Unit]]])
+case class SideSet[T](side: SideInput[Set[T]])
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideSet.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideSet.scala
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2019 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.spotify.scio.values
+
+import scala.collection.mutable.{ArrayBuffer, Map => MMap}
+
+case class SideSet[K](side: SideInput[MMap[K, ArrayBuffer[Unit]]])
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairHashSCollectionFunctions.scala
@@ -36,8 +36,17 @@ class PairHashSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
    */
   def hashJoin[W: Coder](that: SCollection[(K, W)])(implicit koder: Coder[K],
                                                     voder: Coder[V]): SCollection[(K, (V, W))] =
+    hashJoin(SideMap(combineAsMapSideInput(that)))
+
+  /**
+   * Perform an inner join with a SideMap.
+   *
+   * @group join
+   */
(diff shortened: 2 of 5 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SideSet.scala
```diff
@@ -0,0 +1,21 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.spotify.scio.values
+
+import scala.collection.mutable.{ArrayBuffer, Map => MMap}
+
+case class SideSet[K](side: SideInput[MMap[K, ArrayBuffer[Unit]]])
```
anish749: We might use something like this have a separate function `combineAsSetSideInput`
```suggestion
case class SideSet[K](side: SideInput[Set[K]])
```
Just thinking.

spkrka: That would work too, but would mean more code - by defining Set in terms of Map some things become simpler.
This function in particular becomes simple:
`def toSideSet(implicit coder: Coder[T]): SideSet[T] = SideSet(self.map((_, ())).toSideMap.side)`

spkrka:
I think the latest commit with the suggested changes caused this error:
```
[info] - should support hashIntersectByKey() with empty RHS *** FAILED *** (220 milliseconds)
[info]   org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.util.NoSuchElementException: Empty PCollection accessed as a singleton view.
[info]   at org.apache.beam.runners.direct.DirectRunner$DirectPipelineResult.waitUntilFinish(DirectRunner.java:332)
[info]   at org.apache.beam.runners.direct.DirectRunner$DirectPipelineResult.waitUntilFinish(DirectRunner.java:302)
[info]   at org.apache.beam.runners.direct.DirectRunner.run(DirectRunner.java:197)
[info]   at org.apache.beam.runners.direct.DirectRunner.run(DirectRunner.java:64)
[info]   at org.apache.beam.sdk.Pipeline.run(Pipeline.java:313)
[info]   at org.apache.beam.sdk.testing.TestPipeline.run(TestPipeline.java:350)
[info]   at org.apache.beam.sdk.testing.TestPipeline.run(TestPipeline.java:331)
[info]   at com.spotify.scio.ScioContext$$anonfun$close$1.apply(ScioContext.scala:489)
[info]   at com.spotify.scio.ScioContext$$anonfun$close$1.apply(ScioContext.scala:476)
[info]   at com.spotify.scio.ScioContext.requireNotClosed(ScioContext.scala:542)
[info]   ...
[info]   Cause: java.util.NoSuchElementException: Empty PCollection accessed as a singleton view.
[info]   at org.apache.beam.sdk.values.PCollectionViews$SingletonViewFn.getDefaultValue(PCollectionViews.java:181)
[info]   at org.apache.beam.sdk.values.PCollectionViews$SingletonViewFn.apply(PCollectionViews.java:209)
[info]   at org.apache.beam.sdk.values.PCollectionViews$SingletonViewFn.apply(PCollectionViews.java:146)
[info]   at org.apache.beam.runners.direct.SideInputContainer$SideInputContainerSideInputReader.get(SideInputContainer.java:261)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.SimpleDoFnRunner.sideInput(SimpleDoFnRunner.java:304)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.SimpleDoFnRunner.access$900(SimpleDoFnRunner.java:77)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.SimpleDoFnRunner$DoFnProcessContext.sideInput(SimpleDoFnRunner.java:593)
[info]   at com.spotify.scio.values.SingletonSideInput.get(SideInput.scala:95)
[info]   at com.spotify.scio.values.SideInput$class.getCache(SideInput.scala:45)
[info]   at com.spotify.scio.values.SingletonSideInput.getCache(SideInput.scala:93)
```

I guess that means that
```scala
  private def combineAsSet[A: Coder](c: SCollection[A]): SideInput[Set[A]] =
    c.aggregate[Set[A], Set[A]](Aggregator.prepareSemigroup(x => Set(x))).asSingletonSideInput
```
does not work for empty SCollections?

PR #1644, opened 2019-01-31
Improve Bloom Filter aggregations in Sparse Joins

anish749:
BloomFilters used in Sparse Join implementations use Algebird aggregators. This step incurs a good cost in terms of memory and cpu when using `.aggregate(com.twitter.algebird.Aggregator)` because of the way BloomFilters are modelled in Algebird. The aggregate goes as prepare -> sum -> present. For BloomFilterAggregator, the prepare step creates a BFMonoid for each element, and then the sum adds up the monoids. This means that all the keys becomes its own monoid, which takes up a lot of intermediate space. (Around 8 times more space than the keys itself). This can be seen when we expand the transforms in Dataflow.

screenshot 2019-01-31 at 19 48 32
Prepare step of aggregation

screenshot 2019-01-31 at 19 50 24
Sum operation of aggregation

The vCPU time for the sum operation is around 14 hrs, which is quite high.

Changing this to use our own combine functions, provides significant performance improvements.
`.aggregate(bfAggregator.monoid.zero)(_ + _, _ ++ _)`
screenshot 2019-01-31 at 19 54 09

By removing prepare, the intermediate step which was having an output more than 8 times the size of the keys is now removed. The vCPU time has also come down to 5 hrs. (35% of what it was previously)

We noticed up 20% less overall CPU usage on one of our jobs with this change.

Also in this PR.

Added the sparse lookups in a transform to make it look pretty in DataFlow UI.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -345,8 +345,8 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
     val thatBfSIs = that.optimalKeysBloomFiltersAsSideInputs(thatNumKeys, fpProb)
     val n = thatBfSIs.size
 
-    val thisParts = self.partition(n, { case ( key, _ ) => key.hashCode() % n })
-    val thatParts = that.partition(n, { case ( key, _ ) => key.hashCode() % n })
+    val thisParts = self.partition(n, { case (key, _) => key.hashCode() % n })
+    val thatParts = that.partition(n, { case (key, _) => key.hashCode() % n })
 
     thisParts.zip(thatParts).zip(thatBfSIs).map {
       case ((lhs, rhs), bfsi) =>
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -488,14 +487,14 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
     val numKeysPerPartition = if (bfSettings.numBFs == 1) thisNumKeys.toInt else bfSettings.capacity
     val n = bfSettings.numBFs
+    val width = BloomFilter.optimalWidth(numKeysPerPartition, fpProb).get
+    val numHashes = BloomFilter.optimalNumHashes(numKeysPerPartition, width)
+    val bfAggregator = BloomFilterAggregator[K](numHashes, width)
     self
-      .partition(n, _._1.hashCode() % n)
+      .partition(n, { case (key, _) => key.hashCode() % n })
       .map { me =>
-        val width = BloomFilter.optimalWidth(numKeysPerPartition, fpProb).get
```
regadas: 👍 moving this out

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -488,14 +487,14 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
-        val numHashes = BloomFilter.optimalNumHashes(numKeysPerPartition, width)
-        val bfAggregator = BloomFilterAggregator[K](numHashes, width)
         me.keys
-          .aggregate(bfAggregator)
+          .aggregate(bfAggregator.monoid.zero)(_ + _, _ ++ _)
```
regadas: this is a really nice optimization! 👍

anish749:
There are some more bugs.. I'll create another PR soon..!

regadas:
@anish749 interesting ... around this PR or on other stuff?

PR #1646, opened 2019-01-31
resolves #1396

### LINKED ISSUE #1396: store more build info metadata in benchmark datastore

andrisnoko:
Add scio and beam version


scio-test/src/it/scala/com/spotify/ScioBatchBenchmark.scala
```diff
@@ -366,7 +366,9 @@ abstract class Benchmark(val extraConfs: Map[String, Array[String]] = null) {
 
   def run(projectId: String,
           prefix: String,
-          args: Array[String]): Iterable[Future[BenchmarkResult]] = {
+          args: Array[String],
+          scioVersion: String,
```
ClaireMcGinty: `scioVersion`/`beamVersion` probably don't even have to be passed in as function args, they can be computed in `run()` itself like `username` is. wdyt?

PR #1654, opened 2019-02-01
Add HashPartitioning and fix negative hashCode

anish749:
For large SCollections, we create multiple Bloom Filters for some operations. This means partitioning SCollections into multiple smaller ones. This is achieved using keys.hashCode % n, which is a 💣. It blows up because hashCode can be negative, and the partitioning function expects something in the range [0, n) . This is fixed using a `nonNegativeMod` much like Spark. 

Also added APIs for Hash Partitioning a SCollection because that helped remove some of the duplicate code.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -96,14 +96,4 @@ private[scio] object ScioUtil {
     }
   }
 
-  /**
-   * Modulo operation, This always return a non negative value in the range [0, mod)
-   * Returns incorrect value for `mod(Int.MinValue, Int.MaxValue)` as Int.MaxValue - 1
-   * but remainder of Int.MinValue % Int.MaxValue is -1
-   */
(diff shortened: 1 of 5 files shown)


scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -96,4 +96,14 @@ private[scio] object ScioUtil {
     }
   }
 
+  /**
+   * Modulo operation, This always return a non negative value in the range [0, mod)
+   * Returns incorrect value for `mod(Int.MinValue, Int.MaxValue)` as Int.MaxValue - 1
+   * but remainder of Int.MinValue % Int.MaxValue is -1
+   */
+  @inline def mod(value: Int, mod: Int): Int = {
```
regadas: @anish749 couldn't we also consider using `Math.floorMod` since partitions can never be negative?

anish749: I wasn't aware of that. But looks like it can be the best solution. I'll try that out and push. Thanks for pointing out 👍

PR #1663, opened 2019-02-06
Add BinaryIO output.

psobot:
(cc @regadas)

This PR adds `BinaryIO` (name up for debate) which allows us to write `SCollection[Array[Byte]]` to files. Think of this as `TextIO`, but with no delimiters, any arbitrary bytes rather than strings, and write-only (for now).

More tests to be added once I confirm that this might be something we want to move forward with. My team has found that certain Dataflow jobs produce output that's ~60% smaller than flat text files with this change, and while this might not be widely useful, it seems like something that could be used by other Scio users.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
@@ -40,7 +40,7 @@ final case class BinaryIO(path: String) extends ScioIO[Array[Byte]] {
   override def testId: String = s"BinaryIO($path)"
 
   override def read(sc: ScioContext, params: ReadP): SCollection[Array[Byte]] =
-    throw new IllegalStateException("BinaryIO is read-only")
+    throw new IllegalStateException("BinaryIO is write-only")
 
   override def write(data: SCollection[Array[Byte]], params: WriteP): Future[Tap[Nothing]] = {
     data.applyInternal(
--- a/scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2019 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.io
+
+import java.io.OutputStream
+import java.nio.channels.{Channels, WritableByteChannel}
+
+import com.spotify.scio.ScioContext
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
```diff
@@ -0,0 +1,97 @@
...
+  override def tap(params: Nothing): Tap[Nothing] = EmptyTap
+}
+
+object BinaryIO {
+  final case class WriteParam(suffix: String = ".bin",
+                              numShards: Int = 0,
+                              compression: Compression = Compression.UNCOMPRESSED)
+
+  private final class BytesSink extends FileIO.Sink[Array[Byte]] {
+    @SuppressWarnings(Array("org.wartremover.warts.Var"))
```
regadas: @psobot I think we can remove this, we don't use wartremover.

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
```diff
@@ -0,0 +1,97 @@
...
+    def write(datum: Array[Byte]): Unit = {
+      if (this.channel == null) {
+        throw new IllegalStateException("Trying to write to a BytesSink that has not been opened")
+      }
+
+      this.channel.write(datum)
+    }
+  }
+
+  def writeBinary(
```
regadas: @psobot I think I would move this logic to the `BinaryIO.write` it would be consistent with all IO's which normally will expose read/write

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
```diff
@@ -0,0 +1,97 @@
...
+object BinaryIO {
+  final case class WriteParam(suffix: String = ".bin",
+                              numShards: Int = 0,
+                              compression: Compression = Compression.UNCOMPRESSED)
+
+  private final class BytesSink extends FileIO.Sink[Array[Byte]] {
+    @SuppressWarnings(Array("org.wartremover.warts.Var"))
+    @transient private var channel: OutputStream = _
+
+    def open(channel: WritableByteChannel): Unit =
```
regadas: @psobot for these methods we also tend to add `override`

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
psobot: Whoops - accidentally copied from my project where we did use wartremover. 👍

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
```diff
@@ -0,0 +1,90 @@
...
+ */
+final case class BinaryIO(path: String) extends ScioIO[Array[Byte]] {
+  override type ReadP = Nothing
+  override type WriteP = BinaryIO.WriteParam
+  override final val tapT = EmptyTapOf[Array[Byte]]
+
+  override def testId: String = s"BinaryIO($path)"
+
+  override def read(sc: ScioContext, params: ReadP): SCollection[Array[Byte]] =
+    throw new IllegalStateException("BinaryIO is read-only")
```
brianmartin: typo: s/read-only/write-only

regadas: good catch 😄

nevillelyh:
Looks like this might concatenate multiple `Array[Byte]` in non-deterministic order (depending on worker, shuffle, etc.). So the user must explicitly include a trailing delimiter to each byte array, or else the output files can't be read back properly. Is that by design?

PR #1666, opened 2019-02-07
Remove the usage of Future around ScioContext and Tap's

### LINKED ISSUE #1624: Avoid using ExecutionContext.Implicits.global in ContextScioResult
According to [this line](https://github.com/spotify/scio/blob/v0.7.0/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L507), Scio is using `ExecutionContext.Implicits.global` when running the Future `f` on next line, which wraps a blocking call `internal.waitUntilFinish()`. This may starve the global execution context since `ExecutionContext.Implicits.global` only contains limited and uncustomizable number of threads.

Would it be possible to make the ExecutionContext into an implicit parameter and let users decide what ExecutionContext they want to use?

regadas:
This introduces the concept of `ClosedScioContext` and `ClosedTap`

Fixes #1624

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -27,7 +27,7 @@ import java.nio.file.Files
 
 import com.google.datastore.v1.{Entity, Query}
 import com.spotify.scio.coders.{Coder, CoderMaterializer}
-import com.spotify.scio.io.{Tap, _}
+import com.spotify.scio.io._
 import com.spotify.scio.metrics.Metrics
 import com.spotify.scio.options.ScioOptions
 import com.spotify.scio.testing._
@@ -50,7 +50,6 @@ import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.collection.mutable.{Buffer => MBuffer}
 import scala.concurrent.duration.Duration
-import scala.concurrent.{Future, Promise}
 import scala.io.Source
 import scala.reflect.ClassTag
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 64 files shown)


scio-test/src/test/scala/com/spotify/scio/io/TapTest.scala
```diff
@@ -80,29 +79,29 @@ class TapTest extends TapSpec {
...
+//    val sc = ScioContext()
+//    val f = sc
+//      .parallelize(Seq(1, 2, 3))
+//      .map(newSpecificRecord)
+//      .saveAsAvroFile(dir.getPath)
+//    f.isCompleted shouldBe false
+//    sc.close().waitUntilFinish() // block non-test runner
+//    f.isCompleted shouldBe true
+//    FileUtils.deleteDirectory(dir)
+//  }
```
jto: Are those tests broken ?

regadas: @jto nice catch. Forgot to remove them, since they don't make much sense given the new changes.

jto:
@ClaireMcGinty WDYT ?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -294,6 +293,68 @@ object ScioContext {
...
+      val wait = duration match {
+        case Duration.Inf => 0
+        case d            => d.toMillis
+      }
+      pipelineResult.waitUntilFinish(time.Duration.millis(wait))
+    } catch {
+      case _: InterruptedException =>
+        if (cancelJob) {
+          pipelineResult.cancel()
+          throw new PipelineExecutionException(
```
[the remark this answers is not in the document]
regadas: Hm, yeah good point, previously this exception was encapsulated in the Future. Given the new semantics makes sense we throw this.

regadas:
@ClaireMcGinty `mkTap` is still used.

jto:
I think you can merge this to the `0.8` branch :)

PR #1671, opened 2019-02-07
Add simple toString to coders

### LINKED ISSUE #1667: Implement proper toString methods on Coder
For easier debuging when something fails

regadas:
Fixes #1667


scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -60,23 +60,41 @@ Cannot find an implicit Coder instance for type:
...
                                     construct: Seq[Any] => T,
                                     destruct: T => Array[Any])
-    extends Coder[T]
+    extends Coder[T] {
+  override def toString: String = {
+    val str = cs
+      .map {
+        case (k, v) => s"($k, $v)"
+      }
+    s"Record($typeName, ${str.mkString(", ")})"
```
ClaireMcGinty: nice!

PR #1702, opened 2019-02-25
AvroIO to work with SpecificRecordBase case classes

andrisnoko:


jto:
I'm wondering if we should split `AvroIO` in 2 exclusive classes. 

Currently it mixes 2 more or less incompatible use cases:
- reading / writing a class by providing a schema explicitly
- reading / writing a class that `<: SpecificRecordBase`

The current design is a bit odd since if you use a class that extends `SpecificRecordBase` AND you pass a schema explicitly, the schema will simply be ignored. If I understand corretlly, it is what this PR is meant to fix.

How about we make the weird use case (instance of `SpecificRecordBase` AND explicit schema) impossible by design instead ?

smth like:

```scala
sealed trait AvroIO[T] extends ScioIO[T] { ... }
final case class SpecificRecordIO[T <: SpecificRecordBase](path: String) extends AvroIO[T]  { ... }
final case class SchemaAvroIO[T <: GenericRecord](path: String, schema: Schema) extends AvroIO[T] { ... }

object AvroIO {
    def apply[T <: SpecificRecordBase](path: String) = SpecificRecordIO[T](path)
    def apply[T <: GenericRecord](path: String, schema: Schema) = SchemaAvroIO[T](path, schema)
}
```

Note how with this design, the behaviour becomes obvious, either you pass a `GenericRecord` and a `Schema` and that's what's used OR you pass an instance of `SpecificRecord`.
I think it also removes the need for a `ClassTag`.

scio-macros/src/main/scala/com/spotify/scio/coders/AvroCoderMacros.scala
```diff
@@ -37,7 +37,10 @@ private[coders] object AvroCoderMacros {
     _root_.com.spotify.scio.coders.Coder.beam(
       _root_.org.apache.beam.sdk.coders.AvroCoder.of[$companioned](
         classOf[$companioned],
-        ${companionType}.getClassSchema()
+        classOf[$companioned]
```
sisidra: `getSchema` is a public member, so no need to use reflection here - `new $companioned().getSchema()` should work.

jto: Yes indeed

sisidra:
I think separating `SpecificRecordIO` and `GenericRecordIO` makes sense.
Tackling avrohugger compatibility within `SpecificRecordIO` would still be needed.

sisidra:
When it comes to the original issue, I managed to hack avro schema resolution implementation:
https://gist.github.com/sisidra/486ae72f738cf5e7f50d37e3020d260d
I'm not suggesting to use it in scio tho :)
Given that beam AvroIO depends on this Avro functionality, I would suggest to not use beam implementation of `AvroIO.read()` but build new `AvroIO.Read` object in Scio and use schema resolution methods that work for setting schema there.
Or try contribute upstream (beam or avro)...

wdyt?

regadas:
The main purpose of this use case was to provide support for `avrohugger` case classes.

@jto i think you are on point with the split ... that removes the `schema` null check. let's do a separate PR with it!
 
@sisidra yeah having our one is smth i want to avoid 😄 I'm actually thinking in submitting a proposal to open up the `AvroIO.Read` and i think it's the best way to go about it.

@andrisnoko I think we should just keep the changes regarding the usage of `getSchema` because with that we avoid reflection.

Given the subtle specific changes need for `avrohugger` I think this should probably be implemented in a contrib module.

scio-core/src/main/scala/com/spotify/scio/coders/instances/kryo/AvroSerializer.scala
```diff
@@ -58,7 +58,11 @@ private[coders] class SpecificAvroSerializer[T <: SpecificRecordBase] extends KS
 
   private def getCoder(cls: Class[T]): AvroCoder[T] =
     cache.getOrElseUpdate(cls,
-                          Try(cls.getMethod("getClassSchema").invoke(null).asInstanceOf[Schema])
+                          Try(
+                            cls
+                              .getMethod("getSchema")
+                              .invoke(cls.newInstance())
```
regadas: reflection can be avoided here as well

andrisnoko:
Thank you all for the feedback. I will break this PR to small components.

PR #1704, opened 2019-02-26
Add spanner MutationGroup coder

### LINKED ISSUE #1703: scio-spanner: KryoException when using MutationGroup
```
com.esotericsoftware.kryo.KryoException: java.lang.UnsupportedOperationException
Serialization trace:
columns (com.google.cloud.spanner.Mutation)
mutations (org.apache.beam.sdk.io.gcp.spanner.MutationGroup)
Disconnected from the target VM, address: '127.0.0.1:51278', transport: 'socket'
```
      val spannerOutput = SpannerIO.write()
        .withProjectId(projectId)
        .withInstanceId(spannerInstanceId)
        .withDatabaseId(spannerDatabaseId)
        .grouped()

      val list = sc.parallelize(0 to 10)

      list
      .flatMap(p => Try {
        val taskMutation = Mutation.newInsertOrUpdateBuilder("Tasks")
          .set("TaskId").to(UUID.randomUUID().toString)          
          .set("ReceivedAt").to(Timestamp.now())
          .build()

        val tests = (0 to 5).map( i => {
          Mutation.newInsertOrUpdateBuilder("test")
            .set("id").to(i)
            .set("time").to(Timestamp.now())
            .set("value").to(i)
            .build()
        })

        MutationGroup.create(taskMutation, tests.asJava)

      } match {
        case Success(value) => Some(value)
        case Failure(exception) =>
          println(exception)
          None
      })
      .saveAsCustomOutput("spannerOut", spannerOutput)

      sc.close().waitUntilFinish()

regadas:
Fixes #1703

### DIFF
--- a/scio-spanner/src/main/scala/com/spotify/scio/spanner/package.scala
+++ b/scio-spanner/src/main/scala/com/spotify/scio/spanner/package.scala
@@ -22,7 +22,7 @@ import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.Tap
 import com.spotify.scio.values.SCollection
 import org.apache.beam.sdk.{coders => bcoders}
-import org.apache.beam.sdk.io.gcp.spanner.{ReadOperation, SpannerConfig}
+import org.apache.beam.sdk.io.gcp.spanner.{MutationGroup, ReadOperation, SpannerConfig}
 import org.apache.beam.sdk.io.gcp.spanner.SpannerIO.FailureMode
 
 import scala.concurrent.Future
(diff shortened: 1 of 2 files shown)


scio-spanner/src/main/scala/com/spotify/scio/spanner/package.scala
```diff
@@ -22,7 +22,7 @@ import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.Tap
 import com.spotify.scio.values.SCollection
 import org.apache.beam.sdk.{coders => bcoders}
-import org.apache.beam.sdk.io.gcp.spanner.{ReadOperation, SpannerConfig}
+import org.apache.beam.sdk.io.gcp.spanner.{MutationGroup, ReadOperation, SpannerConfig}
```
ClaireMcGinty: wdyt about adding an implicit coder for [Transaction](https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/Transaction.java), too? It's the only other `Serializable` class left in Beam's Spanner package that we're not including here :)

regadas: I was thinking about this and I'm not sure that it would be used. The only way to supply `Transaction` is through a `PCollectionView` and i don't think the users should be moving around `Transaction`'s because the only way to create one is to actually connect to spanner and get one.

PR #1711, opened 2019-02-28
Rework print metrics

regadas:


scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
```diff
@@ -232,8 +232,11 @@ object DatastoreLogger {
 
   lazy val Storage: Datastore = DatastoreHelper.getDatastoreFromEnv
   val Kind = "Benchmarks"
-  val OrderByBuildNumQuery = s"SELECT * from ${Kind}_%s ORDER BY buildNum DESC LIMIT 2"
-  val WhereBuildNumQuery = s"SELECT * from ${Kind}_%s WHERE buildNum = %s"
+
+  def latestBuildNumQuery(benchmarkName: String) =
```
ClaireMcGinty: can these be private? the stylecheck is failing since they're public with no return type :)

scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
```diff
@@ -267,112 +270,128 @@ class DatastoreLogger(metricsToCompare: Set[String]) extends BenchmarkLogger[Try
...
-                )
-                .build())
-          }
-          .map(_.getBatch.getEntityResults(0).getEntity)
-      case None =>
-        val comparisonMetrics = Storage.runQuery(
+  def latestBenchmark(benchmarkName: String): Option[Entity] =
+    benchmarksByBuildNum(benchmarkName, Nil).headOption
+
+  def benchmarksByBuildNum(benchmarkName: String, buildNums: List[Long]): List[Entity] = {
```
ClaireMcGinty: maybe `benchmarksByBuildNums`? or `benchmarksForBuildNums`?

scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
```diff
@@ -267,112 +270,128 @@ class DatastoreLogger(metricsToCompare: Set[String]) extends BenchmarkLogger[Try
...
+        (if (delta > 0) "+" else "") + "%.2f%%".format(delta)
+      }
+      PrettyPrint.print(k, "%15.2f%15.2f%15s".format(prev, curr, signed))
     }
+  }
 
   // TODO: move this to email generator
-  def printMetricsComparison(benchmarks: Iterable[String],
-                             buildNums: Option[(String, String)] = None): Unit = {
+  def printMetricsComparison(benchmarks: Iterable[String], buildNums: List[Long]): Unit =
```
ClaireMcGinty: I feel like these two `def printMetricsComparison` methods should have clearer function names to differentiate them? since this method is about comparing Benchmarks, whereas the function on line 354 is directly comparing two datastore Entities. 

also on that note, if we're comparing benchmarks across multiple builds, I think my assumption would be that the delta percentages are all relative to the oldest build. I might be alone in that though :P

scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
regadas: ¯\_(ツ)_/¯

scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
regadas: the idea is to in a follow up PR not expose the Entity ... coz the way it is right now with the usage of `BenchmarkResult` is just messy. So:

```
def printMetricsComparison(prev: Entity, curr: Entity)

becomes

def printMetricsComparison(prev: BenchmarkResult, curr: BenchmarkResult)
```

well, i actually like the idea of seeing the deltas between consecutive builds and this doesn't invalidate the idea of adding the deltas relative to the older one.

ClaireMcGinty: oh, cool, didn't know there would be a followup PR. makes sense! excited to see the benchmark code cleaned up a bit 💯

PR #1712, opened 2019-03-01
Use BenchmarkResult as entity

regadas:


scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
```diff
@@ -148,74 +127,111 @@ final case class ScioBenchmarkLogger[F[_]](loggers: BenchmarkLogger[F]*) {
...
                 createTime: String,
                 jobMetrics: JobMetrics): BenchmarkResult = {
-    val startTime: LocalDateTime = DateTimeParser.parseLocalDateTime(createTime)
-
     val metrics = jobMetrics.getMetrics.asScala
       .filter(metric => StreamingMetrics.contains(metric.getName.getName))
-      .map(m => (m.getName.getName, m.getScalar.toString))
-      .sortBy(_._1)
-      .toMap
+      .map(m => Metric(m.getName.getName, m.getScalar.toString))
```
ClaireMcGinty: maybe we should just represent `Metric.value` as a scalar, not a string? seems like there's a lot of conversions and the `compareMetrics` method assumes they can all be converted to Doubles anyway

scio-test/src/it/scala/com/spotify/scio/benchmarks/ScioBenchmark.scala
```diff
@@ -148,74 +127,111 @@ final case class ScioBenchmarkLogger[F[_]](loggers: BenchmarkLogger[F]*) {
...
+                           state: String,
+                           extraArgs: Array[String],
+                           metrics: List[Metric],
+                           scioVersion: String,
+                           beamVersion: String) {
+
+  def compareMetrics(other: BenchmarkResult): List[(String, Double, Double, Double)] = {
+    val otherMetrics = other.metrics.map(m => (m.name, m.value)).toMap
+    metrics.map { m =>
+      val prev = otherMetrics(m.name).toDouble
```
ClaireMcGinty: this would break if we add a new metric right? can we return 0/inf/NaN or something if `otherMetrics` doesn't have `m.name`?

regadas: yeah good point ... I'm not really happy with this actually. try something else.

PR #1714, opened 2019-03-01
Remove deprecation warning on tfRecordExampleFileWithSchema

andrewsmartin:


regadas:
@andrewsmartin so this will not be deprecated after all?

andrewsmartin:
@regadas correct, it's still valid to want to read schema within scio job. We only want to deprecate the schema inference logic. Scio shouldn't be concerned with writing schemas as all, we'll leave that to TFDV.

regadas:
@andrewsmartin yeah makes sense!

PR #1715, opened 2019-03-01
Use Combine By Key for DistinctBy

anish749:
Follows #1710 and addresses review comments.
`groupBy` is not necessary as the results are being ignored, and hence we can use `combineByKey`


regadas:
This now matches the previous implementation

PR #1720, opened 2019-03-04
Clarify error message when a job uses an input multiple times.

psobot:
Given a Scio job that uses the same input multiple times:

```scala
object JobWithDuplicateInput {
    sc.textFile(args("input"))
    sc.textFile(args("input"))
    sc.close()
  }
}
```

...Scio's `TestDataManager` currently throws an error in test that reads: `There already exists test input for...`, which implies that the test itself has multiples of the same test input. This is unclear as the issue is actually with the job itself, not the test.

This PR clarifies this error message by changing it to `s"Test input $key has already been used once."`, which is hopefully clearer for users that encounter this error.

(cc @ClaireMcGinty, @regadas)

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
@@ -32,7 +32,7 @@ private[scio] class TestInput(val m: Map[String, Iterable[_]]) {
     val key = io.testId
     require(m.contains(key),
             s"Missing test input: $key, available: ${m.keys.mkString("[", ", ", "]")}")
-    require(!s.contains(key), s"Test input $key has already been used once.")
+    require(!s.contains(key), s"Test input $key has already been read from once.")
     s.add(key)
     m(key).asInstanceOf[Iterable[T]]
   }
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -32,9 +32,7 @@ private[scio] class TestInput(val m: Map[String, Iterable[_]]) {
     val key = io.testId
     require(m.contains(key),
             s"Missing test input: $key, available: ${m.keys.mkString("[", ", ", "]")}")
-    require(!s.contains(key),
-            s"There already exists test input for $key, currently " +
-              s"registered inputs: ${s.mkString("[", ", ", "]")}")
+    require(!s.contains(key), s"Test input $key has already been used once.")
```
ClaireMcGinty: 🎨 "used once" -> "read once"?

while you're at it, would you want to make a matching change to `TestOutput`? :)

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -196,7 +196,7 @@ object JobWithoutClose {
   }
 }
 
-object JobWitDuplicateInput {
+object JobWithDuplicateInput {
```
ClaireMcGinty: heh thanks for fixing the typos!

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
psobot: 👍

ClaireMcGinty:
thanks for the contribution! looks great 👍

Issue #1723, opened 2019-03-06
SparkeyPairSCollection#asSparkey does not create files for empty SCollection


sisidra:
If `SCollection[String, String]` is empty then calling `asSparkey` will not result in any error and will not create an empty Sparkey file.
If that is expected behaviour, can it be documented in scaladocs?

ClaireMcGinty:
As someone who has never personally used Sparkey, I think writing an empty file seems like a reasonable default behavior and would match the behavior of other sinks (text, avro, and BQ I/Os will all create empty files).  Seems like this could easily be achieved by moving `SparkeyWriter`'s instantiation & closing outside of the Transform itself. @jto , @andrisnoko , do either of you have strong opinions on this...?

sisidra:
I agree that empty file should be expected behavior.
Moving SparkeyWriter outside should not work tho as those handles will be accessed on different machines :)
Something like this might work:
```scala
in
  .map(Some(_))
  .union(sc.parallelize[Option[(String, String)]](None))
  .groupBy(...)
  .map(
    // ...
    val it = xs.iterator.flatten
    // ...
  )
```
https://github.com/spotify/scio/blob/master/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala#L127-L140

Maybe I can prepare PR.

ClaireMcGinty:
ah good point, I was assuming some different behavior of the SparkeyWriter! looking at your PR now.

nevillelyh:
Fixed in #1834

PR #1726, opened 2019-03-07
Fix SparkeyPairSCollection#asSparkey to create Sparkey files even when SCollection is empty

sisidra:
Solution for https://github.com/spotify/scio/issues/1723

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
@@ -124,12 +124,16 @@ package object sparkey {
       val uri = SparkeyUri(basePath, self.context.options)
       require(!uri.exists, s"Sparkey URI ${uri.basePath} already exists")
       logger.info(s"Saving as Sparkey: $uri")
+      val atLeastOneElement: SCollection[Option[(K, V)]] =
+        self.context.parallelize(Seq[Option[(K, V)]](None))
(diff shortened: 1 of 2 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -124,12 +124,16 @@ package object sparkey {
       val uri = SparkeyUri(basePath, self.context.options)
       require(!uri.exists, s"Sparkey URI ${uri.basePath} already exists")
       logger.info(s"Saving as Sparkey: $uri")
+      val atLeastOneElement: SCollection[Option[(K, V)]] =
```
ClaireMcGinty: this solution seems a bit hacky to me -- how would you feel about implementing, for example, a `FileBasedSink` for Sparkey output? it supports default destinations for empty PCollections so we wouldn't have to add these extra transforms to work around it. @jto - wdyt?

jto: Same. This solution really feels hacky.

sisidra: It is hacky!
I would also love a specialized IO for Sparkey instead of current implementation.
I'm not sure how it plays with sideInputs where Sparkey is used.

regadas:
Closed via #1834

Issue #1718, opened 2019-03-02
BigQueryConfig is not respected (caching and priority)


bbhoss:
I have tried everything under the sun to get the BigQuery functionality to respect my wishes as documented, which is to set
```
  System.setProperty("bigquery.cache.enabled", "false")
  System.setProperty("bigquery.priority", "INTERACTIVE")
```

BigQueryConfig seems to think that it's interactive when interrogated:
```
scala> com.spotify.scio.bigquery.client.BigQueryConfig.priority.map(_ == "INTERACTIVE")
res2: Option[Boolean] = Some(true)
```
Yet, jobs are still scheduled as batch.

ClaireMcGinty:
hi @bbhoss . Can you share some more information about the kind of BQ job you're writing (code snippet would be great) and how you're running it?

bbhoss:
Thanks @ClaireMcGinty. Here's a small snippet of the code
```scala
        val patients = sc
          .typedBigQuery[Member](
            s"SELECT * from `${c.productionProject}.gold_claims.Member` WHERE identifier.patientId IS NOT NULL")
          .flatMap(member => member.MEMBER_ID)
```

`Member` is a case class that's been enhanced with typed bigquery abilities using `@BigQueryType.toTable`. Doesn't seem too exotic, so I'm wondering if it's some weird deps issue. I am happy to dig in further to debug, just not sure where to start.

bbhoss:
Also, I will add that this happens with DirectRunner and DataflowRunner

ClaireMcGinty:
Thanks! I haven't tested this but my hypothesis is that we need to set [.withQueryPriority(...)](https://github.com/apache/beam/blob/f6fdeaa1a65b690f1c876f77802279a23096f4c2/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L1016) on the [TypedRead](https://github.com/spotify/scio/blob/68d22b04a0d1dff4e621f5204851ce1d3f573ef3/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L75) transform in `BigQuerySelect`. Would you be interested in trying this out/submitting a pull request? Otherwise we can add it to our backlog :)

bbhoss:
I will certainly try that if I can remember how to run snapshots with our
project

bbhoss:
[...]

PR #1736, opened 2019-03-09
make BigQuery priority sysprop case-insensitive

bbhoss:
Prevent others from being confused like I was in #1718. Basically, just convert the sysprop value to upper case, which is checked everywhere else. I considered just making the check itself in `isInteractive` case insensitive, but then I thought that this problem could regress if anyone ever checks this elsewhere in new code.


bbhoss:
Edit: Decided to go the enum route :man_shrugging:. Can revert back if unwanted, I can see not wanting to tie the config up with beam stuff, but I don't see that being a huge downside.

bbhoss:
Done

PR #1737, opened 2019-03-11
avro io with type bound

andrisnoko:
This breaks backward compatibility with test and `saveAsAvroFile` as we can't over load the method with default parameters.

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -27,6 +27,7 @@ import com.spotify.scio.{avro, ScioContext}
 import org.apache.avro.Schema
 import org.apache.avro.file.CodecFactory
 import org.apache.avro.generic.GenericRecord
+import org.apache.avro.specific.SpecificRecordBase
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
 import org.apache.beam.sdk.transforms.{DoFn, SerializableFunction}
 import org.apache.beam.sdk.{io => beam}
@@ -131,6 +132,7 @@ object ProtobufIO {
   val WriteParam = AvroIO.WriteParam
 }
 sealed trait AvroIO[T] extends ScioIO[T] {
+  override final val tapT = TapOf[T]
 
   protected def avroOut[U](sc: SCollection[T],
                            write: beam.AvroIO.Write[U],
@@ -148,10 +150,11 @@ sealed trait AvroIO[T] extends ScioIO[T] {
 
 }
 
-final case class SpecificRecordIO[T: ClassTag: Coder](path: String) extends AvroIO[T] {
+final case class SpecificRecordIO[T: ClassTag: Coder](path: String)(
+  implicit ev: T <:< SpecificRecordBase)
+    extends AvroIO[T] {
   override type ReadP = Unit
   override type WriteP = AvroIO.WriteParam
-  override final val tapT = TapOf[T]
 
   override def testId: String = s"AvroIO($path)"
 
@@ -181,11 +184,11 @@ final case class SpecificRecordIO[T: ClassTag: Coder](path: String) extends Avro
     SpecificRecordAvroTap[T](ScioUtil.addPartSuffix(path))
 }
 
-final case class SchemaAvroIO[T: ClassTag: Coder](path: String, schema: Schema = null)
+final case class SchemaAvroIO[T: Coder](path: String, schema: Schema)(
+  implicit ev: T <:< GenericRecord)
     extends AvroIO[T] {
   override type ReadP = Unit
   override type WriteP = AvroIO.WriteParam
-  override final val tapT = TapOf[T]
 
   override def testId: String = s"AvroIO($path)"
 
@@ -194,7 +197,6 @@ final case class SchemaAvroIO[T: ClassTag: Coder](path: String, schema: Schema =
    * [[org.apache.avro.generic.GenericRecord GenericRecord]].
    */
   override def read(sc: ScioContext, params: ReadP): SCollection[T] = {
-    val cls = ScioUtil.classOf[T]
     val t = beam.AvroIO
       .readGenericRecords(schema)
       .from(path)
@@ -207,7 +209,6 @@ final case class SchemaAvroIO[T: ClassTag: Coder](path: String, schema: Schema =
    * [[org.apache.avro.generic.GenericRecord GenericRecord]].
    */
   override def write(data: SCollection[T], params: WriteP): Future[Tap[T]] = {
-    val cls = ScioUtil.classOf[T]
     val t = beam.AvroIO.writeGenericRecords(schema).asInstanceOf[beam.AvroIO.Write[T]]
     data.applyInternal(
       avroOut(data, t, path, params.numShards, params.suffix, params.codec, params.metadata))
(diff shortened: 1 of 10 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/AvroSCollection.scala
```diff
@@ -37,18 +40,31 @@ final class AvroSCollection[T](@transient val self: SCollection[T]) extends Seri
    * @param schema must be not null if `T` is of type
    *               [[org.apache.avro.generic.GenericRecord GenericRecord]].
    */
+  // scalastyle:off parameter.number
   def saveAsAvroFile(path: String,
                      numShards: Int = AvroIO.WriteParam.DefaultNumShards,
                      schema: Schema = null,
```
jto: Unless I'm missing something Schema should never be null here.

andrisnoko: That's correct. I will fix that.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -131,42 +131,40 @@ object ProtobufIO {
...
       .to(ScioUtil.pathWithShards(path))
       .withNumShards(numShards)
       .withSuffix(suffix)
       .withCodec(codec)
       .withMetadata(metadata.asJava)
 
+}
+
+final case class SpecificRecordIO[T: ClassTag: Coder](path: String)(
+  implicit ev: T <:< SpecificRecordBase)
```
regadas: @andrisnoko couldn't we have this as `SpecificRecordIO[T <: SpecificRecordBase : ClassTag: Coder]`?

andrisnoko:
@jto @regadas 
`SchemaAvroIO[T]` is left unbounded to take GenericRecords and this [use-case](https://github.com/spotify/scio/blob/master/scio-test/src/test/scala/com/spotify/scio/io/TapTest.scala#L138). There seem to be no breaking changes to users; I added `TestIO` in `AvroIO` to resolve issues in tests where types are specified e.g. `AvroIO[SomeType]`. I think we only need testId in JobTest.

wdyt?

scio-avro/src/main/scala/com/spotify/scio/avro/SpecificRecordAvroSCollection.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.spotify.scio.values.SCollection
+import org.apache.avro.file.CodecFactory
+import org.apache.avro.specific.SpecificRecordBase
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
+final class SpecificRecordAvroSCollection[T <: SpecificRecordBase](
+  @transient val self: SCollection[T])
+    extends {
```
ClaireMcGinty: is this meant to be here? `extends {` just defaults to `extends AnyRef {` so it's a bit redundant :)

scio-avro/src/main/scala/com/spotify/scio/avro/taps.scala
```diff
@@ -26,22 +26,36 @@ import com.spotify.scio.values._
...
 import scala.reflect.runtime.universe._
 
 /**
- * Tap for Avro files.
- * @param schema must be not null if `T` is of type
- *               [[org.apache.avro.generic.GenericRecord GenericRecord]].
+ * Tap for [[org.apache.avro.generic.GenericRecord GenericRecord]] Avro files.
  */
-case class AvroTap[T: ClassTag: Coder](path: String, @transient private val schema: Schema = null)
+case class SchemaAvroTap[T: ClassTag: Coder](path: String,
```
ClaireMcGinty: i kind of feel like this should be named `GenericRecordTap` to match `SpecificRecordTap` ? what do you think?

scio-avro/src/main/scala/com/spotify/scio/avro/SpecificRecordAvroSCollection.scala
andrisnoko: Good catch! supposed to extend Serializable.

scio-avro/src/main/scala/com/spotify/scio/avro/taps.scala
andrisnoko: Yeah, makes sense.

andrisnoko:
@regadas rebased.

scio-avro/src/main/scala/com/spotify/scio/avro/taps.scala
```diff
@@ -26,22 +26,36 @@ import com.spotify.scio.values._
...
-  override def open(sc: ScioContext): SCollection[T] =
-    sc.avroFile[T](path, s.get)
+
+  override def open(sc: ScioContext): SCollection[T] = sc.avroFile[T](path, s.get)
+}
+
+/**
+ * Tap for [[org.apache.avro.specific.SpecificRecordBase SpecificRecordBase]] Avro files.
+ */
+case class SpecificRecordTap[T <: SpecificRecordBase: ClassTag: Coder](path: String)(
```
regadas: make it final

scio-avro/src/main/scala/com/spotify/scio/avro/taps.scala
```diff
@@ -26,22 +26,36 @@ import com.spotify.scio.values._
...
 import scala.reflect.runtime.universe._
 
 /**
- * Tap for Avro files.
- * @param schema must be not null if `T` is of type
- *               [[org.apache.avro.generic.GenericRecord GenericRecord]].
+ * Tap for [[org.apache.avro.generic.GenericRecord GenericRecord]] Avro files.
  */
-case class AvroTap[T: ClassTag: Coder](path: String, @transient private val schema: Schema = null)
+case class GenericRecordTap[T: ClassTag: Coder](path: String,
```
regadas: make it final

scio-avro/src/main/scala/com/spotify/scio/avro/taps.scala
```diff
@@ -26,22 +26,36 @@ import com.spotify.scio.values._
...
-    sc.avroFile[T](path, s.get)
+
+  override def open(sc: ScioContext): SCollection[T] = sc.avroFile[T](path, s.get)
+}
+
+/**
+ * Tap for [[org.apache.avro.specific.SpecificRecordBase SpecificRecordBase]] Avro files.
+ */
+case class SpecificRecordTap[T <: SpecificRecordBase: ClassTag: Coder](path: String)(
+  implicit ev: T <:< SpecificRecordBase)
```
regadas: you can remove the generalized type constraint

scio-avro/src/main/scala/com/spotify/scio/avro/SpecificRecordAvroSCollection.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.spotify.scio.coders.Coder
+import com.spotify.scio.io.Tap
+import com.spotify.scio.values.SCollection
+import org.apache.avro.file.CodecFactory
+import org.apache.avro.specific.SpecificRecordBase
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
+final class SpecificRecordAvroSCollection[T <: SpecificRecordBase](
```
regadas: I think we can avoid having this class ... we can make a gen type constrain on the method `saveAsAvroFile` in `AvroSCollection`

regadas: Actually, since we care about backwards compatibility, we cant do it as i thought because we will end up having method overload with default params. I think the current solution is ok and i think we should revise the other methods.

PR #1738, opened 2019-03-11
avro io

andrisnoko:
This just splitting the AvroIO without type bound, seems to work and has not breaking changes.


andrisnoko:
Alternate approach [here](https://github.com/spotify/scio/pull/1737)

PR #1739, opened 2019-03-11
Bump beam to 2.11

regadas:


scio-bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/PatchedBigQueryTableRowIterator.java
```diff
@@ -502,4 +504,34 @@ public void close() {
     }
 
   }
+
+  private static final DateTimeFormatter DATE_AND_SECONDS_FORMATTER =
+      DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss").withZoneUTC();
+
+  private static String formatTimestamp(String timestamp) {
```
ClaireMcGinty: do we eventually want to add these new [formatDate/formatTime](https://github.com/apache/beam/commit/0025a63cc2a233e80999138bcf14573fc6a610a7#diff-44025ee9b9c94123967e1df92bfb1c04R124) methods too?

regadas: hm, not sure, but probably not. So I was intrigued why we have this class in the first place when it was removed a long time go in this [commit](https://github.com/apache/beam/commit/18d7c2969eac36c0cb02cf8869c299eb334fbe80) we probably could do the same...

regadas: but probably in another PR

ClaireMcGinty: good find! 💯 i'm in favor of deleting it... although it might end up being a breaking change since they don't seem to have a 1:1 replacement for it

PR #1741, opened 2019-03-11
1734: Limit job graph size by not wrapping native beam coders

### LINKED ISSUE #1734: The size of the serialized JSON representation of the pipeline exceeds the allowable limit for the API
That error has been reported by several users after migrating to 0.7.2.
Might be caused by a change in Scio or by the new Beam version.

jto:
fixes #1734.

I only removed wrapping aroung coders natively provided by beam. It's a somewhat conservative fix but should reduce the graph size significantly for most users.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
@@ -31,12 +31,19 @@ object CoderMaterializer {
                                o: PipelineOptions = PipelineOptionsFactory.create()): BCoder[T] =
     beam(r, o, coder)
 
+  @inline private def nullCoder[T](o: PipelineOptions, c: BCoder[T]) = {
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -33,6 +33,11 @@ object CoderMaterializer {
 
   final def beam[T](r: CoderRegistry, o: PipelineOptions, coder: Coder[T]): BCoder[T] = {
     coder match {
+      // #1734: do not wrap native beam coders
+      case Beam(c) if c.getClass.getPackage.getName.startsWith("org.apache.beam") =>
+        val nullableCoder = o.as(classOf[com.spotify.scio.options.ScioOptions]).getNullableCoders()
```
regadas: @jto 🎨 can we import this since it's repeated?

regadas: or even move this logic to a separate function

jto:
Fixed according to your comments :)

PR #1746, opened 2019-03-12
Make stacktrace transient.

regadas:


regadas:
Per discussion IRL this would not give the desired effect.

PR #1747, opened 2019-03-12
Explicit reset position on SeekableInput

regadas:


regadas:
@ClaireMcGinty I can, but most of our tests use the `Matchers` dsl

Issue #1659, opened 2019-02-04
tensorflow: sequenceExamples


ckchow:
I've been able to write / read SequenceExamples to / from TFRecords in the same way that regular Examples currently are handled. It would be nice to have library support for both IO and prediction with SequenceExamples.

regadas:
@andrewsmartin @brianmartin Do you have any thoughts on this?

brianmartin:
Yes. @andrewsmartin and I actually just heard from a few internal customers this week that this feature would be useful. I was actually considering picking this up. Let me touch base with @andrewsmartin offline and I'll post an update here.

brianmartin:
Looks fairly straightforward to implement. The implicits for a `SCollection[SequenceExample]` should look very close to the existing ones for `SCollection[Example]`. 

The only slight wrinkle is that `SCollection[Seq[Example]]` is automatically merged into `SCollection[Example]`. But that still seems reasonable because if you actually want `SequenceExample`, which would be somewhat rare, then you would explicitly create a `SCollection[SequenceExample]`.

I haven't investigated whether the existing predict function would work as is, but.. should be possible.

@ckchow Were you interested in producing a PR? :)

ckchow:
I have an implementation that is working. wouldn't mind tidying it up and submitting it when a little less busy.

brianmartin:
Awesome. That'd be most welcome @ckchow!

jbx:
@ckchow - did you have a chance to look at this?

brianmartin:
This is awesome @martinbomio! We had, here on ml-infra, been considering adding this but it was perpetually stuck in the backlog.

(cc @ckchow: I'm assuming PR #1757 satisfies your use case as well?)

brianmartin:
I think we're all good here, right @regadas?

regadas:
@brianmartin yup, we can close this.

PR #1757, opened 2019-03-17
Add sequence example support to scio-tensorflow

martinbomio:
Adds support for 9SequenceExample](https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/core/example/example.proto) to scio tensorflow.
Addresses https://github.com/spotify/scio/issues/1659.

### DIFF
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordIO.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordIO.scala
@@ -18,13 +18,12 @@
 package com.spotify.scio.tensorflow
 
 import com.spotify.scio.ScioContext
-import com.spotify.scio.io.{ScioIO, Tap, TapOf}
+import com.spotify.scio.io.{ScioIO, Tap, TapOf, TapT}
 import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.SCollection
 import org.apache.beam.sdk.io.Compression
 import org.apache.beam.sdk.{io => beam}
-import org.tensorflow.example.Example
-
+import org.tensorflow.example.{Example, SequenceExample}
 import scala.concurrent.Future
 
 final case class TFRecordIO(path: String) extends ScioIO[Array[Byte]] {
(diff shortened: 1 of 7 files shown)


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TFRecordIO.scala
```diff
@@ -88,6 +87,26 @@ object TFExampleIO {
   val WriteParam = TFRecordIO.WriteParam
 }
 
+final case class TFSequenceExampleIO(path: String) extends ScioIO[SequenceExample] {
```
martinbomio: created a new `ScioIO` for the sequence example following how it was done with the exampleIO and the tfrecordIO, but both the `TFSequenceExampleIO ` and `TFExampleIO` can be deleted and the `fRecordSequenceExampleFile` and `fRecordExampleFile` methods can use the `TFRecordIO ` and just map and create the *Examples from bytes. Same applies to the write functions.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -449,3 +449,23 @@ final class TFRecordSCollectionFunctions[T <: Array[Byte]](private val self: SCo
...
+  private val self: SCollection[T])
+    extends AnyVal {
+
+  /**
+   * Saves this SCollection of `org.tensorflow.example.SequenceExample` as a TensorFlow
+   * TFRecord file.
+   *
+   * @return
+   */
+  def saveAsTfSequenceExampleFile(
```
regadas: this is in line with the rest of the stuff but I'm getting the idea that we probably could rename this methods to just `saveAsTfRecordFile`. This would be on another PR but @martinbomio  wdyt?

brianmartin: +1

martinbomio: sounds good to me!

martinbomio: created https://github.com/spotify/scio/pull/1762

PR #1758, opened 2019-03-18
Support Avrohugger SpecificRecord case class in AvroIO

andrisnoko:


andrisnoko:
Closing this as it is too specific to Avrohugger. We will think about coming up with a separate library if necessary.

PR #1759, opened 2019-03-18
Minor doc site fix

regadas:


scripts/make-site.sh
```diff
@@ -5,4 +5,4 @@
 # SOCCO=true sbt scio-examples/clean scio-examples/compile site/makeSite
 
 # generate new site and push to GitHub
-SOCCO=true sbt scio-examples/clean scio-examples/compile site/ghpagesPushSite
+SOCCO=true sbt scio-examples/clean scio-examples/compile #site/ghpagesPushSite
```
ClaireMcGinty: `# generate new site and push to GitHub` is no longer completely accurate with this commented out?

regadas: yeah that was committed by mistake 😄

PR #1762, opened 2019-03-19
Rename unify tf save function names

martinbomio:
Follow up from https://github.com/spotify/scio/pull/1757


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowFunctions.scala
```diff
@@ -377,14 +389,28 @@ final class SeqTFExampleSCollectionFunctions[T <: Example](private val self: SCo
    *
    * @group output
    */
+  @deprecated("saveAsTfExampleFile is deprecated: use saveAsTfRecordFile instead", "Scio 0.7.3")
```
ClaireMcGinty: @regadas ... since these deprecations are new AFAIK shouldn't these be since `Scio 0.7.4` ?

regadas: I'll fix it in an upcoming PR

PR #1765, opened 2019-03-19
Rename Avro ScioContext ops class

regadas:


scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
```diff
@@ -31,7 +31,7 @@ import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
 
 /** Enhanced version of [[ScioContext]] with Avro methods. */
-final class AvroScioContext(private val self: ScioContext) extends AnyVal {
+final class ScioContextOps(private val self: ScioContext) extends AnyVal {
```
andrisnoko: @regadas why rename? `AvroScioContext` seems to be appropriately named as in others like `BigQueryScioContext`

regadas: these are `Ops` that are added to `ScioContext` and the package is `com.spotify.scio.avro.syntax`. Avro prefix is a bit redundant.

regadas: #1763 is another example

regadas: in regards to `BigQueryScioContext` will be proposing the same thing.

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
```diff
@@ -80,5 +80,5 @@ final class AvroScioContext(private val self: ScioContext) extends AnyVal {
 
 /** Enhanced with Avro methods. */
 trait ScioContextSyntax {
-  implicit def toAvroScioContext(c: ScioContext): AvroScioContext = new AvroScioContext(c)
+  implicit def toScioContextOps(c: ScioContext): ScioContextOps = new ScioContextOps(c)
```
jto: Would that not break the implicit conversion if `BigQueryScioContext` and `AvroScioContext` have the implicit conversion with the same name that are  both in scope ?

regadas: @jto 🤦‍♂️ yes! 💣

regadas: @jto I'll try to go with this convention `def <main package><ops class>( ... )`

jto: sounds good

PR #1767, opened 2019-03-20
Simplify query row transform

regadas:
After some digging I noticed that the SqlTransform underlying impl's do exactly the same thing, query parsing and set the row schema. By just doing this we also avoid using the `BeamSqlEnv` which is also marked as `@Internal`.

@jto am I missing something that would prevent us from just doing this?

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
@@ -25,7 +25,6 @@ import com.spotify.scio.schemas.{PrettyPrint, Record, ScalarWrapper, Schema, Sch
 import org.apache.beam.sdk.values._
 import org.apache.beam.sdk.extensions.sql.SqlTransform
 import org.apache.beam.sdk.extensions.sql.impl.{BeamSqlEnv, ParseException}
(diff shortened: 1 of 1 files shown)


jto:
Unless I'am missing something, you're not setting the Coder for the output of the query.
@ClaireMcGinty wdyt ?

regadas:
@jto @ClaireMcGinty ah yes! good point. It's not needed, because `PCollection` creates a `Coder` from schema every time you `setRowSchema` or `setSchema`.

jto:
Yes but we're not calling `setSchema`. AFAIK the schema of the resulting `Row` is never set if we merge that PR.

regadas:
@jto the schema that is set now is the same schema as it was set before 😄

jto:
Once you run the query the schema changes.
Am I missing something here ?

regadas:
@jto  i.e https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamValuesRel.java#L91

Beam has more cases like this.

jto:
Oh. Indeed. I get it now. 
That's a good catch  👍

regadas:
@jto yeah this is very hidden to be honest :(

PR #1768, opened 2019-03-21
GenericRecord fallback warning

jto:
Add compile time warning when Coder fallsback to Kryo on GenericRecord


scio-macros/src/main/scala/com/spotify/scio/coders/CoderMacros.scala
```diff
@@ -30,6 +30,17 @@ private[coders] object CoderMacros {
 
   private[this] val BlacklistedTypes = List("org.apache.beam.sdk.values.Row")
 
+  private[this] val avroGenRecWarn =
```
regadas: 🎨 name does not follow same pattern

regadas:
@jto I think this needs some rebase.

PR #1769, opened 2019-03-22
BeamSQL join support

regadas:
rough proposal for discussion

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
@@ -116,13 +116,16 @@ object Query {
 
 trait Query2[A, B, C] extends ((SCollection[A], SCollection[B]) => SCollection[C]) {
   def query: String
+
+  def leftTag: String
+
+  def rightTag: String
 }
 
 object Query2 {
 
-  def typecheck[A: Schema, B: Schema, O: Schema](q: Query2[A, B, O],
-                                                 left: String,
-                                                 right: String): Either[String, Query2[A, B, O]] = {
+  def typecheck[A: Schema, B: Schema, O: Schema](
+    q: Query2[A, B, O]): Either[String, Query2[A, B, O]] = {
     val schemaA: BSchema = SchemaMaterializer.fieldType(Schema[A]).getRowSchema
     val schemaB: BSchema = SchemaMaterializer.fieldType(Schema[B]).getRowSchema
     val expectedSchema: BSchema =
@@ -134,7 +137,7 @@ object Query2 {
       }
 
     QueryUtils
-      .typecheck(q.query, List((left, schemaA), (right, schemaB)), expectedSchema)
+      .typecheck(q.query, List((q.leftTag, schemaA), (q.rightTag, schemaB)), expectedSchema)
       .map(_ => q)
   }
 
@@ -143,7 +146,11 @@ object Query2 {
                                 right: String,
                                 udfs: Udf*): Query2[A, B, Row] =
     new Query2[A, B, Row] {
-      override def query: String = q
+      override val query: String = q
+
+      override val leftTag: String = left
+
+      override val rightTag: String = right
 
       override def apply(a: SCollection[A], b: SCollection[B]): SCollection[Row] =
         a.context.wrap {
@@ -152,8 +159,8 @@ object Query2 {
           val sqlTransform = SqlTransform.query(query).registerUdf(udfs: _*)
 
           PCollectionTuple
-            .of(new TupleTag[A](left), collA.internal)
-            .and(new TupleTag[B](right), collB.internal)
+            .of(new TupleTag[A](leftTag), collA.internal)
+            .and(new TupleTag[B](rightTag), collB.internal)
             .apply(s"${collA.tfName} join ${collB.tfName}", sqlTransform)
         }
     }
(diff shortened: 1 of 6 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -131,6 +131,8 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   /** A friendly name for this SCollection. */
   def name: String = internal.getName
 
+  def setName(name: String): SCollection[T] = context.wrap(internal.setName(name))
```
regadas: this should be removed 😄

scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
```diff
@@ -207,6 +114,178 @@ object Query {
 
 }
 
+trait Query2[A, B, C] extends ((SCollection[A], SCollection[B]) => SCollection[C]) {
+  def query: String
+}
+
+object Query2 {
```
jto: Does this mean that we can only join 2 `SCollections` ?

scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
```diff
@@ -207,6 +114,178 @@ object Query {
...
+        tag -> new BaseBeamTable(schema) {
+          override def buildIOReader(begin: PBegin): PCollection[Row] = ???
+
+          override def buildIOWriter(input: PCollection[Row]): POutput = ???
+
+          override def isBounded: PCollection.IsBounded = PCollection.IsBounded.BOUNDED
+        }
+    }.toMap
+
+    silence { () =>
```
jto: It's out of the scope of that PR but I think `silence` actually does not work, or a leat not constently. We may remove it.

scio-core/src/main/scala/com/spotify/scio/sql/syntax/SqlTransformSyntax.scala
```diff
@@ -0,0 +1,22 @@
...
+        st.registerUdf(x.fnName, x.clazz)
+      case (st, x: UdfFromSerializableFn[_, _]) =>
+        st.registerUdf(x.fnName, x.fn)
+      case (st, x: UdafFromCombineFn[_, _, _]) =>
+        st.registerUdaf(x.fnName, x.fn)
+    }
+}
+
+trait SqlTransformSyntax {
+  implicit def sqlSqlTransformOps(t: SqlTransform): SqlTransformOps = new SqlTransformOps(t)
```
jto: Why should we introduce an implicit conversion here ?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -197,6 +190,18 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
     }
 
+  /**
+   * Apply BeamSQL query to this SCollection
+   */
+  // this method is not strictly necessary but using a invariant type instead of
+  // simple (SCollection[T] => SCollection[U]) helps with type inference
+  def sql[U](q: Query[T, U]): SCollection[U] = transform(q.query)(q)
+
+  def sqlJoin[B, U](other: SCollection[B])(q: Query2[T, B, U]): SCollection[U] =
```
jto: I don't like that. It feels really odd to have to use a different function just because there's a `JOIN` in your query.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -179,15 +181,6 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
-   * Apply BeamSQL query to this SCollection
-   */
-  // this method is not strictly necessary but using a invariant type instead of
-  // simple (SCollection[T] => SCollection[U]) helps with type inference
-  def sql[U](q: Query[T, U]): SCollection[U] = transform(q.query)(q)
-
-  def to[U](to: To[T, U]): SCollection[U] = transform(to)
-
   @experimental
   def transform[U](name: String)(f: SCollection[T] => SCollection[U]): SCollection[U] =
```
jto: 👍

scio-core/src/main/scala/com/spotify/scio/sql/Query.scala
regadas: yeah, currently beam only allow's join of 2 collections

Bounded input JOIN bounded input
Unbounded input JOIN unbounded input
Unbounded input JOIN bounded input

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: yeah agreed, I was getting some implicit divergence here ... so I took the easy route just to get this out of the door

scio-core/src/main/scala/com/spotify/scio/sql/syntax/SqlTransformSyntax.scala
regadas: hm, yeah it sounded nice to have, coz SqlTransform offers the methods to register the udf's so why not offer another method that supports our type ¯\_(ツ)_/¯

regadas:
Closing in favor of #1778

PR #1773, opened 2019-03-25
Make BQ annotations serializable

anish749:
I was trying to write Scollections to BQ, and for some generic derivations where a nested type contained a class with a big query annotation, I was finding that the annotations are not serializable. I was thinking of making them seralizable because I think that is an easy fix and hopefully wont break anything.

My use case is something like this:
```
//.groupBy(r => r.userId) - usual Scio code.
.bqTrace("gcpproj.dataset.error_records", predicate = _.size > 1) // Trace errors to BQ.
//.flatMap(...) // usual scio code.
```

now my SCollection is of type `[(String, Iterable[CaseClassWithBQAnnotation])]`

### DIFF
--- a/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/Schemas.scala
+++ b/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/Schemas.scala
@@ -71,4 +71,8 @@ object Schemas {
   case class RepeatedNested(required: List[Required],
                             optional: List[Optional],
                             repeated: List[Repeated])
+
+  case class User(@description("user name") name: String, @description("user age") age: Int)
+  case class Account(@description("account user") user: User,
+                     @description("in USD") balance: Double)
 }
--- a/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
+++ b/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
@@ -82,6 +82,10 @@ class TypeProviderTest extends FlatSpec with Matchers {
     SerializableUtils.ensureSerializable(S1(1))
   }
 
+  "BigQueryTag" should "be a serializable annotation" in {
+    SerializableUtils.ensureSerializable[BigQueryTag](new BigQueryTag())
+  }
+
   @BigQueryType.fromSchema("""{"fields": [{"name": "f1", "type": "INTEGER"}]}""")
   class MissingMode
 
--- a/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
+++ b/scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
@@ -18,6 +18,7 @@
 package com.spotify.scio.bigquery.types
 
 import com.google.api.services.bigquery.model.TableRow
+import org.apache.beam.sdk.util.SerializableUtils
 import org.joda.time.Instant
 import org.scalatest.{Assertion, FlatSpec, Matchers}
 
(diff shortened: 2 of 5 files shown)


jto:
Hi @anish749. Can you add a test to ensure that the 2 annotations do indeed serialize ?

anish749:
sure.. I will add them.. Thanks for pointing out.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -28,12 +29,6 @@ import scala.reflect.runtime.universe._
...
   @BigQueryType.toTable
   case class RefinedClass(a1: Int)
-}
-
-// TODO: mock BigQueryClient for fromTable and fromQuery
-class TypeProviderTest extends FlatSpec with Matchers {
-
-  val NOW = Instant.now()
 
   @BigQueryType.fromSchema(
```
anish749: I moved these classes to the companion object because these annotated classes  should be in a companion object (?)

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
```diff
@@ -74,6 +78,10 @@ class TypeProviderTest extends FlatSpec with Matchers {
     S4.tableDescription shouldBe "Table S4"
   }
 
+  it should "be serializable" in {
+    SerializableUtils.ensureSerializable(S1(1))
+  }
```
anish749: Is this the correct place for this test? 
Is this the right way to test?
Beam is running this before converting, so I thought it would be nice to use this.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/Schemas.scala
```diff
@@ -71,4 +71,8 @@ object Schemas {
   case class RepeatedNested(required: List[Required],
                             optional: List[Optional],
                             repeated: List[Repeated])
+
+  case class User(@description("user name") name: String, @description("user age") age: Int)
+  case class Account(@description("account user") user: User,
+                     @description("in USD") balance: Double)
```
anish749: I moved these to the companion object because the `class SchemaProviderTest extends FlatSpec with Matchers` wont be serializable because `org.scalatest.FlatSpec` isn't serializable. Not sure if this is the right way to do this. What are your thoughts.

I was getting the following exception if I had in in the class and not in the companion object: 
```
[info]   Cause: java.io.NotSerializableException: org.scalatest.Assertions$AssertionsHelper
[info]  - field (class "org.scalatest.FlatSpec", name: "assertionsHelper", type: "class org.scalatest.Assertions$AssertionsHelper")
[info]  - object (class "com.spotify.scio.bigquery.types.TypeProviderTest", TypeProviderTest)
[info]  - field (class "com.spotify.scio.bigquery.types.TypeProviderTest$S1", name: "$outer", type: "class com.spotify.scio.bigquery.types.TypeProviderTest")
[info]  - root object (class "com.spotify.scio.bigquery.types.TypeProviderTest$S1", S1(1))
[info]   at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1182)
```
I enabled `-Dsun.io.serialization.extendedDebugInfo=true` to trace this.

anish749:
@regadas @jto I added a test, but I am not very sure it is in the right class. Please let me know if this is otherwise, I would change it accordingly. I am not very familiar with this part of the code.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/TypeProviderTest.scala
anish749: Please wait. I don't think I am doing this correctly here.

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/Schemas.scala
jto: LGTM

Issue #1748, opened 2019-03-12
Question: Allow PrioritizedImplicits Construction for pipelines using Scio


anish749:
I might be wrong here, or there might be better solution, but I was trying out some extensions for joins and in most of our pipelines we use extensive implicits on top of SCollection, to add helper methods / wrappers. Sometimes these helper function names start conflicting with the ones provided by implicit conversions in Scio. Separating implicits in a `trait SCollectionImplicits` from `object SCollection` would help in prioritising these implicits.

in user's code we can then have:
```
object ScioHelpers extends SCollectionImplicits with LowPrioProjectSpecificImplicits
...

import ScioHelpers._
```

An example of a conflicting naming: We have a scenario where having counters during development, makes designing pipelines very easy.
Currently adding counters is a very manual step. Eg. We would like to inner join two PairSCollections, but closely keep track of values which get filtered in the inner join / track duplicates which cause partial cross joins. This leads to either using a cogroup / full outer join and then flatMap to update counters. The counters are a very appropriate way of monitoring how and where bugs introduced after pipeline development in upstreams start affecting data quality adversely beyond a certain user defined threshold. This also makes it easy while troubleshooting errors.

To solve this we would like to have an extension like:
```
def join[W: Coder](that: SCollection[(K, W)], joinMetrics: JoinMetrics*): SCollection[(K, (V, W))] = self.transform { me =>
    me.cogroup(that)
      .flatMap {
        case (k, (iv, iw)) =>
          joinMetrics.foreach(_.updateCounters(iv, iw))
          for {
            v <- iv
            w <- iw
          } yield (k, (v, w))
      }
  }
```
but the `left.join(right, JoinMetrics.metricType(...))` would conflict because of a multiple implicits in scope, and hence we need to use a different name like `left.joinWCounts(right, JoinMetrics.countUnMatched("a", "b"))`.

I also think better integration with counters and the scenario described above can be a good feature for Scio as well.

regadas:
@anish749 Thanks for raising this. yeah currently in `scio` we do have have a lot of `implicits` that are brought into scope through package objects or objects. The reason behind this decision was to make it easier to use without requiring too much `import`. (important for ppl new to scala) 

However, we are in the process of uplifting some of this things and with that decrease the priority on some of them, so expect some changes in the upcoming versions.

PR #1774, opened 2019-03-25
Separate Implicits for SCollections

anish749:
I was trying to have an implicit on SCollection which would have an overloaded method already defined in SCollection. I was thinking about this in #1748 

I want have an implicit with a overloaded `join` on SCollection where the join also takes another argument defining counters on unmatched inputs.

```
def join[W](that: SCollection: [(K, W)], unMatchedCounterName = "") {
   // increment counter when a key in this doesn't match with a key in that.
}
```

With this change I would be able to add my implicits with a lower priority than the `makePairSCollectionFunctions` implicit.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -49,15 +49,7 @@ import scala.concurrent._
 import scala.reflect.ClassTag
 
 /** Convenience functions for creating SCollections. */
-object SCollection {
-
-  /**
-   * Create a union of multiple [[SCollection]] instances.
-   * Will throw an exception if the provided iterable is empty.
-   * For a version that accepts empty iterables, see [[ScioContext#unionAll]].
-   */
-  def unionAll[T: Coder](scs: Iterable[SCollection[T]]): SCollection[T] =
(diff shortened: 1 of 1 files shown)


jto:
I have mixed feelings about this. 

It seems to me that you're essentially trying to monkey patch `SCollection` which quite honestly is something I would rather discourage. IMHO Scio should rely less on implicit conversions.Making changes and adding complexity in the resolution chain to allow more implicit conversions to be created does not really go into the right direction.

There's a much easier solution to your problem: name your function something else than `join`. Given that it's not purely a `join`, I think it only makes sense.

@regadas, @nevillelyh wdyt ?

anish749:
@jto Yes, right now I am using a different name than join.

regadas:
I tend to agree as well to should be using less implicit conversions, or at least be more mindful on when to use them. We need to start thinking more around this topic because I think we will soon be bitten by the same issue.

Regardless of this, `scio` does need some cleanup around the current implemented implicit conversions, which I'm doing and goes in a different direction from this one, so I would prefer to waot on this change.

Regarding @anish749 issue, i also think that since it's a `join` with a different behavior, it should have a different name, it's clearer.

jto:
Ok. I'm closing the PR for now then. We may re-evaluate in the future.

PR #1775, opened 2019-03-25
Add a hand written Coder for pairs

jto:
GIven that `Tuple2` are extremely common, it make sense to provide a manually implemented Coder and get a bit of a performance boost. Most notably, this makes `groupByKey` a bit faster. (see added bench)

Before:
```
[info] Result "com.spotify.scio.jmh.GroupByBenchmark.testScioGroupByKey":
[info]   0.352 ±(99.9%) 0.010 s/op [Average]
[info]   (min, avg, max) = (0.324, 0.352, 0.370), stdev = 0.011
[info]   CI (99.9%): [0.342, 0.362] (assumes normal distribution)
```

After:
```
[info] Result "com.spotify.scio.jmh.GroupByBenchmark.testScioGroupByKey":
[info]   0.309 ±(99.9%) 0.004 s/op [Average]
[info]   (min, avg, max) = (0.299, 0.309, 0.320), stdev = 0.005
[info]   CI (99.9%): [0.304, 0.313] (assumes normal distribution)
[info] # Run complete. Total time: 00:10:18
```


nevillelyh:
@jto So tuple 3-22 are doing the `Array` <-> `Seq` conversion you mentioned? Does it make sense to auto-gen (like `MultiJoin`) optimized coders as well?

regadas:
@nevillelyh good point! I think we could auto-gen them.

jto:
@nevillelyh @regadas Yes but it may not be worse the effort. Tuple > 3 elements are way less common than pairs so it may be overkill.

PR #1778, opened 2019-03-26
Another take on sql syntax

regadas:
@jto @nevillelyh lmk wdyt


scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -0,0 +1,341 @@
...
+        st.registerUdf(x.fnName, x.fn)
+      case (st, x: UdafFromCombineFn[_, _, _]) =>
+        st.registerUdaf(x.fnName, x.fn)
+    }
+
+}
+
+final class SqlSCollection[A: Schema](sc: SCollection[A]) {
+
+  def queryRaw(q: String, udfs: Udf*): SCollection[Row] = query(Query(q, udfs.toList))
```
jto: I'm not a fan of the `queryRaw*` methods. IMHO they make the API more confusing with the only benefit of saving a few keystrokes.
Using the `Query` object explicitly on the other hand makes it obvious that an instance of `Query` can be shared, for example in a library.

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -0,0 +1,341 @@
...
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamRelNode
+import org.apache.beam.sdk.extensions.sql.impl.{BeamSqlEnv, ParseException}
+import org.apache.beam.sdk.extensions.sql.impl.schema.BaseBeamTable
+import org.apache.beam.sdk.extensions.sql.impl.utils.CalciteUtils
+
+import scala.language.experimental.macros
+import scala.util.Try
+import scala.collection.JavaConverters._
+
+final case class Query[A, B](query: String, udfs: List[Udf] = Nil)
```
jto: Would it be possible to also support a TupleTag with a default value here ? 

```scala
final case class Query[A, B](query: String, 
                             aTag: TupleTag[A] = new TupleTag[A]("SCOLLECTION"), 
                             udfs: List[Udf] = Nil)
```

That way we get rid of the awkward `PCOLLECTION` in queries which does not really make sense in Scio

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -0,0 +1,341 @@
...
+                                                  bTag: TupleTag[B],
+                                                  udfs: Udf*): Query2[A, B, R] =
+    macro QueryMacros.typedQuery2Impl[A, B, R]
+}
+
+object Sql {
+
+  def from[A: Schema](sc: SCollection[A]): SqlSCollection[A] = new SqlSCollection(sc)
+
+  def from[A: Schema, B: Schema](a: SCollection[A], b: SCollection[B]): SqlSCollection2[A, B] =
```
jto: Maybe the tag should be there ? Something like this would make sense to me:

```scala
Sql.from("a" -> sc1, "b" -> sc2)
  .query(Query("..."))
```

wdyt ?

jto: Ideally we'd keep track of the required binding in the type, but I'm not sure that's possible:

```scala
val q: Query[('a, User), ('b, Playlist), Int]("SELECT foo FROM a INNER JOIN b on ...)
```

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -0,0 +1,341 @@
...
+    QueryUtils
+      .typecheck(q.query, List((q.aTag.getId, schemaA), (q.bTag.getId, schemaB)), expectedSchema)
+      .right
+      .map(_ => q)
+  }
+
+  def typedQuery[A: Schema, B: Schema](query: String, udfs: Udf*): Query[A, B] =
+    macro QueryMacros.typedQuerylImpl[A, B]
+
+  def typedQuery[A: Schema, B: Schema, R: Schema](query: String,
```
jto: It's already in the `Queries` object, maybe we can just call it `typed` ?

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
regadas: Yup, good idea! also will allow uses to name their tables

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
regadas: yeah that would be nice, but it's not really possible to have `Symbol`'s like that :( also literal types (using shapeless) can be too much?

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
regadas: I don't have a strong opinion here. @nevillelyh in favor as well in removing these?

nevillelyh: :+1: on removing. What about a `def query(q: String)` for convenience when there's no UDFs?

nevillelyh: Or just rename `queryRaw` to `query` and keep the `Udf*`.

regadas: Yeah convenience was what led me to keep them

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
jto: Definitelly too much. I wish there was a way to statically check that the name you give to the `SCollection` matches the SQL query but it's probably impossible.

PR #1780, opened 2019-03-26
use camelCase for typed arguments, fix #1770

### LINKED ISSUE #1770: ContextAndArgs.typed should support camelCase arguments
ContextAndArgs.typed parses camelCase properties only if they are supplied in lisp-case form.
In contrast Beam options always use camelCase.
Different conventions make argument passing weird and confusing for the user, for example:
```bash
myJob --maxNumWorkers=20 --my-input-data=/path/to/data
```

Requests:
1. Add camelCase as an option for TypedParser in addition with lisp-case.
2. Make camelCase default option.
3. Optional: deprecate lisp-case.

nevillelyh:


nevillelyh:
Yeah it should be for 0.8, but do we still need a branch for it?

PR #1798, opened 2019-03-27
Make ensureSerializable return Either

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -145,13 +146,15 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     } else this
   }
 
-  private def checkCoderSerialization[A](coder: BCoder[A]): Unit =
+  private def ensureSerializable[A](coder: BCoder[A]): Either[Throwable, BCoder[A]] =
     coder match {
-      case _ if !context.isTest => ()
+      case c if !context.isTest =>
```
ClaireMcGinty: jw, why is this special case necessary?

regadas: tbh this is just keeping the behavior consistent to preivous versions. We can start checking every time. This will only called when the job graph is being built, so no pref impact.

PR #1799, opened 2019-03-28
Fix AvroIO type bounds

jto:
`0.7.4` accidentally introduced a breaking change because of stricter than necessary type bounds on `AvroIO`. This hsould fix it.


andrisnoko:
@jto could you add classTag and code [here](https://github.com/spotify/scio/blob/72c24ce43a7863890b82f5b9eb61f128c5730481/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala#L236) I think that breaks those that do `AvroIO[T](path)(t, c))` in tests

regadas:
@andrisnoko that's a good point coz AvroIO used to be `AvroIO[T: ClassTag: Coder]`

regadas:
After some thought and not being able to keep backwards compatibility on some other things, we will release 0.7.5 with avro changes reverted.

PR #1800, opened 2019-03-29
Add minimal beam sql word count example

regadas:


scio-examples/src/main/scala/com/spotify/scio/examples/MinimalBeamSqlWordCountExample.scala
```diff
@@ -0,0 +1,46 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Example: Minimal BeamSQL Word Count Example
+// Usage:
+
+// `sbt runMain "com.spotify.scio.examples.BeamSqlExample
```
ClaireMcGinty: `com.spotify.scio.examples.BeamSqlExample` -> `com.spotify.scio.examples. MinimalBeamSqlWordCountExample`

regadas: we can add a test

Issue #1686, opened 2019-02-18
Change Bloom Filter implementation for Sparse Joins


anish749:
In the current version of Sparse Joins and lookups, we are using Algebird's Bloom Filters Aggregators.
This works well and is very easy and elegant to express in Scala, but its not performant. It takes a long time and CPU to build this filter (~3 hrs wall time for 3M entries). For smaller datasets (~5M entries) I found a simple HashSet filter to work better, though that starts to go near the limits of SideInput size. For entries around 50M, it starts taking longer than a shuffle join. 
The Bloom Filters from Breeze as well as ClearSpring Stream lib however gets constructed within 3 to 4 min wall time for 3M entries. 

**Suggestion 1:**
I feel using Stream Lib from AddThis / ClearSpring instead of Algebird BloomFilter would allow us to optimise a lot more jobs. I am not sure why, but I was having a very high false positives with Breeze BFs (may be I was making a mistake, as it seems weird)

Breaking API changes(?)
Algebird provides us with a Hash128[K], however the BF in StreamLib takes a byte array / String, which means we would need to add an implicit ByteArrayEncoder for primitive types which are present in algebird's Hash128 to make sure we are not breaking any API.
Can we go ahead with adding this additional implicit, and move away from Algebird's BF for sparse functions?

Also because this adds an additional dependency, should this be a part of `scio-extra` which would again means removing those methods from `scio-core` thus breaking again? I am a bit confused about adding an extra dependency.

I have some code in an internal repository using StreamLib and here are some benchmarks. Based on suggestions on the above, I can create a PR fixing this. 

Here is a Dataflow join of ~4M with two datasets of 1B each.
Using Algebird Bloom Filters
![screenshot 2019-02-15 at 16 57 41](https://user-images.githubusercontent.com/15274312/52946405-37637a80-3374-11e9-88a2-1865c6f27645.png)

Using ClearSpring / AddThis BloomFilters
![screenshot 2019-02-15 at 16 57 30](https://user-images.githubusercontent.com/15274312/52946362-2155ba00-3374-11e9-8d64-f84fd4319e88.png)

**Suggestion 2:**
Optimise Algebird Bloom Filters, and make it run faster. The Bloom Filter in Algebird is running very slow, but it can surely be fixed to make this work faster.

regadas:
@anish749 interesting stuff! I'll have a look at this and I'll come back to you.

anish749:
[...]

PR #1806, opened 2019-04-02
Change BloomFilter implementation for Sparse Joins

anish749:
I am trying to add the optimisations from  #1686 and https://github.com/twitter/algebird/pull/673.

This involves using a `java.util.BitSet` instead of a custom `Array[Long]` to back the BloomFilter, and not reallocating the BitSet after adding each element. Since adding elements to a BloomFilter are idempotent, and we do not need a reference to the old BloomFilter after it has been created, using a mutable BloomFilter works faster with less things to copy. 

Here is what I did:
Use `java.util.BitSet` backed bloom filter from https://github.com/twitter/algebird/pull/673.
It works but the memory allocation is based on the width of the filter which is derived from the number of keys hint passed in sparse joins.
This means if the hint is misleading (abnormally high) we end up with OOM. Also we have test cases which pass 1Billion elements but doesn't actually add so many elements. This means allocating memory for for the estimated width. The original Algebird BF uses a compressed bitmap and has a sparse bloom filter implementation when less than 10% of the bits are used. This reduces memory usage.
Now, EWAHCompressedBitMaps, used by Algebird, is immutable and is very slow when inserting, because it again copies and reallocates when inserting data. Using this bitmap would again reduce the throughput compared to ClearSpring's BF that I used in #1686. Also this would use Kryo serialization since this bitmap doesn't have a Beam Coder, but that is probably not a problem.
So now we write another compressed mutable bitmap which essentially is a `mutable.Set[Int]`. This works better, and has BeamCoders. With this we don't have any OOM issues for sparse filters and is also mutable, hence we are not reallocating.
But this is slower than ClearSpring. I didn't quite understand why. But it was 50% slower in terms of throughput compared to `util.BitSet`. 
Then we build a `DelayedBloomFilter` (in this version) where we just calculate the hashes(as an `Array[Int]`), and keep the references in a `mutable.Buffer[Array[Int]]` when inserting values.
We wait till this is filled up by the aggregator to a point where this no longer offers any benefit of not allocation the complete memory of `util.BitSet` and then we change to the `MutableBFInstance[T]`. When the first query happens, we materialise the `mutable.Buffer[Array[Int]]` into a `Set[Int]` and store that, to answer queries. For multiple queries, we don't re materialise this. If a new element is inserted, we discard the previously materialised `Set[Int]` and wait till we see the first query. We use aggregators here, which are nice to us and they don't interleave inserts and reads from the filter. This means we can delay the Set allocation till we have a query. Also this BF is not meant to be in caches and only in ephemeral pipelines, which means that we have clearly separated steps for creation and usage of the bloom filter. 

With the above implementation,
we have the following results for putting 1M elements in the BloomFilter:
_Original Implementation_ (v0.7.4)
Screenshot 2019-04-25 at 14 28 44
_Current Branch_ (v0.7.5-SNAPSHOT)
Screenshot 2019-04-25 at 14 28 58

JMH Benchmark (locally outside this repo)
```
Benchmark                         (falsePositiveRate)  (nbrOfElements)   Mode  Cnt      Score       Error  Units
algebirdBF                 0.01              100  thrpt    3  13146.325 ± 11543.779  ops/s
algebirdBF                 0.01            10000  thrpt    3      6.977 ±     4.191  ops/s
clearSpringBF                 0.01              100  thrpt    3  65031.775 ±  5328.316  ops/s
clearSpringBF                 0.01            10000  thrpt    3    575.296 ±   544.399  ops/s
scioBF                        0.01              100  thrpt    3  53440.614 ± 17581.014  ops/s
scioBF                        0.01            10000  thrpt    3    499.989 ±   312.770  ops/s
```


scio-core/src/main/scala/com/spotify/scio/coders/instances/AlgebirdCoders.scala
```diff
@@ -18,11 +18,10 @@
 package com.spotify.scio.coders.instances
 
 import com.spotify.scio.coders.Coder
-import com.twitter.algebird.{BF, Batched, CMS, TopK}
+import com.twitter.algebird.{Batched, CMS, TopK}
 
 trait AlgebirdCoders {
   implicit def cmsCoder[K]: Coder[CMS[K]] = Coder.kryo
-  implicit def bfCoder[K]: Coder[BF[K]] = Coder.kryo
```
anish749: This might be a breaking change, since for Bloom Filters which are less than 10% filled, Algebird uses a Java EWAH Compressed Bitmap and we won't be having a Beam Coder for that. This means that if some one is using a Algebird Bloom Filter in their pipeline along with Scio, they would get an exception and coder derivation would fail. 

Should we keep this as is?

anish749: it won't interfere. I'll revert this change.

anish749:
Thanks for the review @nevillelyh .
I'll look into the comments by tomorrow and fix them, and also resolve conflicts.
I tried the mutation detection, and it works. I'll add that and a `verifyDeterministic` test on the coder as a unit test:
```
val bf: MutableBF[String] = BloomFilter[String](10, 0.1).create(items: _*)
val beamMutationDetector = MutationDetectors
  .forValueWithCoder(
      bf,
      CoderMaterializer
        .beam(ScioContext.forTest(), Coder[MutableBF[String]]) // Create Coder for MutableBF.
    )
beamMutationDetector.verifyUnmodified()
```

For your second point, any pointers about comparing the efficiency of coders?

scio-core/src/main/scala/com/spotify/scio/coders/instances/AlgebirdCoders.scala
regadas: +1 in reverting it.

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+      // p = (1 - e^(-numHashes * setCardinality / width))^numHashes
+      // See: http://en.wikipedia.org/wiki/Bloom_filter#Probability_of_false_positives
+      //
+      // However, the true set cardinality may not be known. From empirical evidence, though,
+      // it is upper bounded with high probability by 1.1 * estimatedCardinality (as long as the
+      // Bloom filter is not too full), so we plug this into the formula instead.
+      // TODO: investigate this upper bound and density more closely (or derive a better formula).
+      // TODO: The following logic is same for immutable Bloom Filters and may be referred here.
+      val fpProb =
+        if (density > 0.95)
```
regadas: 🎨 use some `{}` just because it's a multiline

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+  def hammingDistance(that: MutableBF[A]): Int =
+    (this, that) match {
+      // Comparing with empty filter should give number
+      // of bits in other set
+      case (x: MutableBFZero[A], y: MutableBFZero[A]) => 0
+      case (x: MutableBFZero[A], y: MutableBF[A])     => y.numBits
+      case (x: MutableBF[A], y: MutableBFZero[A])     => x.numBits
+
+      // Otherwise compare as bit sets
+      case (_, _) =>
```
regadas:
```suggestion
      case _ =>
```

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+        (a eq b) || ((a.numHashes == b.numHashes) &&
+          (a.width == b.width) &&
+          a.toBitSet.equals(b.toBitSet))
+    }
+}
+
+/**
+ * A Mutable Bloom Filter data structure
+ */
+private[scio] sealed abstract class MutableBF[A] extends java.io.Serializable {
```
regadas:
```suggestion
private[scio] sealed abstract class MutableBF[A] extends Serializable {
```

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+    }
+
+}
+
+/**
+ * Empty bloom filter.
+ */
+private[scio] final case class MutableBFZero[A](hashes: KirMit32Hash[A]) extends MutableBF[A] {
+
+  lazy val toBitSet: util.BitSet = new util.BitSet()
```
regadas: does this need to be `lazy`?

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+
+  lazy val numHashes: Int = hashes.numHashes
+
+  lazy val width: Int = hashes.width
+
+  lazy val numBits: Int = 0
```
regadas: does this need to be lazy?

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+  // scalastyle:on method.name
+
+  def checkAndAdd(other: A): (MutableBF[A], ApproximateBoolean) =
+    (this += other, ApproximateBoolean.exactFalse)
+
+  override def contains(item: A): ApproximateBoolean = ApproximateBoolean.exactFalse
+
+  def maybeContains(item: A): Boolean = false
+
+  lazy val size: Approximate[Long] = Approximate.exact[Long](0)
```
regadas: does this need to be lazy?

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+
+  def toBitSet: util.BitSet = bits
+
+  // scalastyle:off method.name
+  def ++=(other: MutableBF[A]): MutableBF[A] = {
+    require(this.width == other.width)
+    require(this.numHashes == other.numHashes)
+
+    other match {
+      case MutableBFZero(_)                         => this
```
regadas:
```suggestion
      case _: MutableBFZero                         => this
```

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+}
+
+/**
+ * Empty bloom filter.
+ */
+private[scio] final case class MutableBFZero[A](hashes: KirMit32Hash[A]) extends MutableBF[A] {
+
+  lazy val toBitSet: util.BitSet = new util.BitSet()
+
+  lazy val numHashes: Int = hashes.numHashes
```
regadas: does this need to be lazy?

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+  // scalastyle:off method.name
+  def ++=(other: MutableBF[A]): MutableBF[A] = {
+    require(this.width == other.width)
+    require(this.numHashes == other.numHashes)
+
+    other match {
+      case MutableBFZero(_)                         => this
+      case MutableSparseBFInstance(_, otherSetBits) =>
+        // This is MutableBFInstance, hence not sparse, so don't convert output to sparse.
+        val it = otherSetBits.flatten.iterator
```
regadas: invert this to avoid extra copy `.iterator.flatten`

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+    (this += other, doesContain)
+  }
+
+  // scalastyle:off return
+  def maybeContains(item: A): Boolean = {
+    val il = hashes(item)
+    var idx = 0
+    while (idx < il.length) {
+      val i = il(idx)
+      if (!bits.get(i)) return false
```
regadas: `return` should be avoided

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+ * EWAHCompressedBitmap is not used because OR operations are immutable and copies the underlying
+ * bitmap. Also Apache Beam doesn't have a Coder for EWAHCompressedBitmap, and it would fallback
+ * to Kryo
+ */
+final case class MutableSparseBFInstance[A](
+  hashes: KirMit32Hash[A],
+  allHashes: mutable.Buffer[Array[Int]]
+) extends MutableBF[A] {
+
+  lazy val numHashes: Int = hashes.numHashes
```
regadas: remove `lazy`?

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+
+  /**
+   * The number of bits set to true
+   */
+  def numBits: Int = setBits.size
+
+  lazy val width: Int = hashes.width
```
regadas: remove `lazy`?

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+
+  private def allSeenBit: mutable.Buffer[Int] = allHashes.flatten
+
+  private var set: Set[Int] = _
```
regadas: private[this]

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+  // Keeps a state if elements were added after `allHashes` was converted into a set.
+  private var setIsStale: Boolean = true
```
regadas: private[this]

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+
+  /**
+   * Merge two mutable BFs. Assumes both have same width and hashing strategy.
+   *
+   * Return a new [[MutableBFInstance]] if we should no longer stay sparse.
+   */
+  // scalastyle:off method.name
+  def ++=(other: MutableBF[A]): MutableBF[A] = {
+    other match {
+      case MutableBFZero(_) => this
```
regadas: `case _: MutableBFZero[A]`

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+  }
+
+  // scalastyle:off return
+  def maybeContains(item: A): Boolean = {
+    val sb = setBits // eval setBits only once.
+    val il = hashes(item)
+    var idx = 0
+    while (idx < il.length) {
+      val i = il(idx)
+      if (!sb.contains(i)) return false
```
regadas: avoid `return`

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+    BloomFilter.sizeEstimate(numBits, numHashes, width, 0.05)
+
+  def copy: MutableBF[A] = MutableSparseBFInstance(hashes, allHashes.clone)
+}
+
+/**
+ * Constructors for mutable bloom filters
+ */
+object MutableBFInstance {
+  def apply[A](hashes: KirMit32Hash[A], firstElement: A): MutableBF[A] = {
```
regadas: `final`

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+    val bf = MutableBFInstance.empty(hashes)
+    bf += firstElement
+  }
+
+  def apply[A](hashes: KirMit32Hash[A]): MutableBF[A] =
```
regadas: `@inline final`

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+    empty(hashes)
+
+  // Always Start with a Sparse BF Instance
+  def empty[A](hashes: KirMit32Hash[A]): MutableBF[A] =
```
regadas: `@inline final`

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
anish749: it doesn't, I was trying to some tests on memory allocations, but it wasn't fruitful. I'll revert these.

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
anish749: Thanks, this is hard to notice. :)

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,576 @@
...
+      None
+    } else {
+      // We start with an empty instance here.
+      // An empty instance is always Sparse, and doesn't allocate the complete
+      // memory of the underlying bit map. When adding BFs slowly, at one point
+      // it is no longer sparse and hence `++=` returns a BF with the underlying
+      // bitmap allocated. For later operations we must hold on to the new BF,
+      // and hence we use var here, even though `++=` mutates the instance.
+      var outputInstance = MutableBFInstance.empty(hashes)
+      as.foreach { bf =>
```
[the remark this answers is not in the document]
anish749: I tried this out, and and the benchmarks have improved a bit as well. Thanks for pointing out.

anish749:
@regadas thanks for reviewing. I've made the changes.

anish749:
Here are the updated benchmarks
```
Benchmark      (falsePositiveRate)  (nbrOfElements)   Mode  Cnt     Score     Error  Units
clearSpringBF                 0.01             1000  thrpt    7  5987.608 ± 1363.882  ops/s
clearSpringBF                 0.01            10000  thrpt    7   652.844 ±   63.079  ops/s
scioBF                        0.01             1000  thrpt    7  5155.060 ±  311.114  ops/s
scioBF                        0.01            10000  thrpt    7   425.954 ±   99.796  ops/s
```

scio-core/src/main/scala/com/spotify/scio/util/BloomFilter.scala
```diff
@@ -0,0 +1,600 @@
...
+ */
+private[scio] final case class MutableBFZero[A](hashes: KirMit32Hash[A]) extends MutableBF[A] {
+
+  def toBitSet: util.BitSet = new util.BitSet()
+
+  def numHashes: Int = hashes.numHashes
+
+  def width: Int = hashes.width
+
+  def numBits: Int = 0
```
regadas: Hi @anish749 you could have kept all of this as `val`. the important part was to remove the `lazy`

anish749: umm, I was using `def` first then moved to `lazy val` to not re-evaluate. I am a bit hesitant with `val` because it would be allocating values which are copies of other `Int` and it might increase memory footprint. Also in most cases here, they are not called (hence `lazy val` didn't improve speed compared to `def`)

regadas: It's a valid concern, however it seems to me that some of them will be re-evaluated multiple times and I'm not sure if this cost is lower compared to the usage of `val` given that some of them are actually just ref's and also that there's higher change of getting inlining by the VM. Also note that you were comparing with `lazy val` and it has hidden cost as well.

Another thing is that `val` conveys a better meaning since these values will not actually change.

This interesting stuff and It would be awesome if we could bench it 😄

regadas: on a second thought, the usage of `val` here would need to be carefully applied, because it can really increase object size. Let's keep this as is for now! numbers look pretty good already!

anish749: Yes, it's a bit of a trade off.. I did try different approaches, but the jmh benchmarks doesn't give very accurate results on these changes, so its harder to benchmark.

PR #1808, opened 2019-04-03
Add support for Udf and joins in typed queries

jto:


jto:
Tests fail with a weird error on my machine. Want to see if CI also fails.

scio-test/src/test/scala/com/spotify/scio/sql/TypedBeamSQLTest.scala
```diff
@@ -0,0 +1,121 @@
...
+
+import com.spotify.scio.bean.UserBean
+import com.spotify.scio.schemas.To
+import com.spotify.scio.testing.PipelineSpec
+import org.apache.beam.sdk.values.TupleTag
+
+class TypedBeamSQLTest extends PipelineSpec {
+  import TestData._
+
+  new IsOver18UdfFn()
```
regadas: remove?

scio-test/src/test/scala/com/spotify/scio/sql/BeamSQLTest.scala
```diff
@@ -34,6 +34,14 @@ import org.scalatest.Assertion
 import scala.collection.JavaConverters._
 import com.spotify.scio.avro
 
+class IsOver18UdfFn extends SerializableFunction[Integer, Boolean] {
```
regadas: wondering why move outside of `TestData`? Could we move the hole `TestData` outside of `BeamSQLTest`?

since we have more UDF/s inside TestData ... this seems a bit weird?

jto: I was just testing if this it what makes the compiler crash. Apparently not...

regadas: @jto yeah I realized this PR as a bunch of test stuff after reading your comments on slack ... all makes sense now 😄

jto:
I removed the support of `Udf` from typed queries.

What happens is basically a compilation order issue. The compilation of the implementation of the `Udf` and the expansion of the macro happen in the same compilation unit. Because the macro tries to `eval` its parameters, it ends up in a situation where the `Udf` it's trying to `eval` have not been compiled yet, which causes a `ClassNotFoundException`.

If the `Udf` is built in another compilation unit, the macro works fine. I don't think it's a good idea to trust our users to not make that mistake and I can't think of a way to make that work reliably, so I just removed the feature.

@regadas @nevillelyh WDYT ?

jto:
Yeah so I spent most of yesterday and a few hours today looking into this, and apart from losing a bit of my will to live I don't think I've acheived much. I think we can just merge this PR as it is and unless someone else feels like spending (wasting ?) some time trying to make this work, just accept that scalac can't do that.

I'm removing the WIP tag. feel free to review / merge.

regadas:
@jto I'm ok with removing UDF's from `typed`. Maybe we should can mark it as `@experimental`. That way if we find a solution all good otherwise we can always decide to even remove this or keep it.

I think there's some stuff remaining from the tests

jto:
> I think there's some stuff remaining from the tests
Not sure what you're talking about here.

scio-test/src/test/scala/com/spotify/scio/sql/BeamSQLTest.scala
regadas: @jto can we move back these classes?

PR #1820, opened 2019-04-08
Fix NPE in coders with nullableCoders=true

jto:


scio-test/src/test/scala/com/spotify/scio/coders/coders.scala
```diff
@@ -396,6 +396,26 @@ class CodersTest extends FlatSpec with Matchers {
...
+    type T = (String, Int, Top)
+    val example: T = ("Hello", 42, TA(1, "World"))
+    val nullExample1: T = ("Hello", 42, TA(1, null))
+    val nullExample2: T = ("Hello", 42, null)
+    example coderShould roundtrip(opts)
+    nullExample1 coderShould roundtrip(opts)
+    nullExample2 coderShould roundtrip(opts)
+
+    val nullBCoder = CoderMaterializer.beamWithDefault(Coder[T], o = opts)
+    nullBCoder.isRegisterByteSizeObserverCheap(nullExample1)
```
regadas: @jto why do we need this?

scio-test/src/test/scala/com/spotify/scio/coders/coders.scala
```diff
@@ -396,6 +396,26 @@ class CodersTest extends FlatSpec with Matchers {
...
+    nullBCoder.isRegisterByteSizeObserverCheap(nullExample2)
+
+    val noopObserver =
+      new org.apache.beam.sdk.util.common.ElementByteSizeObserver {
+        def reportElementSize(s: Long) = ()
+      }
+
+    nullBCoder.registerByteSizeObserver(nullExample1, noopObserver)
```
regadas: @jto same here, since we are not checking the expected byte count.

jto: Its not about checking the size, it about checking that it does not throw a NPE

regadas: @jto i see, so in that case I'm a bit confused. As far as I can tell both implementations (`roundtrip` matcher and `registerByteSizeObserver`) will try to encode the `elem` and they will throw an exception in the presence of `null` if `NullEncoder` is not used. IMO I find the `registerByteSizeObserver` a bit confusing, since it's kinda testing the same thing and it's not very explicit about it.

jto: There was an issue on the internal channel about someone haveing a issue specifically with `registerByteSizeObserver` throwing a NPE. Given that the test is a 2 liner, I prefer to be explicit about it.

Issue #1805, opened 2019-04-01
Add gRPC/HTTP examples


regadas:

nevillelyh:
We have something working internally now, not sure if it's generalizable for public consumption though. Closing.

PR #1824, opened 2019-04-09
Add HTTP example (#1805)

andrisnoko:


scio-examples/src/main/scala/com/spotify/scio/examples/extra/HttpAsyncLookupExample.scala
```diff
@@ -0,0 +1,128 @@
...
+    HttpAsyncLookupExample.okHttpClient
+      .newCall(request)
+      .enqueue(new Callback() {
+        def onFailure(request: Request, exception: IOException): Unit = {
+          result.setException(exception)
+          ()
+        }
+
+        def onResponse(response: Response): Unit = {
+          if (response.isSuccessful) {
```
nevillelyh: Why `Option` if both if/else branches are `Some`? Unless if `response.body.string` can be `null` but in that case `input.get` on #79 will NPE?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/HttpAsyncLookupExample.scala
```diff
@@ -0,0 +1,128 @@
...
+    extends AsyncLookupDoFn[Account, String, AsyncHttpClient](100, new LookupCacheSupplier) {
+
+  override def asyncLookup(client: AsyncHttpClient, input: Account): ListenableFuture[String] = {
+    Futures.transform(
+      client.get(url + "?id=" + input.id),
+      new com.google.common.base.Function[Option[String], String] {
+        override def apply(input: Option[String]): String =
+          input.get
+      },
+      MoreExecutors.directExecutor
```
nevillelyh: Direct executor is the default if this argument is not supplied?

scio-examples/src/test/scala/com/spotify/scio/examples/extra/HttpAsyncLookupExampleTest.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    JobTest[com.spotify.scio.examples.extra.HttpAsyncLookupExample.type]
+      .args("--output=out.text", "--url=http://localhost:9000/exception")
+      .output(TextIO("out.text"))(_ should containInAnyOrder(expected))
+      .run()
+  }
+}
+
+object SimpleHttpServer {
+  lazy val server = HttpServer.create(new InetSocketAddress(9000), 0)
```
nevillelyh: Usually a good idea to randomize ports in tests to avoid collisions.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/HttpAsyncLookupExample.scala
andrisnoko: The `transform` that didn't take an executor was deprecated.

scio-examples/src/test/scala/com/spotify/scio/examples/extra/HttpAsyncLookupExampleTest.scala
andrisnoko: Noted.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/HttpAsyncLookupExample.scala
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.beam.sdk.transforms.ParDo
+import org.apache.beam.sdk.values.KV
+
+import scala.util.Success
+
+object HttpAsyncLookupExample {
+
+  case class Account(id: String, amount: Double)
+
+  lazy val okHttpClient = new OkHttpClient
```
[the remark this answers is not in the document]
andrisnoko: Good catch.

build.sbt
```diff
@@ -735,6 +735,7 @@ lazy val scioExamples: Project = Project(
       "mysql" % "mysql-connector-java" % "8.0.15",
       "com.google.cloud.sql" % "mysql-socket-factory" % "1.0.13",
       "org.slf4j" % "slf4j-simple" % slf4jVersion,
+      "com.squareup.okhttp3" % "okhttp" % "3.8.0",
```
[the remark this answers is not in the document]
regadas: +1 on this ... I would like to see a more idiomatic client here.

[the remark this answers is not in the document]
andrisnoko: I agree with having idiomatic client, but we should have Scala version of `AsyncLookupDoFn` so we don't have to be converting from Scala `Future` to Java `Future` and possible to `ListenableFuture`.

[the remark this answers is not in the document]
nevillelyh: Yeah maybe time to have those counterparts now. All 3 futures behavior slightly differently and could cause subtle bugs if not used correctly, especial the way we're using them.

jto:
What's the status of this PR @andrisnoko ?

nevillelyh:
Another abandon-ware.

PR #1826, opened 2019-04-09
Add fluent backoff support for ES

regadas:


soundofjw:
Woo! Glad to see this come in.
Also, hilarious this is exactly 200 PRs after my initial implementation in #1626 

Thank you everyone!

Issue #1813, opened 2019-04-04
Review BigQuery data type support


nevillelyh:
Looking at the documentation there's at least `GEOGRAPHY` that's not covered in our code base. Maybe it's time to do another round of review.
https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types

nevillelyh:
So there're some edge cases in the new `NUMERIC` type coverage. I have a PR coming.

`GEOGRAPHY` is a bit messy. There're various JSON/stringly representations so I'm tempted to map `GEOGRAPHY` to `String` in Scala but that means 1-to-2 mapping and we'll need some way to mark a `String` field (maybe annotation?) for `@BigQueryType.toTable`.
https://cloud.google.com/bigquery/docs/gis-data

nevillelyh:
Created #1882 for GEOGRAPHY. Closing this since all other types are consistent.

PR #1827, opened 2019-04-09
cover BigQuery NUMERIC missing cases #1813

nevillelyh:


jto:
Is that a potentially breaking change ?
If not, LGTM.

nevillelyh:
The NUMERIC one definitely is. The legacy to standard SQL one should be too but I need to test more thoroughly.

nevillelyh:
Looks like BQ is still using `INTEGER`, `FLOAT`, `BOOLEAN`, etc. for field types. I'm removing the last commit.

PR #1834, opened 2019-04-11
Create empty Sparkey file on empty collection

regadas:
Given the properties of sparkey we can't really implement it as a `FileBasedSink` (AFAICS). Also, given the semantic of our current API we couldn't actually implementent it as a sink. 

This PR is an alternative solution to #1726

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
@@ -17,14 +17,13 @@
 
 package com.spotify.scio.extra
 
-import java.util.UUID
+import java.util.{UUID, List => JList}
 
 import com.spotify.scio.ScioContext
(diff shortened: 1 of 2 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -124,18 +123,23 @@ package object sparkey {
...
-              }
-              writer.close()
-              uri
+
+      val coder = CoderMaterializer.beam(self.context, Coder[JList[(K, V)]])
+      self.transform { coll =>
+        coll.context
+          .wrap {
+            val view = coll.applyInternal(View.asList[(K, V)]())
+            coll.internal.getPipeline.apply(Reify.viewInGlobalWindow(view, coder))
```
nevillelyh: `Reify.viewInGlobalWindow` these API names are so cryptic... :man_facepalming:

PR #1835, opened 2019-04-11
Remove references of Future[Tap[T]]

regadas:


ClaireMcGinty:
I think there are some outstanding usage “s of `Future[Tap]` in some of the dynamic io packages. Just wondering if there’s a reason they haven’t been moved over or if it’s on the to do list :)

regadas:
yup I noticed that when i found out about these ... there's no reason just missed them out.

regadas:
#1838 should address this ... hopefully we didn't forget anything else 😄

PR #1838, opened 2019-04-11
Add bigquery syntax package

regadas:


nevillelyh:
Still in progress?

regadas:
yup there are few more things that I want to do.

PR #1843, opened 2019-04-15
Use java.io.tmpdir and username as cache dir

regadas:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
```diff
@@ -26,7 +29,10 @@ import scala.util.Try
 object BigQueryConfig {
 
   /** Default cache directory. */
-  private[this] val CacheDirectoryDefault: String = sys.props("user.dir") + "/.bigquery"
+  private[this] val CacheDirectoryDefault: Path = Paths
+    .get(CoreSysProps.TmpDir.value)
+    .resolve(CoreSysProps.User.value)
```
nevillelyh: This will be something like `/tmp/$USER/`? What about `/tmp/scio-bigquery-$USER` just to be more explicit?

regadas: yeah we can do that

PR #1850, opened 2019-04-16
Fix sql typecheck message

regadas:


scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -238,9 +238,12 @@ object Queries {
...
-           |${inferredSchemas.map(i => PrettyPrint.prettyPrint(i._2.getFields.asScala.toList))}
-           |Query result schema (infered) is unknown
+           |SCOLLECTION schema:
+           |${inferredSchemas
+             .map(i => PrettyPrint.prettyPrint(i._2.getFields.asScala.toList))
+             .mkString("\n")}
+           |
+           |Query result schema (infered) is unknown.
            |Expected schema:
            |${PrettyPrint.prettyPrint(expectedSchema.getFields.asScala.toList)}
```
ClaireMcGinty: Should we have the same formatting changes for the "expected schema" error message (just adding the mkString delimiter) so it's easier for the user to compare them? actually it looks like Beam's `Schema::toString` method does something really similar to our implementation.

regadas: the expected schema is just one. `Schema::toString` and `Field::toString` don't output a nice string. That's why we have the pretty print. 

beams:

```
Fields:
Field{name=value, description=, type=FieldType{typeName=INT32, nullable=false, collectionElementType=null, mapKeyType=null, mapValueType=null, rowSchema=null, metadata=null}}
```

our:

```
Expected schema:
┌──────────────────────────────────────────┬──────────┬──────────┐
│ NAME                                     │ TYPE     │ NULLABLE │
├──────────────────────────────────────────┼──────────┼──────────┤
│ value                                    │ INT32    │ NO       │
└──────────────────────────────────────────┴──────────┴──────────┘
```

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -238,9 +238,12 @@ object Queries {
...
            |
-           |PCOLLECTION schema:
-           |${inferredSchemas.map(i => PrettyPrint.prettyPrint(i._2.getFields.asScala.toList))}
-           |Query result schema (infered) is unknown
+           |SCOLLECTION schema:
+           |${inferredSchemas
+             .map(i => PrettyPrint.prettyPrint(i._2.getFields.asScala.toList))
+             .mkString("\n")}
+           |
+           |Query result schema (infered) is unknown.
```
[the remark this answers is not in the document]
regadas: nice catch 😄

PR #1856, opened 2019-04-18
warn combine*/sum* transform usage, fix #1620

### LINKED ISSUE #1620: None.get errors when aggregating in a window
Hi!

I have been running into issues in a streaming pipeline while draining on the dataflow runner. This is the exception I have been receiving: 
```
java.util.NoSuchElementException: None.get
at scala.None$.get(Option.scala:347)
at scala.None$.get(Option.scala:345)
at com.spotify.scio.util.Functions$$anon$5.extractOutput(Functions.scala:134)
at com.spotify.scio.util.Functions$$anon$5.extractOutput(Functions.scala:96)
at org.apache.beam.runners.dataflow.worker.WindmillStateInternals$WindmillCombiningState.read(WindmillStateInternals.java:920)
at org.apache.beam.runners.core.SystemReduceFn.onTrigger(SystemReduceFn.java:125)
at org.apache.beam.runners.core.ReduceFnRunner.onTrigger(ReduceFnRunner.java:1060)
at org.apache.beam.runners.core.ReduceFnRunner.onTimers(ReduceFnRunner.java:768)
at org.apache.beam.runners.dataflow.worker.StreamingGroupAlsoByWindowViaWindowSetFn.processElement(StreamingGroupAlsoByWindowViaWindowSetFn.java:95)
at org.apache.beam.runners.dataflow.worker.StreamingGroupAlsoByWindowViaWindowSetFn.processElement(StreamingGroupAlsoByWindowViaWindowSetFn.java:42)
at org.apache.beam.runners.dataflow.worker.GroupAlsoByWindowFnRunner.invokeProcessElement(GroupAlsoByWindowFnRunner.java:115)
at org.apache.beam.runners.dataflow.worker.GroupAlsoByWindowFnRunner.processElement(GroupAlsoByWindowFnRunner.java:73)
at org.apache.beam.runners.core.LateDataDroppingDoFnRunner.processElement(LateDataDroppingDoFnRunner.java:80)
at org.apache.beam.runners.dataflow.worker.GroupAlsoByWindowsParDoFn.processElement(GroupAlsoByWindowsParDoFn.java:135)
at org.apache.beam.runners.dataflow.worker.util.common.worker.ParDoOperation.process(ParDoOperation.java:45)
at org.apache.beam.runners.dataflow.worker.util.common.worker.OutputReceiver.process(OutputReceiver.java:50)
at org.apache.beam.runners.dataflow.worker.util.common.worker.ReadOperation.runReadLoop(ReadOperation.java:202)
at org.apache.beam.runners.dataflow.worker.util.common.worker.ReadOperation.start(ReadOperation.java:160)
at org.apache.beam.runners.dataflow.worker.util.common.worker.MapTaskExecutor.execute(MapTaskExecutor.java:77)
at org.apache.beam.runners.dataflow.worker.StreamingDataflowWorker.process(StreamingDataflowWorker.java:1226)
at org.apache.beam.runners.dataflow.worker.StreamingDataflowWorker.access$1000(StreamingDataflowWorker.java:141)
at org.apache.beam.runners.dataflow.worker.StreamingDataflowWorker$6.run(StreamingDataflowWorker.java:965)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
at java.lang.Thread.run(Thread.java:745)
```
This exception occurs when I try to drain a streaming pipeline containing a combineByKey on some tuples. My pipeline looks something like this:

```scala
sc.customInput("Input", pubsubIn(args("inputSub")))
      .map(unmarshalJson)
      .filter{!_.isEmpty}
      .withFixedWindows(
        duration = Duration.standardSeconds(windowSizeSeconds),
        options = WindowOptions(
          trigger = AfterWatermark.pastEndOfWindow()
          .withEarlyFirings(AfterProcessingTime.pastFirstElementInPane()
              .plusDelayOf(Duration.standardSeconds(earlyFiringSeconds)))
            .withLateFirings(AfterProcessingTime.pastFirstElementInPane()
              .plusDelayOf(Duration.standardSeconds(lateFiringSeconds))),
          accumulationMode = ACCUMULATING_FIRED_PANES,
          allowedLateness = Duration.standardDays(allowedLatenessDays)
        )
      )
      .distinctBy[String](f = getId)
      .filter{_!=null} // I also have noticed quite a few nulls when draining the pipeline
      .flatMap{m: Map[String, Any]=> 
         val k = (m.get("user").get.asInstanceOf[String], m.get("customer").get.asInstanceOf[String])
         val v = (m.getOrElse("label1", ""), m.getOrElse("label2", ""))
         Some((k, v)).toTraversable
      // This combines labels from the same key and customer into a set. So all the different 
      // label1s for a user are aggregated to a set of unique values and all the label2s for a user
      // are also aggregated into a set
      .combineByKey((v: (Any, Any))=>(Set(v._1), Set(v._2)))((s: (Set[Any], Set[Any]),v: (Any, Any))=>(s._1+v._1, s._2+v._2))((sa: (Set[Any], Set[Any]),sb: (Set[Any], Set[Any]))=>(sa._1++sb._1, sa._2++sb._2))
      // I have been using this map as a way of checking the integrity of the post aggregated values.
      // I have noticed that if I remove this map step I don't see issues but I don't see how this 
      // could cause the exception thrown.
      .map{kv: ((Any, Any), (Set[Any], Set[Any]))=>
        kv match {
          case null => {
            logger.info(s"kv was null or none! $kv")
          }
          case _ => {
            val k = kv._1
            val v = kv._2
            (k, v) match {
              case (null, _) | (_, null) => {
                logger.warn(s"k or v was null or none! ($k, $v)")
              }
              case _ => {
                kv
              }
            }
          }
        }
      .map{m=>
          mapper.writeValueAsString(m)
      }
      .saveAsPubsub(args("outputTopic"))
```
I am windowing into 60 second windows with 30 second early and late firings and an allowed lateness of 3 days. All the test data I am sending through is not delayed.

I have been reading through the scio code and attempting to reproduce with as simple a pipeline as possible. Although I could very well be mistaken I believe there may be an issue in the [mergeAccumulators](https://github.com/spotify/scio/blob/v0.7.0-beta1/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala#L136) function. As far as I understand it, mergeAccumulators will return a tuple of `None` and an empty arrayList when `flattened` is empty. However if the last accumulator after all the accumulators have been merged is also empty, then a call to `extractOutput` would result in the exception I am receiving. It appears that when I drain my pipeline I am getting a lot of empty accumulators or none at all.

Instead of returning `None` for the value of `combined`, wouldn't we want to return an empty instance of the accumulator the user specifies (in my case it would be a `(Set[Any], Set[Any])`)? I could be misunderstanding the code as I am new to both scio and Beam but this is what seems to be causing this issue to me.

### Version
Scio: `0.7.0-beta1 `
Beam: `2.8.0`

### LINKED ISSUE #1811: BigQuery Storage API support (direct access)
This is to track the progress of "direct access" Storage API, which may allow us to access granular BQ data without performing a query, mirroring projection/predicate functionality of Parquet.

https://cloud.google.com/bigquery/docs/reference/storage/
https://beam.apache.org/documentation/io/built-in/google-bigquery/#storage-api

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -55,6 +56,8 @@ import scala.util.Try
 /** Convenience functions for creating SCollections. */
 object SCollection {
 
+  private[values] val logger = LoggerFactory.getLogger(this.getClass)
```
regadas:
```suggestion
  private[values] val Logger = LoggerFactory.getLogger(this.getClass)
```

`PairSCollectionFunctions` is also lower case can you change it?

PR #1862, opened 2019-04-22
Add BigQuery Storage tap

### LINKED ISSUE #1858: Implement BigQuery Storage Tap
These are not implemented right now. We probably need to use the native BQ client to fetch raw elements as Avro and apply conversion to either `TableRow` for the untyped API or `T` for the typed one.

regadas:
Fixes #1858

### DIFF
--- a/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
+++ b/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
@@ -206,6 +206,37 @@ class StorageIT extends FlatSpec with Matchers {
     PAssert.that(p).containsInAnyOrder(expected)
     sc.close()
   }
+
+  import com.spotify.scio.io.Taps
+  it should "read from tap" in {
+    import com.spotify.scio.bigquery.BigQueryTaps._
+    import org.apache.beam.sdk.io.gcp.{bigquery => beam}
+    import com.google.cloud.bigquery.storage.v1beta1.ReadOptions.TableReadOptions
+
+    val tableRef = beam.BigQueryHelpers.parseTableSpec("data-integration-test:storage.required")
+    val futureTap = Taps().bigQueryStorage(tableRef, TableReadOptions.newBuilder().build())
+
+    import scala.concurrent.Await
+    import scala.concurrent.duration.Duration
+    val res = Await.result(futureTap, Duration.Inf).value.toList
+
+    res should not be (empty)
+  }
+
+  it should "read from typed tap" in {
+    import com.spotify.scio.bigquery.BigQueryTaps._
+    import org.apache.beam.sdk.io.gcp.{bigquery => beam}
+    import com.google.cloud.bigquery.storage.v1beta1.ReadOptions.TableReadOptions
+
+    val tableRef = beam.BigQueryHelpers.parseTableSpec("data-integration-test:storage.required")
+    val futureTap = Taps().bigQueryStorage(tableRef, TableReadOptions.newBuilder().build())
+
+    import scala.concurrent.Await
+    import scala.concurrent.duration.Duration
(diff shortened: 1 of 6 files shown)


scio-bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryAvroUtilsWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+
+package org.apache.beam.sdk.io.gcp.bigquery;
+
+import com.google.api.services.bigquery.model.TableRow;
+import com.google.api.services.bigquery.model.TableSchema;
+import org.apache.avro.generic.GenericRecord;
+
+public final class BigQueryAvroUtilsWrapper {
```
regadas: @nevillelyh I'm unsure about this ... I don't think we should really reimplement this conversion.

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
```diff
@@ -218,18 +245,24 @@ object StorageIT {
   @BigQueryType.fromStorage("data-integration-test:storage.repeated")
   class Repeated
 
-  @BigQueryType.fromStorage("data-integration-test:storage.nested",
-                            selectedFields = List("required", "optional.int"))
+  @BigQueryType.fromStorage(
+    "data-integration-test:storage.nested",
+    selectedFields = List("required", "optional.int")
+  )
```
jto: Still have indentation issues  ?

jto:
@regadas can you resolve the conflicts ?

scio-bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryAvroUtilsWrapper.java
nevillelyh: I'm fine with monkey patching it this way. At least when they change API it breaks our code, instead of us sticking with a potentially outdated implementation.

scio-bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryServicesWrapper.java
```diff
@@ -48,30 +48,21 @@ public void createTable(Table table) throws IOException, InterruptedException {
     bqServices.getDatasetService(bqOptions).createTable(table);
   }
 
-  public boolean isTableEmpty(TableReference tableReference)
-      throws IOException, InterruptedException {
+  public boolean isTableEmpty(TableReference tableReference) throws IOException, InterruptedException {
```
nevillelyh: Are changes in this file format only?

scio-bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryServicesWrapper.java
```diff
@@ -48,30 +48,21 @@ public void createTable(Table table) throws IOException, InterruptedException {
...
-          InsertRetryPolicy.alwaysRetry(),
-          new ArrayList<>(),
-          ErrorContainer.TABLE_ROW_ERROR_CONTAINER,
-          false,
-          false);
+    return bqServices.getDatasetService(bqOptions).insertAll(ref, rows, null, InsertRetryPolicy.alwaysRetry(),
+        new ArrayList<>(), ErrorContainer.TABLE_ROW_ERROR_CONTAINER, false, false);
+  }
+
+  public Table getTable(TableReference tableReference, List<String> selectedFields)
```
regadas: @nevillelyh this method was added also. the file format was a mistake.

PR #1867, opened 2019-04-24
Expose BigQuery Failed inserts

regadas:


anish749:
This looks great :) Thanks for the prompt fix 👍

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -63,7 +66,13 @@ final class DynamicBigQueryOps[T](private val self: SCollection[T]) extends AnyV
       if (writeDisposition != null) {
         transform = transform.withWriteDisposition(writeDisposition)
       }
-      self.applyInternal(transform)
+      val wr = self.applyInternal(transform)
+      if (self.internal.isBounded() == IsBounded.UNBOUNDED) {
+        insertErrorTransform(self.context.wrap(wr.getFailedInsertsWithErr()))
```
anish749: I haven't tested this but from what I read, I think we have to call [`withExtendedErrorInfo`](https://beam.apache.org/releases/javadoc/2.11.0/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.Write.html#withExtendedErrorInfo--) first to make BQ write extended info and then we can access [`getFailedInsertsWithErr`](https://beam.apache.org/releases/javadoc/2.11.0/org/apache/beam/sdk/io/gcp/bigquery/WriteResult.html#getFailedInsertsWithErr--)

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -171,7 +172,7 @@ final case class BigQueryTable(tableSpec: String) extends BigQueryIO[TableRow] {
     Reads.bqReadTable(sc)(beam.BigQueryIO.readTableRows(), table)
 
   override def write(data: SCollection[TableRow], params: WriteP): Tap[TableRow] = {
-    var transform = beam.BigQueryIO.writeTableRows().to(table)
+    var transform = beam.BigQueryIO.writeTableRows().to(table).withExtendedErrorInfo()
```
anish749:
```suggestion
    var transform = beam.BigQueryIO.writeTableRows().to(table)
```

I think `transform.withExtendedErrorInfo()` should be only set when `params.extendedErrorInfo` is `Enabled`?

And it is there in line 195.

regadas: Ahh good catch! that wasn't supposed to be there 😄

jto:
@regadas can you resolve the conflicts ?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -57,13 +80,20 @@ final class DynamicBigQueryOps[T](private val self: SCollection[T]) extends AnyV
         .write()
         .to(destinations)
         .withFormatFunction(Functions.serializableFn(formatFn))
+        .withExtendedErrorInfo()
```
byungk: If we set `withExtendedErrorInfo()` here, does it matter which `ExtendedErrorInfo` is actually passed in for `extendedErrorInfo` param?

Wouldn't it defeat the purpose of subsequent pattern match on `extendedErrorInfo: ExtendedErrorInfo`? _(Line 92-95)_:
```
transform = extendedErrorInfo match {
  case Disabled => transform
  case Enabled  => transform.withExtendedErrorInfo()
}
```

Was this an artifact of incorrect merge conflict resolution?

regadas: Hi @byungk Nice catch!  not sure what really happen but probably was what you mentioned.

PR #1873, opened 2019-04-26
Initialize counters from ScioContext 

anish749:
Add a function to initialise multiple counters from ScioContext.

We mostly have this code copied in multiple internal repos and thought of moving it upstream.


scio-test/src/test/scala/com/spotify/scio/ScioContextTest.scala
```diff
@@ -175,4 +175,42 @@ class ScioContextTest extends PipelineSpec {
     the[IllegalArgumentException] thrownBy { ScioContext.apply(invalidOpts) } should have message
       s"blockFor param foo cannot be cast to type scala.concurrent.duration.Duration"
   }
+
```
anish749: are these test cases needed? I saw that the part was already covered in codecov, but couldn't find where the test case was !!

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -757,13 +757,19 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
...
+   * name.
+   * */
   def initCounter(namespace: String, name: String): Counter = {
     val counter = ScioMetrics.counter(namespace, name)
     _counters.append(counter)
     counter
   }
 
+  /** Initialize a given [[org.apache.beam.sdk.metrics.Counter Counter]] metric. */
+  def initCounter(counter: Counter*): Unit =
```
regadas: @anish749 I think i would do this

```suggestion
  def initCounter(counter: Counter*): List[Counter] =
```

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -757,13 +757,19 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
...
   }
 
-  /** Initialize a new [[org.apache.beam.sdk.metrics.Counter Counter]] metric. */
+  /**
+   * Initialize a new [[org.apache.beam.sdk.metrics.Counter Counter]] metric from namespace and
+   * name.
+   * */
   def initCounter(namespace: String, name: String): Counter = {
     val counter = ScioMetrics.counter(namespace, name)
     _counters.append(counter)
```
regadas: you can replace this with the new call

scio-test/src/test/scala/com/spotify/scio/ScioContextTest.scala
regadas: I don't think we need them. If you replace the internal call with the new one.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
anish749: Yes, thats a good idea. I'll do that

regadas:
Thx @anish749!

PR #1876, opened 2019-04-26
Update scalafmt

regadas:
So apparently we were using a very old version of scalafmt. Newer version of scalafmt has new config defaults I kinda like them.

Tried to replicate the previous config to avoid all these changes but was a bit unsuccessful 😞 I think too much has changed ...


project/plugins.sbt
```diff
@@ -12,7 +12,7 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.3")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.6")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.3.0")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
-addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.0.0")
```
regadas: this is the new maintained plugin

PR #1879, opened 2019-04-29
scio-tensorflow cleanup

andrewsmartin:
This PR:

1. removes an un-needed coder instance for `FeatureType`
2. removes some un-needed test code.


andrewsmartin:
@regadas yeah my thoughts exactly, I don't see a need for Schema to go through ser/de. There's a single instance per dataset and lookup via `DistCache` should be sufficient.

regadas:
@andrewsmartin hm, yeah I guess we can always add it if the use case pops up

PR #1886, opened 2019-05-01
Refactor bigquery schema/table cache

regadas:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/Cache.scala
```diff
@@ -27,51 +27,91 @@ import com.spotify.scio.bigquery.BigQueryUtil
 import org.apache.beam.sdk.io.gcp.{bigquery => bq}
 
 import scala.util.Try
+import org.apache.avro.Schema
 
 private[client] object Cache {
 
+  sealed trait Show[T] {
```
nevillelyh: LOL there isn't a cats version?

regadas: There is 😄 at some point we need to really start using it.

PR #1890, opened 2019-05-02
Compile scala snippets in doc using mdoc

jto:
~Small caveat: annotation macro do not expand properly so they're not compiled :/
see: https://github.com/scalameta/mdoc/issues/164~


build.sbt
```diff
@@ -862,14 +860,26 @@ lazy val siteSettings = Def.settings(
...
   addMappingsToSiteDir(mappings in (ScalaUnidoc, packageDoc), siteSubdirName in ScalaUnidoc),
   gitRemoteRepo := "git@github.com:spotify/scio.git",
   mappings in makeSite ++= Seq(
     file("scio-examples/target/site/index.html") -> "examples/index.html"
   ) ++ SoccoIndex.mappings,
+  // pre-compile md using mdoc
+  mdocIn := baseDirectory.value / "src" / "paradox",
+  mdocExtraArguments ++= Seq(
+    "--no-link-hygiene",
+    "--scalac-options -Ydelambdafy:inline"
```
olafurpg: What happens if this line here is removed?

olafurpg:
🎉

PR #1895, opened 2019-05-03
Fix schema inference for scalar

### LINKED ISSUE #1894: Sql interpolator throws NPE
```scala
    """
    def coll: SCollection[Int] = sc.parallelize((1 to 10).toList)
    val r: SCollection[Int] = tsql"SELECT `value` FROM $coll".as[Int]
    """ should compile
```

regadas:
Fixes #1894

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
@@ -157,17 +157,14 @@ object Queries {
    * If it fails, a error message is returned in a [[Left]].
    */
   def typecheck[A: Schema, B: Schema](q: Query[A, B]): Either[String, Query[A, B]] = {
-    val schema: BSchema = SchemaMaterializer.fieldType(Schema[A]).getRowSchema
-    // TODO: add null check on schema. If A is a scalar (Int, String,...) schema will be null
-    val expectedSchema: BSchema =
-      Schema[B] match {
-        case s: Record[B] =>
-          SchemaMaterializer.fieldType(s).getRowSchema
-        case _ =>
-          SchemaMaterializer.fieldType(Schema[ScalarWrapper[B]]).getRowSchema
-      }
+    def beamSchema[T](implicit schema: Schema[T]): BSchema = schema match {
+      case s @ (_: Record[T] | _: RawRecord[T]) =>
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -245,8 +242,8 @@ object Queries {
 
   private[this] def printInferred(inferredSchemas: List[(String, BSchema)]): String =
     inferredSchemas
-      .map {
-        case (name, schema) =>
+      .collect {
+        case (name: String, schema: BSchema) =>
```
jto: Why the type matching ? The type are concrete. `inferredSchemas: List[(String, BSchema)]`

regadas: yeah feels meh, but `BSchema` is a java type it can be null and without type it will match. 

```scala
scala> List[java.lang.Integer](null).collect { case x => x }
res3: List[Integer] = List(null)

scala> List[java.lang.Integer](null).collect { case x: java.lang.Integer => x }
res4: List[Integer] = List()
```

¯\_(ツ)_/¯

jto: I would rather null check and tell in the message that the schema was not indfred properly than just swallow it.

regadas: we can do that!

scio-core/src/main/scala/com/spotify/scio/sql/Sql.scala
```diff
@@ -157,17 +157,14 @@ object Queries {
...
-    val schema: BSchema = SchemaMaterializer.fieldType(Schema[A]).getRowSchema
-    // TODO: add null check on schema. If A is a scalar (Int, String,...) schema will be null
-    val expectedSchema: BSchema =
-      Schema[B] match {
-        case s: Record[B] =>
-          SchemaMaterializer.fieldType(s).getRowSchema
-        case _ =>
-          SchemaMaterializer.fieldType(Schema[ScalarWrapper[B]]).getRowSchema
-      }
+    def beamSchema[T](implicit schema: Schema[T]): BSchema = schema match {
```
regadas: I think we should apply the same logic in `typecheck` for `Query2`

PR #1896, opened 2019-05-03
Add beam sql interpolation example

regadas:
Depends on #1895 #1893 #1917


scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
```diff
@@ -139,27 +139,28 @@ object To {
...
+    val h = new { val ctx: c.type = c } with SchemaMacroHelpers
     import c.universe._
 
     val tpeI = weakTypeOf[I]
     val tpeO = weakTypeOf[O]
 
-    val sInTree = c.untypecheck(iSchema.tree.duplicate)
-    val sOutTree = c.untypecheck(oSchema.tree.duplicate)
+    val sOut = c.eval(c.Expr[Schema[O]](c.untypecheck(oSchema.tree.duplicate)))
+    val sIn = c.eval(c.Expr[Schema[I]](c.untypecheck(iSchema.tree.duplicate)))
```
jto: Does this really work ? I was doing the 2 evals in one pass because otherwise scalac would randomly crash.

scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
```diff
@@ -139,27 +139,28 @@ object To {
 }
 
 object ToMacro {
-  import scala.reflect.macros.blackbox
+  import scala.reflect.macros._
   def safeImpl[I: c.WeakTypeTag, O: c.WeakTypeTag](
-    c: blackbox.Context
+    c: whitebox.Context
```
jto: Why does this need to be whitebox ?

regadas: ah should be reverted now, I was just playing around with some stuff.

scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
regadas: hm interesting, i haven't seen any inconsistency between compilations with this.

PR #1899, opened 2019-05-03
Use canonical name to allow inner object jobs

regadas:


regadas:
This is not actually a proper fix

PR #1900, opened 2019-05-06
Remote sys.exit from code path which is used in tests

sisidra:
`sys.exit` makes user code fail with unpredictable and uninterpretable results.
This PR replaces it with exceptions.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -198,6 +198,7 @@ object ContextAndArgs {
           // scalastyle:off regex
           Console.out.println(usageOrHelp)
           // scalastyle:on regex
+          UsageOrHelpException.attachUncaughtExceptionHandler()
           throw new UsageOrHelpException()
         case Success(Right((_opts, _args))) =>
           (new ScioContext(_opts, Nil), _args.asInstanceOf[T])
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -191,16 +192,15 @@ object ContextAndArgs {
...
         case Failure(exception) =>
-          Console.err.println(exception.getMessage)
-          sys.exit(1)
+          throw exception
         case Success(Left(usageOrHelp)) =>
+          // scalastyle:off regex
           Console.out.println(usageOrHelp)
-          sys.exit(0)
+          // scalastyle:on regex
+          throw new UsageOrHelpException()
```
sisidra: We could attach `UncaughtExceptionHandler` to current thread here to skip printing `UsageOrHelpException` line from the output.
Do we want to?

sisidra:
@regadas that's the point. The parse exceptions (or help cases) are matched and scio calls `sys.exit` - that leads to jmv to exit. Nobody (read: testing frameworks) do not like that. Your output is cut short (no further tests are executed) and/or fails with exception "jvm exitted". That's not what you expect for a test on our invalid input.

regadas:
@sisidra I understand, i see it makes sense during tests to be able to control bad argument parsing. I mean even for testing the arg parsing it self makes sense.

I'm just not really onboard with `exception` control flow in general. I would want another direction for this and push those effects to user land  ... I understand this might be one of the backward compatible ways to do it.

sisidra:
I see this case as application error, which for most users should be handled as `fail as fast as possible with as much as possible information, so user can fix the cause`. I don't see exception here as a controlling the flow. I see this as non-recoverable invalid application input case.

jto:
@regadas what would be your suggestion to handle this ? An Exception makes sense to me here. It's not like there's something you can do to recover from invalid parameters.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -20,28 +20,34 @@
 package com.spotify.scio
 
 import java.beans.Introspector
-import java.io.{ByteArrayOutputStream, File, PrintStream}
+import java.io.ByteArrayOutputStream
```
regadas: @sisidra can you just undo these changes?

regadas:
@jto i had something in mind but it would require more changes. Not sure we should tackle it right now since we kinda plan to revisit the CLI in a near future. 

Let's go with this fix now!

sisidra:
Fixed. Thank you!  :)

PR #1901, opened 2019-05-06
Implement saveAsDynamicProtobufFile

xBATx:
I would find useful to have this function implemented in scio library. I didn't add it directly to the `DynamicIoSCollection` because `T` in this case needs to have super type `Message` so I created new type class `DynamicProtoSCollection` for this purpose. 

I will add tests if this solution is approved by @nevillelyh.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/AvroBytesUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/AvroBytesUtil.scala
@@ -43,7 +43,7 @@ private[scio] object AvroBytesUtil {
     s
   }
 
-  def encode[T](coder: BCoder[T], obj: T, schema: ASchema = schema): GenericRecord = {
+  def encode[T](coder: BCoder[T], obj: T): GenericRecord = {
     val bytes = CoderUtils.encodeToByteArray(coder, obj)
     val record = new GenericData.Record(schema)
     record.put("bytes", ByteBuffer.wrap(bytes))
--- a/scio-core/src/main/scala/com/spotify/scio/coders/AvroBytesUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/AvroBytesUtil.scala
@@ -43,7 +43,7 @@ private[scio] object AvroBytesUtil {
     s
   }
 
-  def encode[T](coder: BCoder[T], obj: T): GenericRecord = {
+  def encode[T](coder: BCoder[T], obj: T, schema: ASchema = schema): GenericRecord = {
     val bytes = CoderUtils.encodeToByteArray(coder, obj)
     val record = new GenericData.Record(schema)
     record.put("bytes", ByteBuffer.wrap(bytes))
--- a/scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
@@ -17,15 +17,17 @@
 
 package com.spotify.scio.io
 
+import com.google.protobuf.Message
+import com.spotify.scio.coders.AvroBytesUtil
 import com.spotify.scio.util.Functions
 import com.spotify.scio.values.SCollection
 import org.apache.avro.Schema
 import org.apache.avro.file.CodecFactory
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecordBase
-import org.apache.beam.sdk.coders.StringUtf8Coder
-import org.apache.beam.sdk.io.AvroIO.RecordFormatter
+import org.apache.beam.sdk.coders.{Coder, StringUtf8Coder}
 import org.apache.beam.sdk.io.{Compression, FileIO}
+import org.apache.beam.sdk.transforms.Contextful
 import org.apache.beam.sdk.{io => beam}
 
 import scala.collection.JavaConverters._
@@ -69,10 +71,7 @@ package object dynamic {
             beam.AvroIO.sink(cls)
           } else {
             beam.AvroIO
-              .sinkViaGenericRecords(schema, new RecordFormatter[GenericRecord] {
-                override def formatRecord(element: GenericRecord, schema: Schema): GenericRecord =
-                  element
-              })
+              .sinkViaGenericRecords(schema, (element: GenericRecord, _: Schema) => element)
               .asInstanceOf[beam.AvroIO.Sink[T]]
           }
         }.withCodec(codec)
@@ -120,6 +119,46 @@ package object dynamic {
       )
     }
 
+  }
+
+  implicit class DynamicProtoSCollection[T <: Message](private val self: SCollection[T])
+      extends AnyVal {
+
+    def saveAsDynamicProtobufFile(
+      path: String,
+      numShards: Int = 0,
+      suffix: String = ".protobuf.avro",
+      codec: CodecFactory = CodecFactory.deflateCodec(6),
+      metadata: Map[String, AnyRef] = Map.empty
+    )(destinationFn: T => String)(implicit ct: ClassTag[T], coder: Coder[T]): Future[Tap[T]] = {
+      if (self.context.isTest) {
+        throw new NotImplementedError(
+          "Protobuf file with dynamic destinations cannot be used in a test context"
+        )
+      } else {
+        val sink = beam.AvroIO
+          .sinkViaGenericRecords[GenericRecord](
+            AvroBytesUtil.schema,
+            (element: GenericRecord, _: Schema) => element
+          )
+          .withCodec(codec)
+          .withMetadata(com.google.common.collect.Maps.newHashMap(metadata.asJava))
+
+        val write = writeDynamic(path, numShards, suffix, destinationFn)
+          .via(
+            Contextful.fn(
+              Functions.serializableFn[T, GenericRecord](AvroBytesUtil.encode(coder, _))
+            ),
+            sink
+          )
+
+        self.applyInternal(write)
+      }
+
+      Future.failed(
+        new NotImplementedError("Protobuf file future with dynamic destinations not implemented")
+      )
+    }
   }
 
   private def writeDynamic[A](
--- a/scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
@@ -138,10 +138,10 @@ package object dynamic {
       suffix: String = ".protobuf",
       codec: CodecFactory = CodecFactory.deflateCodec(6),
       metadata: Map[String, AnyRef] = Map.empty
-    )(destinationFn: T => String)(implicit cd: ClassTag[T]): Future[Tap[T]] = {
+    )(destinationFn: T => String)(implicit cd: ClassTag[T], coder: ScioCoder[T]): Future[Tap[T]] = {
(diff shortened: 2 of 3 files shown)


xBATx:
Hi @regadas, I tried it this way as well, but I couldn't figure out the solution that would be suitable for me, the only one that I found out (not really powerful) is below:

```
  implicit class DynamicProtobufSCollection[T <: Message](private val self: SCollection[T])
      extends AnyVal {

    def saveAsDynamicProtobufFile(
      path: String,
      numShards: Int = 0,
      suffix: String = ".protobuf.avro",
      codec: CodecFactory = CodecFactory.deflateCodec(6),
      metadata: Map[String, AnyRef] = Map.empty
    )(destinationFn: T => String)(coder: Coder[T]): Future[Tap[GenericRecord]] =
      self
        .map(AvroBytesUtil.encode(coder, _))
        .saveAsDynamicAvroFile(path, numShards, AvroBytesUtil.schema, suffix, codec, metadata)(
          avro => destinationFn(AvroBytesUtil.decode(coder, avro))
        )
  }
```

The issue here is wrapping protobuf to avro and then back to compute the destination function. That's why I proceed with suggested solution which is little bit more uglier but I think more powerful from performance perspective.

I have also the other potential one to add some mapping function to the `saveAsDynamicAvroFile` but I didn't want to touch existing code - and after adding this parameter, function would have 9 parameters which exceeds allowed number of arguments (8).

Do you think that there is any other/better solution?

regadas:
@xBATx I was actually thinking smth along the lines (your second point)
```scala
    def saveAsDynamicAvroFile(
      path: String,
      numShards: Int = 0,
      schema: Schema = null,
      suffix: String = ".avro",
      codec: CodecFactory = CodecFactory.deflateCodec(6),
      metadata: Map[String, AnyRef] = Map.empty
    )(
      destinationFn: T => String,
      recordFormatter: (T, Schema) => GenericRecord = (t, _) => t.asInstanceOf[GenericRecord]
    )(
      implicit ct: ClassTag[T]
    ): Future[Tap[T]]
```

you can ignore the `scalastyle` warn with `// scalastyle:off parameter.number`

xBATx:
Ok, got it. I wrote simple test that should work but I have some issues with protobuf serialization. I would assume that the issue could be in my function here:

 ```
(t, _) => AvroBytesUtil.encode(CoderMaterializer.beam(self.context, coder), t)
```

Please check CI output - I would appreciate some tips that could help here - where am I wrong? Actually I don't have enough time to investigate this in detail, sorry for that 😞.

File: scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala [hunk 1]
```diff
@@ -122,6 +129,29 @@ package object dynamic {
...
+      self.saveAsDynamicAvroFile(
+        path,
+        numShards,
+        AvroBytesUtil.schema,
+        suffix = suffix,
+        codec,
+        metadata
+      )(
+        destinationFn,
+        (t, _) => AvroBytesUtil.encode(CoderMaterializer.beam(self.context, coder), t)
```
regadas: @xBATx you need to materialize the coder outside.

```scala
val c = CoderMaterializer.beam(self.context, coder)
```

scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
```diff
@@ -122,6 +129,29 @@ package object dynamic {
...
+      codec: CodecFactory = CodecFactory.deflateCodec(6),
+      metadata: Map[String, AnyRef] = Map.empty
+    )(destinationFn: T => String)(implicit cd: ClassTag[T], coder: ScioCoder[T]): Future[Tap[T]] =
+      self.saveAsDynamicAvroFile(
+        path,
+        numShards,
+        AvroBytesUtil.schema,
+        suffix = suffix,
+        codec,
+        metadata
```
regadas: here we also need to attach `Map("protobuf.generic.schema" -> schema)` to be consistent with `ProtobufIO`

xBATx:
@regadas Comments addressed, but still having issues with some protobuf serialization when executing created test. This is little bit strange stack trace that I'm getting, it doesn't look like it's related to scio specific protobuf and `GenericRecord` encoding in `AvroBytesUtil` (I spent some time playing with it) it's probably related to something inside `beam.sdk.io.FileIO` as you can see in logs below:

```
java.lang.IllegalArgumentException: unable to serialize org.apache.beam.sdk.io.FileIO$Write$ViaFileBasedSink@1d6ed082
[info]   at org.apache.beam.sdk.util.SerializableUtils.serializeToByteArray(SerializableUtils.java:55)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation.toProto(WriteFilesTranslation.java:110)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation.toProto(WriteFilesTranslation.java:101)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation.access$000(WriteFilesTranslation.java:58)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation$1.translateSink(WriteFilesTranslation.java:72)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation.payloadForWriteFilesLike(WriteFilesTranslation.java:321)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation.payloadForWriteFiles(WriteFilesTranslation.java:67)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.WriteFilesTranslation$WriteFilesTranslator.translate(WriteFilesTranslation.java:291)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.PTransformTranslation$KnownTransformPayloadTranslator.translate(PTransformTranslation.java:336)
[info]   at org.apache.beam.repackaged.beam_runners_direct_java.runners.core.construction.PTransformTranslation.toProto(PTransformTranslation.java:149)
[info]   ...
[info]   Cause: java.io.NotSerializableException: org.apache.avro.Schema$RecordSchema
[info]   at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
[info]   at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
[info]   at java.util.HashMap.internalWriteEntries(HashMap.java:1790)
[info]   at java.util.HashMap.writeObject(HashMap.java:1363)
[info]   at sun.reflect.GeneratedMethodAccessor659.invoke(Unknown Source)
[info]   at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[info]   at java.lang.reflect.Method.invoke(Method.java:498)
[info]   at java.io.ObjectStreamClass.invokeWriteObject(ObjectStreamClass.java:1140)
[info]   at java.io.ObjectOutputStream.writeSerialData(ObjectOutputStream.java:1496)
[info]   at java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1432)
[info]   ...
```

I've implemented just simple test that tries to save `SimplePB` as dynamic proto file and then read the output - I assume you saw it.  Do you have any ideas what could cause that? The only added test  is failing.

xBATx:
Ah, got it. It's related to schema serialization, as the `org.apache.avro.Schema` does not implement `Serializable`. The issue was resolved by passing schema explicitly to `AvroBytesUtil.encode`, would it be OK like this @regadas?

scio-core/src/main/scala/com/spotify/scio/coders/AvroBytesUtil.scala
```diff
@@ -43,7 +43,7 @@ private[scio] object AvroBytesUtil {
     s
   }
 
-  def encode[T](coder: BCoder[T], obj: T): GenericRecord = {
+  def encode[T](coder: BCoder[T], obj: T, schema: ASchema = schema): GenericRecord = {
```
regadas: @xBATx There's no need to pass the `schema` here and also we don't want to allow overriding.

scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
```diff
@@ -122,6 +129,34 @@ package object dynamic {
...
+        self.context,
+        coder
+      )
+      self.saveAsDynamicAvroFile(
+        path,
+        numShards,
+        AvroBytesUtil.schema,
+        suffix,
+        codec,
+        metadata + ("protobuf.generic.schema" -> AvroBytesUtil.schema.toString)
```
regadas: @xBATx you were get serde issues because avro `Schema` is not serializable. Doing a `toString` did the trick. However for consistency we need it in `json` format.

```scala
import me.lyh.protobuf.generic
val schemaJson = generic.Schema.of[Message](ct.asInstanceOf[ClassTag[Message]]).toJson
```

scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
```diff
@@ -122,6 +129,34 @@ package object dynamic {
...
+    def saveAsDynamicProtobufFile(
+      path: String,
+      numShards: Int = 0,
+      suffix: String = ".protobuf",
+      codec: CodecFactory = CodecFactory.deflateCodec(6),
+      metadata: Map[String, AnyRef] = Map.empty
+    )(destinationFn: T => String)(implicit cd: ClassTag[T], coder: ScioCoder[T]): Future[Tap[T]] = {
+      val elemCoder = CoderMaterializer.beam(
+        self.context,
+        coder
```
regadas: we know already that this is a proto so we don;t need to infer the coder. We can:

```scala
val protoCoder =
  Coder.protoMessageCoder[Message](ct.asInstanceOf[ClassTag[Message]]).asInstanceOf[Coder[T]]
val elemCoder = CoderMaterializer.beam(self.context, protoCoder)
```

scio-test/src/test/scala/com/spotify/scio/io/dynamic/DynamicFileTest.scala
```diff
@@ -98,7 +99,7 @@ class DynamicFileTest extends PipelineSpec {
   it should "support generic Avro files" in {
     val tmpDir = Files.createTempDirectory("dynamic-io-")
     val sc1 = ScioContext()
-    implicit val coder = Coder.avroGenericRecordCoder(schema)
+    implicit val coder: Coder[GenericRecord] = Coder.avroGenericRecordCoder(schema)
```
regadas: undo?

xBATx: I prefer explicit types on implicit values, but I will revert that 😺 .

scio-core/src/main/scala/com/spotify/scio/io/dynamic/package.scala
```diff
@@ -17,8 +17,11 @@
 
 package com.spotify.scio.io
 
+import com.google.protobuf.Message
+import com.spotify.scio.coders.{AvroBytesUtil, CoderMaterializer, Coder => ScioCoder}
```
regadas: 🎨 I would just keep the name `Coder`

regadas:
Thanks @xBATx

PR #1905, opened 2019-05-06
(fix #1891) Allow TestStream as JobTest input

### LINKED ISSUE #1891: Support Beam TestStream in JobTest
From this Stack Overflow question:
https://stackoverflow.com/questions/55763770/is-it-possible-to-control-input-processing-time-with-scio-jobtest

We can probably support this by allowing `TestStream` as input in `JobTest`.

ClaireMcGinty:
I wasn't able to overload the existing [`def input[T](io: ScioIO[T], value: Iterable[T]): Builder =...`](https://github.com/spotify/scio/blob/f4ff6be66116f0268797fe43107fe2f09a451154/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala#L105) method without breaking non-explicitly-typed IO construction (like `BigQueryIO("foo")` or `ReadIO("bar")`) so I created a new method `pInput` ... not super happy with the clarity of that method name so any suggestions appreciated!

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
@@ -23,6 +23,8 @@ import com.spotify.scio.ScioResult
 import com.spotify.scio.io.ScioIO
 import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.transforms.PTransform
+import org.apache.beam.sdk.values.{PBegin, PCollection, PInput}
 import org.apache.beam.sdk.{metrics => beam}
 
 import scala.reflect.ClassTag
@@ -73,7 +75,7 @@ object JobTest {
   private case class BuilderState(
     className: String,
     cmdlineArgs: Array[String] = Array(),
-    input: Map[String, Iterable[_]] = Map.empty,
+    input: Map[String, JobInputSource[_]] = Map.empty,
     output: Map[String, SCollection[_] => Unit] = Map.empty,
     distCaches: Map[DistCacheIO[_], _] = Map.empty,
     counters: Map[beam.Counter, Long => Unit] = Map.empty,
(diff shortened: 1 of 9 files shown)


scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +100,24 @@ object JobTest {
...
      */
-    def input[T](io: ScioIO[T], value: Iterable[T]): Builder = {
+    def input[T](io: ScioIO[T], value: Iterable[T]): Builder =
+      input(io, IterableInputSource(value))
+
+    /**
+     * Feed an input in the form of a PTransform[PBegin, PCollection[T] to the pipeline being
+     * tested. Note that PTransform inputs may not be supported for all TestIO[T] types.
+     */
+    def pInput[T](
```
nevillelyh: What's the reason you can't overload here?
What about `def input[T](io)(transform)`? Also if you make `transform` a `fn: TestStream => TestStream`, user code could look like this:
```scala
.input(TextIO[String]) { _.addElements("a b b c").... }
```

WDYT?

ClaireMcGinty: The reason is that the Scala compiler can't infer which method to use when the generic type is not known:

```
def input[T](io: ScioIO[T], value: Iterable[T]): Builder = ...
def input[T](io: ScioIO[T], transform: PTransform[_ >: PBegin <: PInput, PCollection[T]]): Builder = ...

final case class ReadIO[T](id: String) extends TestIO[T]

JobTest[ReadAllJob.type]
  .input(ReadIO("a"), Seq("a1", "a2")) // fails, scala cannot resolve overloaded method

JobTest[ReadAllJob.type]
  .input(ReadIO[String]("a"), Seq("a1", "a2")) // succeeds
```

I don't know of a way around this without changing the original signature of `def input`, or requiring users to be explicit about `[T]` every time they use `ScioIO`s :|

ClaireMcGinty: for the second point - you mean replacing the signature ` PTransform[PBegin, ...` with `TestStream` specifically--or adding a new method for `TestStream` type? IMO the signature `.input(TextIO[String]) { _.addElements("a b b c").... }` reads a little unclear, maybe `.inputStream(TextIO[String]) { _.addElements("a b b c").... }` ?

nevillelyh: :+1: on `.inputStream`. In that case I guess we can just make the 2nd arg `TestStream`, and it doesn't matter whether it's curried or not.

ClaireMcGinty: cool, i'll just make this specific to TestStream, then! I think currying could have been a little awkward since `.addElements(...)` returns a `TestStream.Builder`, which only gets converted to `TestStream` when `advanceWatermarkToInfinity()` is called on it.

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +99,21 @@ object JobTest {
     }
 
     /**
-     * Feed an input to the pipeline being tested. Note that `TestIO[T]` must match the one used
-     * inside the pipeline, e.g. `AvroIO[MyRecord]("in.avro")` with
-     * `sc.avroFile[MyRecord]("in.avro")`.
+     * Feed an input in the form of a raw Iterable[T] to the pipeline being tested. Note that
```
nevillelyh:
```suggestion
     * Feed an input in the form of a raw `Iterable[T]` to the pipeline being tested. Note that
```

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +99,21 @@ object JobTest {
...
+     * `TestIO[T]` must match the one used inside the pipeline, e.g. `AvroIO[MyRecord]("in.avro")`
+     * with `sc.avroFile[MyRecord]("in.avro")`.
      */
-    def input[T](io: ScioIO[T], value: Iterable[T]): Builder = {
+    def input[T](io: ScioIO[T], value: Iterable[T]): Builder =
+      input(io, IterableInputSource(value))
+
+    /**
+     * Feed an input in the form of a PTransform[PBegin, PCollection[T] to the pipeline being
```
nevillelyh:
```suggestion
     * Feed an input in the form of a `PTransform[PBegin, PCollection[T]]` to the pipeline being
```

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -98,11 +99,21 @@ object JobTest {
...
+     * tested. Note that PTransform inputs may not be supported for all TestIO[T] types.
```
nevillelyh:
```suggestion
     * tested. Note that `PTransform` inputs may not be supported for all `TestIO[T]` types.
```

Issue #1929, opened 2019-05-22
NullPointerException thrown when running private method "QueryOps.isInvalidQuery"


martellj:
Periodically when running a query without either the `#standardSql` or `#legacySql` comments in the query, we are receiving the following stack trace:

```
java.lang.NullPointerException
	at com.spotify.scio.bigquery.client.QueryOps.com$spotify$scio$bigquery$client$QueryOps$$isInvalidQuery(QueryOps.scala:245)
	at com.spotify.scio.bigquery.client.QueryOps.isLegacySql(QueryOps.scala:258)
	at com.spotify.scio.bigquery.client.QueryOps.run(QueryOps.scala:193)
	at com.dowjones.dna.dataflow.snapshot.Query$$anonfun$1.apply(Query.scala:89)
	at com.dowjones.dna.dataflow.snapshot.Query$$anonfun$1.apply(Query.scala:89)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:40)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:44)
	at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at akka.dispatch.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at akka.dispatch.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at akka.dispatch.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

One of the following getters is returning `null`:

`e.getDetails.getErrors.get(0).getReason`

This issue appears to be the result of intermittent failures from BigQuery. Looking at the documentation for `GoogleJsonError`.  It appears `getDetails`, `getErrors` and `getReason` can return `null`.

martellj:
PR: https://github.com/spotify/scio/pull/1930

nevillelyh:
Can you provide a repro-able query using a public dataset?

martellj:
@nevillelyh unfortunately I can't.  The issue only appears intermittently and the NPE is suppressing the underlying exception.

martellj:
[...]

PR #1930, opened 2019-05-22
Wrap unsafe access of nullable fields and methods in try

martellj:
The purpose of this PR is to address a NPE that can occur when the method `isInvalidQuery` is called in response to a failed dry run query after determining neither the standard or legacy Sql comments have been set in the query.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
@@ -253,7 +253,7 @@ private[client] final class QueryOps(client: Client, tableService: TableOps, job
   }
 
   private def isInvalidQuery(e: GoogleJsonResponseException): Boolean =
-    e.getDetails.getErrors.get(0).getReason == "invalidQuery"
+    Try(e.getDetails.getErrors.get(0).getReason).map(_ == "invalidQuery").getOrElse(false)
 
   private[scio] def isLegacySql(sqlQuery: String, flattenResults: Boolean = false): Boolean = {
     val dryRun =
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
@@ -254,8 +254,8 @@ private[client] final class QueryOps(client: Client, tableService: TableOps, job
 
   private def isInvalidQuery(e: GoogleJsonResponseException): Boolean =
     Option(e.getDetails)
-      .flatMap(details => Option(details.getErrors.asScala))
-      .flatMap(errors => errors.headOption)
+      .flatMap(details => Option(details.getErrors))
+      .flatMap(_.asScala.headOption)
       .exists(_.getReason == "invalidQuery")
 
   private[scio] def isLegacySql(sqlQuery: String, flattenResults: Boolean = false): Boolean = {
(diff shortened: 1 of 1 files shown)


martellj:
Fixes: https://github.com/spotify/scio/issues/1929

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
```diff
@@ -253,7 +253,7 @@ private[client] final class QueryOps(client: Client, tableService: TableOps, job
   }
 
   private def isInvalidQuery(e: GoogleJsonResponseException): Boolean =
-    e.getDetails.getErrors.get(0).getReason == "invalidQuery"
+    Try(e.getDetails.getErrors.get(0).getReason).map(_ == "invalidQuery").getOrElse(false)
```
jto: Can you do an explicit null check or wrap the call in an `Option` instead of `Try` ?
This code may swallow other exceptions.

martellj: `isInvalidQuery` is only called when a `Failure` condition has already occurred. If I call in an `Option` the underlying failure will still be suppressed by an NPE (unless I wrap every call in an `Option`).  I will suggest some alternatives even though I still think the solution I originally proposed is the best since what is most important is that the underlying failure is reported as opposed to the exact reason the Google client is returning null.

martellj:
```suggestion
    Option(e.getDetails)
      .map(_.getErrors)
      .filterNot(_.isEmpty)
      .map(_.get(0))
      .exists(_.getReason == "invalidQuery")
```

martellj: I actually changed my mind.  I like the approach of wrapping in an `Option` despite its verbosity.

regadas: with a small tweak to your suggestion we should be good.

```suggestion
    Option(e.getDetails)
      .flatMap(details => Option(details.getErrors))
      .flatMap(_.headOption)
      .exists(_.getReason == "invalidQuery")
```

martellj: Thanks @regadas. I'm assuming you intended `Option(details.getErrors.asScala)`?

regadas: ahh yes! however it needs to be in the `.flatMap(_.asScala.headOption)`

martellj: @regadas  I will change it to that, but may I ask why `.flatMap(_.asScala.headOption)` as opposed to `Option(details.getErrors.asScala)`.  It seems to me to be the same result but would interested in learning why the former is preferred.

regadas: yeah no worries! the first suggestion only works under `2.12. With `2.11` you would still get NPE if `getErrors` returned `null`.

martellj: That makes sense! Thanks.  The change has been pushed.

PR #1934, opened 2019-05-23
Scala AsyncLookupDoFn

andrisnoko:


scio-core/src/main/scala/com/spotify/scio/transforms/ScalaAsyncLookupDoFn.scala
```diff
@@ -0,0 +1,230 @@
...
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Await, Future, Promise}
+import scala.concurrent.duration.Duration
+import scala.util.{Failure, Success, Try}
+import scala.language.implicitConversions
+import scala.util.control.NonFatal
+
+object ScalaAsyncLookupDoFn {
```
nevillelyh: Abstract Java `AsyncDoFn` as a base class to avoid copy-paste?

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaAsyncLookupDoFn.scala
```diff
@@ -0,0 +1,230 @@
...
+  val Client = Maps.newConcurrentMap[UUID, Any]
+  val Caches = Maps.newConcurrentMap[UUID, Cache[_, _]]
+
+  implicit def toJavaFunction[A, B](f: Function1[A, B]): JFunction[A, B] =
```
nevillelyh: https://github.com/spotify/scio/pull/1824/files#discussion_r274093420

You said we want a Scala AsyncDoFn to avoid converting between Java and Scala futures, and yet we're doing here again?

Also avoid implicit conversion unless absolute necessary.

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaAsyncLookupDoFn.scala
```diff
@@ -0,0 +1,230 @@
...
+      outputFn.apply(r)
+      ResultCount += 1
+      r = Results.poll
+    }
+  }
+
+  case class Result(input: A, output: Try[B], timestamp: Instant, window: BoundedWindow)
+}
+
+abstract class GuavaCacheSupplier[A, B, K] extends Serializable {
```
nevillelyh: Reuse the Java version?

andrisnoko:
@nevillelyh  PTAL

build.sbt
```diff
@@ -717,11 +717,20 @@ lazy val scioTensorFlow: Project = Project(
...
-  .settings(protobufSettings)
   .settings(
+    PB.protocVersion := "-v371",
+    PB.targets in Compile := Seq(
+      PB.gens.java -> (sourceManaged in Compile).value,
+      scalapb.gen(javaConversions = true) -> (sourceManaged in Compile).value
+    ),
+    libraryDependencies ++= Seq(
+      "io.grpc" % "grpc-netty" % grpcVersion,
+      "com.thesamet.scalapb" %% "scalapb-runtime-grpc" % scalapb.compiler.Version.scalapbVersion
```
nevillelyh: Do we REALLY need ScalaPB here? Can't use the gPRC native java proto?

scio-core/src/main/java/com/spotify/scio/transforms/AsyncDoFn.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.google.common.cache.Cache;
+import com.google.common.collect.Maps;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Semaphore;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.values.KV;
+
+public abstract class AsyncDoFn<A, B, C, D, FutureT> extends DoFn<A, KV<A, D>> {
```
nevillelyh: There's already a bunch of `AsyncDoFn`s and this will just add confusion. It should probably be named `BaseAsyncDoFn` to be consistent with the existing ones.

scio-core/src/main/java/com/spotify/scio/transforms/AsyncLookupDoFn.java
```diff
@@ -78,25 +78,19 @@ public AsyncLookupDoFn(int maxPendingRequests) {
...
+    super(maxPendingRequests, cacheSupplier);
   }
 
-  protected abstract C newClient();
-
   @Setup
   public void setup() {
-    client.computeIfAbsent(instanceId, instanceId -> newClient());
-    cache.computeIfAbsent(instanceId, instanceId -> cacheSupplier.createCache());
+   super.setup();
```
nevillelyh: Probably don't need this if super already has a `@Setup` method.

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaAsyncLookupDoFn.scala
```diff
@@ -0,0 +1,159 @@
...
+import scala.concurrent.{Await, Future, Promise}
+import scala.concurrent.duration.Duration
+import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
+
+object ScalaAsyncLookupDoFn {
+
+  private val Logger = LoggerFactory.getLogger(this.getClass)
+
+  implicit private class FutureExtension[A](val future: Future[A]) {
```
nevillelyh: This is not resolved. Why are we converting between Scala and Java future?

scio-core/src/main/java/com/spotify/scio/transforms/AsyncLookupDoFn.java
```diff
@@ -30,29 +46,13 @@
...
-  // same Client and Cache.
-  private static final ConcurrentMap<UUID, Object> client = Maps.newConcurrentMap();
-  private static final ConcurrentMap<UUID, Cache> cache = Maps.newConcurrentMap();
-  private final UUID instanceId;
-
-  private final CacheSupplier<A, B, ?> cacheSupplier;
-
   // Data structures for handling async requests
-  private final Semaphore semaphore;
   private final ConcurrentMap<UUID, ListenableFuture<B>> futures = Maps.newConcurrentMap();
```
nevillelyh: You can move this to parent with `FutureT`.

scio-core/src/main/java/com/spotify/scio/transforms/AsyncLookupDoFn.java
```diff
@@ -30,29 +46,13 @@
...
   private final ConcurrentLinkedQueue<Result> results = Queues.newConcurrentLinkedQueue();
```
nevillelyh: Is `Result` the `case class Result` in Scala? In that case this Java class is hard to use in pure Java Beam. Also it could be abstracted away as a type param in parent, or just a pure Java class.

scio-core/src/main/java/com/spotify/scio/transforms/AsyncLookupDoFn.java
```diff
@@ -78,25 +78,19 @@ public AsyncLookupDoFn(int maxPendingRequests) {
...
-
   @Setup
   public void setup() {
-    client.computeIfAbsent(instanceId, instanceId -> newClient());
-    cache.computeIfAbsent(instanceId, instanceId -> cacheSupplier.createCache());
+   super.setup();
   }
 
   @StartBundle
   public void startBundle() {
```
nevillelyh: Once you abstract away both `futures` & `results` this method can be removed.

build.sbt
andrisnoko: Yes we can use the gPRC-java native plugin, the idea was to provide the Scala example without converting futures from Java to Scala.

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaAsyncLookupDoFn.scala
andrisnoko: This is not converting between Scala and Jaba future, it providing the transform functionality available in Scala 2.12 since this code will also be compiled for 2.11. This is needed so transform both success and failure to Success because we are call Await.result in finishBundle which assumes that a future is successful.

scio-core/src/main/java/com/spotify/scio/transforms/AsyncLookupDoFn.java
andrisnoko: Result is Java, private Result class in AsyncLookupDoFn

nevillelyh:
https://github.com/spotify/scio/blob/neville/future/scio-core/src/main/java/com/spotify/scio/transforms/AsyncLookupDoFn.java

This is how I envisioned it should work. Once you abstract away Guava/Java8/Scala future handling in `FutureHandlers` interface and subtypes with default methods, each specialized version is just a mixed in with different type params.

nevillelyh:
And for the gPRC case we're serving mostly backend Java developers, so they're better off using the plain Java Protobuf interface & Guava AsyncLookupDoFn.

nevillelyh:
I'm gonna clean up the AsyncDoFn stuff. Closing this for now. Revisit later?

PR #1941, opened 2019-05-28
Add docs for CoderGrammar

anish749:
I was trying to speed up my compilation and resorted to adding implicit definitions explicitly using `Coder.gen`, `Coder.kryo` and `Coder.xmap` which means I am somewhat relying on this api.

I am adding a few docs, so that someone reading our code can debug / understand what these functions are doing.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
@@ -20,7 +20,7 @@ package com.spotify.scio.coders
 private object Derived extends Serializable {
   import magnolia._
 
-  @inline private def catching[T](msg: => String, stack: Array[StackTraceElement])(v: => T): T =
+  @inline private def catching[T](msg: => String, stack: => Array[StackTraceElement])(v: => T): T =
     try {
       v
     } catch {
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
```diff
@@ -41,7 +41,7 @@ private object Derived extends Serializable {
       i = i + 1
     }
 
-    val stack = CoderException.prepareStackTrace
+    lazy val stack = CoderException.prepareStackTrace
```
jto: Why did you change that ?

anish749: I was trying to delay the stack trace preparation to see if the speed improves. its negligible, shall i remove this?

jto: Yes. The purpose is to pre-generate the stack trace to display the call site instead of beam's internal when an exception occurs. Maky it lazy is a bug.

anish749: I see now. Thanks for explaining.

anish749:
Thanks for the reviews..
I was wondering if the information in this PR is correct. I kind of guessed my way through so not very sure if everything is correct and if they advise best practices.

jto:
Everything is correct. Thanks for the PR @anish749 :)

PR #1945, opened 2019-06-03
Add WrappedArray coder

sisidra:


sisidra:
Nested data structures with some arrays in them are not comparable without wrapping in something else then plain array, ergo it makes testing much harder with a lot of boilerplate required...
Duplicating an effort of implementing my own wrapper that mimics WrappedArray seems highly redundant.
Using anything but WrappedArray (like Seq, Vector, List...) seems highly wasteful.

Given that `WrappedArray` will be deprecated indeed and `ÀrraySeq` seems not to be well suited for long lived object type, is user defined wrapper case class with overridden `equals` the golden recommendation?

regadas:
When working with arrays *normally* is good to define your own `equals`, because you can also take advantage of `java.util.Arrays.equals(a, b)` or just use the `sameElements`. I do see your point that when used in nested types things become a bit cumbersome.

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -29,6 +29,7 @@ import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 import scala.collection.{BitSet, SortedSet, TraversableOnce, mutable => m}
 import scala.collection.convert.Wrappers
+import scala.collection.mutable
```
regadas: this import is already defined

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -454,6 +455,11 @@ trait ScalaCoders {
   implicit def arrayByteCoder: Coder[Array[Byte]] =
     Coder.beam(ByteArrayCoder.of())
 
+  implicit def wrappedArrayCoder[T: Coder: ClassTag](
+    implicit wrap: Array[T] => mutable.WrappedArray[T]
```
regadas:
```suggestion
    implicit wrap: Array[T] => m.WrappedArray[T]
```

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -454,6 +455,11 @@ trait ScalaCoders {
...
+  ): Coder[mutable.WrappedArray[T]] =
```
regadas:
```suggestion
  ): Coder[m.WrappedArray[T]] =
```

sisidra:
@regadas imagine case class with 20 fields and one is `Array`. User would need to ether wrap it in duplicate implementation of something similar to `WrappedArray` or overwrite case class `equals` (and `hashCode` as it's a good pattern to overwrite those in pairs) and do the comparison for all 20 fields. :(

PR #1946, opened 2019-06-03
Update avro to 1.9.0

regadas:
This PR is just to raise awareness around some potential breaking changes.

### DIFF
--- a/scio-schemas/src/main/avro/schema.avsc
+++ b/scio-schemas/src/main/avro/schema.avsc
@@ -56,6 +56,6 @@
         {"name": "double_field", "type": ["null", "double"], "default": null},
         {"name": "boolean_field", "type": ["null", "boolean"], "default": null},
         {"name": "string_field", "type": ["null", "string"], "default": null},
-        {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": null}
+        {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": []}
     ]
 }]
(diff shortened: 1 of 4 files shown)


scio-schemas/src/main/avro/schema.avsc
```diff
@@ -56,6 +56,6 @@
         {"name": "double_field", "type": ["null", "double"], "default": null},
         {"name": "boolean_field", "type": ["null", "boolean"], "default": null},
         {"name": "string_field", "type": ["null", "string"], "default": null},
-        {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": null}
+        {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": []}
```
regadas: schemas became a bit more strict apparently

sisidra: IMO this was a runtime error in avro, when trying to init value with not matching default value. Good if it is now compile-time :)

syodage:
Avro 1.9 has been defaulted to use jsr310 implementation instead of Joda for time logical types, Joda is still an option but needs to pass/add configuration to set it active.

jto:
Should we close this PR ? It seems that we need to wait for Beam to upgrade and that upgrading to 1.9 introduces compatibility issues.

See https://github.com/apache/beam/pull/9779 for context

regadas:
Yeah we can close it ... this will take awhile!

PR #1947, opened 2019-06-03
Use composite transform for sparse transforms

anish749:
The sparse transforms are spreading out into multiple blocks which makes it hard to understand in the dataflow UI.

This pr aims to put things in blocks which can be expanded later.

Please view using Files Changed -> Settings -> Hide WhiteSpaceChanges

Eg output in Datafow UI with this change:
Screenshot 2019-06-03 at 17 04 18

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -313,14 +313,15 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
     implicit hash: Hash128[K],
     koder: Coder[K],
     voder: Coder[V]
-  ): SCollection[(K, (Option[V], Option[W]))] =
+  ): SCollection[(K, (Option[V], Option[W]))] = self.transform { me =>
     SCollection.unionAll(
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -344,12 +345,14 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
   )(implicit hash: Hash128[K], koder: Coder[K], voder: Coder[V]): SCollection[(K, (V, W))] =
-    SCollection.unionAll(
-      splitSelfUsing(that, thatNumKeys, fpProb).map {
-        case (_, lhsOverlap, rhs) =>
-          lhsOverlap.join(rhs)
-      }
-    )
+    self.transform { me =>
+      SCollection.unionAll(
+        splitThisUsingThat(me, that, thatNumKeys, fpProb).map {
```
regadas: I would probably just rename this to `split`

anish749: Thanks for the feedback. It makes sense, I was a bit confused myself with the earlier name.
I'll make the change and push a new commit by tomorrow.

PR #1966, opened 2019-06-10
Set `location` when fetching Job in `JobOps`

seanhagen:
For any job not in the locations "US" or "EU", the location is required. This happens when a table is not in one of the multi-regional dataset locations ( such as "northamerica-northeast1" ).

This information is contained in `jobReference`, so pass it along when fetching the job.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/JobOps.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/JobOps.scala
@@ -94,7 +94,11 @@ private[client] final class JobOps(client: Client) {
         case (bqJob, jobReference) =>
           val jobId = jobReference.getJobId
           try {
-            val poll = client.underlying.jobs().get(client.project, jobId).execute()
+            val poll = client.underlying
+              .jobs()
+              .get(client.project, jobId)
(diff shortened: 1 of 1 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/JobOps.scala
```diff
@@ -94,7 +94,11 @@ private[client] final class JobOps(client: Client) {
         case (bqJob, jobReference) =>
           val jobId = jobReference.getJobId
           try {
-            val poll = client.underlying.jobs().get(client.project, jobId).execute()
+            val poll = client.underlying
+              .jobs()
+              .get(client.project, jobId)
+              .set("location", jobReference.getLocation)
```
regadas: wouldn't this work as well?

```suggestion
              .setLocation(jobReference.getLocation)
```

nevillelyh: @seanhagen can you revise?

seanhagen: Sure thing! ( Sorry, busy day yesterday )

PR #1968, opened 2019-06-11
Add partitionToMap function

freyrsae:
It can be annoying to be forced to work with Integers while doing partitions. This is a helper method that takes care of that annoyance by allowing the user to give a set of possible outcomes and a function into that set instead.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -295,9 +295,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     val partitionKeysIndexed = partitionKeys.toIndexedSeq
 
     partitionKeysIndexed
-      .zip(
-        partition(partitionKeys.size, t => partitionKeysIndexed.indexOf(f(t)))
-      )
+      .zip(partition(partitionKeys.size, t => partitionKeysIndexed.indexOf(f(t))))
       .toMap
   }
 
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -282,6 +282,26 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   }
 
   /**
+   * Partition this SCollection into a map from possible key values to an SCollection of
+   * corresponding elements based on the provided function .
+   *
+   * @param partitionKeys The keys for the output partitions
+   * @param f function that assigns an output partition to each element, should be in the range
+   * of `partitionKeys`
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -281,6 +281,26 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   * Partition this SCollection into a map from possible key values to an SCollection of
+   * corresponding elements based on the provided function .
+   *
+   * @param partitionKeys The keys for the output partitions
+   * @param f function that assigns an output partition to each element, should be in the range
+   * of `partitionKeys`
+   * @return partitioned SCollections in a `Map`
+   * @group collection
+   */
+  def partitionToMap[U: Coder](partitionKeys: Set[U], f: T => U): Map[U, SCollection[T]] = {
```
regadas: rename to `partition`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -281,6 +281,26 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+    val partitionKeysIndexed = partitionKeys.toIndexedSeq
+
+    partitionKeysIndexed
```
regadas: nit 🎨  indentation looks a bit weird

```scala
    partitionKeysIndexed
      .zip(partition(partitionKeys.size, f.andThen(partitionKeysIndexed.indexOf)))
      .toMap
```

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
freyrsae: that is what I wanted to do originally but the compiler didn't like it.

Caused errors for example in the partition by predicate method it didn't know if to use my new one or the original method. Can that be avoided without breaking backwards compatibility?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
freyrsae: will fix

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: yeah, it will require type ascription, it's ok to add it in this version since already introduces some breaking changes. TBH `partition` should have been curried

regadas: on a second thought, it might be better to just use similar function signature to what you had.

```scala
def partitionByKey[U: Coder](partitionKeys: Set[U])(f: T => U): Map[U, SCollection[T]] 
```

PR #1971, opened 2019-06-12
refactor async DoFns

nevillelyh:


scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
```diff
@@ -64,9 +61,10 @@ public void finishBundle(FinishBundleContext c) {
         waitForFutures(futures.values());
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
+        LOG.error("Failed to process futures", e);
         throw new RuntimeException("Failed to process futures", e);
       } catch (ExecutionException e) {
-        throw new RuntimeException("Failed to process futures", e);
+        LOG.error("Failed to process futures", e);
```
ClaireMcGinty: Why take away the RuntimeException? Shouldn't the bundle get failed ?

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -250,14 +255,15 @@ public boolean isSuccess() {
   }
 
   /**
-   * Supplier for {@link AsyncLookupDoFn} with caching logic.
+   * Supplier for {@link BaseAsyncLookupDoFn} with caching logic.
```
ClaireMcGinty: Isn't this technically supplying a `Cache` not a `BaseAsyncLookupDoFn` ? 🤔 Also -- it seems a bit strange to me that this class is both a Supplier for new `Cache`s while also handling get/put logic for a cache of `Cache`s? It seems like `CacheManager` might be a more accurate name... it's not a big deal though.

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+
+/** Utility to abstract away Guava, Java 8 and Scala future handling. */
+public class FutureHandlers {
+
+  /**
+   * Base interface for future handling.
+   * @param <F> future type.
+   * @param <T> value type.
```
ClaireMcGinty: jw, why is this parameterized as `<T>` ? It seems like `<B>` would be more consistent with the rest of the package unless I'm misreading this!

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
nevillelyh: You're right.
I had to remove it for the `BaseAsyncLookupDoFn` case since we don't want to fail a bundle if there're lookup errors. In that case the callback appends into the error queue and we surface it to user as a failed `Try`.
In this case we should rethrow.

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
nevillelyh: Agree, was just being lazy :sob:
Although `B` is usually for cases like `<A, B, C>`. Call it `V` instead?

ClaireMcGinty: `V` sounds good! 👍

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
nevillelyh: Maybe rephrase it as `{@link Cache} supplier for {@link BaseAsyncLookupDoFn}`? Supplier is from the Java naming convention for `() => T`.

ClaireMcGinty: oh, yeah my thought here was more that the scope of this class seems to be greater than just being a `Supplier`! I think your rephrasing makes sense though.

PR #1973, opened 2019-06-13
Expose Beam's AvroIO.parseGenericRecords

anish749:
Beam has an Api to apply a parseFn on Generic Records before reading them to a type `T`. Currently Scio's `AvroIO` doesn't provide a way to access this API and also benefit from `JobTest`.

This PR adds access to `AvroIO.parseGenericRecords` as a Scio `AvroIO`

Motivation is [here](https://docs.google.com/document/d/1m5M0JgjrmRca8qWtzsJa5lJoCGBqRqRcv5h0JaE3NYE/edit#heading=h.q4tiuatwvivr).

TODO
 - Find out if this is the appropriate way to expose this API.
 - ~Add unit tests.~

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -21,7 +21,7 @@ import com.google.protobuf.Message
 import com.spotify.scio.avro.types.AvroType.HasAvroAnnotation
 import com.spotify.scio.coders.{AvroBytesUtil, Coder, CoderMaterializer}
 import com.spotify.scio.io._
-import com.spotify.scio.util.ScioUtil
+import com.spotify.scio.util.{Functions, ScioUtil}
 import com.spotify.scio.values._
 import com.spotify.scio.{avro, ScioContext}
 import org.apache.avro.Schema
@@ -218,6 +218,45 @@ final case class GenericRecordIO[T: ClassTag: Coder](path: String, schema: Schem
 
   override def tap(read: ReadP): Tap[T] =
     GenericRecordTap[T](ScioUtil.addPartSuffix(path), schema)
+}
+
+/**
+ * Given a parseFn, read [[org.apache.avro.generic.GenericRecord GenericRecord]]
+ * and apply a function mapping [[GenericRecord => T]] before producing output.
+ * This IO applies the function at the time to de-serialising Avro GenericRecords.
+ *
+ * This IO doesn't define write, and should not be used to write Avro GenericRecords.
+ */
+final case class GenericRecordParseIO[T](path: String, parseFn: GenericRecord => T)(
+  implicit coder: Coder[T]
+) extends AvroIO[T] {
+  override type ReadP = Unit
+  override type WriteP = Nothing // Output is not defined for Avro Generic Parse IO.
+
+  override def testId: String = s"AvroIO($path)"
+
+  /**
+   * Get an SCollection[T] by applying the [[parseFn]] on
+   * [[org.apache.avro.generic.GenericRecord GenericRecord]]
+   * from an Avro file.
+   */
+  override protected def read(sc: ScioContext, params: Unit): SCollection[T] = {
+    val t = beam.AvroIO
+      .parseGenericRecords(
+        Functions
+          .serializableFn(parseFn)
+      )
+      .from(path)
+      .withCoder(CoderMaterializer.beamWithDefault(coder))
+
+    sc.wrap(sc.applyInternal(t))
+  }
+
+  /**
+   * Writes are undefined for [[GenericRecordParseIO]] since it is used only for reading.
+   */
+  override protected def write(data: SCollection[T], params: Nothing): Tap[T] = ???
+  override def tap(read: Unit): Tap[T] = ???
 }
 
 object AvroIO {
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -242,9 +242,7 @@ final case class GenericRecordParseIO[T](path: String, parseFn: GenericRecord =>
    */
   override protected def read(sc: ScioContext, params: Unit): SCollection[T] = {
     val t = beam.AvroIO
-      .parseGenericRecords(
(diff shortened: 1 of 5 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -220,6 +220,45 @@ final case class GenericRecordIO[T: ClassTag: Coder](path: String, schema: Schem
...
+   * from an Avro file.
+   */
+  override protected def read(sc: ScioContext, params: Unit): SCollection[T] = {
+    val t = beam.AvroIO
+      .parseGenericRecords(
+        Functions
+          .serializableFn(parseFn)
+      )
+      .from(path)
+      .withCoder(CoderMaterializer.beamWithDefault(coder))
```
jto: Please pass the correct `ScioContext` (`sc`) here and not the default one.

regadas:
@anish749 any progress on this?

anish749:
I wasn't sure if it would be a good idea to have this. I can update this and add some tests to it in a few days / later this week.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
anish749: Thanks for the review. I've fixed this.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -21,12 +21,11 @@ import com.google.common.collect.ImmutableMap
 import com.google.datastore.v1.Entity
 import com.google.datastore.v1.client.DatastoreHelper.{makeKey, makeValue}
 import com.spotify.scio._
-import com.spotify.scio.coders.Coder
 import com.spotify.scio.avro.AvroUtils.{newGenericRecord, newSpecificRecord}
-import com.spotify.scio.avro._
+import com.spotify.scio.avro.{AvroUtils, _}
```
regadas: `AvroUtils.newGenericRecord` is alread in imported in L:24

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
```diff
@@ -49,6 +50,14 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
   def avroFile[T: ClassTag: Coder](path: String, schema: Schema): SCollection[T] =
     self.read(GenericRecordIO[T](path, schema))
 
+  /**
+   * Get an SCollection of type [[T]] for data stored in Avro format after applying
+   * parseFn to map a serialized [[org.apache.avro.generic.GenericRecord GenericRecord]]
+   * to type [[T]]
+   */
+  def avroFile[T: Coder](path: String, parseFn: GenericRecord => T): SCollection[T] =
```
regadas: 🎨 can we curry this?

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -220,6 +220,46 @@ final case class GenericRecordIO[T: ClassTag: Coder](path: String, schema: Schem
...
+  override def testId: String = s"AvroIO($path)"
+
+  /**
+   * Get an SCollection[T] by applying the [[parseFn]] on
+   * [[org.apache.avro.generic.GenericRecord GenericRecord]]
+   * from an Avro file.
+   */
+  override protected def read(sc: ScioContext, params: Unit): SCollection[T] = {
+    val t = beam.AvroIO
+      .parseGenericRecords(
```
regadas: nit 🎨 can this fit in one line?

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
anish749: I can try this.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
anish749: Will fix this.

anish749:
Hey, do you know if the test cases are complete? Should there be any extra test cases for the Taps? I am not familiar with this part.

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
anish749: I tried a few things, but I am unable to figure out the correct syntax.
 - `def avroFile[T: Coder](path: String)(parseFn: GenericRecord => T)`
This one is not working because reading `SpecificRecord` is by using `def avroFile[T: <: SpecificRecord: Coder](path: String)` and hence the first function fails because the compiler wont be able to resolve the overloaded method.
 - Shall we try and rename this api? something like `parsedAvroFile` ?
 - ~The other option is to add "not a sub type" implicit~
```
import shapeless.<:!<
def avroFile[T](path: String)(parseFn: GenericRecord => T)(
    implicit toder: Coder[T],
    notSpecificRecord: T <:!< SpecificRecord
  )
```
or
```
import shapeless._
def avroFile[T: |¬|[SpecificRecord]#λ: Coder](path: String)(parseFn: GenericRecord => T)
```
This actually brings in an extra requirement that `T` should not be a SpecificRecord, which can be a little too restricting, considering that the api doesn't need to impose that. But still this doesn't compile with currying.

regadas: Had a brief look and yes we do have ambiguous resolution here when we curry. I’m leaning towards renaming it to your suggestion or just `parseAvroFile`. I think this gives a better syntax when expressing the parsing function

nevillelyh:
LGTM but again not sure if we really want to expose this, since it can be done with a custom IO easily? It's a power user case and adding it might just lead to confusion?

regadas:
indeed it's a power user case, but with this I guess we kinda make it a little bit more well known? Also we get a `Tap` for it as well.

i don't think it can hurt us in the future ... but i do see a concern on wether we expose something or not

regadas:
talked with @nevillelyh hence the merge

PR #1978, opened 2019-06-18
PairSCollectionFunctions: recover from a Failure coder as well as from a bad coder

cchepelov:
While trying to implement `SCollection[(K,V)].scanLeft[U: Coder](init: U)(op: (U, V) => V)`, I needed to use `applyPerKeyDoFn`. While running that, I encountered a problem in PairSCollectionFunctions.scala#L107

The intent of the original code appeared to be, if there is a coder with the proper shape, keep using it, otherwise overwrite with a code made for KV[K,V]. This would have worked pre-BEAM 2.0 where .getCoder didn't throw an exception if the coder was a failure.

This patch fixes that, by swallowing exceptions related to .getCoder following the `coderOrFailure.failure` path and handling that as a mis-shaped coder.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -25,11 +25,9 @@ import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.spotify.scio.util._
 import com.spotify.scio.util.random.{BernoulliValueSampler, PoissonValueSampler}
 import com.twitter.algebird.{Aggregator, Hash128, Monoid, Semigroup}
-import org.apache.beam.sdk.coders.KvCoder
 import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.values.{KV, PCollection, PCollectionView}
 import org.slf4j.LoggerFactory
-import scala.util.Try
 
 private object PairSCollectionFunctions {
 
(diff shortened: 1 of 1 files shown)


cchepelov:
(for reference, attempting to achieve collection.scanLeft as we find it in https://github.com/twitter/scalding this way: https://gist.github.com/cchepelov/5c6ad17c95515d82c3ff1211c1568c37 — initially hoped to have that as part of SCollection[T] but reluctantly accepting it must logically be part of SCollection[(K, V)] )
/cc @jto @nevillelyh

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -104,7 +105,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
           .apply("TupleToKv", toKvTransform)
           .setCoder(CoderMaterializer.kvCoder[K, V](context))
           .apply(t)
-        if (!kv.getCoder.isInstanceOf[KvCoder[_, _]]) {
+        if (!(Try(kv.getCoder.isInstanceOf[KvCoder[_, _]]).getOrElse(false))) {
```
nevillelyh: Just curious, what exception are you trying to swallow?

cchepelov:
IllegalArgument or State, whatever happens when on calls  .getCoder on a 
PCollection and the coder is a failure.

nevillelyh:
Is it possible that coder inference failed because you didn't pass the sufficient `: Coder` context bound in your wrapper code? @jto WDYT?

cchepelov:
I think the coder instance is right there in the gist, but maybe I forgot 
something else?

jto:
It seems to me that the `if` should should be removed and `kv.setCoder(CoderMaterializer.kvCoder[K, UI](context))` should just always be called.

jto:
Thanks for the PR @cchepelov :)

cchepelov:
thanks for the fast feedback and turnaround!

PR #1993, opened 2019-06-21
Support passing Assertions directly to `JobTest` `output`

NJAldwin:
The instructions and examples for `JobTest` `output` assertions warn if
run with `-Ywarn-value-discard`, because the compiler sees the `Unit` as
being discarded.  Most of the tests I can see fix this by returning `()`
_after_ the assertion, but this shouldn't be necessary.  This commit
changes the `assertion` to be `SCollection[T] => _`, where `_` is
(obviously) discarded.  The `assertion` is still converted using
`.asInstanceOf[SCollection[_] => Unit]` (so there's no actual behavior
change), but now passing `Assertion`s will work without a hitch.

Also see: [relevant discussion in
Gitter](https://gitter.im/spotify/scio?at=5d0bf10e4b0b7b477b2c14f2)

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
@@ -127,7 +127,7 @@ object JobTest {
      * @param assertion assertion for output data. See [[SCollectionMatchers]] for available
      *                  matchers on an [[com.spotify.scio.values.SCollection SCollection]].
      */
-    def output[T](io: ScioIO[T])(assertion: SCollection[T] => _): Builder = {
+    def output[T](io: ScioIO[T])(assertion: SCollection[T] => Any): Builder = {
       require(!state.output.contains(io.toString), "Duplicate test output: " + io.toString)
       state = state.copy(
         output = state.output + (io.testId -> assertion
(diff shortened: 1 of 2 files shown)


NJAldwin:
Hm, for some reason the `SCollection[T] => _` change, which I'd hoped would prevent any sort of signature changes being needed, has caused some tests to fail.

Example:
```scala
    val builder = com.spotify.scio.testing
      .JobTest("null")
      .input(ioFn(in), xs)
      .output(TextIO("out")) { coll =>
        coll should containInAnyOrder(xs.map(_.toString))
        ()
      }
```
It's erroring on
```
[error] /home/circleci/scio/scio-test/src/main/scala/com/spotify/scio/testing/ScioIOSpec.scala:66:32: missing parameter type
[error]       .output(TextIO("out")) { coll =>
```

I'm not sure why it isn't able to figure out that `coll` is a `String` given the `TextIO[String]` passed in the first argument set.

nevillelyh:
What's the missing parameter type it's complaining about? I wonder if you can make `assersion` a `SCollection[T] => Any` since the return type is covariant?

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -127,7 +127,7 @@ object JobTest {
      * @param assertion assertion for output data. See [[SCollectionMatchers]] for available
      *                  matchers on an [[com.spotify.scio.values.SCollection SCollection]].
      */
-    def output[T](io: ScioIO[T])(assertion: SCollection[T] => Unit): Builder = {
+    def output[T](io: ScioIO[T])(assertion: SCollection[T] => _): Builder = {
```
nevillelyh:
```suggestion
    def output[T](io: ScioIO[T])(assertion: SCollection[T] => Any): Builder = {
```

NJAldwin: Ah, good idea.  Seems to have worked!

nevillelyh:
Awesome thanks!

PR #2001, opened 2019-06-25
more BigtableDoFn fixes

nevillelyh:


scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -17,10 +17,7 @@
 package com.spotify.scio.transforms;
 
 import com.google.common.collect.Iterables;
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.*;
```
ClaireMcGinty: is the wildcard import intended?

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -252,6 +253,24 @@ public boolean isSuccess() {
...
+    public boolean equals(Object obj) {
+      if (obj == this) {
+        return true;
+      }
+      if (!(obj instanceof Try)) {
+        return false;
+      }
+      Try<?> that = (Try<?>) obj;
+      return (this.isSuccess == that.isSuccess) && Objects.equals(this.get(), that.get()) &&
+          Objects.equals(this.getException(), that.getException());
```
ClaireMcGinty: do we need a special `equals()` override for this `Throwable` equality check, too? for example `Objects.equals(new RuntimeException("foo"), new RuntimeException("foo"))` returns false

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
nevillelyh: IntelliJ did it but we don't enforce Java style so I didn't bother to fix.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
nevillelyh: Same behavior with scala `Try` I guess that's acceptable?
This is mainly to avoid tripping mutation detector, but probably not 100% necessary.

PR #2003, opened 2019-06-26
Use BigtableOptions param for BigTable utils

kiarash-rezahanjani:
Currently, it is not possible to use local BigTable emulator for integration testing as the current set up ignores BigtableOptions passed as parameters from test to ChannelPoolCreator.createPool method. I changed the method signature rather than adding a method but so not backward comp for users using ChannelPoolCreator.java but I didn see any value in keeping the old method.
Please advice if adding rather changing is prefered?


nevillelyh:
This will be part of 0.8 and the breakage is small enough. I don't have a problem with that :)

kiarash-rezahanjani:
Awesome! do we have estimtaed time when we release 0.8?

PR #2011, opened 2019-07-01
Cleanup around applyPerKey

regadas:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -82,35 +82,26 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
 
   import TupleFunctions._
 
-  private val context: ScioContext = self.context
+  private[this] val context: ScioContext = self.context
 
-  private val toKvTransform =
-    ParDo.of(Functions.mapFn[(K, V), KV[K, V]](kv => KV.of(kv._1, kv._2)))
+  private[this] val toKvTransform =
```
nevillelyh: This is used in `toKV` only now right? Make it a local `val`?

regadas: It's still used in more places

Issue #1848, opened 2019-04-16
Question: Support for ScioContext.onClose()


anish749:
I saw that the Scio Context has a function called `onClose(() => ())` [here](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L498-L502) which allows someone to add callbacks to the pipeline before it is executed.

However this is private and is used by BQ reads internally in Scio to run queries before starting a pipeline.

I was wondering if it would be a good idea to make this API public and have broader use cases for it.

Example use cases:
I want to run some code before starting a pipeline, but after defining it, eg: Make sure replication in some database slave is complete, before querying via Scio Jdbc, and stuff like that. Pretty similar to how it is used to run a query for BQ.

Another example that I can think of would be to check if dependencies exists and only then start a dataflow job.
If it is not an expected thing, I am happy to have my common components in `com.spotify.scio` package and then have my pipelines call my common code. (I do expect to have some common code for checking database replication, and things like that as add-ons to a pipeline IO.)

regadas:
Hi @anish749 the **TL;DR** it that `onClose` is meant to be used internally, if you use it ... you shouldn't! 

From your examples those checks/resource managers could easily be done outside of `scio` and possibly provided as libs.

anish749:
It is definitely possible to do this outside of Scio, and we do this today as well. However it requires a lot of layers of boilerplate code.

_From a user code perspective:_
We can make decisions and checks about whether to submit a job to dataflow or not. A job submissions can have a lot of preconditions:
 - are the files present / complete?
 - does the service account / user have correct permissions? (we don't do this, and jobs fail late)
 - does the output bucket exist? (we don't do this, and jobs fail late, after processing n-1 stages)
 - is there enough external resources available (eg: writing to cloudsql from multiple independent jobs, we don't want to overload the database with too many jobs / connections, we might just wait for this to finish)

Of course all of the above can be done in luigi, but that needs an extra layer, and more things to keep in mind while maintaining code.

Here are some inspirations / attempts / thoughts on making the whole structure simple.
[RFC1](https://docs.google.com/document/d/1JGiurBmefY5yHPgOvj2TeEZD2FHcw6-bbG5PDPYeXcc/edit#), [RFC2](https://docs.google.com/document/d/1U99vQMsXL2gAM9IG-035PtGM1mH6sAPAXxKl4vMqLas/edit#)


_From an extra lib perspective_ (not scio-core):
[...]

PR #2021, opened 2019-07-03
Adding onJobStart hook to ScioContext

d4nilos:
We have a use case for something similar to https://github.com/spotify/scio/issues/1848 but our handler needs the ClosedScioContext. 

The reason is that we need to get the Dataflow Job ID at the start of a job, even if --blocking is used.

Thanks and have a nice day

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -519,6 +519,7 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
   private var _isClosed: Boolean = false
   private val _counters: MBuffer[Counter] = MBuffer.empty
   private var _onClose: Unit => Unit = identity
+  private var _onJobStart: ClosedScioContext => Unit = (_ => ())
 
   /** Wrap a [[org.apache.beam.sdk.values.PCollection PCollection]]. */
   def wrap[T](p: PCollection[T]): SCollection[T] =
@@ -529,6 +530,16 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
    */
   private[scio] def onClose(f: Unit => Unit): Unit =
     _onClose = _onClose compose f
+
+  /**
+   * Add callbacks calls when the ClosedScioContext is created.
+   * This happens when the job starts.
+   */
+  def onJobStart(f: ClosedScioContext => Unit): Unit =
+    _onJobStart = closedContext => {
+      _onJobStart(closedContext)
+      f(closedContext)
+    }
 
   // =======================================================================
   // States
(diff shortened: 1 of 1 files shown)


gizem969:
Hi @regadas ,

We are using the [bigtable autoscaler](https://github.com/spotify/bigtable-autoscaler) to manage our bigtable clusters. If we use the utility functions in `scio-bigtable` to manually scale up before the pipeline, it clashes with the autoscaler, so instead we have a separate system that monitors the pipelines running on our clusters and informs the autoscaler as described [here](https://github.com/spotify/bigtable-autoscaler#does-it-handle-sudden-load-spikes-for-instance-dataflow-jobs-readingwriting-batch-data).

What we want to do is, inform our system that a pipeline is about to start, sleep, inform the system that the pipeline has been submitted with `dataflow jobid`. Afterwards our system will monitor its status from the Dataflow API and make necessary calls to the autoscaler.
In essense, what we need is the ability to execute some piece of code as soon as we get the job id from dataflow. We don't want to wait for the pipeline to finish. Theoretically, this could be achievable right after calling `close()`, however some pipelines configure `isBlocking`( https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L574 ), so it doesn't work.

regadas:
Hi @gizem969 Thanks for the explanation! 

> Theoretically, this could be achievable right after calling close(), however some pipelines configure isBlocking

Yes, i think this is the main issue and it should be addressed! Given the new semantic brought by `ClosedScioContext` this option does not make that much sense and it should be removed. With its removal there shouldn't be a need for hooks.

gizem969:
Hi @regadas , yes if it can be ensured that `close()` is always non-blocking, then it'd work. One advantage of having hooks might be that for cases like these, where external scio plug-ins change the state of a resource or do anything else that needs to be reverted/updated depending on the state of the pipeline, they can simply add these actions as callbacks rather than relying on clients to explicitly adding those steps to their pipelines.

Bj0rnen:
@gizem969 I don't know the Scala syntax, but you can most likely write a higher order function that "wraps around" the pipeline code (sort of like a try-with-resources block in Java). So something like this:
```scala
val sc = ...
val closedContext = runJobWithExtraBigtableNodes(
    instanceId = "foo",
    clusterId = "c1",
    extraNodes = 25,
    waitTimeMinutes = 20,
    pipeline = () => {
  // 20 minutes will pass before this block of code runs,
  // so that Bigtable has time to scale up.

  // Do stuff with sc...
  sc.close()  // The ClosedScioContext is picked up by runJobWithExtraBigtableNodes,
              // which can find out the dataflow jobid.
})

// The ClosedScioContext is also passed through runJobWithExtraBigtableNodes,
// and can be used for further steps.
closedContext.waitUntilDone()
```
The type of `runJobWithExtraBigtableNodes` in this example is (sorry, Haskell syntax)
```haskell
String -> String -> Int -> Int -> (() -> ClosedScioContext) -> ClosedScioContext
```

regadas:
Hi @d4nilos @gizem969 so we are deprecating `.close()` and with it the `blocking` behavior in favor of:

 https://github.com/spotify/scio/blob/ca8d49c16adf30ce41e9d27e6729d4ee9422e360/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L637

This should effectively remove the need for these kinda of hooks.

Issue #1682, opened 2019-02-13
Executing code after closeAndWait


jwiklund:
It would be useful to be able to register to execute code after close (and wait), currently it is hard to write a utility method that would for example resize a bigtable cluster after a scio job has been completed and make it composable with and without metrics (eg you would need closeAndWaitAndResizeCluster as well as closeAndPublishMetricsAndResizeCluster) and so on.

regadas:
Hi @jwiklund! I think it will be better if you expand your comment with an example of how you are doing things right now. IMO composing actions after a `sc.close().waitUntilDone()` can be achieved with the help of other data types like `Try` or even `Future` depending on what you are trying to achieve. i.e:

```scala
Try(sc.close().waitUntilDone())
  .map { sr =>
    Metrics.publish(...) // some metrics lib
  }
  .map { sr =>
    BigtableUtil.updateNumberOfBigtableNodes(...) // provided by scio-bigtable
  }
```

guidj:
@regadas I think one (bad) pattern I've seen libraries using is to wrap the close function for themselves, thus creating the `closeAndX` and `closeAndZ` situation, which is ok if there is the only thing you want, but it's problematic when you do want to execute multiple steps before/after running the job, and you have different libraries wrapping `sc.close*` themselves

One way scio could aid is to provide a goodies class that one could use to wrap the context with pre-process and post-process ops. For example:

```
class ContextWrapper( preOps: Seq[ScioContext => Try[_], 
                                        postOps: Seq[(ScioContext, ScioResult) => Try[_]]) {

  def closeAndWait(): Try = {
     // gracefully chain pre ops checking failure after each one
     val scioResult = sc.close().waitUntilDone()
     // gracefully chain post ops checking failure after each one
  }
}
```

That way, libraries would provide a these functions for you to chain yourself.  What do you think?

Of course, one can always create such class for their own project.

anish749:
Related:
https://github.com/spotify/scio/issues/1848

mewppis:
Being tracked in #2030

PR #2022, opened 2019-07-04
Job finish hook

gizem969:
Hi,

We have a use case where we want to provide a library function that does some pre work before the pipeline is run and registers a callback to revert those actions when the job is complete.

This branch is based on https://github.com/spotify/scio/pull/2021
And it also addresses https://github.com/spotify/scio/issues/1682

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -519,6 +519,7 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
   private var _isClosed: Boolean = false
   private val _counters: MBuffer[Counter] = MBuffer.empty
   private var _onClose: Unit => Unit = identity
+  private var _onJobStart: ClosedScioContext => Unit = (_ => ())
 
   /** Wrap a [[org.apache.beam.sdk.values.PCollection PCollection]]. */
   def wrap[T](p: PCollection[T]): SCollection[T] =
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -565,6 +590,13 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
 
     val closedContext = ClosedScioContext(this.pipeline.run(), this)
 
+    _onJobStart(closedContext)
+
+    if (_hasJobFinishHook) {
+      val result = closedContext.waitUntilFinish(closedContext.getAwaitDuration)
```
Bj0rnen: But won't this now make `close()` turn into a blocking call simply by using `onJobFinish` once? I think we need to look deeper for a more appropriate place to trigger the `_onJobFinish` call. We might have to make a choice between:
1. Only trigger the `_onJobFinish` call if the user actually blocks on the job to finish, either by calling `close()` with `isBlocking == true` or explicitly calling `waitUntilDone` or `waitUntilFinish` on the ClosedScioContext. If they don't do anything like that, the event never triggers, even if they `sleep` for a year after calling (non-blocking) `close()`.
2. Use a background thread to poll job progress.

I'm don't know much about the code base, so I can't say what's the best way. There could be other ways to do it. But I think turining `close()` into a blocking call simply because there's an even handler registered will be surprising to users; it's too implicit, even if documented.

gizem969: true that, but isn't a registered callback never being called unless they explicitly wait also kind of surprising?

gizem969: then again, i think we technically don't really need this, it is possible to have a `.close().waitUntilDone().doWhatever()` in the pipeline. Closing this PR.

Issue #2024, opened 2019-07-04
Exception-wrapping in Coder breaks SynchronousChainedCombinedDriver under Flink


cchepelov:
While trying to do a fairly simple outer join job:
```scala
val left = readCSV[LeftRecord](leftInputPath)
val right = readCSV[RightRecord](leftInputPath)

val leftByKey = left.map(v => v.key -> v)
val rightByKey = right.flatMap(v => v.maybeKey.map(key => key -> v))

// key is of type "case class Isrc(value: String)" which has appropriate Kryo

val joined = leftByKey.fullOuterJoin(rightByKey)

joined
   .map { case (key, pair) => pair.toString }
   .saveAsTextFile("foo.txt")
sc.close()
```
on very small inputs (40-70K lines each, 0 to 1 lines per key, very rarely a couple more), the job crashes quickly under Flink (both local and remote)

The immediately apparent error is an `EOFException`, even though the fact is not immediately obvious upon inspection:

```
Caused by: org.apache.flink.runtime.client.JobExecutionException: Job execution failed.
	at org.apache.flink.runtime.jobmaster.JobResult.toJobExecutionResult(JobResult.java:146)
	at org.apache.flink.client.program.rest.RestClusterClient.submitJob(RestClusterClient.java:265)
	... 22 common frames omitted
Caused by: com.spotify.scio.coders.CoderException: null
	at org.apache.flink.runtime.io.disk.SimpleCollectingOutputView.nextSegment(SimpleCollectingOutputView.java:79)
```  

> (yes, the report says it's a `CoderException` yet the code at https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/SimpleCollectingOutputView.java#L79 clearly says `EOFException` )

```text
	at org.apache.flink.runtime.memory.AbstractPagedOutputView.advance(AbstractPagedOutputView.java:137)
	at org.apache.flink.runtime.memory.AbstractPagedOutputView.write(AbstractPagedOutputView.java:198)
	at org.apache.beam.runners.flink.translation.wrappers.DataOutputViewWrapper.write(DataOutputViewWrapper.java:48)
	at com.esotericsoftware.kryo.io.Output.flush(Output.java:163)
	at com.esotericsoftware.kryo.io.OutputChunked.flush(OutputChunked.java:45)
	at com.twitter.chill.TraversableSerializer.$anonfun$write$1(Traversable.scala:31)
[...]

PR #2025, opened 2019-07-05
flink: don't wrap EOFException in WrappedBCoder 

cchepelov:
This is a stab at addressing #2024 
In case an EOFException is caught by WrappedBCoder, it is not wrapped but rethrown, as an outer layer of Flink expects to catch it to complete some try-and-recover step.
Wrapping into `CoderException` prevents the outer layer of Flink from catching its signal from the innermost layer.

The base functionality of `CoderException` is preserved by appending the "materialization point" call stack to the actual exception call stack.

Note: I'm thinking about deprecating `CoderException` and applying this stack-decoration procedure to all exceptions, but I am not in a position to assess the potential consequences to existing clients of scio (there doesn't appear any place where CoderException is caught or matched against in the scio repository properly)

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -17,7 +17,7 @@
 
 package com.spotify.scio.coders
 
-import java.io.{EOFException, InputStream, OutputStream}
+import java.io.{InputStream, OutputStream}
 
 import com.spotify.scio.coders.instances.Implicits
 import org.apache.beam.sdk.coders.Coder.NonDeterministicException
@@ -151,29 +151,6 @@ private final case class DisjunctionCoder[T, Id](
     coders.values.forall(_.consistentWithEquals())
 }
 
-final case class CoderException private[coders] (
-  stacktrace: Array[StackTraceElement],
-  cause: Throwable,
-  message: String = ""
-) extends RuntimeException {
-  override def getMessage: String = cause.getMessage
-  override def getCause: Throwable = new RuntimeException {
-    override def getMessage: String = s"$message - Coder was materialized at"
-    override def getStackTrace: Array[StackTraceElement] = stacktrace
-    override def getCause: Throwable = cause.getCause
-  }
-  override def getStackTrace: Array[StackTraceElement] = cause.getStackTrace
-}
-
-private[coders] object CoderException {
-  def prepareStackTrace: Array[StackTraceElement] =
-    Thread
-      .currentThread()
-      .getStackTrace()
-      .dropWhile(!_.getClassName.contains(CoderMaterializer.getClass.getName))
-      .take(10)
-}
-
 // XXX: Workaround a NPE deep down the stack in Beam
 // info]   java.lang.NullPointerException: null value in entry: T=null
 private[scio] case class WrappedBCoder[T](u: BCoder[T]) extends BCoder[T] {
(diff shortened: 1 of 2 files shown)


cchepelov:
(test failure due to "# Blocked due to plan-no-credits-available Please purchase a new credit block then push a new commit or call the API to run a new build.")

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -224,12 +202,42 @@ private[scio] case class WrappedBCoder[T](u: BCoder[T]) extends BCoder[T] {
...
 
-private object WrappedBCoder {
+private[scio] object WrappedBCoder {
   def create[T](u: BCoder[T]): BCoder[T] =
     u match {
       case WrappedBCoder(_) => u
       case _                => new WrappedBCoder(u)
     }
+
+  def appendMaterializationStack[X <: Throwable](
```
regadas: I think I would keep this logic in a separate `object` ... I kinda like to keep the decoupling from `WrappedBCoder`

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -224,12 +202,42 @@ private[scio] case class WrappedBCoder[T](u: BCoder[T]) extends BCoder[T] {
...
+      additionalMessage.map(msg => new StackTraceElement("Due to " + msg, "", "", 0)).toArray
+
+    val adjustedStack = messageItem ++ existingStack ++ baseStack
+    cause.setStackTrace(adjustedStack)
+    cause
+  }
+
+  def prepareStackTrace: Array[StackTraceElement] =
+    Array(
+      new StackTraceElement(
```
nevillelyh: The parameters are class, method, file name and line number. This is a little hacky? How does it look in an actual stacktrace? I wonder if there're better ways to cascade these, like caused by or suppressed?

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
cchepelov: Good point, will pull that away in a StackTraceUtil of sorts

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
cchepelov: Yes, the call to StackTraceElement does look a bit ugly (hence the "dash-scissor-dash" thingy). Let's see if we can do better.

I don't have a screenshot at the moment, since the whole point was to let an EOFException unchanged (and it ends up being caught by the Flink runtime as expected), but the prototype looked clear enough ☺ I'll whip up something ASAP for demonstration.

The problem with the other options:

* to generate a a"Caused By:", I would need to wrap the exception into another exception… Which is the problem in the first place. There might have been a way to wrap into an exception of the same type using reflection, if we could rely on the knowledge that every exception type must have a ctor that takes a Throwable or a String and a Throwable, but there is no such rule in the Java spec. 

* I don't know that we can say "suppressed" here. In the specific case that led me here, the EOFException ends up being suppressed, indeed. Let's say (hypothetically) that the coder must throw an exception such as "NoReallyICannotSerializeThisException", that the Runtime is _not_ aware of. In that hypothetical case, we probably (i) don't want to suppress (current behaviour: `WrappedException(NoReallyICannotSerializeThisException(original stack), materialization stack)` surfaces into a Job Failed, with this PR: `NoReallyICannotSerializeThisException(original stack / "—✂—" / materialization stack)` surfaces into the JobFailed. And _maybe_ the client is the owner of the Coder[T] involved and knows what to do with it.

cchepelov: (at the very minimum, I'd name the parameters and put an explanatory comment)

Issue #2019, opened 2019-07-02
REPL ignores jobName DataFlow parameter


arturroo:
When I run REPL on DataFlow it sets jobName based on my username and not on jobName parameter for DataFlow. My Google IAM username consists '_' underscore sign and those signs are not allows in DataFlow job names, so REPL starts, but any action that needs DF jobs breaks with error: 
```
  at org.apache.beam.sdk.util.InstanceBuilder.buildFromMethod(InstanceBuilder.java:214)
  ... 62 more
Caused by: java.lang.IllegalArgumentException: JobName invalid; the name must consist of only the characters [-a-z0-9], starting with a letter and ending with a letter or number
  at org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
```
Workaroud is to set jobName manually when REPL is running:
```
scio> sc.setJobName("scio-repl-af")
scio> sc.options.getJobName
res2: String = scio-repl-af
scio> val mytable = sc.bigQueryTable("precog:streaming.mytab")
...
mytable: com.spotify.scio.values.SCollection[com.spotify.scio.bigquery.TableRow] = com.spotify.scio.values.SCollectionImpl@1263664f
scio>
```

Please allow overwriting Scio jobName with specified --jobName parameter to DataFlow.

arturroo:
I think this might be fixed by checking if jobName is set in options during Scio Context initialization here
https://github.com/spotify/scio/blob/master/scio-repl/src/main/scala/com/spotify/scio/repl/ReplScioContext.scala
```Scala
class ReplScioContext(options: PipelineOptions, artifacts: List[String])
    extends ScioContext(options, artifacts) {

  this.setAppName("sciorepl")
  this.setJobName(s"""sciorepl-${CoreSysProps.User.value}-${System.currentTimeMillis()}""")
```
So I will try to fix it.

regadas:
Thanks for reporting this! Can you normalize the user as well?

arturroo:
Hi, 

Yes, I would like to normalize the username at the beginning. This would remove need for workaround.
Next I would like to add additional parameter --jobNamePostfix.
So in https://github.com/spotify/scio/blob/master/scio-repl/src/main/scala/com/spotify/scio/repl/ReplScioContext.scala
I would change line 27
```
this.setJobName(s"""sciorepl-${CoreSysProps.User.value}-${System.currentTimeMillis()}""")
```
to
```

[...]

PR #2028, opened 2019-07-05
Normalizes username in jobName for Scio Repl session.

arturroo:
This is to remove need for workaround when one must always manually specify new JobName after Scio Repl is started by sc.setJobName("my-job-name").
This will keep stiff jobName as
"sciorepl-%s-%d".format(
    CoreSysProps.User.value.replaceAll("[^a-z0-9]", "0"),
    System.currentTimeMillis()
But users that's username consist of characters not allowed in jobName will be able to save time and not do the workaround.


arturroo:
Hi @regadas 

Done, lines removed I confirm that jobName parameter is interpreted again on my snapshot version, and I have describe it in the issue #2019.
Changes to the source pushed in second commit.

regadas:
Hi @arturroo Thanks for this! unfortunately this is a bit incomplete because we want to set the app name to `sciorepl` if it wasn't set and allow override. I took the liberty to extend this PR in this one #2046. Hope you don't mind!

PR #2032, opened 2019-07-09
Add support for recursive Coders

### LINKED ISSUE #1992: Coder derivation for nested case class
Here is my simplified case class, when I try to generate Coder for it, different behaviours are observed, can someone help explain them ? Thanks!
```scala

case class SampleField(name: String, fieldType: SampleFieldType)

sealed trait SampleFieldType
case object IntegerType extends SampleFieldType
case object StringType extends SampleFieldType
case class RecordType(fields: Vector[SampleField]) extends SampleFieldType
```
**Compiler error when trying to derive coders**

```scala
implicit val fieldTypeCoder = Coder[SampleFieldType]  
implicit val fieldCoder= Coder[SampleField]
// Cannot find an implicit Coder instance for type:
// [error] 
// [error]   >> pipeline.SampleField
//...
```
**Coders are null when type annotations are added**

```scala
implicit val fieldTypeCoder: Coder[SampleFieldType] = Coder[SampleFieldType]  // fieldTypeCoder: null
implicit val fieldCoder: Coder[SampleField] = Coder[SampleField] // fieldCoder: null
```

**Coders derived successfully when a random class added in between**
```scala
implicit val fieldTypeCoder = Coder[SampleFieldType]  // fieldTypeCoder: Disjunction(pipeline.SampleFieldType, Map(0 -> Record(pipeline.IntegerType, ), 1 -> Record(pipeline.RecordType, (fields, Fallback(scala.collection.immutable.Vector))), 2 -> Record(pipeline.StringType, )))

case class Foo()  
implicit val fooCoder = Coder[Foo]  // fooCoder: Record(pipeline.Pipeline.Foo, )
  
implicit val fieldCoder= Coder[SampleField] // fieldCoder: Record(pipeline.SampleField, (name, Beam(StringUtf8Coder)), (fieldType, Disjunction(pipeline.SampleFieldType, Map(0 -> Record(pipeline.IntegerType, ), 1 -> Record(pipeline.RecordType, (fields, Fallback(scala.collection.immutable.Vector))), 2 -> Record(pipeline.StringType, )))))
```

jto:
fixes #1992

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -60,6 +60,18 @@ Cannot find an implicit Coder instance for type:
 """
 )
 sealed trait Coder[T] extends Serializable
+
+private[scio] final class Ref[T](val typeName: String, c: => Coder[T]) extends Coder[T] {
+  def value: Coder[T] = c
+
+  override def toString(): String = s"""Ref($typeName)"""
+}
+
+private[scio] object Ref {
+  def apply[T](t: String, c: => Coder[T]): Ref[T] = new Ref[T](t, c)
+  def unapply[T](c: Ref[T]): Option[(String, Coder[T])] = Option((c.typeName, c.value))
+}
+
 // scalastyle:on line.size.limit
 final case class Beam[T] private (beam: BCoder[T]) extends Coder[T] {
   override def toString: String = s"Beam($beam)"
@@ -172,6 +184,20 @@ private[coders] object CoderException {
       .getStackTrace()
       .dropWhile(!_.getClassName.contains(CoderMaterializer.getClass.getName))
       .take(10)
+}
+
+private[scio] final class RefCoder[T](val typeName: String, c: => BCoder[T]) extends BCoder[T] {
+  def value: BCoder[T] = c
+
+  def decode(inStream: InputStream): T = c.decode(inStream)
+  def encode(value: T, outStream: OutputStream): Unit = c.encode(value, outStream)
+  def getCoderArguments(): java.util.List[_ <: BCoder[_]] = c.getCoderArguments()
+  def verifyDeterministic(): Unit = c.verifyDeterministic()
+}
+
+private[scio] object RefCoder {
+  def apply[T](t: String, c: => BCoder[T]): RefCoder[T] = new RefCoder[T](t, c)
+  def unapply[T](c: RefCoder[T]): Option[(String, BCoder[T])] = Option((c.typeName, c.value))
 }
 
 // XXX: Workaround a NPE deep down the stack in Beam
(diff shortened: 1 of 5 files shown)


scio-test/src/test/scala/com/spotify/scio/coders/coders.scala
```diff
@@ -455,4 +455,17 @@ class CodersTest extends FlatSpec with Matchers {
     val wrappedArray: mut.WrappedArray[String] = Array("foo", "bar")
     wrappedArray coderShould notFallback()
   }
+
+  it should "support derivation of recursive types" in {
+    case class SampleField(name: String, fieldType: SampleFieldType)
+    sealed trait SampleFieldType
+    case object IntegerType extends SampleFieldType
+    case object StringType extends SampleFieldType
+    case class RecordType(fields: Vector[SampleField]) extends SampleFieldType
```
regadas: in this test replacing `Vector` with `List` and getting as well a Coder[RecordType] should tigger the issue

jto: It does. I'm working on that example and trying to come up with a fix.

scio-macros/src/main/scala/com/spotify/scio/MagnoliaMacros.scala
```diff
@@ -34,13 +34,7 @@ private[scio] object MagnoliaMacros {
...
     }
 
-    val magTree = magnolia.Magnolia.gen[T](c)
-
-    def getLazyVal =
-      magTree match {
-        case q"val $name = $body; $rest" =>
-          body
-      }
+    val getLazyVal = magnolia.Magnolia.gen[T](c)
```
regadas: 🎨 should we rename this ... or even move it to `removeAnnotations.transform( ... )`?

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -60,6 +60,18 @@ Cannot find an implicit Coder instance for type:
 """
 )
 sealed trait Coder[T] extends Serializable
+
+private[scio] final class Ref[T](val typeName: String, c: => Coder[T]) extends Coder[T] {
```
regadas: 🎨 replace `c` with `coder`?

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -60,6 +60,18 @@ Cannot find an implicit Coder instance for type:
...
+  def value = c
+
+  override def toString(): String = s"""Ref($typeName)"""
+}
+
+private[scio] object Ref {
+  def apply[T](t: String, c: => Coder[T]) = new Ref[T](t, c)
```
regadas: 🎨 better arg names here as well?

PR #2036, opened 2019-07-09
Little rework of ClosedScioContext

### LINKED ISSUE #2030: Remove isBlocking option
`isBlocking` option under `ScioOptions` should be removed in favor of explicitly adding `waitUntilDone()`. This option also doesn't work well with the new semantic that `ClosedScioContext` introduces.

As deprecation policy, 2 options:
* the presence of this option in the args should throw an error before running the pipeline.
* rename `.close()` to `.run()`

leaning towards point 2

regadas:
Fixes #2030

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -261,6 +261,8 @@ object ScioRunner {
 
 trait ScioExecutionContext {
 
+  def pipelineResult: PipelineResult
+
   /** Whether the pipeline is completed. */
   def isCompleted: Boolean
 
@@ -289,10 +291,12 @@ trait ScioExecutionContext {
 }
 
 object ScioExecutionContext {
-  private[scio] def default(pipelineResult: PipelineResult, context: ScioContext) =
+  private[scio] def default(pr: PipelineResult, sc: ScioContext) =
     new ScioExecutionContext {
-      private[this] val DefaultAwaitDuration = context.awaitDuration
+      private[this] val DefaultAwaitDuration = sc.awaitDuration
       private[this] val DefaultCancelJob = true
+
+      override def pipelineResult: PipelineResult = pr
 
       override def isCompleted: Boolean = state.isTerminal()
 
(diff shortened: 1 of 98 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -550,8 +611,7 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
     options.setJobName(name)
   }
 
-  /** Close the context. No operation can be performed once the context is closed. */
-  def close(): ClosedScioContext = requireNotClosed {
+  private[scio] def prepare(): ScioContext = requireNotClosed {
```
ClaireMcGinty: any reason this logic can't go in `ScioRunner` directly?

regadas: reason is to not leak `ScioContext` internals

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -456,7 +517,7 @@ class ScioContext private[scio] (val options: PipelineOptions, private var artif
     }
 
   /** Amount of time to block job for. */
-  private[scio] val awaitDuration: Duration = {
+  val awaitDuration: Duration = {
```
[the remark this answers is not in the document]
regadas: it's not used anywhere but we were offering `getAwaitDuration` before ... so simply offering a way to get it

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -233,6 +233,137 @@ object ContextAndArgs {
   }
 }
 
+trait ScioExecutor {
+  def run(sc: ScioContext): ScioExecutionContext
+}
+
+sealed trait ScioRunner extends ScioExecutor
```
nevillelyh: What's the reason for having `ScioRunner` and `ScioExecutor`? They can be the same trait/object?

regadas: the idea is to allow the following:

```scala
val sc: ScioContext = ???
val testExecutor: ScioExecutor = ???
val runner = ScioRunner(testExecutor)

// case 1
sc.run(runner)

// case 2
runner.run(sc)
```

in either case inner context logic before `run` is preserved. User is only allowed to override the `ScioExecutor`

Issue #2038, opened 2019-07-10
InMemorySink throws exception on empty SCollection


davidvspot:
The cache will not be populated if the SCollection is empty.

PR #2039, opened 2019-07-10
InMemorySink will now create an empty collection in the cache so empty SCollections also return an iterable instead of throwing an exception

davidvspot:
fixes https://github.com/spotify/scio/issues/2038


scio-test/src/test/scala/com/spotify/scio/io/InMemorySinkTest.scala
```diff
@@ -0,0 +1,24 @@
+package com.spotify.scio.io
```
regadas: Can you add a *Copyright* header?

PR #2042, opened 2019-07-10
Add bq not implemented tap when write append

regadas:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -199,7 +199,7 @@ final case class BigQueryTable(table: Table) extends BigQueryIO[TableRow] {
     params.insertErrorTransform(params.extendedErrorInfo.coll(data.context, wr))
 
     if (params.writeDisposition == WriteDisposition.WRITE_APPEND) {
-      throw new NotImplementedError("BigQuery future with append not implemented")
+      NotImplementedTap("BigQuery with append does not support tap")
```
ClaireMcGinty: I kind of feel like this error message is somewhat conflicting with the class name `NotImplementedTap` ? To me `NotImplementedTap` implies that in a future version we might implement it, whereas `"does not support tap"` implies it's not supported by design, and never will be... what do you think?

regadas: Yeah you might be right, renamed this.

regadas:
@ClaireMcGinty maybe! some og the IO's probably need some double check

PR #2054, opened 2019-07-15
bigtable: Add setCellExpiration

brianmartin:
When creating new tables it is common to configure "cell expiration" / "garbage collection" as well.

---

I'm looking for some advice on testing strategy. The tests in `scio-bigtable` are currently pretty sparse.

I've tested against my own Bigtable instance with the following code:

```scala
package com.spotify.scio.bigtable

import com.google.cloud.bigtable.config.BigtableOptions
import org.joda.time.Duration
import org.scalatest._


class TableAdminTest extends FlatSpec with Matchers {

  private val bigtableOptions = new BigtableOptions.Builder()
    .setProjectId("sp-ml-infra")
    .setInstanceId("test-tfrecords-loader")
    .build

  private val duration = Duration.standardDays(7)

  private def set(xs: Map[String, List[String]]): Unit = {
    TableAdmin.setCellExpiration(bigtableOptions, xs, duration)
    TableAdmin.setCellExpiration(bigtableOptions, xs, duration)
  }

  private def ensure(xs: Map[String, List[String]], duration: Option[Duration]): Unit = {
    TableAdmin.ensureTables(bigtableOptions, xs, duration)
  }

  "TableAdmin.setCellExpiration" should "support setCellExpiration" in {
    set(Map("brianm-test" -> List("family1", "family2")))
  }

  "TableAdmin.setCellExpiration" should "skip setCellExpiration when family doesn't exist" in {
    set(Map("brianm-test" -> List("does-not-exist")))
  }

  "TableAdmin.setCellExpiration" should "skip setCellExpiration when table doesn't exist" in {
    set(Map("does-not-exist" -> List("does-not-exist")))
  }

  "TableAdmin.ensureTables" should "skip setCellExpiration when not provided" in {
    ensure(Map("new-table-1" -> List("does-not-exist")), None)
  }

  "TableAdmin.ensureTables" should "set setCellExpiration when provided" in {
    ensure(Map("new-table-1" -> List("does-not-exist")), Some(duration))
  }

}
```

```
sbt:scio-bigtable> testOnly com.spotify.scio.bigtable.TableAdminTest
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Updating gcRule for column families List(family1, family2) in projects/sp-ml-infra/instances/test-tfrecords-loader/tables/brianm-test
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Updating gcRule for column families List(family1, family2) in projects/sp-ml-infra/instances/test-tfrecords-loader/tables/brianm-test
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Skipping modification for non-existent column family does-not-exist in table projects/sp-ml-infra/instances/test-tfrecords-loader/tables/brianm-test
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Skipping modification for non-existent column family does-not-exist in table projects/sp-ml-infra/instances/test-tfrecords-loader/tables/brianm-test
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Skipping modification for non-existent table projects/sp-ml-infra/instances/test-tfrecords-loader/tables/does-not-exist
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Skipping modification for non-existent table projects/sp-ml-infra/instances/test-tfrecords-loader/tables/does-not-exist
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Ensuring tables and column families exist in instance test-tfrecords-loader
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Table new-table-1 exists
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Column families List(does-not-exist) exist in projects/sp-ml-infra/instances/test-tfrecords-loader/tables/new-table-1
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Ensuring tables and column families exist in instance test-tfrecords-loader
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Table new-table-1 exists
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Column families List(does-not-exist) exist in projects/sp-ml-infra/instances/test-tfrecords-loader/tables/new-table-1
[pool-87-thread-4-ScalaTest-running-TableAdminTest] INFO com.spotify.scio.bigtable.TableAdmin$ - Updating gcRule for column families List(does-not-exist) in projects/sp-ml-infra/instances/test-tfrecords-loader/tables/new-table-1
[info] TableAdminTest:
[info] TableAdmin.setCellExpiration
[info] - should support setCellExpiration (4 seconds, 314 milliseconds)
[info] TableAdmin.setCellExpiration
[info] - should skip setCellExpiration when family doesn't exist (1 second, 388 milliseconds)
[info] TableAdmin.setCellExpiration
[info] - should skip setCellExpiration when table doesn't exist (514 milliseconds)
[info] TableAdmin.ensureTables
[info] - should skip setCellExpiration when not provided (611 milliseconds)
[info] TableAdmin.ensureTables
[info] - should set setCellExpiration when provided (2 seconds, 350 milliseconds)
[info] ScalaTest
[info] Run completed in 9 seconds, 367 milliseconds.
[info] Total number of tests run: 5
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 5, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 5, Failed 0, Errors 0, Passed 5
[success] Total time: 14 s, completed Jul 16, 2019, 2:21:55 PM
```

### DIFF
--- a/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
+++ b/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
@@ -78,10 +78,13 @@ object TableAdmin {
    *
    * @param tablesAndColumnFamilies A map of tables and column families.  Keys are table names.
    *                                Values are a list of column family names.
+   * @param cellExpiration The duration before which garbage collection of a cell may occur.
+   *                       Note: minimum granularity is second.
    */
   def ensureTables(
     bigtableOptions: BigtableOptions,
-    tablesAndColumnFamilies: Map[String, List[String]]
+    tablesAndColumnFamilies: Map[String, List[String]],
+    cellExpirationDuration: Option[Duration] = None
   ): Unit = {
     val project = bigtableOptions.getProjectId
     val instance = bigtableOptions.getInstanceId
@@ -109,6 +112,12 @@ object TableAdmin {
         }
 
         ensureColumnFamilies(client, tablePath, columnFamilies)
+      }
+    }.flatMap { _ =>
+      cellExpirationDuration match {
+        case None => Try {}
+        case Some(duration) =>
+          setGcRule(bigtableOptions, tablesAndColumnFamilies, gcRuleFromDuration(duration))
       }
     }.get
   }
@@ -166,9 +175,12 @@ object TableAdmin {
   def setCellExpiration(bigtableOptions: BigtableOptions,
                         tablesAndColumnFamilies: Map[String, List[String]],
                         cellExpiration: Duration): Unit = {
-    val protoDuration = ProtoDuration.newBuilder.setSeconds(cellExpiration.getStandardSeconds)
-    val gcRule = GcRule.newBuilder.setMaxAge(protoDuration).build
-    setGcRule(bigtableOptions, tablesAndColumnFamilies, gcRule)
+    setGcRule(bigtableOptions, tablesAndColumnFamilies, gcRuleFromDuration(cellExpiration)).get
+  }
+
+  private def gcRuleFromDuration(duration: Duration): GcRule = {
+    val protoDuration = ProtoDuration.newBuilder.setSeconds(duration.getStandardSeconds)
+    GcRule.newBuilder.setMaxAge(protoDuration).build
   }
 
   /**
@@ -181,7 +193,7 @@ object TableAdmin {
    */
   private def setGcRule(bigtableOptions: BigtableOptions,
                         tablesAndColumnFamilies: Map[String, List[String]],
-                        gcRule: GcRule): Unit = {
+                        gcRule: GcRule): Try[Unit] = {
     val project = bigtableOptions.getProjectId
     val instance = bigtableOptions.getInstanceId
     val instancePath = s"projects/$project/instances/$instance"
@@ -232,7 +244,7 @@ object TableAdmin {
               .build)
         }
       }
-    }.get
+    }
   }
 
   /**
--- a/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
+++ b/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
@@ -102,7 +102,7 @@ object TableAdmin {
     bigtableOptions: BigtableOptions,
     tablesAndColumnFamiliesWithExpiration: Map[String, List[(String, Option[Duration])]]
   ): Unit = {
-    val tablesAndColumnFamilies = tablesAndColumnFamiliesWithExpiration.mapValues { _.unzip._1 }
+    val tablesAndColumnFamilies = tablesAndColumnFamiliesWithExpiration.mapValues(_.unzip._1)
     ensureTablesImpl(bigtableOptions, tablesAndColumnFamilies).flatMap { _ =>
       setCellExpiration(bigtableOptions, tablesAndColumnFamiliesWithExpiration)
     }.get
--- a/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
+++ b/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
@@ -78,14 +78,48 @@ object TableAdmin {
    *
    * @param tablesAndColumnFamilies A map of tables and column families.  Keys are table names.
    *                                Values are a list of column family names.
-   * @param cellExpiration The duration before which garbage collection of a cell may occur.
-   *                       Note: minimum granularity is second.
    */
   def ensureTables(
     bigtableOptions: BigtableOptions,
-    tablesAndColumnFamilies: Map[String, List[String]],
-    cellExpirationDuration: Option[Duration] = None
+    tablesAndColumnFamilies: Map[String, List[String]]
+  ): Unit =
+    ensureTablesImpl(bigtableOptions, tablesAndColumnFamilies).get
+
+  /**
+   * Ensure that tables and column families exist.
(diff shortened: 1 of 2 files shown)


scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -142,6 +143,75 @@ object TableAdmin {
...
+    val instancePath = s"projects/$project/instances/$instance"
+
+    adminClient(bigtableOptions) { client =>
+      tablesAndColumnFamilies.foreach { case (table, columnFamilies) =>
+        val tablePath = s"$instancePath/tables/$table"
+        val tableInfo = client.getTable(GetTableRequest.newBuilder.setName(tablePath).build)
+
+        val modifications: List[Modification] =
+          columnFamilies
+            .filter { cf =>
```
regadas: nit 🎨 i guess you could use partition here and that would push the logging to a better place?

brianmartin:
Thanks for the review @regadas!

One question here after getting some great input from a teammate:

- The better user experience may actually be to have `cellExpirationDuration` as an optional argument to `ensureTables`. Any opposition to this? (as it would technically be a change to an existing API).

regadas:
@brianmartin I like that I idea and I think we can still make it backwards compatible 👍

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -50,17 +50,41 @@ object TableAdmin {
...
    *
    * @param tablesAndColumnFamilies A map of tables and column families.  Keys are table names.
    *                                Values are a list of column family names.
+   * @param cellExpiration The duration before which garbage collection of a cell may occur.
+   *                       Note: minimum granularity is second.
    */
   def ensureTables(
     bigtableOptions: BigtableOptions,
-    tablesAndColumnFamilies: Map[String, List[String]]
+    tablesAndColumnFamilies: Map[String, List[String]],
```
regadas: @brianmartin would it make sense to have Duration per column family? `Map[String, List[(String, Option[Duration])]]`?

brianmartin: Hm, interesting.. Definitely more flexibility there. For our immediate use case a single `cellExpiration` makes sense. I'll defer to you here.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
brianmartin: I gave it a shot in https://github.com/spotify/scio/pull/2054/commits/a774601c9208afc6f6612ccfb4809293faaf4d3e. Is that what you had in mind?

brianmartin:
I've updated description to reflect more tests and new changes. Note: I couldn't have a default value of None because of this constraint: https://stackoverflow.com/questions/4652095/why-does-the-scala-compiler-disallow-overloaded-methods-with-default-arguments

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
regadas: That's what I was thinking initially and maybe it's easier to get backward compatibility working?

brianmartin:
I've updated the interface to have a per-column-family cell expiration.

The only issue was that we can't do dispatch on `ensureTable` because of:

```
scala> object X {
     |   def x(a: Map[String, String]) = ???
     |   def x(a: Map[String, Int]) = ???
     | }
         def x(a: Map[String, Int]) = ???
             ^
On line 3: error: double definition:
       def x(a: Map[String,String]): Nothing at line 2 and
       def x(a: Map[String,Int]): Nothing at line 3
       have same type after erasure: (a: scala.collection.immutable.Map)Nothing
```

So I've added a couple `ensureTableWithExpiration` functions.

brianmartin:
Sidenote: Looks like the build is running out of memory. Has this started happening on other PRs as well?

```
/opt/sbt/bin/sbt-launch-lib.bash: line 59:   196 Killed                  "$@"
Exited with code 137

Hint: Exit code 137 typically means the process is killed because it was running out of memory
Hint: Check if you can optimize the memory usage in your app
Hint: Max memory usage of this container is 4282101760
 according to /sys/fs/cgroup/memory/memory.max_usage_in_bytes
```

brianmartin:
@regadas Wow, blowing my mind with that GH feature for suggesting changes. Committed those.

brianmartin:
@regadas I really like your approach: it simplifies and also reduces that chaining of Try's I had.

regadas:
@brianmartin ok so I think we can close this in favor of that one?

brianmartin:
Closing in favor of #2067

PR #2060, opened 2019-07-17
Reuse intersect by key in sparse intersect by key

anish749:
This change makes the code more readable and reuses code from `intersectByKey` in `sparseIntersectByKey`

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -870,11 +870,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
 
               if (computeExact) {
                 approxResults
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -870,11 +870,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
 
               if (computeExact) {
                 approxResults
-                  .cogroup(rhs.map((_, ())))
-                  .flatMap { t =>
-                    if (t._2._1.nonEmpty && t._2._2.nonEmpty) t._2._1.map((t._1, _))
-                    else Seq.empty
-                  }
+                  .intersectByKey(rhs)
```
regadas: nit 🎨 can you inline? `approxResults.intersectByKey(rhs)`

PR #2066, opened 2019-07-19
Fix a bunch of issues in `To`

jto:


scio-core/src/main/scala/com/spotify/scio/schemas/instances/JavaInstances.scala
```diff
@@ -69,4 +70,17 @@ trait JavaInstances {
   implicit def javaBeanSchema[T: IsJavaBean: ClassTag]: RawRecord[T] =
     RawRecord[T](new JavaBeanSchema())
 
+  implicit def javaEnumSchema[T <: java.lang.Enum[T] : ClassTag]: Schema[T] =
```
regadas: TBH i think would prefer to use the straightforward way of using the `String` representation here. I see some problematic use case like when you are persist or have multiple pipelines running (i.e streaming) with "somehow" different Enum.

scio-core/src/main/scala/com/spotify/scio/schemas/SchemaMaterializer.scala
```diff
@@ -115,7 +116,18 @@ object SchemaMaterializer {
     }
     record.construct(values)
   }
-  private def decode[A](schema: Type[A])(v: schema.Repr): A = v
+
+  private[scio] def decode[A](schema: Type[A])(v: schema.Repr): A = {
+    schema match {
+      case Type(t) if t.getTypeName() == BSchema.TypeName.LOGICAL_TYPE =>
+        // XXX: The Beam API is shitty
```
regadas: can we reword this? 😄

scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
```diff
@@ -35,13 +47,14 @@ object To {
       case (BSchema.TypeName.ARRAY, BSchema.TypeName.ARRAY, _) =>
         areCompatible(t0.getCollectionElementType, t1.getCollectionElementType)
       case (BSchema.TypeName.MAP, BSchema.TypeName.MAP, _) =>
-        areCompatible(t0.getMapKeyType, t1.getMapKeyType)
+        areCompatible(t0.getMapKeyType, t1.getMapKeyType) &&
```
regadas: whaa nice catch!

scio-core/src/main/scala/com/spotify/scio/schemas/instances/JavaInstances.scala
sisidra: I agree, I think it also helps for use cases when writing beam sql query? I.e. user wants to write enum name, not ordinal. If I understand how this whole thing works :)

scio-core/src/main/scala/com/spotify/scio/schemas/instances/JavaInstances.scala
```diff
@@ -69,4 +70,21 @@ trait JavaInstances {
...
+      private val className = clazz.getCanonicalName
+      private val values: Map[String, T] =
+        clazz.getMethod("values")
+          .invoke(null).asInstanceOf[Array[T]]
+          .map{ e => (e.name(), e) }
+          .toMap
+      override def getIdentifier: String = className
+      override def getBaseType: FieldType = FieldType.STRING
+      override def toBaseType(input: T): String = input.name()
+      override def toInputType(base: String): T = values(base)
```
regadas: awesome! I think we can even reduce this by just using:
 
```suggestion
      override def toInputType(base: String): T = java.lang.Enum.valueOf(clazz, base)
```

jto: Good catch. Apparently I can't Java 😅

scio-test/src/test/java/com/spotify/scio/sql/j/Level.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.spotify.scio.sql.j;
+
+public enum Level {
+  LOW, MEDIUM, HIGH
+}
```
regadas:
```suggestion
}

```

Issue #1864, opened 2019-04-23
Expose Beam's BigQuery WriteResult


anish749:
Beam's BigQuery Write Transform gives our the errors and failures as a [`WriteResult`](https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/WriteResult.java) output when writing a PCollection to BigQuery.

Currently Scio Exposes a Tap of the output table and the `WriteResult` is not accessible any more.

It would be nice to preserve this functionality of Beam and expose the error records as an `SCollection` to the user.

I am not sure if `saveAsBigQuery` should be modified, but maybe we can have something like `writeToBigQuery` which returns `(SCollection[TableRow], SCollection[BigQueryInsertError])`

One side effect of not having this is that this results in unconsumed reads and a `DropInputs` transform appearing in Dataflow UI which is hard to understand and reason about when writing to Big Query

regadas:
Hi @anish749 we are aware of this and we do have plans in future to address it. Unfortunately, I can 't really give you a timeline for this.

anish749:
As a short term solution do you think it is a good idea to have a no-op after the transform? This would make sure we don't have an `UnconsumedRead` and the name in the Dataflow is more understandable. Right now it comes up as `DropInputs1` which is kind of misleading.

Beam is running [this](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/UnconsumedReads.java) after creating the pipeline to add these annoying and misleading drop inputs.

This becomes especially confusing when the write happens from inside a PTransform (`sColl.transform`)

I can try this in a PR.

anish749:
> yeah I can see this happening if you have failed inserts when using unbounded collection.

I am doing this for a bounded SCollection. The Drop Inputs is added by this transform [here](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/UnconsumedReads.java) which is executed by beam when the graph creation is complete but before sending to Dataflow.

> if you are referring to adding a transform that consumes those errors, we could add it and at leas log them.

[...]

PR #2075, opened 2019-07-24
Default insert error transform for BQ IO

anish749:
By default Beam adds a DropInput transform outside the transform writing to BQ. This leads to some not-so-easy-to-reason-about boxes in the Dataflow graph. This eagerly adds a no-op by default instead of relying on Beam to use Drop Inputs.

Follow up on #1864 

TODO
 - ~Set this if `ExtendedErrorInfo` is disabled, else leave it as is.~
 - ~Test this via some integration test (?) / Check the UI on an integration test.~

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -230,7 +230,7 @@ object BigQueryTable {
       : SCollection[DefaultExtendedErrorInfo.Info] => Unit =
       // A NoOp on the extended error output, so that we don't have DropInputs
       // in the pipeline graph.
-      _.withName("Ignore ExtendedErrorInfo").map(_ => ())
+      _.withName("DropFailedInserts").map(_ => ())
 
     @inline final def apply(
       s: TableSchema,
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -228,7 +228,8 @@ object BigQueryTable {
     private[bigquery] val DefaultExtendedErrorInfo: ExtendedErrorInfo = ExtendedErrorInfo.Disabled
     private[bigquery] val DefaultInsertErrorTransform
       : SCollection[DefaultExtendedErrorInfo.Info] => Unit =
-      // A NoOp on the extended error output, so that we don't have DropInputs in the pipeline graph.
+      // A NoOp on the extended error output, so that we don't have DropInputs
+      // in the pipeline graph.
       _.withName("Ignore ExtendedErrorInfo").map(_ => ())
 
     @inline final def apply(
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -228,7 +228,7 @@ object BigQueryTable {
     private[bigquery] val DefaultExtendedErrorInfo: ExtendedErrorInfo = ExtendedErrorInfo.Disabled
     private[bigquery] val DefaultInsertErrorTransform
       : SCollection[DefaultExtendedErrorInfo.Info] => Unit =
-      // A NoOp on the extended error output, so that we don't have DropInputs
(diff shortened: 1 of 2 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -227,7 +227,10 @@ object BigQueryTable {
     private[bigquery] val DefaultTimePartitioning: TimePartitioning = null
     private[bigquery] val DefaultExtendedErrorInfo: ExtendedErrorInfo = ExtendedErrorInfo.Disabled
     private[bigquery] val DefaultInsertErrorTransform
-      : SCollection[DefaultExtendedErrorInfo.Info] => Unit = _ => ()
+      : SCollection[DefaultExtendedErrorInfo.Info] => Unit =
+      // A NoOp on the extended error output, so that we don't have DropInputs
+      // in the pipeline graph.
+      _.withName("Ignore ExtendedErrorInfo").map(_ => ())
```
nevillelyh: Trying to understand this. How is adding a transform getting rid of it from the graph?

anish749: We have a step in Beam, where before running (submitting) the job to the runner, it is traversing the graph and consuming all transforms which are Reads without a Sink. This code is [here](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/UnconsumedReads.java)

The BQ write transform outputs errors by creating a read of insert errors.

This adds a [NoOp](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/UnconsumedReads.java#L67-L69) transform to consume this read.


When we organize the pipeline in multiple composable transforms, having a bq write from within a transform results in the `NoOp` lying outside of transform.

Also if we write to BQ, this transform shows up as [DropInput](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/UnconsumedReads.java#L63) which can be confusing for someone not very familiar with this.

nevillelyh: Might be hard to test plus a bit overkill. LGTM as is.

regadas: @anish749 
This is actually independent of the `ExtendedErrorInfo` value so no need to.
> Set this if ExtendedErrorInfo is disabled, else leave it as is.

I would also rename as:

```suggestion
      _.withName("DropFailedInserts").map(_ => ())
```

there's another place where this is required.

scio-test/src/test/scala/com/spotify/scio/io/ScioIOTest.scala
```diff
@@ -98,6 +106,62 @@ class ScioIOTest extends ScioIOSpec {
...
+      new PipelineVisitor.Defaults {
+
+        override def visitPrimitiveTransform(node: TransformHierarchy#Node): Unit =
+          consumedOutputs ++= node.getInputs.values().asScala
+
+        override def visitValue(
+          value: PValue,
+          producer: TransformHierarchy#Node
+        ): Unit = {
+          if (producer.getTransform.isInstanceOf[Read.Bounded[_]] ||
```
regadas: 🎨 `match` would have been more clean 😄

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -227,7 +227,10 @@ object BigQueryTable {
     private[bigquery] val DefaultTimePartitioning: TimePartitioning = null
     private[bigquery] val DefaultExtendedErrorInfo: ExtendedErrorInfo = ExtendedErrorInfo.Disabled
     private[bigquery] val DefaultInsertErrorTransform
-      : SCollection[DefaultExtendedErrorInfo.Info] => Unit = _ => ()
+      : SCollection[DefaultExtendedErrorInfo.Info] => Unit =
```
regadas: @anish749 There's another default function here https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L496

anish749:
Thanks for the reviews :) I've updated the code.

PR #2082, opened 2019-07-25
Another take @ simplifying scio context execution

regadas:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -411,8 +403,8 @@ object ScioContext {
     new ScioContext(options, artifacts)
 
   /** Create a new [[ScioContext]] instance. */
-  def apply(options: PipelineOptions, runner: ScioRunner): ScioContext =
-    new ScioContext(options, Nil, runner)
+  def apply(options: PipelineOptions, executor: ScioExecutor): ScioContext =
```
anish749: I now noticed that this constructor is public.
Do you think this should also be propagated to `ContextAndArgs` as well?
So we can do something like 
```
val (sc: ScioContext, args: Args) = ContextAndArgs(cmdlineArgs, executor)
```

jto:
We're starting to have a lot of breaking changes in `0.8`. 
Can you write add Scalafix to this PR ?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
regadas: @anish749 tbh i think for the time I'll keep it only on `ScioContext` still a bit unsure about it's full usefulness

regadas:
Closing in favor of #2092

PR #2083, opened 2019-07-25
Improve doc for deprecated close method

regadas:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -653,13 +653,52 @@ class ScioContext private[scio] (
   /**
    * Runs the underlying pipeline according to the options used to create this context.
    * Running closes the context and no operation can be performed once the context is closed.
+   * This method does not respect the deprecated `--blocking` flag.
+   *
+   * This method does not respect the deprecated `--blocking` flag.
```
brianmartin: Did you intend both line 656 and 658?

regadas: oh nice catch! Thanks @brianmartin

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -653,13 +653,43 @@ class ScioContext private[scio] (
   /**
    * Runs the underlying pipeline according to the options used to create this context.
    * Running closes the context and no operation can be performed once the context is closed.
+   *
+   * This operation is non-blocking and no operation can be performed once the context is closed.
```
nevillelyh: "according to the options used to create this context." is a bit redundant and implied?

Clarify `no further operation`? No further transformations?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -653,13 +653,43 @@ class ScioContext private[scio] (
...
+   * Runs the underlying pipeline according to the options used to create this context.
+   * Running closes the context and no operation can be performed once the context is closed.
+   *
+   * This operation is non-blocking and no operation can be performed once the context is closed.
+   *
+   * @param runner [[ScioRunner]] responsible for the execution for this context.
+   *
+   * @return the [[ScioExecutionContext]] for the underlying job execution.
+   */
   def run(runner: ScioRunner): ScioExecutionContext = runner.run(this)
```
nevillelyh: Add something on how this is different from the default `run()`, like explain the purpose of the runner?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -653,13 +653,43 @@ class ScioContext private[scio] (
...
+   * This method is deprecated and with it the `--blocking` flag.
+   *
+   * Use [[ScioContext#run]].
+   *
+   * To achieve the same behaviour when `--blocking` was enabled use:
+   *
+   * {{{
+   * val coll: SCollection = ???
+   *
+   * coll.run().waitUntilDone(Duration.Inf)
```
nevillelyh: Should be `sc.run()`?

regadas: right

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
regadas: I plan to remove this in the stable release ... so I'll not bother with it 😄

PR #2084, opened 2019-07-26
Change not to call String.format when fromQuery has only single argument

mrkm4ntr:
I tried to compile following code, but it was not able to be compiled because `%Y(/m/d)` was handled as format string.
```
@BigQueryType.fromQuery("""#standardSQL
SELECT id, created, updated, FORMAT_TIMESTAMP("%Y-%m-%d", created, 'Asia/Tokyo') AS created_day FROM `table`""")
  class Row
```

Next, I changed like this, so it could be compiled. But actual executed query had double `%`. 
```
@BigQueryType.fromQuery("""#standardSQL
SELECT id, created, updated, FORMAT_TIMESTAMP("%%Y-%%m-%%d", created, 'Asia/Tokyo') AS created_day FROM `table`""")
  class Row
```

So, I took this workaround.
```
@BigQueryType.fromQuery("""#standardSQL
SELECT id, created, updated, FORMAT_TIMESTAMP("%s", created, 'Asia/Tokyo') AS created_day FROM `table`""", "%Y-%m-%d")
  class Row
```
and interpolated at runtime.
```
sc.typedBigQuery[Row](Row.query("%y-%m-%d"))
``` 

I think this behavior of fromQuery is not intuitive.

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
@@ -407,7 +407,11 @@ private[types] object TypeProvider {
   }
 
   private def formatString(xs: List[Any]): String =
-    xs.head.asInstanceOf[String].format(xs.tail: _*)
+    xs match {
+      case (head: String) :: Nil  => head
+      case (head: String) :: tail => head.format(tail: _*)
+      case _                      => ""
+    }
 
   /** Generate a case class. */
   private def caseClass(
--- a/site/src/paradox/io/Type-Safe-BigQuery.md
+++ b/site/src/paradox/io/Type-Safe-BigQuery.md
@@ -55,7 +55,7 @@ import com.spotify.scio.bigquery.types.BigQueryType
 class Row
 
 // generate new query strings at runtime
-val newQuery = Row.query.format(args(0))
+val newQuery = Row.query(args(0))
 ```
 
 There's also a `$LATEST` placeholder for table partitions. The latest common partition for all tables with the placeholder will be used.
@@ -70,7 +70,7 @@ import com.spotify.scio.bigquery.types.BigQueryType
 class Row
 
 // generate new query strings at runtime
-val newQuery = Row.query.format(args(0), args(0))
+val newQuery = Row.query(args(0), args(0))
 ```
 
 ### BigQueryType.fromSchema


site/src/paradox/io/Type-Safe-BigQuery.md
```diff
@@ -55,7 +55,7 @@ import com.spotify.scio.bigquery.types.BigQueryType
 class Row
 
 // generate new query strings at runtime
-val newQuery = Row.query.format(args(0))
+val newQuery = Row.query(args(0))
```
nevillelyh: Why removing `format`? `query` is `String` and `String()` is `charAt`?

mrkm4ntr: If `fromQuery` of Row has multiple arguments, `Row` has a method `query` like below.
```
def query(queryArg$macro$1: String): String = "...".format(queryArg$macro$1)
```
This is generated here.
https://github.com/spotify/scio/blob/a6268507fb86be5fb9233231e3b4320eb351c03d/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala#L139-L144

If `fromQuery` of Row has a single argument, `Row.query(args(0))` will be compile error.  I think it is preferred because the query is not intended to be interpolated in such case.

mrkm4ntr:
> For example: String.format("abc%d") fails and String.format("abc%%d") works.

Yes, that's why my first example is failed to compile.

In second example, it is intuitive that I have to call format explicitly when I execute this query?
```
@BigQueryType.fromQuery("""#standardSQL
SELECT id, created, updated, FORMAT_TIMESTAMP("%%Y-%%m-%%d", created, 'Asia/Tokyo') AS created_day FROM `table`""")
  class Row
```
I just want to execute like below, but Row.query is not formatted.
```
sc.typedBigQuery[Row]()
```
I have to execute like this.
```
sc.typedBigQuery[Row](String.format(Row.query))
```

> It's breaking for those already using %%Y to escape

I don't think so. Those who use %%Y with single argument must call String.format explicitly like above. So, this change doesn't affect to them.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -407,7 +407,10 @@ private[types] object TypeProvider {
   }
 
   private def formatString(xs: List[Any]): String =
-    xs.head.asInstanceOf[String].format(xs.tail: _*)
+    xs.tail match {
```
regadas: @mrkm4ntr Thanks for this! wdyt about having it like 

```scala
xs match {
  case (head: String) :: Nil => head
  case (head: String) :: tail => head.format(xs.tail: _*)
  case _ => ""
}
```

mrkm4ntr: @regadas Thanks, fixed.

mrkm4ntr:
@regadas PTAL.

nevillelyh:
Thanks!

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
regadas: @mrkm4ntr Thanks!

Issue #2085, opened 2019-07-26
Spark waitUntilFinish() does not wait when using Duration.Inf


JozoVilcek:
I have a pipeline Scio: `0.8.0-alpha2` running on `SparkRunner`. To kick off pipeline I was using
`sc.close()` with `blocking = true` in `ScioOptions`. Migrating to `0.8.0-alpha2` I am doing now 

`sc.run().waitUntilFinish()`
but it does not block and wait for pipeline to finish. On the other hand, as a workaround, I am doing

`sc.pipeline.run().waitUntilFinish()`
which behaves as expected and waits untill job is finished.

JozoVilcek:
This does not feels right for `Duration.Inf` wait
https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L337

But even when specifying duration explicitly, it does not wait.

regadas:
Hi @JozoVilcek Which version are you migrating from?

Regarding the `Duration.Inf` it seems that it's adhering to apache beam `PipelineResult`... values `<=0` wait inf. However, from a brief look it seems that `SparkRunner` does not follow it.

JozoVilcek:
Can you share a reference to spark runner code which is casing this not wait problem?

regadas:
@JozoVilcek also is this a streaming pipeline or batch?

JozoVilcek:
I am running batch pipeline

regadas:
@JozoVilcek yeah this is an issue in `Beam` with `SparkPipelineResult` that does not follow `PipelineResult`.

as a quick test I ran our [WordCount](https://github.com/spotify/scio/blob/master/scio-examples/src/main/scala/com/spotify/scio/examples/WordCount.scala) with a big enough duration:

```
val sc: ScioContext = ???

sc.run().waitUntilFinish(1.hour)
```

So make sure you give a large duration for the time being.

regadas:
The fix was merged in `Beam` not sure if it will be in `2.14.0`

PR #2086, opened 2019-07-26
Use lower value to indicate Duration.Inf. Behaviour matches PipelineResult

regadas:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -334,7 +334,8 @@ object ScioExecutionContext {
       override def waitUntilFinish(duration: Duration, cancelJob: Boolean): ScioResult = {
         try {
           val wait = duration match {
-            case Duration.Inf => 0
+            // according to PipelineResult values <= 1 ms mean `Duration.Inf`
```
nevillelyh: Isn't `0 <= 1`?
> The time to wait for the pipeline to finish. Provide a value less than 1 ms for an infinite wait.

Spark runner uses `Long.MAX_VALUE` and is backed by a Java `Future.get(long, TimeUnit)`.
https://github.com/apache/beam/blob/master/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineResult.java#L86
https://github.com/apache/beam/blob/master/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineResult.java#L136

regadas: This is just to be consistent with `DataflowPipelineJob` which uses the same [default value](https://github.com/apache/beam/blob/master/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java#L176).

Regarding `SparkRunner` yes that is the issue for #2085 which I believe it's a bug in `Beam`. Already made a [PR](https://github.com/apache/beam/pull/9170) for it.

PR #2093, opened 2019-07-30
Add SparkRunner context

regadas:
This should make job submission through SBT a little bit easier


nevillelyh:
Files to stage logic might be DF runner specific though especially when submitting jobs inside `scio-repl`. Not sure how it'll work for `SparkRunner`. Also there's a new `SparkPipelineRunner` (portability) now.

regadas:
Hmm, `filesToStage` is need to submit jobs for `SparkRunner`. `SparkPipelineRunner` i can have a look

regadas:
`SparkPipelineRunner` seems to need it as well ...

PR #2102, opened 2019-07-31
0.8 migration rules

jto:


jto:
It seems that there's a weird issue where rule tests a passing but when applying the migration on an existing project, some fix gets applied twice.

scalafix/rules/src/main/scala/fix/0.8/MigrateV0_8.scala
```diff
@@ -0,0 +1,143 @@
...
+private class FixScioIO extends SemanticRule("FixScioIO") {
+
+  // Check that the method is a member of an implementation of ScioIO
+  private def isScioIOMember(t: Tree)(implicit doc: SemanticDocument) =
+    t.parent.collect {
+      case p @ Template(_) =>
+        p.inits.exists(_.tpe.symbol == Symbol("com/spotify/scio/io/ScioIO#"))
+    }.getOrElse(false)
+
+  override def fix(implicit doc: SemanticDocument): Patch = {
```
anish749: Correct me if I am wrong, but I don't think this is modifying imports.

I saw this change of going from `Future` to `ClosedTap`. Doesn't this need a change in the import as well?

[This](https://ghe.spotify.net/payments/pipeline-subs-platform/pull/1168/files#diff-31ada5de4461a63e507abdd6ce0b34fb) is what I was doing manually.

regadas:
@jto sweet! I'm gonna have a look at this tmr! I'm almost done with the doc as well.

scalafix/rules/src/main/scala/fix/0.8/MigrateV0_8.scala
```diff
@@ -0,0 +1,143 @@
...
+}
+
+private class FixWaitForResult extends SemanticRule("FixWaitForResult") {
+  override def fix(implicit doc: SemanticDocument): Patch = {
+    doc.tree.collect {
+      case t @ q"$x.waitForResult()" =>
+        x.symbol.info.get.signature match {
+          // XXX: Hackish wait to check the type of x
+          case ValueSignature(tpe) if tpe.toString.startsWith("WaitableFutureTap") =>
+            Patch.replaceTree(t, x.syntax)
```
regadas: hm, tbh i think this fix will  be a bit more complex and tricky ... we went from `Future[Tap[T]]` => `ClosedTap[T]` removing `waitForResult` is not enough ... we actually need the result from `sc.close()` ie:

```scala
    val materialized = sc.textFile(input).materialize
    val executionContext = sc.close()
    executionContext.waitUntilDone().tap(materialized).value
```

regadas: @jto what do you think of removing this rule for now? and get this merged so we can keep iterating on this?

jto: I will :)

PR #2110, opened 2019-08-01
handle non-overlap keys in hashFullOuterJoin, fix #2109

### LINKED ISSUE #2109: hashFullOuterJoin does not include items from right only collection
The following code demonstrates issue:
```scala
import com.spotify.scio._
val sc = ScioContext.forTest()
val left = sc.parallelize(Seq("a" -> 1))
val right = sc.parallelize(Seq("b" -> 2))
val hashJoined = left.hashFullOuterJoin(right).debug(prefix = "HashJoined: ")
val shuffleJoined = left.fullOuterJoin(right).debug(prefix = "ShuffleJoined: ")
sc.close().waitUntilDone()
```
The expected result - Shuffle and Hash should give exactly the same result - items from both collections:
```
ShuffleJoined: (b,(None,Some(2)))
ShuffleJoined: (a,(Some(1),None))
HashJoined: (b,(None,Some(2)))
HashJoined: (a,(Some(1),None))
```
Actual result - hash join does not return an item from `right` collection:
```
ShuffleJoined: (b,(None,Some(2)))
ShuffleJoined: (a,(Some(1),None))
HashJoined: (a,(Some(1),None))
```

Affected versions:
At least 0.7.4 and 0.8.0-alpha2.

nevillelyh:
Summary:

Beam `Combine.Globally` transform doesn't allow defaults when windowing is not the default `GlobalWindow`, and we erroneously added `.withoutDefaults()` to all `Combine.Globally` instances in `PCollectionWrapper`.

To fix properly, `.withoutDefaults()` should only be set when:
- `CombineFn` doesn't have a logical default, e.g. `reduce`, `Semigroup`, `Aggregator`, or
- `PCollection#getWindowStrategy != WindowingStrategy.globalDefault()`

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/package.scala
@@ -17,7 +17,7 @@
 
 package com.spotify
 
-import com.twitter.algebird.Monoid
+import com.twitter.algebird.Semigroup
 
 import scala.reflect.ClassTag
 
@@ -30,20 +30,20 @@ import scala.reflect.ClassTag
  */
 package object scio {
 
-  /** [[com.twitter.algebird.Monoid Monoid]] for `Array[Int]`. */
-  implicit val intArrayMon: Monoid[Array[Int]] = new ArrayMonoid[Int]
+  /** [[com.twitter.algebird.Semigroup Semigroup]] for `Array[Int]`. */
+  implicit val intArraySg: Semigroup[Array[Int]] = new ArraySemigroup[Int]
 
-  /** [[com.twitter.algebird.Monoid Monoid]] for `Array[Long]`. */
-  implicit val longArrayMon: Monoid[Array[Long]] = new ArrayMonoid[Long]
+  /** [[com.twitter.algebird.Semigroup Semigroup]] for `Array[Long]`. */
+  implicit val longArraySg: Semigroup[Array[Long]] = new ArraySemigroup[Long]
 
-  /** [[com.twitter.algebird.Monoid Monoid]] for `Array[Float]`. */
-  implicit val floatArrayMon: Monoid[Array[Float]] = new ArrayMonoid[Float]
+  /** [[com.twitter.algebird.Semigroup Semigroup]] for `Array[Float]`. */
+  implicit val floatArraySg: Semigroup[Array[Float]] = new ArraySemigroup[Float]
 
-  /** [[com.twitter.algebird.Monoid Monoid]] for `Array[Double]`. */
-  implicit val doubleArrayMon: Monoid[Array[Double]] = new ArrayMonoid[Double]
+  /** [[com.twitter.algebird.Semigroup Semigroup]] for `Array[Double]`. */
+  implicit val doubleArraySg: Semigroup[Array[Double]] = new ArraySemigroup[Double]
 
-  private class ArrayMonoid[@specialized(Int, Long, Float, Double) T: ClassTag: Numeric]
-      extends Monoid[Array[T]] {
+  private class ArraySemigroup[@specialized(Int, Long, Float, Double) T: ClassTag: Numeric]
+      extends Semigroup[Array[T]] {
 
     private val num = implicitly[Numeric[T]]
 
@@ -58,7 +58,6 @@ package object scio {
       l
     }
 
-    override val zero: Array[T] = Array.empty[T]
     override def plus(l: Array[T], r: Array[T]): Array[T] = {
       if (l.length == 0) {
         r
--- a/scio-core/src/main/scala/com/spotify/scio/package.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/package.scala
@@ -59,16 +59,11 @@ package object scio {
     }
 
     override def plus(l: Array[T], r: Array[T]): Array[T] = {
-      if (l.length == 0) {
-        r
-      } else if (r.length == 0) {
-        l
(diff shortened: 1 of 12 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala
```diff
@@ -42,7 +42,8 @@ private[values] trait PCollectionWrapper[T] extends TransformNameable {
     transform: PTransform[_ >: PCollection[T], PCollection[U]]
   ): SCollection[U] = {
     val t =
-      if (classOf[Combine.Globally[T, U]] isAssignableFrom transform.getClass) {
+      if ((classOf[Combine.Globally[T, U]] isAssignableFrom transform.getClass)
+          && internal.getWindowingStrategy != WindowingStrategy.globalDefault()) {
```
ClaireMcGinty: What's the reasoning behind modifying this `PTransform` in Scio, instead of letting Beam throw an error upon [expansion of the Combine.Globally transform](https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Combine.java#L1184)? Since we're doing this modification "behind the scenes" users might get confused why their code isn't behaving like they'd expect?

nevillelyh: Without this, all Scio "global combine" methods will fail when there's windowing and there's no way to fix without dropping to Beam API.

scio-test/src/test/scala/com/spotify/scio/ArraySemigroupTest.scala
```diff
@@ -34,25 +33,17 @@ class ArrayMonoidTest extends FlatSpec with Matchers {
   "ArrayMonoid" should "support plus" in {
```
ClaireMcGinty: `ArrayMonoid` -> `ArraySemigroup`?

scio-core/src/main/scala/com/spotify/scio/package.scala
```diff
@@ -30,20 +30,20 @@ import scala.reflect.ClassTag
  */
 package object scio {
 
-  /** [[com.twitter.algebird.Monoid Monoid]] for `Array[Int]`. */
-  implicit val intArrayMon: Monoid[Array[Int]] = new ArrayMonoid[Int]
+  /** [[com.twitter.algebird.Semigroup Semigroup]] for `Array[Int]`. */
```
regadas: tbh i don't like the fact that this implicits are here with this scope.

Given that they are, this is a breaking change. Is this change really necessary for this PR? Why changing from Monoid to Semigroup?

nevillelyh: `Array.empty` is not the same as array of fixed size. And there's no way to produce a correct `zeroValue` without seeing an array first.

Removing it from the scope is also a breaking change but probably lesser of 2 evils, assuming not many does `.sum` on arrays.

scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala
ClaireMcGinty: cool, I was forgetting that user code doesn't directly interact directly with `pApply`. in that case isn't it redundant to add `.withoutDefaults())` to global combines elsewhere in this PR?

scio-core/src/main/scala/com/spotify/scio/package.scala
regadas: hm, doesn't it only affect the allocation? 

yeah, having this like this leads to that. However, i would probably push this change to another PR.

scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala
nevillelyh: Hmm... They're not exactly the same, depending on windowing.

With windowing:
- `withoutDefaults` always needed, regardless whether the `fn` supports default value or not.

Without windowing:
- `aggregate(Aggregator)`, `combine`, `sum` - always need `withoutDefaults`, otherwise the runner might attempt to get default value and fail the assertion below.
`at com.spotify.scio.util.Functions$$anon$2.extractOutput(Functions.scala:171)`
- `fold`, `aggregate(zeroValue)` provides defaults and should not call `withoutDefaults` unless if there's windowing.

So basically it depends on the type of `CombineFn` which we can't determine here.

scio-core/src/main/scala/com/spotify/scio/package.scala
nevillelyh: Downstream transform will make assumption about the array dimension, i.e. if you're `sum`ming 40 dimension vectors it'll assume the single output has size 40 and AIOB if it's 0.

regadas: I think that's somewhat application specific and I it would need a different Monoid instance. This one is a lawful one (as long the array is not mutated) but also the simplest.

I think it will be better to mark these Monoid's as deprecated and then let users defined their instance according to their needs.

nevillelyh: Empty zero is not a lawful monoid, because  `zero + v` doesn't work if `v` has > 0 dimension and `+` is element-wise plus.

PR #2124, opened 2019-08-02
remove ArrayMonoid

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/package.scala
```diff
@@ -17,76 +17,11 @@
...
-          plusI(s, a)
-        }
-      }
-      Option(s)
-    }
-
-  }
-
-}
+package object scio
```
regadas: :art: we can probably just remove it completely :smile:

PR #2125, opened 2019-08-02
Add Char coder

regadas:


regadas:
Nothing special really, I was doing some tests and breaking down String's into Char's and I noticed that we were fallbacking to kryo. Since this is a primitive type, I think it doesn't hurt to add it.

PR #2134, opened 2019-08-06
add coder for BaseAsyncLookupDoFn.Try

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -378,7 +378,7 @@ trait ScalaCoders {
   implicit def bigDecimalCoder: Coder[BigDecimal] =
     Coder.xmap(Coder.beam(BigDecimalCoder.of()))(BigDecimal.apply, _.bigDecimal)
 
-  implicit def tryCoder[A: Coder, B: Coder]: Coder[Try[A]] =
+  implicit def tryCoder[A: Coder]: Coder[Try[A]] =
```
regadas: nice catch 👍

PR #2135, opened 2019-08-06
fix some compiler warnings

nevillelyh:


jto:
scalafmt is not happy

nevillelyh:
Wow so `scalafmtCheck` failed because of `): Unit =  {` (2 spaces) but didn't fix it in `scalafmt`.

regadas:
that's weird indeed

PR #2142, opened 2019-08-07
Add CombineContext support

### LINKED ISSUE #2140: CombineFn doesn't use NullCoder when set in ScioOptions

regadas:
Fixes #2140

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
@@ -20,9 +20,13 @@ package com.spotify.scio.util
 import java.lang.{Iterable => JIterable}
 import java.util.{ArrayList => JArrayList, List => JList}
 
+import com.spotify.scio.ScioContext
+import com.spotify.scio.options.ScioOptions
 import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.twitter.algebird.{Monoid, Semigroup}
 import org.apache.beam.sdk.coders.{CoderRegistry, Coder => BCoder}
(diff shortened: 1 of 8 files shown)


scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -63,27 +66,48 @@ private[scio] object Functions {
...
-      CoderMaterializer.beamWithDefault(vocoder, registry)
+    ): BCoder[VO] = {
+      val options = PipelineOptionsFactory.create()
+      options.as(classOf[ScioOptions]).setNullableCoders(context.useNullableCoders)
+      CoderMaterializer.beamWithDefault(vocoder, registry, options)
+    }
   }
 
   def aggregateFn[T: Coder, U: Coder](
+    sc: ScioContext,
```
nevillelyh: Should we make these `sc: ScioContext` params (and the one in `PCollectionWrapper`) implicit?

regadas: Thought about it ... to be honest this is only internal so i think it's not a big deal but ¯\_(ツ)_/¯ not sure lol

nevillelyh: Yeah just curious. There must be a reason `val context` isn't implicit in the first place but I don't remember...

regadas: @nevillelyh since we don't any `ScioContext` as implicit anywhere and since this internal api ... i think it's good to merge as is?

PR #2144, opened 2019-08-09
Fix java.time.Instant coder

sisidra:
- Support full Instant precision
- Support full range of Instant values


kanterov:
Why don't we use `org.apache.beam.sdk.coders.InstantCoder`?

sisidra:
beam `InstantCoder` is for Joda.

PR #2145, opened 2019-08-09
clean up coders

nevillelyh:
- Implement structuralValue for map coders
- Delete our FloatCoder now that upstream has it
- Delegate to structuralValue, but only if `consistentWithEquals` is false
- Call Builder#sizeHint whenever available


scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
```diff
@@ -39,9 +39,13 @@ import scala.reflect.ClassTag
 import scala.util.{Failure, Success, Try}
 
 private object VoidCoder extends AtomicCoder[Void] {
+  private val STRUCTURAL_VOID_VALUE = new AnyRef
+
   override def encode(value: Void, outStream: OutputStream): Unit = ()
 
   override def decode(inStream: InputStream): Void = ???
+
+  override def structuralValue(value: Void): AnyRef = STRUCTURAL_VOID_VALUE
```
regadas: I think you can just use `AnyRef` object?

PR #2158, opened 2019-08-13
count should return a single emission of 0 when ran on an empty SCollection

### LINKED ISSUE #2157: count should return a single item of 0 instead of returning an empty SCollection
This bug has been flagged in Beam but has yet to be solved so compensating for this in Scio

https://issues.apache.org/jira/browse/BEAM-7123

davidvspot:
resolves #2157

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -385,14 +385,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @group transform
    */
   def count: SCollection[Long] =
-    context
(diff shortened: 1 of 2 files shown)


regadas:
@nevillelyh 🤔 we could probably reify this instead of using union?

nevillelyh:
@regadas do you mean the Beam `Reify` transform?

davidvspot:
Could this be the source of this issue?
https://github.com/spotify/scio/blob/e10771a93634d04496519ab7488c2cecf0de4268/scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala#L48

I cant test it atm because master wont compile for some reason but disabling defaults for empty collections sounds exactly like something that would break count...

PR #2173, opened 2019-08-21
fix #2169, skip HTTP call if ignoreVersionWarning is set

### LINKED ISSUE #2169: Allow hidden version check HTTP call to be disabled
Each time a `ScioContext` class is initialized, scio makes a HTTP call to GitHub to check if there's a newer version.  If there is, it logs a warning.

The warning itself can be disabled with the system property `scio.ignoreVersionWarning`.  However, this just disables the warning messages -- the HTTP call is still made!

This should absolutely be configurable (ideally from code rather than solely from a global system property).  Anything that goes wrong with an external HTTP call like this could potentially put our production systems at risk, and it's completely unnecessary for teams that are prudent about keeping track of versions.

This should all be documented, as well.  Especially since it currently defaults to always making this call, but there's no indication of it anywhere in the docs or in the logs unless there's a new version available.

In summary:
 - ScioContext makes a hidden HTTP call every time a new one is instantiated
 - This cannot currently be disabled
 - This should be able to be disabled, ideally via a code param rather than a system property
 - The fact that the check is made, as well as the ways to disable the check, should be documented.

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
```diff
@@ -99,7 +99,7 @@ private[scio] object VersionUtil {
 
   def checkVersion(
     current: String,
-    latest: Option[String],
```
regadas: making this a call-by-name would also work `latest: => Option[String]`

nevillelyh: That was my initial thought but feels it could be confusing, and the argument is really only used for testing.

PR #2204, opened 2019-09-03
Add Storage API read from query

regadas:


nevillelyh:
Looks like the Beam `BigQueryIO` calls [`expandForDirectRead`](https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L928) when `DIRECT_READ` is set, but then bypasses [`ReadOptions/SelectedFields/RowRestrictions`](https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L1099) completely and delegate to `BigQueryStorageQuerySource`, which then does `BigQueryQueryHelper.executeQuery`? Does that even leverage the Storage API?

nevillelyh:
https://github.com/apache/beam/commit/2399c72041a74cbf30a565bceeb4ecdebd3fd956#diff-667f600e3c8f86fe0ce171b3915e51ac

Ah looks like it runs query as a temp table like the old `fromQuery` source, but reads the result with Storage API, so at least it works around the export quota limit.

regadas:
@nevillelyh yup Storage API can't handle all anonymous tables right now, so they need to create a temp table and read from that.

PR #2205, opened 2019-09-03
Reduce some guava usages

regadas:


scio-jmh/src/test/scala/com/spotify/scio/jmh/JoinBenchmark.scala
```diff
@@ -34,7 +33,7 @@ import scala.collection.JavaConverters._
 class JoinBenchmark {
 
   private def genIterable(n: Int): JIterable[Int] = {
-    val l = Lists.newArrayList[Int]()
+    val l = List.empty[Int].asJava
```
sisidra: How about `java.util.ArrayList(n)`?
Or whole block `IntStream.rangeClosed(1, 100).boxed().collect(Collectors.toList())`

regadas: search and replace sometimes fails 😄 this can actually just be `(1 to n).asJava`

PR #2235, opened 2019-09-12
Add helpers for tensorflow IO using tfexample-derive

andrewsmartin:
Using https://github.com/spotify/tfexample-derive, converters between case classes and Tensorflow Example protos can be automatically derived. This PR adds some helpers in `scio-tensorflow` for being able to read and write TF record files without users needing to manually do the conversion.


andrewsmartin:
Discussed with @nevillelyh offline - will close this for now and revisit when we have a better long-term plan for replacing `shapeless-datatype`.

PR #2241, opened 2019-09-13
use upstream magnolia for scala >= 2.12, fix #2192

### LINKED ISSUE #2192: Stop using forked magnolia
https://github.com/spotify/scio/blob/master/build.sbt#L375

Just realized that we're still using a forked magnolia, and can't use latest upstream since they dropped 2.11 support. We should figure out a way forward and minimize custom stuff.

https://github.com/propensive/magnolia/issues/171

One solution might be using our fork for 2.11, and latest upstream for 2.12, and drop 2.11 when we move to 2.13.

nevillelyh:


build.sbt
```diff
@@ -104,6 +104,16 @@ val beamSDKIODependencies = Def.settings(
   )
 )
 
+val magnoliaDependencies = Def.settings(
+  libraryDependencies ++= Seq(
+    if (scalaBinaryVersion.value == "2.11") {
```
ClaireMcGinty: maybe `scalaBinaryVersion.value <= "2.11"` in preparation for scala 2.13? 😅

nevillelyh: Talked IRL. This branch is for 2.11 only (we no longer support 2.10), and the `else` branch is for >= 2.12.

nevillelyh:
Well upstream no longer publishes magnolia & mercator 2.11 so we'll have to fork both and backport just for 2.11. Seems more work than it's worth.

This is the easy way and we should push more to move to 2.12, and remove 2.11 support once 2.13 is ready.

PR #2244, opened 2019-09-13
Truncate app args if long (#2172)

### LINKED ISSUE #2172: Add a warning / helpful debug message when app arguments exceed limits of PipelineOptions
When using ContextAndArgs constructor of ScioContext, we convert all the AppArgs as a String [here](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L384-L388)

It works well and shows up in the Dataflow / other runner UI.

However if the there are lots of arguments, it hits the 512K limit. For cases where this limit can exceed we can put a check + warning and optionally shorten it? (maybe this is runner specific).

benkogan:
Prevents long app args from contributing to upload bloat, which can e.g. cause Dataflow to reject a job upload for being too large.

I went with a pretty arbitrary number for the limit, but the limits Dataflow declares in its error messages are thus:
> SDK pipeline options or staging file list exceeds size limit. Please keep their length under 256K Bytes each and 512K Bytes in total.

Will add a test but just want to check first that this is what was intended.

Fixes #2172.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -383,13 +383,26 @@ object ScioContext {
     } else {
       val args = Args(appArgs)
       if (appArgs.nonEmpty) {
+        val argString = args.toString("", ", ", "")
+        val sanitizedArgString =
+          if (argString.length > appArgStringMaxLength) {
+            log.warn("Truncating long app arguments")
+            argString.substring(0, appArgStringMaxLength) + " [...]"
+          } else {
+            argString
+          }
+
         pipelineOpts
           .as(classOf[ScioOptions])
-          .setAppArguments(args.toString("", ", ", ""))
+          .setAppArguments(sanitizedArgString)
       }
       (pipelineOpts, args)
     }
   }
+
+  // Used to trim app args for UI if too long to avoid
+  // contributing to an exceeded upload size limit.
+  private val appArgStringMaxLength = 50000
 
   import scala.language.implicitConversions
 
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -383,14 +383,27 @@ object ScioContext {
     } else {
       val args = Args(appArgs)
       if (appArgs.nonEmpty) {
+        val argString = args.toString("", ", ", "")
+        val sanitizedArgString =
+          if (argString.length > appArgStringMaxLength) {
+            log.warn("Truncating long app arguments")
+            argString.substring(0, appArgStringMaxLength)
```
nevillelyh: Minor but I'd add a suffix like `"..."` to indicate that it'd been truncated.

nevillelyh:
Yeah might be a bit overkill, especially given the current limit 50K. I doubt many will look at the string closely, especially the end of it.

sisidra:
Just another suggestion: you might want to add a checksum from the full args list to the truncated string, so users can easily verify and match which command is running which job.

Idea stolen from Luigi ;)

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
regadas: yeah +1 on this!

benkogan:
Thanks for the feedback, all. I've updated with a truncation indicator (meant to add this originally but forgot!).

@sisidra I'm curious what the use-case is for the checksum. If you're matching the checksum against e.g. one logged by the job, wouldn't it be just as easy to match the job ID to that logged by the driver program when run?

PR #2256, opened 2019-09-20
Allow user to pass in Bigtable GcRule

rculbertson:
Bigtable's GC policy can be configured to use the max age of a cell, the max
number of versions in a cell, or a combination of the two. Currently scio's
API only lets you specify max age. This PR makes it so the user can pass in
a GcRule, so they can configure GC however they like.

### DIFF
--- a/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
+++ b/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
@@ -266,7 +266,6 @@ package object bigtable {
      *                                              garbage collection of a cell may occur.
      *                                              Note: minimum granularity is second.
      */
-    @deprecated("this method will be removed; use ensureTablesWithGcRules instead", "Scio 0.8")
     def ensureTablesWithExpiration(
       bigtableOptions: BigtableOptions,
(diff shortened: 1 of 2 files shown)


rculbertson:
Yea, totally fine with not deprecating. Having a convenience method for max age, but not max versions, felt a little arbitrary....but maybe max age being more common is a good reason to have it.

Happy to go with whatever you all decide.

nevillelyh:
👍 on keeping both.

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -234,6 +235,7 @@ package object bigtable {
      *                                              garbage collection of a cell may occur.
      *                                              Note: minimum granularity is second.
      */
+    @deprecated("this method will be removed; use ensureTablesWithGcRules instead", "Scio 0.8")
```
nevillelyh:
```suggestion
```

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/package.scala
```diff
@@ -265,6 +267,7 @@ package object bigtable {
...
```
nevillelyh:
```suggestion
```

PR #2267, opened 2019-09-25
AvroType macro shouldn't generate duplicate record classes fix #2264)

### LINKED ISSUE #2264: Avro regenerates referenced definitions
In Scio 0.7.4, `@AvroType.fromSchema` regenerates case classes for previously seen definitions instead of reusing them, which breaks types which have multiple members of the same type.

```scala
@AvroType.fromSchema(
  """
  |{
  |  "type" : "record",
  |  "name" : "A",
  |  "fields" : [ {
  |    "name" : "a",
  |    "type" : {
  |      "type" : "record",
  |      "name" : "B",
  |      "fields" : [ ]
  |    }
  |  }, {
  |    "name" : "b",
  |    "type" : "B"
  |  } ]
  |}""".stripMargin
)
class ReuseDef
```

This example fails with:
```
ReuseDef$B is already defined as case class ReuseDef$B
```

`avro-tool idl2schemata` will generate this case from this IDL input:

```
protocol test {
  record B {}

  record A {
    B a;
    B b;
  }
}
```

ClaireMcGinty:
fix #2264

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/types/TypeProvider.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/types/TypeProvider.scala
@@ -285,6 +285,8 @@ private[types] object TypeProvider {
       val recordClasses = f
         .flatMap(_._2)
         .groupBy(_.asInstanceOf[ClassDef].name)
+        // note that if there are conflicting definitions of a nested record type, the Avro schema
+        // parser itself will catch it before getting to this step.
         .map { case (_, cDefs) => cDefs.head } // Don't generate duplicate case classes
         .toSeq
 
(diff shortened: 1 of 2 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/types/TypeProvider.scala
```diff
@@ -279,7 +279,16 @@ private[types] object TypeProvider {
...
       val f = schema.getFields.asScala
         .map(f => extractField(className, f.name, f.schema))
-      (f.map(_._1), f.flatMap(_._2))
+
+      val fields = f.map(_._1)
+
+      val recordClasses = f
+        .flatMap(_._2)
+        .groupBy(_.asInstanceOf[ClassDef].name)
+        .map { case (_, cDefs) => cDefs.head } // Don't generate duplicate case classes
```
ClaireMcGinty: note that if there are conflicting definitions of a nested record type, the Avro schema parser itself will catch it before getting to this step.

nevillelyh:
```suggestion
        // note that if there are conflicting definitions of a nested record type, the Avro schema
        // parser itself will catch it before getting to this step.
        .map { case (_, cDefs) => cDefs.head } // Don't generate duplicate case classes
```

Issue #2272, opened 2019-09-30
Fix contributor experience


nevillelyh:
A couple of things:

- [x] By default IntelliJ uses its own sbt-launcher and tries to build integration tests which will fail due to lack of permission for contributors. We should document that in https://github.com/spotify/scio/blob/master/CONTRIBUTING.md, e.g. use sbt shell for build & tests.
- [x] Add `@BigQueryType.fromStorage` cache in `scio-examples`
- [x] Fix flaky tests, build speed
- [x] Document other prerequisites, e.g. JDK version, etc?
  - [x] JDK version

nevillelyh:
Closing for now. Build has been mostly stable except the JDK 11 one.

PR #2282, opened 2019-10-02
fix BigQuery cache handling when GCP credentials are missing, #2272

nevillelyh:
Turns out we `gen_schemas.sh` set up cache for everything in `scio-examples` already, but `bigquery` instance in `TypeProvider` is not truly lazy even when no BQ requests are needed.

https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala#L48

Adding `-Dbigquery.project=` to `.jvmopts` is the easy solution but may not work for some cases depending on how IntelliJ runs sbt. Alternatively we can make that `bigquery` instance and its usage, i.e. in `BigQueryPartitionUtil` lazier.

cc @psobot

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -8,3 +8,6 @@ SCHEMAS_DIR=/tmp/scio-bigquery-$USER/.bigquery/
 
 mkdir -p $SCHEMAS_DIR
 cp -v scripts/bigquery/* $SCHEMAS_DIR
+
+echo "Adding BigQuery flags"
+echo "-Dbigquery.project=dummy-project" >> .jvmopts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,6 @@ commands:
         keys:
         - v7-dependencies-{{ .Environment.CIRCLE_JOB }}-{{ checksum "/tmp/dependencies.cache.tmp" }}
         - v7-dependencies-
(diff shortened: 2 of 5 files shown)


scripts/bootstrap.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo "Restoring cached BigQuery schemas"
+
+SCHEMAS_DIR=/tmp/scio-bigquery-$USER/.bigquery/
+
+mkdir -p $SCHEMAS_DIR
+cp -v scripts/bigquery/* $SCHEMAS_DIR
+
+echo "Adding BigQuery flags"
+echo "-Dbigquery.project=dummy-project" >> .jvmopts
```
regadas: I like the idea of avoiding setting the sys prop when running sbt. The downside of this is that contributors running this script will have `.jvmopts` staged for commit and it can lead to unwanted changes.

Maybe we need to rethink the default value of `bigquery.project` and have a red warn if this setting is used; or try to make it really lazy as you pointed out.

nevillelyh: What about this:
We move these examples that require GCP creds to a separate source dir, e.g. `/src/{main,test}/scala-gcp`, and conditionally include them in sbt if `bigquery.project` is set?

nevillelyh:
Closing this in favor of #2284.

PR #2284, opened 2019-10-03
fix typed BQ examples build, #2272

nevillelyh:


nevillelyh:
FYI @psobot this addresses the BQ storage compilation issue.

Issue #2101, opened 2019-07-30
SparkeySideInput with cache support


nevillelyh:
Sparkey only supports string or bytes key/value types. In some cases, decoding value bytes can be expensive especially when dealing with hot keys, and it's desirable to have a on-heap cache, i.e. [caffeine](https://github.com/ben-manes/caffeine), to cache decoded objects.

We should abstract that away to simplify complex issues like memory sharing, threading, serialization.

nevillelyh:
@iyahimovitz any updates? @O-OF-N is also interested in this. Would be great to see a branch/PR so we can discuss over there.

ClaireMcGinty:
talked to @iyahimovitz, I'm un-assigning this ticket since he doesn't have bandwidth to work on it right now. if you could push your WIP code to a branch that would be great!

Issue #1967, opened 2019-06-11
Make SideInput extendable with wrapping


sisidra:
It would be nice to support delegating/forwarding pattern in SideInput API for easy extensions?
Currently it is heavily sealed with many member being private.

For example, I want to create a `MySideInput` that reuses `SparkeySideInput`:
```
package com.spotify.scio.values  // Dirty workaround for problematic private members...

class MyLookupSideInput(val delegate: SideInput[SparkeyReader]) extends SideInput[MyLookup] {
  override val view: PCollectionView[_] = delegate.view  // This is problematic because private

  override def get[I, O](context: DoFn[I, O]#ProcessContext): MyLookup = {
    val reader: SparkeyReader = delegate.get(context)  // This is problematic because private
    new MyLookup(reader)
  }
 }
```

I can quickly imagine 2 solutions:
* Remove `private` modifiers from `get` and `view`, as class it is not exposed to pipeline anyway, it should be quite safe to loosen those.
* Provide `WrappingSideInput` for this one specific use case that users can extend.
* or both?

nevillelyh:
Maybe something like a `SideInput[A]#def map(f: A => B): DelegateSideInput[B]`?

sisidra:
Something like that might be an option.

nevillelyh:
Someone else brought up that if `f: A => B` is expensive, or in some cases `A` or `B` itself is a lookup table, e.g. Sparkey where we want to deserialize value bytes to protobuf, then it's desirable to have a on-heap cache to save CPU cycles.

Sounds like 2 related issues?

nevillelyh:
@iyahimovitz any updates? @O-OF-N is also interested in this. Would be great to see a branch/PR so we can discuss over there.

nevillelyh:
Another note, the abstraction should support state for the purpose of caching, see #2101.

PR #2287, opened 2019-10-04
Add DelegatingSideInput and TypedSparkeySideInput.

psobot:
Should solve #1967 and #2101.

This PR adds:
 - a `DelegatingSideInput[A, B]` class, which allows users to do `SideInput[A]#map(f: A => B): DelegatingSideInput[A, B]`
 - a `TypedSparkeyReader[T]` with a `decoder: Array[Byte] => T` and an optional `cache: Cache[String, T]`
 - a `CachedStringSparkeyReader`, with a mandatory `cache: Cache[String, String]`
 - a `MockCache` Java class that implements `caffeine.Cache` for use in tests, to ensure the cache is used as expected.
 - tests for all of the above.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -47,6 +47,11 @@ trait SideInput[T] extends Serializable {
     cache
   }
 
+  /**
+   * Create a new [[SideInput]] by applying a wrapping this SideInput with a mapper.
+   */
+  def map[B](f: T => B): SideInput[B] = new DelegatingSideInput[T, B](this, f)
+
   private[values] val view: PCollectionView[_]
 }
 
@@ -118,6 +123,13 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K, 
     extends SideInput[Map[K, Iterable[V]]] {
   override def get[I, O](context: DoFn[I, O]#ProcessContext): Map[K, Iterable[V]] =
     JMapWrapper.ofMultiMap(context.sideInput(view))
+}
+
+private[scio] class DelegatingSideInput[A, B](val si: SideInput[A], val mapper: A => B)
+    extends SideInput[B] {
+  override def get[I, O](context: DoFn[I, O]#ProcessContext): B = mapper(si.get(context))
+
+  val view: PCollectionView[_] = si.view
 }
 
 /** Encapsulate context of one or more [[SideInput]]s in an [[SCollectionWithSideInput]]. */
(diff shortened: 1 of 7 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -120,6 +125,13 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K,
     JMapWrapper.ofMultiMap(context.sideInput(view))
 }
 
+private[scio] class DelegatingSideInput[A, B](val si: SideInput[A], val mapper: A => B)
+    extends SideInput[B] {
+  override def get[I, O](context: DoFn[I, O]#ProcessContext): B = mapper(si.get(context))
```
nevillelyh: This should match parent trait with `private[values]`?

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -120,6 +125,13 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K,
...
+
+  val view: PCollectionView[_] = si.view
```
nevillelyh: Same for this one, `private[values]`?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -77,21 +79,69 @@ import scala.collection.JavaConverters._
...
+    def sparkeySideInput(basePath: String): SideInput[SparkeyReader] =
+      new SparkeySideInput(viewOf(basePath))
+
+    /**
+     * Create a SideInput of `TypedSparkeyReader` from a [[SparkeyUri]] base path, to be used with
+     * [[com.spotify.scio.values.SCollection.withSideInputs SCollection.withSideInputs]].
+     * The provided decoder function will map from the underlying byte array to a JVM type, and
+     * the optional [[Cache]] object can be used to cache reads in memory after decoding.
+     */
+    def typedSparkeySideInput[T >: Null <: AnyRef](
```
nevillelyh: This is called in `main` so `cache` has to survive `ser/de`? Also before job execution cache should be empty right? What about make it a cache factory method or something like `f: () => Cache[String, T]`? Or even `SparkeyReader => Cache[String, T]`, so the builder can infer cache settings from sparkey header, e.g. num entries, file size.

For user facing APIs I'd also curry the lambda args so one can do `.typedSSI("...") { b => ... }`. You might have to overload the method for the optional `cache` version.

psobot: We could use a cache factory method, but [`caffeine.Cache` itself is serializable](https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.1.0/com/github/benmanes/caffeine/cache/Caffeine.html):

> The caches produced by Caffeine are serializable, and the deserialized caches retain all the configuration properties of the original cache. Note that the serialized form does not include cache contents, but only configuration.

I do like the idea of maybe using the `SparkeyReader` to inform the cache properties, though. 👍

PR #2288, opened 2019-10-04
Wrap PubsubWithAttributes Write in composite PTransform (fix #2286)

### LINKED ISSUE #2286: Cannot control all step names when using saveAsPubsubWithAttributes
I'm writing a streaming job so I need to control all step names to ensure I can perform updates without having to map step names. When using `saveAsPubsubWithAttributes` two steps will be crated and I can only seem to control the first one (which I assume just maps from the `(T, Map[String, String])` tuple to `PubsubMessage`. The second step, which is the real sink is still named based on the source code line.

Example code:

```scala
sc.customInput("ReadPsWireEvent", readPubsubMessages(options))
      .withName("ConvertAnonymize")
      .applyTransform(ParDo.of(new MyDoFn()))
      .withName("WritePsConsumptionEvent")
      .saveAsPubsubWithAttributes[GenericRecord]("MyOutputTopic")
```

This results in the following execution graph:

scio

ClaireMcGinty:
fix #2286 

For a job that looks like: `scoll.parallelize(...).map(...).withName("Custom Pubsub Write Name").saveAsPubsubWithAttributes[Foo](...)`, the resulting top-level graph looks like: 

Screen Shot 2019-10-04 at 11 06 53 AM

And expands into:
Screen Shot 2019-10-04 at 11 07 03 AM

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
@@ -21,12 +21,14 @@ import com.google.protobuf.Message
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.spotify.scio.testing.TestDataManager
-import com.spotify.scio.util.{JMapWrapper, ScioUtil}
+import com.spotify.scio.util.{Functions, JMapWrapper, ScioUtil}
 import com.spotify.scio.values.SCollection
 import org.apache.avro.specific.SpecificRecordBase
-import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessageWithAttributesCoder
 import org.apache.beam.sdk.io.gcp.{pubsub => beam}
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
```diff
@@ -210,13 +212,21 @@ private final case class PubsubIOWithAttributes[T: ClassTag: Coder](
...
     val coder = CoderMaterializer.beam(data.context, Coder[T])
-    data
-      .map { kv =>
-        val payload = CoderUtils.encodeToByteArray(coder, kv._1)
-        val attributes = kv._2.asJava
-        new beam.PubsubMessage(payload, attributes)
-      }
-      .applyInternal(w)
+
+    data.applyInternal(new PTransform[PCollection[WithAttributeMap], PDone]() {
```
nevillelyh: Just curious, does `data.transform` not work for this case?

ClaireMcGinty: no :/ since it's a sink transform that returns a `PDone`, whereas `SCollection#transform` expects something a mapping from `SCollection => SCollection`.

PR #2290, opened 2019-10-07
Kieran/it docs

KieranCTaylor:
Update docs for external users looking to compile the source code. Warns that the default config relies on access to internal Spotify GCP project.


nevillelyh:
Thanks!

PR #2292, opened 2019-10-08
bump beam to 2.16.0

nevillelyh:


nevillelyh:
Possibly because of this behavior change: https://github.com/apache/beam/commit/e3cd3b9f9b

Issue #2294, opened 2019-10-08
SCollectionMatchers.containValue[T] requires T to be Serializable


drobert:
When using the construct `mySColl should inWindow(foo) { containValue[T](bar) }`, the type `T` must be serializable or else the process fails with:
```
unable to serialize DoFnWithExecutionInformation{doFn=org.apache.beam.sdk.testing.PAssert$GroupedValuesCheckerDoFn@6367a688, mainOutputTag=Tag<output>, schemaInformation=DoFnSchemaInformation{elementConverters=[]}}
java.lang.IllegalArgumentException: unable to serialize DoFnWithExecutionInformation{doFn=org.apache.beam.sdk.testing.PAssert$GroupedValuesCheckerDoFn@6367a688, mainOutputTag=Tag<output>, schemaInformation=DoFnSchemaInformation{elementConverters=[]}}
	at org.apache.beam.sdk.util.SerializableUtils.serializeToByteArray(SerializableUtils.java:55)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.translateDoFn(ParDoTranslation.java:563)
...
Caused by: java.io.NotSerializableException: io.chrisdavenport.fuuid.FUUID
```

Here is some code that can reproduce the use case:
```scala

import java.util.UUID

import com.spotify.scio.testing._
import com.spotify.scio.values.WindowOptions
import io.chrisdavenport.fuuid.FUUID
import org.apache.beam.sdk.transforms.windowing.{
  AfterProcessingTime,
  AfterWatermark,
  IntervalWindow,
  Repeatedly
}
import org.apache.beam.sdk.values.TimestampedValue
import org.apache.beam.sdk.values.WindowingStrategy.AccumulationMode
import org.joda.time.{Duration, Instant}

// for holding values while ensuring no closures of test-related context
object SerializableSpec {
  // FUUID is non-serializable
  // class: https://github.com/ChristopherDavenport/fuuid/blob/master/modules/core/src/main/scala/io/chrisdavenport/fuuid/FUUID.scala
  case class Message(id: FUUID, body: String)

  val safeMsg = Message(FUUID.fromUUID(UUID.randomUUID), "one")
}

class SerializableSpec extends PipelineSpec {
  "SCollection[T]" should "not require T to be Serializable" in {
    val inTestMsg =
      SerializableSpec.Message(FUUID.fromUUID(UUID.randomUUID), "one")

    val epoch = new Instant(0)
[...]

PR #2296, opened 2019-10-08
fix lambda ser/de in SCollectionMatchers, #2294

nevillelyh:


scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
```diff
@@ -362,3 +362,10 @@ trait SCollectionMatchers {
   }
 
 }
+
+/** Helper class to ensure test data survives lambda serialization. */
+private class Externalizer[T: Coder](sc: ScioContext, original: T) extends Serializable {
```
nevillelyh: This should probably be public? Could be useful for user code lambda see/de cases.

ClaireMcGinty: just wondering, since this implements Serializable, would it work to just implement the `writeObject`/`readObject` methods of Serializable?

nevillelyh: Talked IRL. Will get this in for now and do a separate PR for a public API.

PR #2300, opened 2019-10-10
Segregate and deprecate avro reflective records support

jto:
This PR is an attempt to discourage the usage of reflective serialization in Avro and avoid its accidental use.

It separates "normal" IOs on `GenericRecord` and avro magic reflection based serialization. Back compatibility is preserved but a deprecation warning is issued.

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
@@ -32,6 +32,30 @@ import scala.language.implicitConversions
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
 
+sealed trait AvroFileImpl[T] {
+  def apply(self: ScioContext)(path: String, schema: Schema): SCollection[T]
+}
+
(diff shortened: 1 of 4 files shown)


jto:
Yes

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
```diff
@@ -32,6 +32,30 @@ import scala.language.implicitConversions
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
 
+sealed trait AvroFileImpl[T] {
+  def apply(self: ScioContext)(path: String, schema: Schema): SCollection[T]
+}
+
+sealed trait LowPriorityAvroFileImpl {
```
regadas: Hi @jto seems that this is not used, `defaultImpl` will not be in scope; is it intended? or should `AvroFileImpl` extend `LowPriorityAvroFileImpl`?

jto: Oh. Good catch. How are the tests compiling ? oO

jto:
@nevillelyh Can we merge this ?

PR #2316, opened 2019-10-23
Fix annoy writer

regadas:


jto:
I have basically zero experience with Annoy. Maybe @nevillelyh should review this :)

PR #2317, opened 2019-10-23
fix NPE in scio-bigtable's ChannelPoolCreator when credentials aren't set

ClaireMcGinty:
Noticed this while running a Bigtable write using the emulator tool, which doesn't set credentials. The Bigtable client lib will just return a null `ClientInterceptor` in that case (which is still valid; the Netty channel creator takes in 0 or more `ClientInterceptor` args), which we weren't handling properly.


scio-bigtable/src/main/java/com/spotify/scio/bigtable/ChannelPoolCreator.java
```diff
@@ -28,19 +28,27 @@
 
 public class ChannelPoolCreator {
   public static final BigtableOptions options = new BigtableOptions.Builder().build();
```
regadas: we could probably make this one private as well

PR #2336, opened 2019-10-25
Add sharded Sparkey support.

psobot:
This PR adds support for sharded Sparkey files ("sharkey," anyone?) which allows for parallelization of the writing of the files. Why do this? Currently, Sparkey files are single-threaded-on-write, meaning we essentially have to do a `GroupByKey` and then block on a single Sparkey writer. For larger Sparkey files (i.e.: 1-100GB), this step can take a huge amount of wall clock time, delaying pipelines.

The approach taken here is as follows:
 - The pipeline can specify the number of shards to use when writing a Sparkey file. By default, the behaviour does not change and the number of shards is 1 - except when using the `.asSparkeySideInput` syntactic sugar, in which case the default number of shards is 64.
 - When writing a sharded Sparkey collection, the key-value pairs have their keys hashed via [MurmurHash](https://en.wikipedia.org/wiki/MurmurHash) and grouped by that hash value. Each group of keys is then writable in parallel to a separate Sparkey file pair, named `part-%05d-of-%05d.sp[il]`.
 - To read the resulting Sparkey collection, a `ShardedSparkeyReader` has been added, which uses the same fast MurmurHash before looking up each key, and combines the iterators of the shards to still allow for unordered iteration over the entire collection.

The performance benefits are pretty clear: on a test dataset of 1 billion elements being written to a 40GB Sparkey file, things look optimal around 64 shards. Sparkey creation is about **12x** faster and uses **6x** less vCPU time, saving money.

image

image


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -0,0 +1,162 @@
...
+        path =>
+          (
+            ShardedSparkeyUri.shardIndexFromPath(path),
+            new ThreadLocalSparkeyReader(new File(path + ".spi"))
+          )
+      )
+      .toMap
+  }
+
+  implicit def coderSparkeyURI: Coder[sparkey.ShardedSparkeyUri] =
```
ClaireMcGinty: just wondering -- could you convert `LocalShardedSparkeyUri` and `RemoteShardedSparkeyUri` to case classes and let Magnolia can derive a coder for them? Looks like all the class members are serializable (unlike the existing `RemoteSparkeyUri` since it stores a `PipelineOptions` and tbh we could probably fix that)

psobot: Great idea! I've just updated this PR to make all of the `SparkeyUri` subclasses into case classes, and I've removed the implicit coders.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  private[sparkey] def exists: Boolean
+  override def toString: String = basePath
+
+  def basePathForShard(shardIndex: Short, numShards: Short): String =
+    f"$basePath/part-$shardIndex%05d-of-$numShards%05d"
+
+  def sparkeyUriForShard(shardIndex: Short, numShards: Short): SparkeyUri
+
+  lazy val globExpression = s"$basePath/part-*"
```
regadas: Hi @psobot seems that this can be a normal `val`?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.beam.sdk.options.PipelineOptions
+
+import scala.collection.JavaConverters._
+
+/**
+ * Sharded Sparkey support
+ */
+trait ShardedSparkeyUri extends SparkeyUri {
+  val basePath: String
+  def getReader: ShardedSparkeyReader
```
regadas: 🎨 `reader`? 😄

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -0,0 +1,157 @@
...
+  private[sparkey] def basePathsAndCount(): (Seq[String], Short) = {
+    val matchResult: MatchResult = FileSystems.`match`(globExpression)
+    val paths = matchResult.metadata().asScala.map(_.resourceId.toString)
+    val indexPaths = paths.filter(_.endsWith(".spi")).sorted
+
+    val allStartParts = indexPaths.map(ShardedSparkeyUri.shardIndexFromPath)
+    val allEndParts = indexPaths.map(ShardedSparkeyUri.numShardsFromPath)
+
+    val distinctNumShards = allEndParts.toSet
+    if (distinctNumShards.isEmpty) {
```
regadas: i would probably have an exception more specific for invalid shards

```scala
distinctNumShards match {
 case Nil => (Seq.empty[String], 0)
 case distinctShard :: Nil =>
  ....
 case _ =>
  throw new InvalidShards
```

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -0,0 +1,157 @@
...
+    path.split("-of-").toList.last.split("\\.").head.toShort
+
+  private[sparkey] def shardIndexFromPath(path: String): Short =
+    path.split("part-").toList.last.split("-of-").head.toShort
+
+  private[sparkey] def localReadersByShard(
+    localBasePaths: Iterable[String]
+  ): Map[Short, SparkeyReader] = {
+    localBasePaths
+      .map(
```
regadas: 🎨 I think that in this case `{}` is better suited

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  private[sparkey] def numShardsFromPath(path: String): Short =
+    path.split("-of-").toList.last.split("\\.").head.toShort
+
+  private[sparkey] def shardIndexFromPath(path: String): Short =
+    path.split("part-").toList.last.split("-of-").head.toShort
+
+  private[sparkey] def localReadersByShard(
+    localBasePaths: Iterable[String]
+  ): Map[Short, SparkeyReader] = {
```
regadas: 🎨 no need for outer `{}`

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
regadas: also not that it's huge deal here but `localBasePaths.iterator.map` to avoid extra copying

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -39,32 +38,32 @@ import scala.collection.JavaConverters._
...
     new ThreadLocalSparkeyReader(new File(basePath))
   override private[sparkey] def exists: Boolean =
     SparkeyUri.extensions.map(e => new File(basePath + e)).exists(_.exists)
 }
 
-private class RemoteSparkeyUri(val basePath: String, options: PipelineOptions) extends SparkeyUri {
+private case class RemoteSparkeyUri(basePath: String, rfu: RemoteFileUtil) extends SparkeyUri {
 
-  val rfu: RemoteFileUtil = RemoteFileUtil.create(options)
+  def this(basePath: String, options: PipelineOptions) =
```
regadas: since this is a `case class` i would prefer this to be moved to a companion object `apply(basePath: String, options: PipelineOptions)`

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -145,6 +171,24 @@ package object sparkey {
       sparkeySideInput(basePath).map(reader => new CachedStringSparkeyReader(reader, cache))
   }
 
+  private val DEFAULT_NUM_SHARDS: Short = 1
```
regadas: 🎨 we are starting to adopt `DefaultNumShards` for constants instead of the java'ish way

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
psobot: Whoops, this overrides `getReader` from `SparkeyUri` - I'm going to leave this as `getReader` for now and add `override` to be explicit.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
psobot: Thanks for keeping me away from the dreaded `if`. 😆

marklaw106:
win!

PR #2340, opened 2019-10-28
Add direct dependencies

regadas:
This PR adds all the dependencies that we use directly in our code; the added libs are mostly the ones we were using transitively from `beam` libs and versions are kept consistent with `beam`;

The purpose of this is to 1) avoid conflicts with transitive deps 2) make clear what we are really using

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -82,6 +82,10 @@ val generatedGrpcGaVersion = "1.43.0"
 val generatedGrpcBetaVersion = "0.44.0"
 val googleClientsVersion = "1.27.0"
 val googleApiServicesBigQuery = "v2-rev20181104-1.27.0"
+val bigdataossVersion = "1.9.16"
+val gaxVersion = "1.38.0"
+val googleAuthVersion = "0.12.0"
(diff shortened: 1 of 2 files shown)


nevillelyh:
How would version management work though if a Beam version bump changed some of these transitive deps? Without an automated way to audit it'll be more work to keep them consistent?

regadas:
All the added dependencies are our direct dependencies and `Beam`'s direct dependencies.
 
This is actually "forcing" us to be a bit more mindful of what we actually use but it will indeed require a bit more carefulness when bumping `Beam` version (we kinda do a bit implicitly anyways for `grpc-*` and friends). Given the state of `Beam` dependencies and it's transitive dependencies knowing what we use and making sure we use the same dependencies across `scio-*` (which we don't 😢 ) is a must.

Automatic way to audit is actually doable and a bit easier since we have all the direct deps declared now. I think we can target it in a next PR?

regadas:
Added as part of this PR an sbt task to audit the dependencies.

PR #2342, opened 2019-10-28
Adds tf prediction with default-sig

guidj:
Zoltar nows loads tensorflow models with their signature definitions
Default one is `serving_default` -- currently, the only one supported
This enables consumers to pass in signature names of tensors and fetchOps

e.g. of signature
```
$ saved_model_cli show --dir scio-tensorflow/src/test/resources/trained_model --all

MetaGraphDef with tag-set: 'serve' contains the following SignatureDefs:

signature_def['classification']:
  The given SavedModel SignatureDef contains the following input(s):
    inputs['inputs'] tensor_info:
        dtype: DT_STRING
        shape: (-1)
        name: input_example_tensor:0
  The given SavedModel SignatureDef contains the following output(s):
    outputs['classes'] tensor_info:
        dtype: DT_STRING
        shape: (-1, 3)
        name: linear/head/Tile:0
    outputs['scores'] tensor_info:
        dtype: DT_FLOAT
        shape: (-1, 3)
        name: linear/head/predictions/probabilities:0
  Method name is: tensorflow/serving/classify

signature_def['predict']:
  The given SavedModel SignatureDef contains the following input(s):
    inputs['examples'] tensor_info:
        dtype: DT_STRING
        shape: (-1)
        name: input_example_tensor:0
  The given SavedModel SignatureDef contains the following output(s):
    outputs['class_ids'] tensor_info:
        dtype: DT_INT64
        shape: (-1, 1)
        name: linear/head/predictions/ExpandDims:0
    outputs['classes'] tensor_info:
        dtype: DT_STRING
        shape: (-1, 1)
        name: linear/head/predictions/str_classes:0
    outputs['logits'] tensor_info:
        dtype: DT_FLOAT
        shape: (-1, 3)
        name: linear/linear_model/weighted_sum:0
    outputs['probabilities'] tensor_info:
        dtype: DT_FLOAT
        shape: (-1, 3)
        name: linear/head/predictions/probabilities:0
  Method name is: tensorflow/serving/predict

signature_def['serving_default']:
  The given SavedModel SignatureDef contains the following input(s):
    inputs['inputs'] tensor_info:
        dtype: DT_STRING
        shape: (-1)
        name: input_example_tensor:0
  The given SavedModel SignatureDef contains the following output(s):
    outputs['classes'] tensor_info:
        dtype: DT_STRING
        shape: (-1, 3)
        name: linear/head/Tile:0
    outputs['scores'] tensor_info:
        dtype: DT_FLOAT
[...]

PR #2343, opened 2019-10-29
Tensorflow example predict

martinbomio:
There are two main changes in this PR:
1) Simplify the tensorflow dofns by removing the resources, since they are not needed.
2) Add a predict function that uses the signature definition to get the input and output tensors.

It is similar to #2342 but are not doing the same thing.

### DIFF
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
@@ -20,7 +20,7 @@ package com.spotify.scio.tensorflow
 import java.time.Duration
 
 import com.spotify.zoltar.tf.{TensorFlowLoader, TensorFlowModel}
-import com.spotify.zoltar.{Model, Models}
+import com.spotify.zoltar.Model
 import org.apache.beam.sdk.transforms.DoFn
 import org.apache.beam.sdk.transforms.DoFn.{ProcessElement, Teardown}
 import org.slf4j.LoggerFactory
@@ -45,7 +45,7 @@ sealed trait PredictDoFn[T, V, M <: Model[_]] extends DoFnWithResource[T, V, M] 
 
   override def createResource(): M
 
-  override def getResourceType: DoFnWithResource.ResourceType = ResourceType.PER_CLASS
+  override def getResourceType: DoFnWithResource.ResourceType = ResourceType.PER_INSTANCE
 
   /**
    * Process an element asynchronously.
@@ -84,16 +84,15 @@ sealed trait PredictDoFn[T, V, M <: Model[_]] extends DoFnWithResource[T, V, M] 
 
 private[tensorflow] abstract class SavedBundlePredictDoFn[T, V](
   uri: String,
+  signatureName: String,
   options: TensorFlowModel.Options
 ) extends PredictDoFn[T, V, TensorFlowModel] {
   @transient private lazy val log = LoggerFactory.getLogger(this.getClass)
 
-  override def createResource(): TensorFlowModel =
-    Models
-      .tensorFlow(uri, options)
-      .get(Duration.ofDays(Integer.MAX_VALUE))
-
   override def withRunner(f: Session#Runner => V): V = f(getResource.instance().session().runner())
+
+  def id(): String =
+    (Seq("tf", uri, signatureName) ++ options.tags.asScala.mkString(";")).mkString(":")
 
   @Teardown
   def teardown(): Unit = {
@@ -105,41 +104,85 @@ private[tensorflow] abstract class SavedBundlePredictDoFn[T, V](
 object SavedBundlePredictDoFn {
   def forInput[T, V](
     uri: String,
+    fetchOps: Option[Seq[String]],
     options: TensorFlowModel.Options,
-    fetchOp: Seq[String],
+    signatureName: String,
     inFn: T => Map[String, Tensor[_]],
     outFn: (T, Map[String, Tensor[_]]) => V
-  ): SavedBundlePredictDoFn[T, V] = new SavedBundlePredictDoFn[T, V](uri, options) {
-    override def extractInput(input: T): Map[String, Tensor[_]] = inFn(input)
+  ): SavedBundlePredictDoFn[T, V] = new SavedBundlePredictDoFn[T, V](uri, signatureName, options) {
+    var exportedFetchOps: Map[String, String] = _
 
-    override def extractOutput(input: T, out: Map[String, Tensor[_]]): V = outFn(input, out)
+    private def requestedFetchOps: Map[String, String] = {
+      fetchOps
+        .map { tensorIds =>
+          tensorIds.map { tensorId =>
+            tensorId -> exportedFetchOps(tensorId)
+          }.toMap
+        }
+        .getOrElse(exportedFetchOps)
+    }
 
-    override def outputTensorNames: Seq[String] = fetchOp
+    override def createResource(): TensorFlowModel = {
+      val model = TensorFlowLoader
+        .create(Id.create(id), uri, options, signatureName)
+        .get(Duration.ofDays(Integer.MAX_VALUE))
+      // by doing it here we make sure we only do it once
+      exportedFetchOps = model.outputsNameMap().asScala.toMap
 
+      model
+    }
+
+    override def extractInput(input: T): Map[String, Tensor[_]] = {
+      val extractedInput = inFn(input)
+      extractedInput.map {
+        case (tensorId, tensor) =>
+          getResource.inputsNameMap().get(tensorId) -> tensor
+      }.toMap
+    }
+
+    override def extractOutput(input: T, out: Map[String, Tensor[_]]): V =
+      outFn(input, requestedFetchOps.mapValues(out(_)))
+
+    override def outputTensorNames: Seq[String] = requestedFetchOps.values.toSeq
   }
 
+  /**
+   * Note: if fetchOps isn't provided, then all outputs defined in the signature of the model
+   * are retrieved. This can be expensive and unwanted, depending on the model.
+   */
   def forTensorFlowExample[T, V](
     uri: String,
     exampleTensorName: String,
-    signatureName: String,
+    fetchOps: Option[Seq[String]],
     options: TensorFlowModel.Options,
+    signatureName: String,
     outFn: (T, Map[String, Tensor[_]]) => V
   )(implicit ev: T <:< Example): SavedBundlePredictDoFn[T, V] =
-    new SavedBundlePredictDoFn[T, V](uri, options) {
+    new SavedBundlePredictDoFn[T, V](uri, signatureName, options) {
+      var exportedFetchOps: Map[String, String] = _
 
-      var fetchOps: Map[String, String] = _
+      /* tensorId refers to the public/export name of the tensor */
+      private def requestedFetchOps: Map[String, String] = {
+        fetchOps
+          .map { tensorIds =>
+            tensorIds.map { tensorId =>
+              tensorId -> exportedFetchOps(tensorId)
+            }.toMap
+          }
+          .getOrElse(exportedFetchOps)
+      }
 
       override def createResource(): TensorFlowModel = {
         val model = TensorFlowLoader
-          .create(Id.create("tensorflow"), uri, options, signatureName)
+          .create(Id.create(id), uri, options, signatureName)
           .get(Duration.ofDays(Integer.MAX_VALUE))
         // by doing it here we make sure we only do it once
-        fetchOps = model.outputsNameMap().asScala.toMap
+        exportedFetchOps = model.outputsNameMap().asScala.toMap
 
         model
       }
 
-      override def outputTensorNames: Seq[String] = fetchOps.values.toSeq
+      override def outputTensorNames: Seq[String] = requestedFetchOps.values.toSeq
 
       override def extractInput(input: T): Map[String, Tensor[_]] = {
         val i = getResource.inputsNameMap().get(exampleTensorName)
@@ -147,6 +190,6 @@ object SavedBundlePredictDoFn {
       }
 
       override def extractOutput(input: T, out: Map[String, Tensor[_]]): V =
-        outFn(input, fetchOps.mapValues(out(_)))
+        outFn(input, requestedFetchOps.mapValues(out(_)))
     }
 }
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
@@ -22,7 +22,7 @@ import java.time.Duration
 import com.spotify.zoltar.tf.{TensorFlowLoader, TensorFlowModel}
 import com.spotify.zoltar.{Model, Models}
 import org.apache.beam.sdk.transforms.DoFn
-import org.apache.beam.sdk.transforms.DoFn.{ProcessElement, Setup, Teardown}
+import org.apache.beam.sdk.transforms.DoFn.{ProcessElement, Teardown}
 import org.slf4j.LoggerFactory
 import org.tensorflow._
 import org.tensorflow.example.Example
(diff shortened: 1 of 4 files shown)


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -75,40 +73,70 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
-        new Function[String, TensorFlowModel] {
-          override def apply(t: String): TensorFlowModel =
-            Models
-              .tensorFlow(uri, options)
-              .get(Duration.ofDays(Integer.MAX_VALUE))
-        }
-      )
+  @Setup
+  def setup(): Unit =
+    model = Models
```
regadas: @martinbomio hm, not sure about this;`DoFnWithResource` was used here to avoid loading the model per **instance**; each worker will create N instances of this `DoFn`. This might cause some memory issues if big models get loaded.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -18,31 +18,29 @@
...
 import scala.collection.JavaConverters._
 
-private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
-  fetchOp: Seq[String],
-  inFn: T => Map[String, Tensor[_]],
-  outFn: (T, Map[String, Tensor[_]]) => V
-) extends DoFnWithResource[T, V, ConcurrentMap[String, M]] {
+import com.spotify.zoltar.Model.Id
+
+private[this] abstract class PredictDoFn[T, V, M <: Model[_]] extends DoFn[T, V] {
```
regadas: I like the idea of refactoring this by moving the arguments to methods;

we can make this a `sealed trait`

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -51,16 +49,16 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
     val result = withRunner { runner =>
       val input = c.element()
-      val i = inFn(input)
+      val i = extractInput(input)
       var result: V = null.asInstanceOf[V]
 
       try {
         i.foreach { case (op, t) => runner.feed(op, t) }
-        fetchOp.foreach(runner.fetch)
+        outputTensorNames.foreach(runner.fetch)
```
regadas: some comments made on #2342 apply here

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -51,16 +49,16 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
         val outTensors = runner.run()
         try {
           import scala.collection.breakOut
-          result = outFn(input, (fetchOp zip outTensors.asScala)(breakOut))
+          result = extractOutput(input, (outputTensorNames zip outTensors.asScala)(breakOut))
```
regadas: same here #2342

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
martinbomio: hmm, ok makes sense. I wanted to simplify the resource logic a bit, I believe we are incorrectly using a Map as a resource, since the `DoFn` only support a single model. I will bring back the `DoFnWithResource `

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
regadas: rigth in this case here is not needed; the thing is that the other PR was doing extra processing on the input; 👍

martinbomio:
@regadas I think this one is ready for another look 👀

guidj:
@martinbomio I still need to add the changes from my side

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -18,31 +18,34 @@
...
-    ResourceType.PER_CLASS
+  def extractInput(input: T): Map[String, Tensor[_]]
+
+  def extractOutput(input: T, out: Map[String, Tensor[_]]): V
+
+  def outputTensorNames: Seq[String]
+
+  override def createResource(): M
+
+  override def getResourceType: DoFnWithResource.ResourceType = ResourceType.PER_INSTANCE
```
guidj: Had to make this run per instance, as we don't have control over resource identification
Without this, whichever model is loaded first in the JVM is always reused for the same class -- so a new instance with a different signature-def won't be loaded

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -75,38 +80,116 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
-              .tensorFlow(uri, options)
-              .get(Duration.ofDays(Integer.MAX_VALUE))
-        }
-      )
+  override def withRunner(f: Session#Runner => V): V = f(getResource.instance().session().runner())
 
-    f(model.instance().session().runner())
-  }
+  def id(): String =
+    (Seq("tf", uri, signatureName) ++ options.tags.asScala.mkString(";")).mkString(":")
```
guidj: This isn't particularly useful, since the name is only internal to the loaded model

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/syntax/SCollectionSyntax.scala
```diff
@@ -54,15 +55,52 @@ final class PredictSCollectionOps[T: ClassTag](private val self: SCollection[T])
    */
   def predict[V: Coder, W](
     savedModelUri: String,
-    fetchOps: Seq[String],
-    options: TensorFlowModel.Options
+    options: TensorFlowModel.Options,
+    fetchOps: Option[Seq[String]] = None,
```
guidj: Note: API breaking change

guidj: An alternative would be to still accept a seq from callers, and check if it's empty to return all ops; but I think that would be a strange behavior

guidj:
@regadas ready

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/syntax/SCollectionSyntax.scala
guidj: Thinking about the issue with memory due to multiple instances on the same JVM... won't there one instance + clones made for each machine? If so, than this should be ok because only one model is loaded per machine-instance. I'm not very familiar with how that works

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
martinbomio: hmm, ok, I think this might actually be an issue I introduced in my changes. Before the resource was a concurrent map of models `Map<String, Model>`, I believe this solves (1) and (2) ]

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -105,48 +104,92 @@ private[tensorflow] abstract class SavedBundlePredictDoFn[T, V](
...
     uri: String,
+    fetchOps: Option[Seq[String]],
     options: TensorFlowModel.Options,
-    fetchOp: Seq[String],
+    signatureName: String,
     inFn: T => Map[String, Tensor[_]],
     outFn: (T, Map[String, Tensor[_]]) => V
-  ): SavedBundlePredictDoFn[T, V] = new SavedBundlePredictDoFn[T, V](uri, options) {
-    override def extractInput(input: T): Map[String, Tensor[_]] = inFn(input)
+  ): SavedBundlePredictDoFn[T, V] = new SavedBundlePredictDoFn[T, V](uri, signatureName, options) {
```
martinbomio: @guidj this is a breaking change. I believe a new function should be added to preserve backward compatibility

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -105,48 +104,92 @@ private[tensorflow] abstract class SavedBundlePredictDoFn[T, V](
...
     outFn: (T, Map[String, Tensor[_]]) => V
   )(implicit ev: T <:< Example): SavedBundlePredictDoFn[T, V] =
-    new SavedBundlePredictDoFn[T, V](uri, options) {
-
-      var fetchOps: Map[String, String] = _
+    new SavedBundlePredictDoFn[T, V](uri, signatureName, options) {
+      var exportedFetchOps: Map[String, String] = _
+
+      /* tensorId refers to the public/export name of the tensor */
+      private def requestedFetchOps: Map[String, String] = {
```
martinbomio: this will be executed for every element. probably a good idea to do it only on setup. The `exportedFetchOps ` are not really used outside of this function, so maybe keep a reference to this map instead of that one?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
guidj: Right, I see

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
guidj: Yes, it would. I'm still curious if the multiple functions spawn on the same worker are clones or not. The resource method specifies 3 options, and talks about clone functions. So I'm curious as to when they would spawn up. If it works the way I suspect it does, then loading per instance shouldn't be an issue

martinbomio: @regadas do you know about this?

martinbomio: bumping this up @regadas ^^

regadas:
@martinbomio @guidj Thanks for this; I'm gonna have a look later today.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -75,38 +80,133 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
+          }.toMap
+        }
+        .getOrElse(exportedFetchOps)
+
+      model
+    }
+
+    override def extractInput(input: T): Map[String, Tensor[_]] = {
+      val extractedInput = inFn(input)
+      extractedInput.map {
```
regadas: `.iterator`

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -75,38 +80,133 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
+        case (tensorId, tensor) =>
+          getResource.inputsNameMap().get(tensorId) -> tensor
+      }.toMap
+    }
+
+    override def extractOutput(input: T, out: Map[String, Tensor[_]]): V =
+      outFn(input, requestedFetchOps.mapValues(out(_)))
```
regadas: These days I'm up for avoid `mapValues`. can we `.iterator.map()` instead?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -75,38 +80,133 @@ private[this] abstract class PredictDoFn[T, V, M <: Model[_]](
...
+
+      override def createResource(): TensorFlowModel = {
+        val model = TensorFlowLoader
+          .create(Id.create(id), uri, options, signatureName)
+          .get(Duration.ofDays(Integer.MAX_VALUE))
+        // by doing it here we make sure we only do it once
+        val exportedFetchOps = model.outputsNameMap().asScala.toMap
+        requestedFetchOps = fetchOps
+          .map { tensorIds =>
+            tensorIds.map { tensorId =>
```
regadas: same thing here

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
regadas: This is bad performance wise ... Why can't we keep the `Map<ModelId, Model>` as resource? `ModelId` can contain be based on the model signture def.

Regarding the `DoFn` `TearDown` to be honest is a bit unnecessary in this case; Models are never hot swapped or anything like that. They will be pretty much closed when pipeline is shutting down or being updated.

guidj: I revered the resource to map. I'm not sure how clean-up will be dealt with though, since the model is loaded in tensorflow -- I guess if the machine/container gets teared down that cleanup might happen automagically, but it might leave behind some garbage/memory leaks

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/syntax/SCollectionSyntax.scala
```diff
@@ -51,18 +52,86 @@ final class PredictSCollectionOps[T: ClassTag](private val self: SCollection[T])
...
    *                 ownership of the [[org.tensorflow.Tensor Tensor]]s.
+   * @param signatureName  name of [[org.tensorflow.framework.SignatureDef]]s to be used
+   *                       to run the prediction.
    */
   def predict[V: Coder, W](
     savedModelUri: String,
     fetchOps: Seq[String],
-    options: TensorFlowModel.Options
+    options: TensorFlowModel.Options,
+    signatureName: String = "serving_default"
```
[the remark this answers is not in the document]
guidj: I created an a companion object with default values

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
guidj: Changed

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
guidj: Changed

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
guidj: Changed

guidj:
@regadas @martinbomio

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
regadas: The container will GC'ed so it should be ok. I'm actually getting an idea how to be a good citizen here ... but we can explore it in another PR.

regadas:
Thank you @guidj and @martinbomio for this!

PR #2345, opened 2019-10-29
fail early on malformed BigQuery spec

nevillelyh:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -53,7 +53,7 @@ object Table {
     override lazy val spec: String = BigQueryHelpers.toTableSpec(ref)
   }
   final case class Spec(spec: String) extends Table {
-    override lazy val ref: GTableReference = BigQueryHelpers.parseTableSpec(spec)
+    override val ref: GTableReference = BigQueryHelpers.parseTableSpec(spec)
```
regadas: @nevillelyh should we make line 53 eager as well?

nevillelyh: We can but the impl won't throw any exception so it doesn't make a difference anyway?

regadas: yeah it kinda makes that it doesn't throw; let's keep it as is

PR #2364, opened 2019-11-01
More schemas

jto:


scio-core/src/main/scala/com/spotify/scio/schemas/instances/ScalaInstances.scala
```diff
@@ -62,6 +64,44 @@ trait ScalaInstances {
...
+  implicit def listBufferSchema[T](implicit s: Schema[T]): Schema[mutable.ListBuffer[T]] =
+    ArrayType(s, _.toList.asJava, xs => mutable.ListBuffer.apply(xs.asScala: _*))
+
+  implicit def vectorSchema[T](implicit s: Schema[T]): Schema[Vector[T]] =
+    ArrayType(s, _.toList.asJava, _.asScala.toVector)
+
   implicit def mapSchema[K, V](implicit k: Schema[K], v: Schema[V]): Schema[Map[K, V]] =
     MapType(k, v, _.asJava, _.asScala.toMap)
+
+  // TODO: WrappedArray ?
```
regadas: we can add it; we also added a Coder instance for it;

scio-core/src/main/scala/com/spotify/scio/schemas/instances/ScalaInstances.scala
```diff
@@ -62,6 +64,44 @@ trait ScalaInstances {
   implicit def listSchema[T](implicit s: Schema[T]): Schema[List[T]] =
     ArrayType(s, _.asJava, _.asScala.toList)
 
+  implicit def seqSchema[T](implicit s: Schema[T]): Schema[Seq[T]] =
```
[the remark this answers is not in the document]
nevillelyh: Can most of these be abstracted away with `CanBuildFrom`? Although `CBF` is not serializable and we'll have to implement our own fork so maybe not worth it.

scio-core/src/main/scala/com/spotify/scio/schemas/instances/ScalaInstances.scala
```diff
@@ -62,6 +64,44 @@ trait ScalaInstances {
...
+  implicit def iterableSchema[T](implicit s: Schema[T]): Schema[Iterable[T]] =
+    ArrayType(s, _.toList.asJava, _.asScala.toList)
+
+  implicit def arrayBufferSchema[T](implicit s: Schema[T]): Schema[mutable.ArrayBuffer[T]] =
+    ArrayType(s, _.toList.asJava, xs => mutable.ArrayBuffer(xs.asScala: _*))
+
+  implicit def bufferSchema[T](implicit s: Schema[T]): Schema[mutable.Buffer[T]] =
+    ArrayType(s, _.toList.asJava, xs => mutable.Buffer(xs.asScala: _*))
+
+  implicit def setSchema[T](implicit s: Schema[T]): Schema[Set[T]] =
```
nevillelyh: Set <-> array mapping isn't guaranteed to round trip, e.g. in the case of duplicates. Also the array ordering isn't deterministic. I'd prefer not to support sets to avoid potential bugs.

sisidra: I support your concern about not equivalent foundtrip transformation, but I do not share sentiment on not supporting `Set`. :)

As data engineers, it is a huge problem to compare (in tests or in pipelines) semantically Set that is quite often expressed as List.

As a solution I would prefer then save set builder that would throw exception on duplicate elements.

nevillelyh: The testing issue feels orthogonal and can be solved in other ways, like customizable comparator/matcher? Neither is simple but I'd rather not introduce mistakes that affect prod code and hard to remove later.

scio-core/src/main/scala/com/spotify/scio/schemas/instances/ScalaInstances.scala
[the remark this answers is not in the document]
regadas: tbh I don't think worth it

scio-core/src/main/scala/com/spotify/scio/schemas/instances/ScalaInstances.scala
regadas: To clarify the `ArrayType` here is a bit misleading; the underlying representation is actually a `List`.

sisidra: I thought a bit, and I came to conclusion that this implementation absolutely makes sense.
In [Java][1], [Scala][2], even [Python][3] and probably many other languages define conversion from collection to set as well defined operation that loses duplicates.

I don't think it is Scio place to teach (CS101? :) ) developers what Set is or paint the picture that Set is dangerous.

Putting aside case for testing and comparisons there, the set is very old and very useful data structure, that is very useful in the data pipeline code.

[1]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/HashSet.html#%3Cinit%3E(java.util.Collection)
[2]: https://www.scala-lang.org/api/2.12.2/scala/collection/immutable/List.html#toSet[B%3E:A]:scala.collection.immutable.Set[B]
[3]: https://docs.python.org/3.7/library/stdtypes.html#set

jto:
@regadas @nevillelyh do we merge this ?

scio-core/src/main/scala/com/spotify/scio/schemas/instances/ScalaInstances.scala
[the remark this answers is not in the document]
jto: `CBF` is going away in Scala 2.13 so I would rather not.

jto:
Merging since the consensus seems to be that the Set implementation is correct.

PR #2366, opened 2019-11-04
improve the schema compatibility error message

jto:
The message now tell exactly what's wrong. See repl session:

```scala

case class Foo(s: String, i: Option[Int], f: Float)
case class Bar(x: String, i: Int, f: String)

case class Super0(f: Foo)
case class Super1(f: Bar)

To.safe[Super0, Super1]

// Schema are not compatible.
// 
// FROM schema:
// ┌──────────────────────────────────────────┬──────────────────────┬──────────┐
// │ NAME                                     │ TYPE                 │ NULLABLE │
// ├──────────────────────────────────────────┼──────────────────────┼──────────┤
// │ f                                        │ ROW                  │ NO       │
// │ f.s                                      │ STRING               │ NO       │
// │ f.i                                      │ INT32                │ YES      │
// │ f.f                                      │ FLOAT                │ NO       │
// └──────────────────────────────────────────┴──────────────────────┴──────────┘
// TO schema:
// ┌──────────────────────────────────────────┬──────────────────────┬──────────┐
// │ NAME                                     │ TYPE                 │ NULLABLE │
// ├──────────────────────────────────────────┼──────────────────────┼──────────┤
// │ f                                        │ ROW                  │ NO       │
// │ f.x                                      │ STRING               │ NO       │
// │ f.i                                      │ INT32                │ NO       │
// │ f.f                                      │ STRING               │ NO       │
// └──────────────────────────────────────────┴──────────────────────┴──────────┘
// ⨯ Field f.x was not found
// ⨯ Field f.i has incompatible NULLABLE. Got NOT NULLABLE expected: NULLABLE
// ⨯ Field f.f has incompatible types. Got STRING expected: FLOAT
```

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
@@ -35,40 +35,90 @@ object To {
     else getBaseType(log.getBaseType())
   }
 
+  // Position API
+  private case class Location(p: List[String])
+  private case class Positional[T](location: Location, value: T)
+
+  private def merge(p0: Location, p1: Location): Location =
+    Location(p0.p ++ p1.p)
+
+  sealed trait Nullable
+  case object `NOT NULLABLE` extends Nullable
+  case object NULLABLE extends Nullable
+  private object NullableBuilder {
+    def fromBoolean(isNullable: Boolean): Nullable =
+      if (isNullable) NULLABLE else `NOT NULLABLE`
+  }
+
+  // Error API
+  private sealed trait Error
+  private case class NullableError(got: Nullable, expected: Nullable) extends Error
+  private case class TypeError(got: BSchema.FieldType, expected: BSchema.FieldType) extends Error
+  private case object FieldNotFound extends Error
+
+  private type Errors = List[Positional[Error]]
+
   /**
    * Test if Rows with Schema t0 can be safely converted to Rows with Schema t1
    */
-  private def areCompatible(t0: BSchema.FieldType, t1: BSchema.FieldType): Boolean = {
+  private def areCompatible(
+    context: Location
+  )(t0: BSchema.FieldType, t1: BSchema.FieldType): Errors = {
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
```diff
@@ -35,40 +35,91 @@ object To {
     else getBaseType(log.getBaseType())
   }
 
+  // Position API
+  private case class Location(p: List[String])
+  private case class Positional[T](location: Location, value: T)
+
+  private def merge(p0: Location, p1: Location): Location =
+    (p0, p1) match {
+      case (Location(ps0), Location(ps1)) => Location(ps0 ++ ps1)
```
regadas: can't we just do `Location(p0.p ++ p1.p)`?

scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
```diff
@@ -35,40 +35,91 @@ object To {
...
+    }
+
+  private def inContext[T](location: Location, value: Positional[T]) =
```
regadas: since this is used in 1 place I would actually suppress it

scio-core/src/main/scala/com/spotify/scio/schemas/To.scala
```diff
@@ -35,40 +35,91 @@ object To {
...
    * Test if Rows with Schema t0 can be safely converted to Rows with Schema t1
    */
-  private def areCompatible(t0: BSchema.FieldType, t1: BSchema.FieldType): Boolean = {
+  private def areCompatible(
+    context: Location
+  )(t0: BSchema.FieldType, t1: BSchema.FieldType): Errors = {
     (t0.getTypeName, t1.getTypeName, t0.getNullable == t1.getNullable) match {
       case (_, _, false) =>
-        false
+        val expected = if (t1.getNullable()) NULLABLE else `NOT NULLABLE`
```
regadas: move this if to an `apply` method? `Nullable(t1.getNullable())`

PR #2404, opened 2019-11-13
Make recursive coders serializable

### LINKED ISSUE #2269: SCollection of recursive types produces NotSerializableException
I'm new to this project, so please bear with me.

I have a `SCollection` with a type referring to a recursive type, it produces this exception when I `flatMap` over it

```
Caused by: java.io.NotSerializableException: PipelineOptions objects are not serializable and should not be embedded into transforms (did you capture a PipelineOptions object in a field or in an anonymous class?). Instead, if you're using a DoFn, access PipelineOptions at runtime via ProcessContext/StartBundleContext/FinishBundleContext.getPipelineOptions(), or pre-extract necessary fields from PipelineOptions at pipeline construction time.
        at org.apache.beam.sdk.options.ProxyInvocationHandler.writeObject(ProxyInvocationHandler.java:176)
```

I reproduced the error in the test for derivation of recursive encoders, please see the linked PR.

This is the problematic path, if it's not obvious to you:

* `com.spotify.scio.values.PCollectionWrapper.parDo` materializes the coder, [link](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala#L58). This gives a `com.spotify.scio.coders.WrappedBCoder`.
* [`CoderTranslation.toProto`](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/CoderTranslation.java#L82) does not recognize `WrappedBCoder` and to triggers serialization through [`toCustomCoder`](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/CoderTranslation.java#L113)

Please note that this path is not caused by the recursive type, but the recursive types fails to serialize with the error above.

jto:
(partially) Fixes #2269
The coder derivation still needs to happen in a "safe" location bc. Magnolia generates a `lazy val`

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -167,26 +167,31 @@ private final case class DisjunctionCoder[T, Id](
     }
 }
 
-private[scio] final class RefCoder[T](val typeName: String, c: => BCoder[T]) extends BCoder[T] {
-  def value: BCoder[T] = c
+private[scio] final case class RefCoder[T](val typeName: String, var coder: BCoder[T])
+    extends BCoder[T] {
+  def setImpl(_c: BCoder[T]): Unit = coder = _c
 
-  def decode(inStream: InputStream): T = c.decode(inStream)
-  def encode(value: T, outStream: OutputStream): Unit = c.encode(value, outStream)
-  def getCoderArguments(): java.util.List[_ <: BCoder[_]] = c.getCoderArguments()
-  def verifyDeterministic(): Unit = c.verifyDeterministic()
+  private def check[A](t: => A): A = {
+    require(coder != null, s"Coder implementation should not be null in ${this}")
+    t
+  }
 
-  override def consistentWithEquals(): Boolean = c.consistentWithEquals()
-  override def structuralValue(value: T): AnyRef =
+  def decode(inStream: InputStream): T = check { coder.decode(inStream) }
+  def encode(value: T, outStream: OutputStream): Unit = check { coder.encode(value, outStream) }
+  def getCoderArguments(): java.util.List[_ <: BCoder[_]] = check { coder.getCoderArguments() }
+  def verifyDeterministic(): Unit = check { coder.verifyDeterministic() }
+
+  override def consistentWithEquals(): Boolean = check { coder.consistentWithEquals() }
+  override def structuralValue(value: T): AnyRef = check {
     if (consistentWithEquals()) {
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -170,21 +170,45 @@ private final case class DisjunctionCoder[T, Id](
...
     } else {
       c.structuralValue(value)
     }
+  }
+
+  override def hashCode(): Int = check {
+    typeName.hashCode + (41 * c.hashCode())
+  }
+
+  override def equals(obj: Any): Boolean = check {
```
regadas: we could make `RefCoder` `case class` and keep it short?

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -87,7 +88,11 @@ object CoderMaterializer {
         refs
           .get(t)
           .getOrElse {
-            lazy val bc: BCoder[T] = beamImpl(r, o, c, refs + (t -> RefCoder(t, bc)))
+            // #2269: _REALLY_ support recursive ADTs
+            val bc = RefCoder[T](t, null) // placeholder
+            refs.put(t, bc)
+            val actualImpl = beamImpl(r, o, c, refs)
+            bc.setImpl(actualImpl)
```
regadas: couldn't we just do?

```
bc.c = beamImpl(r, o, c, refs)
```

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -170,21 +170,45 @@ private final case class DisjunctionCoder[T, Id](
     }
 }
 
-private[scio] final class RefCoder[T](val typeName: String, c: => BCoder[T]) extends BCoder[T] {
+private[scio] final class RefCoder[T](val typeName: String, var c: BCoder[T]) extends BCoder[T] {
   def value: BCoder[T] = c
```
regadas: since it's a `var` now we could remove this and probably have a better name for `c`?

jto:
@regadas updated :)

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
regadas: @jto you’re in favor of keeping setImpl?

PR #2405, opened 2019-11-13
Add file:linenum only to outer transform

### LINKED ISSUE #2301: Enforce explicit naming of execution steps
To make it as easy as possible to keep streaming jobs updatable it would be great if there was a way to enforce explicit naming of all steps.

For example, there could be a way of configuring the Scio context in such a way that the pipeline will fail startup if a step without an explicit name is encountered, or there could be some simple helper mechanism for writing a test that will fail if there are such steps.

### LINKED ISSUE #2396: Cannot upgrade stateful pipeline using Flink backend
Hi,

I am going to describe our problem:

**Context**

We have a streaming (not batch) pipeline (CustomerPipeline) with the following functionality:

1. Union 2 streams of events from Kafka
2. Stateful transform (keyed state) using applyPerKeyDoFn
3. Sink events to Kinesis

The snippet of the pipeline is: 
```
    sc
      .withName("union")
      .unionAll[(String, InputEvent)](List(customerEvents, addressesEvents))
      .withName("customer-state")
      .mapValues(InputEvent.toState)
      .withName("accumulated")
      .applyPerKeyDoFn(CustomerAccumulation(schemaRegistryUrl))
	  .withName("customer-output")
      .map {
        case (key, state) =>
          val metadata = CustomerMetadata(key, "customer", "1.0-rc1", "user")
          val payload = state
            .into[CustomerPayload]
            .withFieldComputed(
              _.facts,
              _.facts.map(
                f =>
                  f.into[FactOutput]
                    .withFieldComputed(_.createdAt, f => Instant.ofEpochMilli(f.createdAt))
                    .transform
              )
            )
            .transform
          CustomerOutput(metadata, payload)
      }
      .withName("avro-customer-output")
      .map(output => CustomerOutputEncoder(schemaRegistryUrl).encode(output))
      .withName("kinesis")
      .saveAsCustomOutput("kinesis-stream", kinesisEndpoint(kinesisConfig))
    sc.run()
  }
}
```
Technologies/Frameworks

Apache Beam: 2.15.0
Scio: 0.8.0-beta1
Apache Flink: 1.8.2
Scala version: 2.12.10

Execution graph in Apache Flink:

```
Source: customer-topic/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> customer-topic/Remove Kafka Metadata/ParMultiDo(Anonymous) -> customer-topic-input-event/ParMultiDo(anon$6)

Source: addresses-topic/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> addresses-topic/Remove Kafka Metadata/ParMultiDo(Anonymous) -> addresses-topic-tuple/ParMultiDo(anon$6) -> addresses-topic-input-event/filter@{SCollection.scala:352}2/ParDo(Anonymous)/ParMultiDo(Anonymous) -> addresses-topic-input-event/map@{SCollection.scala:352}2/ParMultiDo(anon$6)

customer-state/ParMultiDo(anon$6) -> accumulated/TupleToKv/ParMultiDo(anon$6)

accumulated/applyPerKeyDoFn@{CustomerPipeline.scala:93}2/ParMultiDo(CustomerAccumulation) -> accumulated/KvToTuple/ParMultiDo(anon$6) -> customer-output/ParMultiDo(anon$6) -> avro-customer-output/ParMultiDo(anon$6) -> kinesis-stream/ParDo(KinesisWriter)/ParMultiDo(KinesisWriter)
```

**Problem**

After changing the code in the pipeline (even a trivial change such as adding a logger.info in the CustomerPipeline.scala) and deploying it to Flink (cancel/resume process using savingpoints) an exception is thrown in the job manager:

```
org.apache.flink.client.program.ProgramInvocationException: Could not submit job (JobID: 8e2dddfb824b27fff67c32f4f3b9e178)
.........
Caused by: org.apache.flink.runtime.client.JobSubmissionException: Failed to submit JobGraph.
	at org.apache.flink.client.program.rest.RestClusterClient.lambda$submitJob$8(RestClusterClient.java:388)
.........
Caused by: java.lang.IllegalStateException: Failed to rollback to checkpoint/savepoint s3://customer-storage-my-project/aws/savepoints/savepoint-9339d9-30503f3f5804. Cannot map checkpoint/savepoint state for operator 9a90eb3bccad535e025178bf2b20928e to the new program, because the operator is not available in the new program. If you want to allow to skip this, you can set the --allowNonRestoredState option on the CLI.
	at org.apache.flink.runtime.checkpoint.Checkpoints.loadAndValidateCheckpoint(Checkpoints.java:205)
	at org.apache.flink.runtime.checkpoint.CheckpointCoordinator.restoreSavepoint(
```

And the upgraded pipeline is deployed ignoring checkpoints and savepoints as a new pipeline.

The execution graph in Apache Flink of the pipeline is:

```
Source: customer-topic/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> customer-topic/Remove Kafka Metadata/ParMultiDo(Anonymous) -> customer-topic-input-event/ParMultiDo(anon$6)

Source: addresses-topic/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> addresses-topic/Remove Kafka Metadata/ParMultiDo(Anonymous) -> addresses-topic-tuple/ParMultiDo(anon$6) -> addresses-topic-input-event/filter@{SCollection.scala:352}2/ParDo(Anonymous)/ParMultiDo(Anonymous) -> addresses-topic-input-event/map@{SCollection.scala:352}2/ParMultiDo(anon$6)

customer-state/ParMultiDo(anon$6) -> accumulated/TupleToKv/ParMultiDo(anon$6)

accumulated/applyPerKeyDoFn@{CustomerPipeline.scala:94}2/ParMultiDo(CustomerAccumulation) -> accumulated/KvToTuple/ParMultiDo(anon$6) -> customer-output/ParMultiDo(anon$6) -> avro-customer-output/ParMultiDo(anon$6) -> kinesis-stream/ParDo(KinesisWriter)/ParMultiDo(KinesisWriter)
```

Analysing the the execution graphs of both pipelines, the name/id of one of the operators has a difference
Initial version: ...applyPerKeyDoFn@{CustomerPipeline.scala:93}....
Upgraded version: ...applyPerKeyDoFn@{CustomerPipeline.scala:94}....

It looks like the line numbers in transforms names are causing Flink operators to have new names, hence the state for operators cannot be resolved by Flink during the deployment.
We would be really happy to keep working using Beam/Scio as we have the ambition of deploying this pipelines across different cloud providers and backends (Dataflow in the near future), but due to time constraints if we don't get stateful pipelines to work properly on Flink, we are likely to fallback to Flink api and give up on our ambitions.

Please, can you help us with that?

regadas:
Fixes #2396 
Fixes #2301: composite transform names are now stable; naming the outer transform and using `--stableUniqueNames=ERROR` should produce the desire effect of failing the pipeline submission.

The following example will fail now:

```scala
coll1.withName("A").applyPerKeyDoFn(...)
coll2.withName("A").applyPerKeyDoFn(...)
```

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
@@ -69,7 +69,7 @@ private[scio] object CallSites {
     }
   }
 
-  /** Get current call site name in the form of "method@{file:line}". */
+  /** Get current call site name in the form of "(method, file:line, isNested)". */
   def getCurrentName: (String, String, Boolean) = {
     val stack = Thread.currentThread().getStackTrace.drop(1)
     val firstExtIdx = stack.indexWhere(e => isExternalClass(e.getClassName))
--- a/scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
@@ -21,67 +21,70 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.function.BiFunction
 
 private[scio] object CallSites {
-  private val scioNs = "com.spotify.scio."
-  private val beamNs = "org.apache.beam."
+  private val scioNs = "com.spotify.scio"
+  private val beamNs = "org.apache.beam"
 
   private val methodMap = Map("$plus$plus" -> "++")
-  private val nameCache = new ConcurrentHashMap[String, Int]()
+  private val nameCache = new ConcurrentHashMap[(String, String, Boolean), Int]()
 
   private def isExternalClass(c: String): Boolean =
     // Not in our code base or an interpreter
     (!c.startsWith(scioNs) && !c.startsWith("scala.") && !c.startsWith(beamNs)) ||
-      c.startsWith(scioNs + "examples.") || // unless if it's in examples
-      c.startsWith(scioNs + "values.ClosureTest") || // or this test
+      c.startsWith(s"$scioNs.examples.") || // unless if it's in examples
+      c.startsWith(s"$scioNs.values.NamedTransformTest") || // or this test
+      c.startsWith(s"$scioNs.values.SimpleJob") || // or this test
+      c.startsWith(s"$scioNs.values.ClosureTest") || // or this test
       // or benchmarks/ITs
       (c.startsWith(scioNs) && (c.contains("Benchmark$") || c.contains("IT")))
 
   private def isTransform(e: StackTraceElement): Boolean =
-    e.getClassName == scioNs + "values.SCollectionImpl" && e.getMethodName == "transform"
+    e.getClassName == s"$scioNs.values.SCollectionImpl" && e.getMethodName == "transform"
+
+  private def isPCollectionApply(e: StackTraceElement): Boolean =
+    e.getClassName == s"$beamNs.sdk.values.PCollection" && e.getMethodName == "apply"
 
   def getAppName: String = {
     Thread
       .currentThread()
       .getStackTrace
       .drop(1)
-      .find(e => isExternalClass(e.getClassName))
       .map(_.getClassName)
+      .find(isExternalClass)
       .getOrElse("unknown")
       .split("\\.")
       .last
       .replaceAll("\\$$", "")
   }
 
-  /** Get a unique identifier for the current call site. */
   def getCurrent: String = {
-    val name = getCurrentName
-
-    val n = nameCache.merge(name, 1, new BiFunction[Int, Int, Int] {
+    val (method, location, nested) = getCurrentName
+    val idx = nameCache.merge((method, location, nested), 1, new BiFunction[Int, Int, Int] {
       override def apply(t: Int, u: Int): Int = t + u
     })
-    if (n == 1) name else name + n
+
+    if (nested) {
+      s"$method:$idx"
+    } else {
+      s"$method@{$location}:$idx"
+    }
   }
 
   /** Get current call site name in the form of "method@{file:line}". */
-  def getCurrentName: String = {
-    val stack = new Exception().getStackTrace.drop(1)
+  def getCurrentName: (String, String, Boolean) = {
+    val stack = Thread.currentThread().getStackTrace.drop(1)
+    val firstExtIdx = stack.indexWhere(e => isExternalClass(e.getClassName))
+    val scioMethod = stack(firstExtIdx - 1).getMethodName
+    val method = methodMap.getOrElse(scioMethod, scioMethod)
 
     // find first stack outside of Scio or SDK
-    var pExt = stack.indexWhere(e => isExternalClass(e.getClassName))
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
```diff
@@ -21,67 +21,66 @@ import java.util.concurrent.ConcurrentHashMap
...
-      if (_p > 0) {
-        pExt = _p
-      }
-    }
-
-    val k = stack(pExt - 1).getMethodName
-    val method = methodMap.getOrElse(k, k)
-    val file = stack(pExt).getFileName
-    val line = stack(pExt).getLineNumber
-    s"$method@{$file:$line}"
```
jto: No file /line number in `DisplayData` 😿

regadas: Beam still doesn't support it for composite transforms https://issues.apache.org/jira/browse/BEAM-366

scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
```diff
@@ -21,67 +21,70 @@ import java.util.concurrent.ConcurrentHashMap
...
-    if (n == 1) name else name + n
+
+    if (nested) {
+      s"$method:$idx"
+    } else {
+      s"$method@{$location}:$idx"
+    }
   }
 
   /** Get current call site name in the form of "method@{file:line}". */
```
nevillelyh:
```suggestion
  /** Get current call site name in the form of "(method, file:line, isNested)". */
```

scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala
```diff
@@ -21,67 +21,70 @@ import java.util.concurrent.ConcurrentHashMap
...
-      if (_p > 0) {
-        pExt = _p
-      }
-    }
+    val externalCall = stack(firstExtIdx)
+    val location = s"${externalCall.getFileName}:${externalCall.getLineNumber}"
+    // check if transform is nested
+    val transformIdx = stack.indexWhere(isTransform)
+    val pApplyIdx = stack.indexWhere(isPCollectionApply)
+    val isNested = transformIdx > 0 && pApplyIdx > 0
```
nevillelyh: Is it possible for both to be `>0` and outside (lower in value) of `firstExtIdx`? I mean user code can also call `.transform { _.internal.apply(...) }`?

regadas: hmm, no, `.transform` will have a `PCollection.apply` when expanding. So `.transform` stack idx will always be > then `apply`. in your case we will have at least 2 `apply` calls being the most inner one the one generated by `_.internal.apply(...)`.

nevillelyh:
Ready to merge?

PR #2421, opened 2019-11-19
make deprecate "since" field more consistent

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -683,7 +683,7 @@ class ScioContext private[scio] (
    *
    * @see [[ScioContext#run]]
    */
-  @deprecated("this method will be removed in next scio version; use run() instead.", "Scio 0.8.0")
+  @deprecated("this method will be removed in next scio version; use run() instead.", "0.8.0")
```
honnix: This kind of deprecation won't help users much because their already compiled byte code would not work with scio 0.8.0 due the method signature change.

PR #2422, opened 2019-11-19
Remove kryo coder override in intermediate steps

anish749:


jto:
LGTM. Why were we forcing Kryo ?

PR #2436, opened 2019-11-25
scala API for sort merge bucket module

ClaireMcGinty:
putting this up as a draft to get feedback on the user-facing API first before finishing up the to-dos:

- [ ] support TableRow/TFRecord types (right now it's just avro)
- [ ] codegen signatures for SMB MultiJoins similar to multijoin.py currently
- [ ] clean up/improve test coverage
- [x] add example job to scio-examples

example job showing how a user would interact with SMB in a Scio job and a `JobTest`:  https://github.com/spotify/scio/compare/claire/smb_scala#diff-7e82cfe5022eea78d15bf3b13b944501R68

I did have to lift the access level of some of the Java APIs to make them usable in tests (i.e., for unit tests the `testId` passed to the JobTest inputs/outputs corresponds to the value of the `TupleTag` which corresponds to the base path of the input/output).


scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -0,0 +1,126 @@
...
+}
+
+final class SortedBucketGenericAvroSCollection(private val self: SCollection[GenericRecord]) {
+  import com.spotify.scio.avro.{GenericRecordIO, GenericRecordTap}
+  import com.spotify.scio.smb.syntax.SortMergeBucketSCollectionSyntax._
+  import org.apache.avro.Schema
+  import org.apache.avro.file.CodecFactory
+
+  def saveAsSortedBucket[K: Coder](
+    keyClass: Class[K],
```
nevillelyh: This can be replaced by a `K: ClassTag` context bound?

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -0,0 +1,126 @@
...
+    else s"$path/bucket-*-shard-*.avro"
+}
+
+final class SortedBucketGenericAvroSCollection(private val self: SCollection[GenericRecord]) {
+  import com.spotify.scio.avro.{GenericRecordIO, GenericRecordTap}
+  import com.spotify.scio.smb.syntax.SortMergeBucketSCollectionSyntax._
+  import org.apache.avro.Schema
+  import org.apache.avro.file.CodecFactory
+
+  def saveAsSortedBucket[K: Coder](
```
nevillelyh: We'll probably have to rename these methods so they don't cause ambiguous overloading issues.
Either that, or maybe we can have a type class to encapsulate `BucketMetadata`, `FileOps`, etc.?

Something like:
```
trait SmbIO[T] {
  def bucketMetdata: BucketMetadata[T]
  def fileOps: FileOperations[T]
}

object SmbIO {
  def forJson(keyField: String): SmbIO[String]
  def forGenericRecord(...)
  ...
}
```

`schema`, `codec` (specific to Avro, Parquet), `CompressionType` (specific to TFRecord, JSON), `keyField` (might be `List[String]` in the case of feature store?) can be IO specific, while `hashType`, `numBuckets`, `numShards` are common to all IOs right?

nevillelyh: On second though, can this type class encapsulate `SortedBucketIO.{Read,Write}` both?

```
trait SmbIO[T] {
  def read(params ...): SortedBucketIO.Read[T]
  def write(params ...): SortedBucketIO.Write[T]
}
```

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
ClaireMcGinty: Yeah my thought there was to require it as a parameter to force the user to be explicit about the class of the key, since it seems easy to forget/misunderstand what that type parameter is used for. That might just be me though :)

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
ClaireMcGinty: typeclass could work--we could also do something simpler and just put the Avro/Json/TF sinks into separate package objects that a user has to import, to get around any method overloading issues. It would be pretty similar to how they interact with those IOs now actually (like instead of importing `com.spotify.scio.avro._` they'd import `com.spotify.scio.smb.avro._`).

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
nevillelyh: Actually we might even be able to remove this `K` type param if it's encapsulated in `SmbIO[K, T]` type class?

Something like `SmbIO.forAvro[MyClass](codec, schema, etc.).withKey[String]("keyField")`?

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
nevillelyh: Although an uncommon case, but the same method names might still cause problem down the road (say if we add more default args) if one wants to save as Avro + TF records in the same code block. So either the user has to isolate import scopes, or we have to prefix method names to distinguish them.

If the type class approach can parameterize Avro/TF/JSON/etc. and enable a single pair of read/write methods that'd be dope. I wonder if we can also use the type class for test stubbing?

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+  // @Todo: expand signatures in the style of multijoin.py
+  def sortMergeCoGroup[K: Coder, A: Coder, B: Coder](
+    keyClass: Class[K],
+    a: SortedBucketIO.Read[A],
+    b: SortedBucketIO.Read[B]
+  ): SCollection[(K, (Iterable[A], Iterable[B]))] = {
+    if (self.isTest) {
+      val testInputs = TestDataManager.getInput(self.testId.get)
```
nevillelyh: With the `SmbIO[T]` type class approach, we can add a method `SmbIO[K, T]#getInput(testInputs): SCollection[(K, T)]`, that way the user should be able to use `JobTest#input(SmbIO[K, T](path), ...)` without us adding `inputSmb` method?

For the writer part it should be similar:
- `scoll.saveAsSmb(smbIo, other params)` in job code
- `JobTest#output(SmbIO[K, T](path)) { assert }` in test code

nevillelyh: For that to work, `SmbIO[K, T]` needs to extend `ScioIO[(K, T)]` though, not sure if we can contort the types to work.
You might have to make the `read`/`write` methods `???` and add new ones specifically for SMB.

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
ClaireMcGinty: Yeah, I just feel that the typeclasses are reinventing the `SortedBucketIO` java APIs a bit... like `SmbIO.forAvro[MyClass](codec, schema, etc.).withKey[String]("keyField")` isn't that far away from the user just using the Java API directly like `AvroSortedBucketIO.write(classOf[String], "keyField", schema)`... especially if the Scala API is just a thin wrapper around the Java one.

Maybe for the first iteration of this we can just have basic sink and source APIs that take `SortedBucketIO.Write` and `SortedBucketIO.Read` params directly, with no implicit conversions for avro/json/TF collections? (and giving up the extra wiring for unit testing.) That way, it's simple, usable and can easily be extended in future Scio versions?

nevillelyh: 👍 yeah all the extra plumbing is really to make `JobTest` work while it won't really so 🤷‍♂

ClaireMcGinty: cool! I'll fix up the APIs today and add some kind of integration testing example to scio-examples.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  def main(cmdLineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdLineArgs)
+    val outputL = args("outputL")
+    val outputR = args("outputR")
+
+    implicit val coder: Coder[GenericRecord] =
+      Coder.avroGenericRecordCoder(SortMergeBucketExample.UserDataSchema)
+
+    sc.parallelize(0 to 499)
```
nevillelyh: Nit: I do this but that's purely personal preference ;)
```suggestion
    sc.parallelize(0 until 500)
```

build.sbt
```diff
@@ -1016,6 +1018,11 @@ lazy val `scio-smb`: Project = project
   .configs(
     IntegrationTest
   )
+  .dependsOn(
+    `scio-core`,
+    `scio-avro`,
+    `scio-test`
```
nevillelyh: Should it be this?
```suggestion
    `scio-test` % Test
```

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -0,0 +1,144 @@
...
+   * See note on [[SortedBucketScioContext.sortMergeJoin() for information on how an SMB cogroup
+   * differs from a regular [[org.apache.beam.sdk.schemas.transforms.CoGroup]] operation.
+   *
+   * @group cogroup
+
+   * @param keyClass cogroup key class. Must have a Coder in Beam's default
+   *                 [[org.apache.beam.sdk.coders.CoderRegistry]] as custom key coders are not
+   *                 supported yet.
+   */
+  def sortMergeCoGroup[K: Coder, A: Coder, B: Coder, C: Coder](
```
nevillelyh: Let's add a 4-way one to match `PairSCollectionFunctions`, and save `MultiSmb` for later?

build.sbt
```diff
@@ -1016,6 +1018,11 @@ lazy val `scio-smb`: Project = project
   .configs(
     IntegrationTest
   )
+  .dependsOn(
+    `scio-core`,
+    `scio-avro`,
```
nevillelyh: We shouldn't need this dep since SMB has its own Avro IO impl?

build.sbt
ClaireMcGinty: thanks! I forgot to clean up this and `scio-avro` after getting rid of the JobTest helper stuff.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
ClaireMcGinty: TIL!

PR #2446, opened 2019-11-26
Add an ApproxFilter API

anish749:
Things not covered

Guava dependency issues and work arounds
Integrating with Magnolify
Completeness of ScalableBloomFilters
~Delayed Initialization (Not required)~
~Serialization with Delayed Initialization~
Tests
Benchmark
Cuckoo filter and other structures.


scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,53 @@
...
+import com.spotify.scio.annotations.experimental
+import com.spotify.scio.coders.Coder
+
+/**
+ * A read only container for collections which can be used to check approximate
+ * membership queries.
+ *
+ * Constructors for [[ApproxFilter]] are defined using [[ApproxFilterBuilder]]
+ */
+trait ApproxFilter[-T] extends (T => Boolean) {
```
nevillelyh: This is a weird use for `A => B`, is it really necessary vs simply `extends Serializable` and `def apply(t: T): Boolean`?

scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
```diff
@@ -0,0 +1,22 @@
...
+trait GuavaCoders {
+
+  /**
+   * Beam Coder for [[gBloomFilter]]
+   */
+  implicit def guavaBloomFilterCoder[A](implicit fa: Funnel[A]): Coder[gBloomFilter[A]] =
+    Coder.beam(
+      new AtomicCoder[gBloomFilter[A]] {
+        override def encode(value: gBloomFilter[A], outStream: OutputStream): Unit =
+          value.writeTo(outStream)
```
nevillelyh: So guava `BF` already has well-defined and deterministic ser/de methods?

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  override def apply(t: T): Boolean = mayBeContains(t)
+
+  /**
+   * Check if the filter may contain a given element.
+   */
+  def mayBeContains(t: T): Boolean
+
+  /**
```
nevillelyh: Should we add a few more `val`s to reflect filter property? Assuming this abstracts an immutable, read only filter, I'm thinking:

- `numItems`: number of items inserted
- `fpp`: false positive rate at current load (this will be different from the Algebird one which tries to inefficiently compute per element `fpp`)
- `byteSize`: size of the serialized byte array, could be useful for hinting side-input or save methods

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,53 @@
...
+   * Serialize the Filter to an Array[Byte]
+   */
+  def serialize: Array[Byte]
```
nevillelyh: Should probably add a `def writeTo(out: OutputStream)` for the FS API and rename this to `toBytes`?

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,53 @@
...
+    sc.groupBy(_ => ())
+      .values
+      .map(build)
+
+  /**
+   * Read from serialized bytes to this filter.
+   *
+   * The serialization is done using `ApproxFilter[T]#serialize`
+   */
+  def fromBytes(serializedBytes: Array[Byte]): To[T]
```
nevillelyh: I'd add a `readFrom(in: InputStream)` to match `writeTo`

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
```diff
@@ -0,0 +1,127 @@
...
+ * An immutable wrapper over Guava's Bloom Filter.
+ *
+ * Serializable as a field in Avro Schema.
+ *
+ * @see companion object [[BloomFilter]]
+ */
+case class BloomFilter[T] private (private val internal: gBloomFilter[T]
+) extends ApproxFilter[T] {
+
+  def expectedFpp: Double = internal.expectedFpp()
```
nevillelyh: Is `internal.expectedFpp` expensive, like does it do complex math? Since the BF is immutable this can be a `val`, and part of parent trait?

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilterSCollectionOps.scala
```diff
@@ -0,0 +1,93 @@
...
+   * Generic `to` for all [[ApproxFilter]] collections.
+   **/
+  def to[C[B] <: ApproxFilter[B]](
+    builder: ApproxFilterBuilder[T, C]
+  )(implicit coder: Coder[C[T]]): SCollection[C[T]] =
+    builder.build(self)
+
+  def toBloomFilter(
+    fpProb: Double
+  )(implicit f: Funnel[T], coder: Coder[BloomFilter[T]]): SCollection[BloomFilter[T]] =
```
nevillelyh: We should have our own well defined coder for `BloomFilter`, or rather, `Coder[ApproxFilter[_]]`, so shouldn't need this implicit? The `T` can be erased here since we don't need that for ser/de?

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+}
+
+object BloomFilter {
+
+  /**
+   * Coder for [[BloomFilter]]
+   */
+  implicit def coder[T](implicit f: Funnel[T]): Coder[BloomFilter[T]] =
```
nevillelyh: I wonder if we can abstract this so that we only need one `Coder[ApproxFilter[_]]` instance instead of specialized ones for each impl? I'm thinking you can serialize the FQDN of impl class as a string prefix, and have a constructor that takes `InputStream`?

That way:
- `MyBF.writeTo` writes: "com.myorg.MyBF", len(byte array), byte array
- `Coder[ApproxFilter[_]]` reads FQDN, then calls `new MyBF(rest of input stream)`

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilterSCollectionOps.scala
nevillelyh: Also do we want to avoid leaking `Funnel[T]` to users? Like what if we want to swap in a non-guava BF impl?

scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
anish749: Yes, Guava's BF defines it, along with Strategies which defines the implementation. It serializes the strategies as well.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
anish749: I am not sure if we should enforce these at an Abstract level, as it might have different meanings and guarantees for different filters (BloomF/ CuckooF / QuotientF / etc). What do you think?

the `byteSize` or `estimatedByteSize` would be very useful for Side Inputs, I'll add that.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilterSCollectionOps.scala
anish749: This is a bit tricky.
For serialization we don't need the type `T`, however for deserialization we return a `BloomFilter[T]` for which need `T`.

Defining a `Coder[ApproxFilter[_]]` instead of `Coder[BloomFilter[_]]` means that the user won't be able extend `ApproxFilter` to add their own filters, since in that case they need to explicitly inject the coder which covers all implementation. To prevent this we can make ApproxFilter `sealed`, which is kind of restrictive. 

In order to hide the `Funnel[T]` we need to have a fallback for the `Funnel[T]`, and convey in some way that we are using something else as a fallback, which might be rare.
If we want to change the guava impl to something else, it might need a different Typeclass.
To abstract out of the type classes we have
```
  def to[C[B] <: ApproxFilter[B]](
    builder: ApproxFilterBuilder[T, C]
  )(implicit coder: Coder[C[T]]): SCollection[C[T]] =
```  
however to provide a shortcut for the user to direct go a BF, we have `BloomFilter` we have this method, hence this exposes the funnel.

I do think that we should hide the funnel from leaking, however we would need a fallback strategy.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
anish749: I am not sure why I did this. I'll remove this. I think just having apply may be a bit confusing, let's have the `mayBeContains` and be explicit.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
nevillelyh: `estimatedByteSize` sounds better than `byteSize`? I mean I know for a fact for BF, CF it's deterministic, not sure about QF. But again it depends on whether there's compression or not.

`numItems` & `fpp` are still universal across any approx filter? `numItems` may have different semantics depending on deduplication, what about `numInsertions`? That should be universal regardless of dups. Since this is an approximate filter, `fpp` should still be universal and constant? Even `fnp`, which should be `0` for most cases but we probably don't care about that assuming most impls we have guarantee 0 `fnp`.

anish749: I think we can do away with `estimated` and keep it `sizeInBytes`. I realize that this is immutable and if there is compression, that shouldn't change.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilterSCollectionOps.scala
anish749: I changed this a bit. We can have a `Coder[ApproxFilter[T]]` as AprroxFilter are serializable and having the deserialization as a separate trait allows us to always have the Coder if an implicit `Deserializer` is present.

anish749:
@nevillelyh I iterated on this once more,
I have some usage examples [here](https://github.com/spotify/scio/blob/231b16e3a45cddb74789fac21aac50064ea84777/scio-examples/src/main/scala/com/spotify/scio/examples/extra/ApproxFilterExamples.scala).

and the interface defined [here](https://github.com/spotify/scio/blob/231b16e3a45cddb74789fac21aac50064ea84777/scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala).

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Build a `SCollection[To[T]]` from an SCollection[T]
+   *
+   * By default groups all elements and builds the [[To]]
+   */
+  def build(
+    sc: SCollection[T]
+  )(implicit coder: Coder[T], filterCoder: Coder[To[T]]): SCollection[To[T]] =
+    sc.transform(name)(
+      _.distinct
```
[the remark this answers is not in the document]
anish749: for larger collections, it tries to group all everything into one iterable, hence I was removing the duplicates.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,186 @@
...
+ * Defines how to deserialize an [[ApproxFilter]].
+ *
+ * This is defined as in implicit in the companion object of the [[ApproxFilter]]
+ * A deserializer doesn't need any configurations / parameters from the user to
+ * deserialize an [[ApproxFilter]]. Since this is available as an implicit, it might
+ * summon other implicit type class instances before being able to successfully provide
+ * a deserializer instance.
+ */
+@experimental
+trait ApproxFilterDeserializer[T, To[_] <: ApproxFilter[_]] extends Serializable {
```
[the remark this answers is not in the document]
anish749: Yeah, this is primarily to pull in `Funnel` for BF and anything else for others, and not coupling Funnel with ApproxFilter

anish749:
Closing this in favor of #2526

Issue #2447, opened 2019-11-26
`saveAsCustomOutput` and `customInput` shouldn't require a name.


jabrouwer82:
None of the other transformation methods on `SCollection` or `ScioContext` require a name to be used, they're all capable of generating a name using private scio internals. My team is writing an internal library utilizing scio and it would be very useful for us to be able to let the name be generated so that we don't have to pass the burden on to our users.

jabrouwer82:
Alternatively, perhaps additionally, it would be nice if we could access `CallSite` in some way to explicitly generate a unique name the same way everything else does!

jabrouwer82:
Neville Li helped me understand these can be trivially implemented as `SCollection.internal.apply(transform)` in our own code base.

PR #2449, opened 2019-11-27
Automatically generate names for saveAsCustomOutput and customInput.

jabrouwer82:
Also make `ClosedTap`'s constructor public, since its apply was already public.
See https://github.com/spotify/scio/issues/2447

Let me know if there's anything else I need to do on this PR!

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -823,7 +823,7 @@ class ScioContext private[scio] (
       if (this.isTest) {
         TestDataManager.getInput(testId.get)(CustomIO[T](this.tfName)).toSCollection(this)
       } else {
-        wrap(this.applyInternal(name, transform))
+        wrap(this.applyInternal(transform))
       }
     }
 
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -813,6 +813,21 @@ class ScioContext private[scio] (
     }
 
   /**
+   * Get an SCollection with a custom input transform.
+   * @group input
+   */
+  def customInput[T: Coder, I >: PBegin <: PInput](
+    transform: PTransform[I, PCollection[T]]
+  ): SCollection[T] =
+    requireNotClosed {
+      if (this.isTest) {
+        TestDataManager.getInput(testId.get)(CustomIO[T](this.tfName)).toSCollection(this)
+      } else {
+        wrap(this.applyInternal(name, transform))
+      }
+    }
+
+  /**
    * Generic read method for all `ScioIO[T]` implementations, which will invoke the provided IO's
    * [[com.spotify.scio.io.ScioIO[T]#readWithContext]] method along with read configurations
    * passed in. The IO class can delegate test-specific behavior if necessary.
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -812,6 +812,21 @@ class ScioContext private[scio] (
       }
     }
 
+  /**
+   * Get an SCollection with a custom input transform.
+   * @group input
+   */
+  def customInput[T: Coder, I >: PBegin <: PInput](
```
nevillelyh: Have you considered adding these to your own implicit class, like:

```
implicit class MyScioContext(self: ScioContext) extends AnyVal {
  def customInput...
}
```

I'm hesitant to have this in the public API because `JobTest` is still the most common way of testing and `CustomIO[T](this.tfName)` isn't very user friendly. If you need to open some private/protected member to make implicit class easier in your own library I'm open for it.

jabrouwer82: I'm not very familiar with `JobTest`, but how is this sort of thing different than all the other transforms that apply `tfName` as their name? I think the simplest alternative would be to make `SCollection.applyInternal` and `ScioContext.applyInternal` public, or give them public aliases.

nevillelyh: `{SCollection,ScioContext}#applyInternal` doesn't go through `TestManager` so you can't mock input/output data with `JobTest#{input,output}` with them. Also `JobTest` requires a unique ID (usually some string param like path) for each IO but `tf.name` is not user defined and not deterministic (it depends on the line #). So for average job that wants to leverage `JobTest` we recommend `customInput` & `saveAsCustomOutput`, which require the string ID for `CustomIO` stubbing in `JobTest`. The `applyInternal` ones are usually for those wanting to integrate Scio with vanilla Beam code (which doesn't have a notion of end-to-end `JobTest`.

In your case if you know you won't need `JobTest` then it's pretty straightforward to have an implicit class and syntactic sugar around `wrap(sc.applyInternal(mySink))`? Adding itto the public API will likely make it easier for users to pick it over the "named" version by mistake and then be confused about `JobTest` stubbing.

jabrouwer82: That makes perfect sense! Would you prefer that I rename `applyInternal` in both `ScioContext` and `PCollectionWrapper` and all their uses and make them public, or just add a new public alias for both of them?

nevillelyh: I just remembered that `ScioContext#pipeline: Pipeline` and `SCollection#internal: PCollection` are public, so you can use them directly in your wrapper?

jabrouwer82: I think my team's lack of familiarity with beam might be a part of the gap for us. If we use `SCollection.internal.apply(transform)` as opposed to `SCollection.applyInternal(transform)`, then we still get a unique name for free, but we'll get the beam default naming rather than the scio default naming?

nevillelyh: Yeah but if you don’t care about streaming update then the name is nothing but UI aesthetic (if you’re on DF runner). You can even use your own naming convention in that case torepresent your custom IO.

jabrouwer82: We'll go ahead and do that, thanks!

PR #2454, opened 2019-11-27
Fix non deterministic hashCode for sparse transforms

### LINKED ISSUE #2450: Non primitive keys result in incorrect sparse transforms for more than 125M sparse keys
For cases when the smaller side of the sparse transforms have more than 125M keys, we partition the data and use multiple BloomFilters. This partitioning logic relies on Java's `.hashCode()` on the Key of the Pair SCollection.

Java's `.hashCode()` is non deterministic (in its API) and returns non primitive types (arrays and tuples).

Hence sparse do not work for arrays and tuples + more than 125M keys.

It is fairly easy for the user to make this mistake :(

anish749:
Fixes #2450 

This is definitely not the most elegant fix, but is one hacky solution.

We would want to move to an implicit Hash fn from https://github.com/spotify/magnolify soon


nevillelyh:
Let's talk about this in depth later. Composite key with non-deterministic hash is not that common? Case classes with "good" fields are OK right? Only those with array or java POJO are problematic. Reflection has a cost too, and adding `Hash[T]` context bound will break API and possibly cause other complication?

Issue #2368, opened 2019-11-04
Better way to customize equality in unit tests and better failure message


nevillelyh:
Right now most test assertions delegate to Beam's `PAssert` and hamcrest. The most common feedback about writing tests are:

- Hard to apply custom comparison logic, e.g. comparing `List[T]` fields as sets
- Hard to parse failure message, e.g. `containsInAnyOrder` should contain _very long list of records_ but does not ...

For the 1st, cats `Eq[T]` seems like an idiomatic choice but we can't inject it into Beam/hamcrest easily. We either have to hack the internals of these java libs or add our own matchers?

2 is only doable if we add our own matchers instead of using Beam/hamcrest?

anish749:
For number 1:

Scala test supports `NormalizingEquality` using scalactic, however we don't support it.

Our current example is:
```
x:SCollection[T] should containInAnyOrder(expected: Iterable[T])
```
here `containInAnyOrder` returns a `org.scalatest.matchers.Matcher` which gets a bit hacky with Coders as our current signature is `def containInAnyOrder[T: Coder](value: Iterable[T]): IterableMatcher[SCollection[T], T]`
Scalatest supports type classes using MatcherFactory, however we are not using this.

Ideally we want to use `MatcherFactory2[T, Coder, Equality]` instead of `Matcher`, this way we can inject an equality (we need to move away from hamcrest in favor of scalatest to support this)

In that case we can support the syntax:
```
x should containInAnyOrder (Seq("hi", "hello")) (after being convertedToSet)
```

where `(after being convertedToSet)` returns a `NormalizingEquality[T]` for comparing Sequences.

This follows the scalatest and scalatic syntax like:
```
"Hi" should equal ("hi") (after being lowerCased)
```

jto:
Re-implementing our own custom matchers sounds like a lot of works and it might be tricky to handle windowing correctly.

It might be possible to simply wrap each element of the `SCollection` into a wrapper class that overrides `toString` and `equals` ? 

```scala
case class TestWrapper[T: Eq](t: T) {
   override def toString = prettyPrint(this)
   override def equals(o: Any) = Eq[T].equals(t, o.t)
   override def hashCode = ???
}
```

Then matchers like `containInAnyOrder` would start by internally wrapping the elements in `value` and the elements in the tested `SCollection` in `TestWrapper`.

wdyt @nevillelyh @anish749 ?

regadas:
Closed via #2733

PR #2464, opened 2019-12-02
Use Prettifiers for Test outputs

anish749:
I tried to solve the 2nd problem in https://github.com/spotify/scio/issues/2368.

This attempts to improve the assertion failed message on the `containInAnyOrder`. 

~This PR is a very rough draft with printlns, I'll clean it up today.
The basic idea is to use scalatest instead of hamcrest / beam without introducing our own assert primitives and relying completely on Beam's. With scalatest we can inject scalatic prettifiers, which by default calls `toString`. Since the prettifier is an implicit, we can override this for types which have a `Schema` defined. With the help of the derived `Schema[T]` we can produce a nice output to help users debug easily.~
 
Previous output:
```
"apply@{Matchers.scala:6721}:6/ParMultiDo(anon$7).output: Wrappers.JIterableWrapper({"int_field": 1, "long_field": null, "float_field": null, "double_field": null, "boolean_field": null, "string_field": null, "array_field": []}) did not contain the same elements as Vector({"int_field": 1, "long_field": null, "float_field": null, "double_field": null, "boolean_field": null, "string_field": null, "array_field": []}, {"int_field": 2, "long_field": null, "float_field": null, "double_field": null, "boolean_field": null, "string_field": null, "array_field": []})" did not include substring "Vector({"int_field": 1, "long_field": null, "float_field": null, "double_field": null, "boolean_field": null, "string_field": null, "array_field": []})"
```
to:

```
java.lang.AssertionError: apply@{Matchers.scala:6721}:6/ParMultiDo(anon$7).output: 
┌──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┐
│int_field                     │long_field                    │float_field                   │double_field                  │boolean_field                 │string_field                  │array_field                   │
├──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│1                             │null                          │null                          │null                          │null                          │null                          │[]                            │
└──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┘
 did not contain the same elements as 
┌──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┐
│int_field                     │long_field                    │float_field                   │double_field                  │boolean_field                 │string_field                  │array_field                   │
├──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│1                             │null                          │null                          │null                          │null                          │null                          │[]                            │
│2                             │null                          │null                          │null                          │null                          │null                          │[]                            │
└──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┘

```


nevillelyh:
Just curious how does this look for complex types like heavily nested repeated records? Most of our schemas are like that and might be hard to pretty print in a 2D table with limited width?

anish749:
> Just curious how does this look for complex types like heavily nested repeated records? Most of our schemas are like that and might be hard to pretty print in a 2D table with limited width?

As of now I am just calling toString, but if using scalactic is a good idea, I think we can do something similar to BQ.
So array of struct would be supported, but array of struct of array would fall back to a toString.

Here is an example of what I've done so far:
```
    prettyString should be
      """
        |┌──────────────────────────────┬──────────────────────────────┐
        |│a                             │b                             │
        |├──────────────────────────────┼──────────────────────────────┤
        |│1                             │Row:[one]                     │
        |└──────────────────────────────┴──────────────────────────────┘
      """.stripMargin
```

sisidra:
I'm curious about how does it look on not infinite wide screens, is there a better way than default wrapping?

anish749:
@sisidra 
In this PR, I hard coded a width of 30 for each column. Some inspirations for Spark would be to shorten the column with a `...` to maintain the structure of the table, and then allow the user to explicitly specify that they want to see the full column. Or they might just select specific column to make sure it fits in their screen. 

Another option would be to specify a setting where each cell is printed in a new line.
Eg:
```
--- Row 1
  -- NameOfField1 : <string representation of value of field1>
  -- NameOfField2 : <string representation of value of field2>
  -- NameOfField3 : <string representation of value of field3>
--- Row2
  -- NameOfField1 : <string representation of value of field1>
  -- NameOfField2 : <string representation of value of field2>
  -- NameOfField3 : <string representation of value of field3>
```
This is useful for debugging when we have more columns.

What do you think is a good idea? and what's the best way to specify these settings explicitly when debugging / running tests?

nevillelyh:
Another thing, IIRC most common errors are from `containInAnyOrder` in which case the _expected_ can be a very long list of records. In that case, individual records not being pretty printed isn't the root cause of usability, but rather `hamcrest` not able to highlight the mismatch record & field. To put it another way, if `expected` is 10 records of 30 fields each, the pretty print output will be 300+ lines, it's still hard to eyeball which record/field fails the test?

Also the `Schema` approach will only work for types with well-defined `Schema[T]`, and not for a lot of common ones used in IO, like protobuf, Datastore `Entity`, TF `Example`, etc.? Using `Schema` will actually make it less usable for those types.

I wonder if there's an easier way, assuming `hamcrest` output is based on `toString` and most objects being tests have well defined `toString`, is it sufficient to just capture the `expected: $X`, `actual: $Y` part of output, diff strings, and highlight?

jto:
Not a fan. Ideally I would rather have pretty toString like Ammonite does (see how res0 is rendered):
ammonite

plus structured diffs would be awesome (example from [jest](https://jestjs.io/)):

jest

anish749:
@jto That would be cool.. This is what I have now. The code looks slightly hacky. I am relying on intellij show the diff, you can do side by side / unified and stuff.
Screenshot 2019-12-11 at 13 25 12

This is using scalatest instead of hamcrest.

anish749:
> Not a fan. Ideally I would rather have pretty toString like Ammonite does (see how res0 is rendered):
> ammonite

I tried out ammonite, it seems to take the screen width in account when generating the `toString`
I tried it out by resizing my terminal in two weird shapes. (`res7` and `res8`)
Screenshot 2019-12-11 at 13 42 28
Screenshot 2019-12-11 at 13 42 46

anish749:
@regadas I removed the coupling with the `Schema` thing, so right now it works as `toString` and for Avro specific records, it shows a table.

I don't want something specific for IntelliJ as that looks very ugly.

I was trying to find a way to do structured diff in sbt / error log, and I think it should be possible without too many hacks / custom stuff, but I need some time to work on it.

regadas:
> Avro specific records, it shows a table.

hm, I don't like the idea of having different ways (some tabular others not) of pretty printing. I think we should be consistent. It would be pretty confusing to the user to see different output styles just by changing datatypes.

jto:
Should we close this PR ?

regadas:
yeah I think we need to give it a little bit more thought. @anish749 do you mind if we close this from now an track this in a separate issue?

anish749:
I am not working on this now. Let's close this for now.

PR #2466, opened 2019-12-02
Rename that method argument in join functions to rhs

syodage:
Renamed `that` with `rhs` in all join functions arguments.


syodage:
Scala fix rules were added for the migration https://github.com/spotify/scio/pull/2468

PR #2483, opened 2019-12-03
improve iterable equality

anish749:
`kvListToTuple` used by transforms like `topByKey` etc, returns a `JIterableWrapper` instead of a Scala Iterable, which means equality comparisons fail. A workaround was to use `Arrays.asList(1, 2, 3).asInstanceOf[java.lang.Iterable[Int]].asScala` which is particularly inconvenient for writing test cases.

This PR allows the user to use `Iterable(...)` directly for comparison while maintaining backward compatibility with users who checked equality using `JIterableWrapper`s.
However from future versions, it is encouraged to use scala `Iterable` directly.

Detailed example:
```
it should "not confuse users" in {
    runWithContext{
      sc =>
        val p = sc.parallelize(Seq(("a", 1)))
        // The following failed previously and would be successful after this change
        p.topByKey(1).values should containInAnyOrder(Seq(Iterable(1))) 
        // The following worked and would continue to work.
        p.topByKey(1).values should containInAnyOrder(Seq(java.util.Arrays.asList(1).asInstanceOf[java.lang.Iterable[Int]].asScala.toIterable))
    }
  }
```


scio-core/src/main/scala/com/spotify/scio/util/TupleFunctions.scala
```diff
@@ -35,5 +35,7 @@ private[scio] object TupleFunctions {
     (kv.getKey, kv.getValue.asScala)
 
   def kvListToTuple[K, V](kv: KV[K, JList[V]]): (K, Iterable[V]) =
-    (kv.getKey, kv.getValue.asInstanceOf[JIterable[V]].asScala)
+    (kv.getKey, new Iterable[V] {
+      override def iterator: Iterator[V] = kv.getValue.iterator().asScala
```
regadas: 🎉 glad to see that this idea works!

scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
```diff
@@ -557,6 +557,18 @@ class PairSCollectionFunctionsTest extends PipelineSpec {
     }
   }
 
+  it should "support topByKey() with Iterable" in {
+    runWithContext { sc =>
+      val p = sc.parallelize(Seq(("a", 1), ("b", 11), ("b", 12), ("c", 21), ("c", 22), ("c", 23)))
+      val r1 = p.topByKey(1)
+      val r2 = p.topByKey(1, Ordering.by(-_))
+      r1 should
+        containInAnyOrder(Seq(("a", Iterable(1)), ("b", Iterable(12)), ("c", Iterable(23))))
```
regadas: in addition to new test ... you could potentially deprecate the `iterable` utility and update the tests that use it.

scio-core/src/main/scala/com/spotify/scio/util/TupleFunctions.scala
nevillelyh: What's the different? Something wrong with `JIterableWrapper`?

scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
anish749: yeah, sure.. I am just compiling and running tests in the cloud to save battery! I'll update the pr soon..

scio-core/src/main/scala/com/spotify/scio/util/TupleFunctions.scala
anish749: Because `JIterableWrapper` is a case class in Scala's collections, it is using case classes's equals which means that a `JIterableWrapper` is not equal to an Iterable defined by `Iterable(...)`

Example:
```
scala> Arrays.asList(1, 2, 3).asInstanceOf[java.lang.Iterable[Int]].asScala.toIterable == Iterable(1,2,3)
res18: Boolean = false
```

This behavior is hacked around in the tests when we have only one element iterables, by using a helper class:
```
def iterable[T](elems: T*): Iterable[T] = {
    val l = new java.util.ArrayList[T]()
    elems.foreach(l.add)
    l.asInstanceOf[JIterable[T]].asScala
  }
```
and then the tests are:
```
r1 should
        containInAnyOrder(Seq(("a", iterable(1)), ("b", iterable(12)), ("c", iterable(23))))
```

This causes confusion and gives rise to a very non-idiomatic way of expressing one element iterables.

Example test case:
```
it should "not confuse users" in {
    runWithContext{
      sc =>
        val p = sc.parallelize(Seq(("a", 1)))
        // The following fails
        p.topByKey(1).values should containInAnyOrder(Seq(Iterable(1))) 
        // The following works
        p.topByKey(1).values should containInAnyOrder(Seq(java.util.Arrays.asList(1).asInstanceOf[java.lang.Iterable[Int]].asScala.toIterable))
    }
  }
```

nevillelyh: Heh right it's been so long I almost forgot having to do that.

scio-test/src/test/scala/com/spotify/scio/testing/TestingUtils.scala
```diff
@@ -23,6 +23,7 @@ import scala.collection.JavaConverters._
 
 private[scio] object TestingUtils {
   // Value type Iterable[T] is wrapped from Java and fails equality check
+  @deprecated // Kept for backward compatibility as a test case only
```
nevillelyh: This is in `private[scio]` so we can safely remove. Those who hack to use it are on their own to refactor.

scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
```diff
@@ -546,14 +544,29 @@ class PairSCollectionFunctionsTest extends PipelineSpec {
...
+      r2 should
+        containInAnyOrder(Seq(("a", Iterable(1)), ("b", Iterable(11)), ("c", Iterable(21))))
+    }
+  }
+
+  // TODO explain this test case
+  // Kept for backward compat
+  it should "support topByKey() with JavaWrappedIterables" in {
```
nevillelyh: You can remove this case if we remove `def iterable()` instead of deprecate?

scio-test/src/test/scala/com/spotify/scio/testing/TestingUtils.scala
regadas: good point ... I didn't realized that it was private[scio]

scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
anish749: yes, we can remove this. I was thinking of keeping backward compat for those who hacked around, but i don't want to, so I'll remove this.

PR #2489, opened 2019-12-04
Remove schema fallback

jto:
- Get rid of the Schema fallback which does not makes sense.
- Simplify the API
- Fix `LogicalType` since they become important now :)


nevillelyh:
@regadas wanna take a closer look?

scio-core/src/main/scala/com/spotify/scio/schemas/instances/LowPrioritySchemaDerivation.scala
```diff
@@ -0,0 +1,57 @@
...
+    }.toArray
+
+    Record(schemas, rawConstruct, destruct)
+  }
+}
+
+trait LowPrioritySchemaDerivation {
+  import magnolia._
+
+  import language.experimental.macros
```
regadas: 🎨 we shouldn't need this now

scio-core/src/main/scala/com/spotify/scio/schemas/instances/LowPrioritySchemaDerivation.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  type Typeclass[T] = Schema[T]
+
+  def combine[T](ctx: CaseClass[Schema, T]): Record[T] = {
+    val ps = ctx.parameters
+    Derived.combineSchema(ps, ctx.rawConstruct)
```
regadas:
```suggestion
    Derived.combineSchema(ctx.parameters, ctx.rawConstruct)
```

scio-core/src/main/scala/com/spotify/scio/schemas/Schema.scala
```diff
@@ -117,14 +102,52 @@ object Schema extends JodaInstances with AvroInstances with LowPriorityFallbackI
...
+
+sealed trait LogicalType[T] extends Schema[T] {
+  def underlying: BSchema.FieldType
+  def toBase(v: T): Repr
+  def fromBase(base: Repr): T
+  override def toString(): String = s"LogicalType($underlying)"
+}
+
+object LogicalType {
+  def apply[T, U](u: BSchema.FieldType, t: T => U, f: U => T) = {
```
regadas:
```suggestion
  def apply[T, U](u: BSchema.FieldType, t: T => U, f: U => T): LogicalType = {
```

scio-core/src/main/scala/com/spotify/scio/schemas/Schema.scala
```diff
@@ -117,14 +102,52 @@ object Schema extends JodaInstances with AvroInstances with LowPriorityFallbackI
...
+    val ct = ClosureCleaner.clean(t)
+    val cf = ClosureCleaner.clean(f)
+    require(
```
regadas: we could move this require up

scio-core/src/main/scala/com/spotify/scio/schemas/Schema.scala
```diff
@@ -117,14 +102,52 @@ object Schema extends JodaInstances with AvroInstances with LowPriorityFallbackI
...
+  final def logicalType[T, U](
+    underlying: BSchema.FieldType,
+    toBase: T => U,
+    fromBase: U => T
+  ): Schema[T] =
+    LogicalType[T, U](underlying, toBase, fromBase)
 }
 
-sealed trait Schema[T] {
+sealed trait Schema[T] extends Serializable {
```
regadas: wondering why we need to make Serializable now? Do you see some use cases where we need to pass it around?

With this we seem to `ClosureClean` in the `LogicalType` but if we go forward with this should we clean also under `Record` and friends?

jto: bc. `SchemaCoder` somehow keeps a reference on `LogicalType` somehow. I'll try to figure out why.

jto: So I think I understand what happens. 

When you materialize a `LogicalType`, the beam Schema itself poses no problem BUT the `fromRow` and `toRow` method need to delegate conversions to `toBase` and `fromBase`. So when materializing a logical type, you need to keep a reference on those, and therefore, `LogicalType` needs to be `Serializable`.

Technically you don't need other `Schema` implementations to be `Serializable` but it's not really a problem to mark all of them `Serializable` since they are case classes anyway.

regadas: hmm interesting 🤔

Issue #2490, opened 2019-12-04
scalafix subproject doesn't have scalafmt configured


syodage:
scalafix subproject's sbt console doesn't have scalafmt tasks.

Let's configure scalafmt for scalfix subproject. Here we can use the same `.scalafmt.conf` file in the root project. 

Also let's add scalafmt check for scalafix subproject in circleci check job.

PR #2468, opened 2019-12-02
Shameera/scalafix that to rhs

syodage:
Scala fix rule to handle
-  `that` -> `rhs` 
- `that1` -> `rhs1`
- `that2` -> `rhs2`
- `that3` -> `rhs3`
- `thatNumKeys` -> `rhsNumKeys`

PR #2491, opened 2019-12-04
Shameera/add scalafmt to scalafix

syodage:
Fixes: https://github.com/spotify/scio/issues/2490

- [x] Add scalafmt support to scalafix subproject
- [x] Use the same `.scalafmt.conf` file in the main project
- [x] Add check to the circleci's check job.

**Note** this PR is based on #2468

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -30,5 +30,7 @@ rewrite {
 
 project.excludeFilters = [
   "MultiJoin.scala",
-  "TupleCoders.scala"
+  "TupleCoders.scala",
+  "scalafix/input/*",
+  "scalafix/output/*"
 ]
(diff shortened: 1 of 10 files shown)


syodage:
@regadas , you are right there were some failures with scalafix, but managed to fix those with little tweaking. In future if we don't need scalafmt to format our testing rules we can just disable the formatting with `// format: off` and enbale it back `// format: on`

regadas:
Yup i know that it's possible, but i still think it's better to just exclude them... the same way we are tweaking it around the imports it will probably happen for other cases and we end up with a bunch of on/off's

.scalafmt.conf
```diff
@@ -25,5 +25,7 @@ rewrite {
 
 project.excludeFilters = [
   "MultiJoin.scala",
-  "TupleCoders.scala"
+  "TupleCoders.scala",
+  "scalafix/input/*",
+  "scalafix/output/*"
```
syodage: These two filter out followings. 
`scalafix/input`
`scalafix/input-0_8`
`scalafix/output`
`scalafix/output-0_8`

syodage:
@regadas updated to run scalafmt only on `rules`

regadas:
@syodage CI still complains about the scalafix output

PR #2499, opened 2019-12-06
Fix bigQuerySelect deprecation message

ClaireMcGinty:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
```diff
@@ -53,7 +53,7 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
    * behavior, start the query string with `#legacysql` or `#standardsql`.
    */
   @deprecated(
-    "this method will be removed; use bigQuery(Query(sql), flattenResults) instead",
+    "this method will be removed; use bigQuerySelect(Query(sql), flattenResults) instead",
```
ClaireMcGinty: on a related note, the new `bigQuerySelect` method doesn't have a default value for `flattenResults` like this method does -- I'm in favor of adding it, thoughts?

regadas: yeah oops; adding default param it's not enough though we will get method overload issue. 

maybe we can add an extra method?
```scala
  def bigQuerySelect(sqlQuery: Query): SCollection[TableRow] = ???
```

ClaireMcGinty: ahhh right we can do an extra method! I'll add it into this PR in a minute 👍

regadas: cool cool

regadas: oh sorry LOL you said another PR

ClaireMcGinty: no worries :)

PR #2518, opened 2019-12-10
Assert FileStorage.isDone in MaterializeTap

ClaireMcGinty:
IMO, since the implementation of `materialize` is pretty opaque to the user, they may not realize it's actually modifying the pipeline graph to add a write operation, and call it after the context is closed (`sc.run().waitUntilDone().tap(data.materialize)`). this adds a check using `FileStorage.isDone` API so the user at least gets a clear error message instead of just an empty tap (actually, the avro file tap implementation has a bug when file list is empty, so user would just get a confusing strack trace)

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
@@ -30,6 +30,7 @@ import org.apache.avro.file.{DataFileReader, SeekableInput}
 import org.apache.avro.generic.GenericDatumReader
 import org.apache.avro.specific.SpecificDatumReader
 import org.apache.beam.sdk.io.FileSystems
+import org.apache.beam.sdk.io.fs.EmptyMatchTreatment
 import org.apache.beam.sdk.io.fs.MatchResult.Metadata
 import org.apache.commons.compress.compressors.CompressorStreamFactory
 import org.apache.commons.io.IOUtils
@@ -43,7 +44,7 @@ private[scio] object FileStorage {
 
 private[scio] final class FileStorage(protected[scio] val path: String) {
   private def listFiles: Seq[Metadata] =
-    FileSystems.`match`(path).metadata().asScala
+    FileSystems.`match`(path, EmptyMatchTreatment.DISALLOW).metadata().asScala
 
   private def getObjectInputStream(meta: Metadata): InputStream =
     Channels.newInputStream(FileSystems.open(meta.resourceId()))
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
```diff
@@ -43,7 +44,7 @@ private[scio] object FileStorage {
 
 private[scio] final class FileStorage(protected[scio] val path: String) {
   private def listFiles: Seq[Metadata] =
-    FileSystems.`match`(path).metadata().asScala
+    FileSystems.`match`(path, EmptyMatchTreatment.DISALLOW).metadata().asScala
```
ClaireMcGinty: I added this because the behavior of `match` for missing files is inconsistent across file system implementations and some of them will just return a status of `NOT_OK` and an empty list... so it's kind of a breaking change but the only way a user would encounter it is by explicitly creating something like a `TextTap("non/existent/path")`.

nevillelyh: And in that case it should fail instead of silently returning `Nil` right? :+1:

scio-core/src/main/scala/com/spotify/scio/io/Tap.scala
```diff
@@ -88,9 +87,18 @@ private[scio] class MaterializeTap[T: Coder] private (val path: String, coder: B
...
-      .avroFile[GenericRecord](AvroBytesUtil.schema)
-      .map(AvroBytesUtil.decode(coder, _))
+    val storage = FileStorage(_path)
+
+    if (storage.isDone) {
+      storage
+        .avroFile[GenericRecord](AvroBytesUtil.schema)
+        .map(AvroBytesUtil.decode(coder, _))
+    } else {
+      throw new RuntimeException(
```
regadas: @ClaireMcGinty when will we hit this case? users can't start reading taps if the pipeline execution is not done.

ClaireMcGinty: yeah the regular ScioIO taps are very safe! But for example, this compiles and might not look obviously wrong to the user if they don't know how `materialize` works under the hood:

```
val data = sc.parallelize(1 to 10)
sc.run().waitUntilDone().tap(data.materialize)
```

regadas: Ahh yes I think they could (somehow I missed the example in the comment) ... can we just reword this and not mention the closing?

ClaireMcGinty: ah yes my bad! Meant to reword that before merging this 😅I'm going to make a PR tomorrow rewording a bunch of the docs/comments that mention closing, I'll include this :)

PR #2520, opened 2019-12-10
delete TensorFlowIT

nevillelyh:


nevillelyh:
Yes it overlaps with several other `TFTap`/`TFExample` tests, except this one uses GCS and is considered IT. It's flaky because GCS list consistency.

PR #2526, opened 2019-12-11
Simplified Approx Filter impl

anish749:


scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,83 @@
...
+  private[values] def setField(name: String, value: Any): Unit = {
+    val f = getClass.getDeclaredField(name)
+    f.setAccessible(true)
+    f.set(this, value)
+  }
+
+  /**
+   * [[Coder]] for [[ApproxFilter]]
+   */
+  implicit def coder[T, AF[_] <: ApproxFilter[T]]: Coder[AF[T]] = {
```
nevillelyh: This is the same signature for any instances. Can it just be an `implicit def` in `object ApproxFilter`? If so you don't need this trait.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * For example usage see [[BloomFilter]]
+ */
+@experimental
+trait ApproxFilter[-T] extends Serializable {
+
+  /**
+   * Check if the filter may contain a given element.
+   */
+  def mayBeContains(t: T): Boolean
```
nevillelyh: Nit: `maybe` one word? `may be` is a different thing. Or just use `mightContain` like the Guava BF.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
```diff
@@ -0,0 +1,83 @@
...
+ * An [[ApproxFilter]] is an abstraction over various Approximate / Probabilistic
+ * data structures used for checking membership of elements.
+ *
+ * This trait defines read-only immutable filters. The filters are primarily aimed
+ * to be used as singleton [[SideInput]]s in Scio pipelines.
+ *
+ * For example usage see [[BloomFilter]]
+ */
+@experimental
+trait ApproxFilter[-T] extends Serializable {
```
nevillelyh: Is variance really necessary? If it's gonna be used in an `SCollection[T]`, `T` is always invariant.

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * Add an element to this [[BloomFilter]]. It creates a copy of the underlying
+   * structure.
+   *
+   * For creating BloomFilters from large collections, use [[BloomFilter#apply(Iterable)]]
+   * instead.
+   */
+  @experimental
+  def put(t: T): BloomFilter[T] = {
```
nevillelyh: Do we really want to expose these 2? They will be misused and sort of defeats the purpose of wrapping mutable Guava BF. I'd at least make them package private and rename to something that indicates copy, like `+/++`.

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
```diff
@@ -0,0 +1,202 @@
...
+   * @param f [[Funnel]] for the type [[T]]
+   * @return A [[BloomFilter]] for the given [[Iterable]] of elements.
+   */
+  def apply[T](
+    iterable: Iterable[T],
+    fpProb: Double
+  )(
+    implicit f: Funnel[T]
+  ): BloomFilter[T] = {
+    val numElements = iterable.size
```
nevillelyh: When used with Scio, `Iterable[T]` can very likely be the lazy ones after `groupBy` so `.size` is O(n).
IMHO the default choice or one most likely picked by user (e.g. least # of args) should minimize risk of misuse.

OTOH `fpp` of 0.01 should be a sane default, since it has the best trade off between size and fpp?

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
```diff
@@ -0,0 +1,202 @@
...
+      settings.numBFs == 1,
+      s"BloomFilter overflow: $numElements elements found, max allowed: ${settings.capacity}"
+    )
+
+    val bf = gBloomFilter.create[T](f, numElements, fpProb)
+    val it = iterable.iterator
+    while (it.hasNext) {
+      bf.put(it.next())
+    }
+    BloomFilter(bf, f)
```
nevillelyh: If we require `expectedInsertions`, we can check `expectedFpp` vs `fpProb` here and warn.

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
```diff
@@ -0,0 +1,202 @@
...
+   *
+   * An empty BloomFilter is particularly useful when used with
+   * [[SCollection.asSetSingletonSideInput(defaultValue)]]
+   *
+   * @param numElements The number of elements that we might later want to insert using
+   *                    [[BloomFilter.put]] to create a new [[BloomFilter]]
+   * @param fpProb Expected false positive probability of the resulting [[BloomFilter]]
+   *               Default 0.01 (1 %)
+   */
+  def empty[T: Funnel](numElements: Int, fpProb: Double = 0.01): BloomFilter[T] = {
```
nevillelyh: If we disallow mutation (`put/putAll`) by default, this can be a `val` and we don't need any args?

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
anish749: Fixed, we don't really need this as a trait.

scio-core/src/main/scala/com/spotify/scio/values/ApproxFilter.scala
anish749: I am not sure if this needs to be related with invariance of an `SCollection[T]`. This is because the types `ApproxFilter` and `SCollection` are never related.

Since `ApproxFilter` is read only and meant to be used as a predicate, the type is contravariant.
When casted to an `ApproxFilter` the contravariant type restricts users from building monoids.
Eg: you cannot have a `putAll( other: ApproxFilterImpl[T]): ApproxFilterImpl[T]` if it is contravariant.

My original intention was to easily support reusing `BloomFilterSideInputs` to do `sparseIntersectByKey` or filtering multiple collections without having to build the bloom filters multiple times, and also keeping the impl as generic as we can. But now that I tried this out, it doesn't need to be contravariant, however the type signatures are a bit complicated.
```
  sealed trait Animal
  case class Cat() extends Animal
  case class Dog() extends Animal
  implicit def f: Funnel[Animal] = ???

  val animals: SCollection[Animal] = ???
  val cats: SCollection[Cat] = ???
  
  val animalsBf = animals.toBloomFilter(0.01)
  cats.appoxFilter(animalsBf)
```

The types end up like this:
```
def appoxFilter[S >: T: Coder, AF[_] <: ApproxFilter[_]](filter: SCollection[AF[S]])(implicit afcoder: Coder[AF[S]]): SCollection[T] = {
```

I'll park this this to implement later and see where we end up after a few iterations.

nevillelyh: Yeah but how common are these cases, when key is not a simple string or case class?

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
anish749: I am not sure I understand this.
So the idea here is that you can have an empty BF and then add an `Iterable` using `putAll` / `++`.

A `val` will mean a wrapper over a `Funnel`. we can't add elements without specifying a `fpProb`. How do you see this being used?

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
anish749: I am not sure if we should keep this. The use case in Scio would be to add more elements at different stages.

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
nevillelyh: The idea is to not expose `putAll`/`++`, and hence the ability to build a mutable empty BF.
In that case the `empty` instance should only be useful as a side input default in case the original collection is empty. Yeah so maybe `val` wouldn't work because of `T` but it might look this:

```
def empty[T] = new BloomFilter {
  override def maybeContains(v: T): Boolean = false
}
```

scio-core/src/main/scala/com/spotify/scio/values/BloomFilter.scala
anish749: Warn as in?

I am relying on the following failure checks from Guava
```
    checkArgument(
        expectedInsertions >= 0, "Expected insertions (%s) must be >= 0", expectedInsertions);
    checkArgument(fpp > 0.0, "False positive probability (%s) must be > 0.0", fpp);
    checkArgument(fpp < 1.0, "False positive probability (%s) must be < 1.0", fpp);
    checkArgument(numHashFunctions > 0, "numHashFunctions (%s) must be > 0", numHashFunctions);
    checkArgument(
        numHashFunctions <= 255, "numHashFunctions (%s) must be <= 255", numHashFunctions);
```

nevillelyh:
```
var i = 0
while (it.hasNext) {
  i += 1
  // ...
}
// actual numbers exceeding user specified.
if (i >= expectedInsertions) // WARN
if (bf.expectedFpp > fpp) // WARN
```

nevillelyh:
I'm re-working this right now. After some tinkering I realized `Funnel[T]` relies on Java ser/de and magnolia internals and will likely break across versions. So we really should only ser/de underlying bits and decouple `Funnel[T]` with implicits. That also means we won't be able to generalize `Funnel[T]`. WIP branch pushed to `neville/bf`.

nevillelyh:
Replaced with #2635.

PR #2533, opened 2019-12-16
deprecate legacy components in scio-extras

nevillelyh:


scio-extra/src/main/scala/com/spotify/scio/extra/checkpoint/package.scala
```diff
@@ -52,8 +51,9 @@ package object checkpoint {
      * @param fn result of this arbitrary => [[com.spotify.scio.values.SCollection SCollection]]
      *           flow is what is checkpointed
      */
+    @deprecated("Checkpoint support is deprecated, " +
+      "use smaller workflows and orchestration framework instead", "0.8.0")
     def checkpoint[T: Coder](fileOrPath: String)(fn: => SCollection[T]): SCollection[T] = {
-      FileSystems.setDefaultPipelineOptions(self.options)
```
regadas: was this remove intended? we can just leave as is since this is deprecated?

nevillelyh: Yes, we added this to `ScioContext` constructor so this work around is no longer needed.

regadas: ahh yes right

PR #2537, opened 2019-12-17
Remove @experimental from transform

regadas:


nevillelyh:
Any particular reason? IIRC these fail more often than the vanilla ones due to lambda ser/de and is only needed for UI aesthetic?

regadas:
They have been in @experimental for a year and they are used everywhere. Also they are not serialized.

PR #2541, opened 2019-12-17
tag SQL API as experimental

nevillelyh:


scio-sql/src/main/scala/com/spotify/scio/sql/SqlSCollections.scala
```diff
@@ -22,10 +22,12 @@
...
 package com.spotify.scio.sql
 
+import com.spotify.scio.annotations.experimental
 import com.spotify.scio.schemas._
 import com.spotify.scio.values.SCollection
 
 trait SqlSCollections {
+  @experimental
   def from[A: Schema](a: SCollection[A]): SqlSCollection1[A] = new SqlSCollection1(a)
   def from[A: Schema, B: Schema](a: SCollection[A], b: SCollection[B]): SqlSCollection2[A, B] =
```
nevillelyh: Might be overkill to annotate all these. `from(a)` is the only one used in an implicit conversion.

Issue #2530, opened 2019-12-13
Add Scio binding for Beam's `SortValues` transform


ClaireMcGinty:
https://github.com/apache/beam/blob/master/sdks/java/extensions/sorter/src/main/java/org/apache/beam/sdk/extensions/sorter/SortValues.java

Possibly we could include the preceding Beam `GroupByKey` transform as well so the signature of the Scio method looks more like `topByKey` and users don't get confused trying to use Scio's `.groupBy` method with it, and fix the Coder specifically as a Beam `KvCoder`.

nevillelyh:
If we follow the `SortValues` interface, it should probably be:
`SCollection[(K1, (K2, V))]#sortByKey: SCollection[(K1, Iterable[(K2, V)])]`.

Or do we want to mimic the Scio ones instead?
`SCollection[(K, V)]#sortByKey[K2](ord: V => SK): SCollection[(K, Iterable[V])]`

The 2nd feels more idiomatic Scio. Another question is how to we ensure this is not misused in place of `topByKey`?

ClaireMcGinty:
I agree, the second one feels more consistent with the rest of the Scio API, although users may be confused anyway we don't support an implicit `Ordering[V]` as argument.

Not sure how to guard against misuse--putting it in a separate module seems like overkill, maybe the best we can do is good documentation/maybe logging a warning.

nevillelyh:
Brought up in internal discussion. `SortValues` doesn't support descending, and lexicographical comparator won't work for types like double. I wonder if there're ways to work around this. We might have to fork the `NativeFileSorter` though?

nevillelyh:
https://issues.apache.org/jira/browse/BEAM-8985
https://issues.apache.org/jira/browse/BEAM-8986
https://issues.apache.org/jira/browse/BEAM-8987

regadas:
Yeah interesting points. If we tackle those I think we can try fix it upstream instead of forking?

Nevertheless, I think we can still provide the api (with some caveats) since it still helps with the `Coder` and `KV` juggling.

nevillelyh:
Would be nice to at least fix the lexicographical sorting issue w.r.t. numbers, otherwise users might get incorrect result by mistake.

regadas:
Trying to get a working example where I can repro that issue

ClaireMcGinty:
[...]

PR #2542, opened 2019-12-17
Add simple sorter method

regadas:


nevillelyh:
Should we discuss issues with `SortValues` in #2530 first?

regadas:
@nevillelyh I think we can merge this one as well

nevillelyh:
I'll look in a bit.

scio-extra/src/main/scala/com/spotify/scio/extra/sorter/syntax/SCollectionSyntax.scala
```diff
@@ -0,0 +1,78 @@
...
+          IterableCoder.of(CoderMaterializer.kvCoder[K2, V](self.context))
+        )
+      )
+      .withName("SortValues")
+      .applyTransform(SortValues.create[K1, K2, V](sorterOptions))
+      .withName("KvToTuple")
+      .map { kv =>
+        (
+          kv.getKey,
+          kv.getValue.asScala.map(xs => (xs.getKey, xs.getValue))
```
nevillelyh: This is `Iterable#map` and eager right? Even though the `JIterable` from `SortValues` is lazy. Do we want a lazy wrapper? Otherwise for the spilled to disk case it can easily OOM.

scio-extra/src/main/scala/com/spotify/scio/extra/sorter/syntax/SCollectionSyntax.scala
```diff
@@ -0,0 +1,78 @@
...
+   * The secondary key needs to be encoded as a [[String]] or [[Array[Byte]].
+   * [[SortValues]] compares bytes lexicographically and may write secondary key-value pairs to disk.
+   *
+   * @note The primary key is explicit here only because this
+   * transform is typically used on a result of a [[PairSCollectionFunctions.groupByKey]].
+   *
+   * @param sorterOptions see [[BufferedExternalSorter.Options]]
+   */
+  @experimental
+  def sortValues(sorterOptions: BufferedExternalSorter.Options)(
```
nevillelyh: Do we want a default options? IMO the Beam default Hadoop `SequenceFile` sorter is 2x slower than the native sorter. The only other option is memory. Should we just have a `memoryMB: Int = 128`?

scio-extra/src/main/scala/com/spotify/scio/extra/sorter/syntax/SCollectionSyntax.scala
regadas: yeah that's a good point ... either that or simply return `Iterator`

scio-extra/src/main/scala/com/spotify/scio/extra/sorter/syntax/SCollectionSyntax.scala
regadas: to be honest , not sure not that keen in setting defaults here. I guess i would prefer users to look at the options and set them according to their use case?

nevillelyh: At least make native sorter a default and exclude Hadoop dependency? The only reason it's the default in Beam is some review prefer it for "backwards compatibility" ;)

We can then expose `memoryMB` as a required option?

scio-extra/src/main/scala/com/spotify/scio/extra/sorter/syntax/SCollectionSyntax.scala
nevillelyh: `Iterator` might not survive ser/de properly though need to double check.

regadas: `Iterator` is a no go we really need a lazy iterable

scio-extra/src/main/scala/com/spotify/scio/extra/sorter/syntax/SCollectionSyntax.scala
regadas: me and @nevillelyh talked in IRL; and @nevillelyh did some benchmarks on the default `SorterType.HADOOP` and concluded that is slower then the `SorterType.NATIVE`. With this in mind we will restrict this to the later and only expose `memoryMB `

PR #2549, opened 2019-12-18
Add support for cleaning up TF models on shutdown

regadas:
It's good to be a good citizen and cleanup resources. I believe this will be primarily important when we add support for reloading models in streaming use cases.


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -33,10 +33,17 @@ import scala.collection.JavaConverters._
...
 
 sealed trait PredictDoFn[T, V, M <: Model[_]]
-    extends DoFnWithResource[T, V, ConcurrentMap[String, M]] {
-  @transient private lazy val log = LoggerFactory.getLogger(this.getClass)
+    extends DoFnWithResource[T, V, ConcurrentMap[String, (AtomicInteger, M)]] {
+  import PredictDoFn.Log
+
+  def modelId: String
+
+  def createModel: M
```
regadas: 🎨 I think this just be renamed to `def loadModel: M`

regadas:
yeah, good to go.

PR #2556, opened 2019-12-24
Add scalafmt SortModifiers

regadas:


scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
```diff
@@ -30,7 +30,7 @@ import scala.reflect.runtime.universe._
 
 class TypeProviderTest extends AnyFlatSpec with Matchers {
   @AvroType.fromSchema(
-    """{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}"""
+    "{\"type\":\"record\",\"name\": \"Record\",\"fields\":[{\"name\":\"f1\",\"type\":\"int\"}]}"
```
nevillelyh: This is not part of `SortModifiers`? Is it from a new scalafmt version? Feels less readable than the triple-double-quoted version.

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaFutureHandlers.scala
```diff
@@ -27,7 +27,7 @@ import scala.concurrent.duration.Duration
 /** A [[FutureHandlers.Base]] implementation for Scala [[Future]]. */
 trait ScalaFutureHandlers[T] extends FutureHandlers.Base[Future[T], T] {
   @transient
-  private lazy implicit val immediateExecutionContext = new ExecutionContext {
+  implicit private lazy val immediateExecutionContext = new ExecutionContext {
```
nevillelyh: This also feels slightly counter intuitive since we usually say "implicit val" and I often grep for that exact match. 🤷‍♂

scio-core/src/main/scala/com/spotify/scio/transforms/package.scala
```diff
@@ -39,16 +39,14 @@ import scala.util.{Failure, Success, Try}
  */
 package object transforms {
   @deprecated(
-    "Class AsyncLookupDoFn was renamed to BaseAsyncLookupDoFn." +
-      "see https://spotify.github.io/scio/migrations/v0.8.0.html#async-dofns for more information",
+    "Class AsyncLookupDoFn was renamed to BaseAsyncLookupDoFn." + "see https://spotify.github.io/scio/migrations/v0.8.0.html#async-dofns for more information",
```
nevillelyh: So these are excluded from line width rule now? the `" + "` in the middle is now redundant.

scio-avro/src/test/scala/com/spotify/scio/avro/types/TypeProviderTest.scala
regadas: yeah I saw this as well and this is caused by the SortModifiers only on annotations. I haven't had time yet to track those down and change them to:

```scala
  @AvroType.fromSchema("""
      |{"type":"record","name": "Record","fields":[{"name":"f1","type":"int"}]}
      |""".stripMargin)
  class StringLiteralRecord
```

scio-core/src/main/scala/com/spotify/scio/transforms/package.scala
regadas: don't like this either I need to change this

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaFutureHandlers.scala
regadas: access modifier aside... I would say `implicit lazy val` is slightly more intuitive 😄 and it's also inline with scalac style. It's because of these preferences/inconsistencies that i want smth consistent.

PR #2589, opened 2020-01-10
Make KryoAtomicCoder.instanceId an incrementing integer, not a UUID

ClaireMcGinty:
The randomly generated UUID breaks streaming pipeline updates for jobs that use Kryo coders, since the coder's serialized payload is different every time the job graph is constructed. Tested this out in a few streaming jobs and didn't see any issues, it's had to fully verify though...


nevillelyh:
It's breaking scalafmt in a different file, sth fishy going on?

ClaireMcGinty:
@nevillelyh fixed! just needed a rebase to pick up a [formatting change](https://github.com/spotify/scio/blame/8e046f3abf72ac5bd78e5bcad612160a64413268/scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala#L92)

PR #2594, opened 2020-01-14
Add test case for avro conversion with bytes

jto:


scio-test/src/test/scala/com/spotify/scio/sql/BeamSQLTest.scala
```diff
@@ -659,6 +659,38 @@ class BeamSQLTest extends PipelineSpec {
...
+    import TypeConvertionsTestData._
+
+    val cat = "lorem"
+    val cid = "42"
+    val devId = "device_id".getBytes()
+
+    val m = com.spotify.scio.avro.Message.newBuilder()
+    val r = AvroWithBytesRecord.newBuilder()
+    m.setCatalogue(cat)
+    m.setClientId("42")
```
regadas:
```suggestion
    m.setClientId(cid)
```

PR #2629, opened 2020-01-23
add Map/MultiMap singleton side input, fix #2628

### LINKED ISSUE #2628: Possibly regression from #2424, map side inputs
In #2424 we replaced `SideMap` with vanilla `asMapSideInput` & `asMultiMapSideInput` which are backed by GCS/external storage on DF and is possibly causing a performance regression.

We should review all `as*SideInput` usages.

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -1156,4 +1158,46 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+   * `Map[key, value]`, to be used with [[SCollection.withSideInputs]]. It is required that each
+   * key of the input be associated with a single value.
+   *
+   * Currently, the resulting map is required to fit into memory. This is preferable to
+   * [[asMapSideInput]] if that's the case.
+   */
+  def asMapSingletonSideInput(implicit koder: Coder[K], voder: Coder[V]): SideInput[Map[K, V]] =
+    self
+      .transform(
+        _.groupByKey
```
regadas: wondering if we could have a combine step here instead?

nevillelyh: Yeah was being lazy. Figured <100MB it might not make a huge difference. I'll take a look.

nevillelyh: Actually I'm not sure how much performance gain we get for the cost of readability. `Iterable` probably has O(n) append & concat so we'll have to use mutable/vector/whatever & cast later, adds a fair amount of cruft.

regadas: In this particular case the combiner would be only `Map[K, V]` and adding to an immutable map is effectively constant so I think it would be a lot faster. Also with this iirc step fusion might not be guaranteed

nevillelyh: The merge fns `(Map[K, V], (K, V)) => Map[K, V]` & `(Map[K, V], Map[K, V]) => Map[K, V])` will still have to check keys with >1 values though.

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -1156,4 +1158,46 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+          .map { kv =>
+            require(kv._2.size == 1, s"Multiple values for key ${kv._1}")
+            (kv._1, kv._2.head)
+          }
+          .groupBy(_ => ())
+          .map(_._2.toMap)
```
rscarvalho: Could a `fold` work better in this case?

```scala
.map { kv =>
  require(kv._2.size == 1, s"Multiple values for key ${kv._1}")
  Map(kv._1 -> kv._2.head)
}
.fold
```

I'm not sure if `fold` would do preagg (and if that would make any real difference) but might be a bit more readable

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -841,8 +841,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       .transform(
         _.distinct
           .groupBy(_ => ())
-          .values
-          .map(_.toSet)
+          .map(_._2.toSet)
```
rscarvalho: Same thing here for `.fold`. Would read something like

```scala
_.distinct.map(Set(_)).fold
```

PR #2631, opened 2020-01-24
Update beam to 2.18

regadas:
Interesting nugget about windows and that might be a user pitfall: https://docs.google.com/document/d/1sr8Y8jgk8TPToJerWgcdmfaRSoTzPCQPnuW9F73yTpE/edit#heading=h.k6vgpjglhg2k

with beam 2.18 top level `GroupBykey` will not allow windows that finish.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,7 @@ import bloop.integrations.sbt.BloopDefaults
 
 ThisBuild / turbo := true
 
-val beamVersion = "2.17.0"
+val beamVersion = "2.18.0"
 
 val algebirdVersion = "0.13.6"
 val annoy4sVersion = "0.9.0"
@@ -81,8 +81,8 @@ val caffeineVersion = "2.8.1"
 val bigtableClientVersion = "1.8.0"
 val generatedGrpcGaVersion = "1.43.0"
 val generatedGrpcBetaVersion = "0.44.0"
-val googleClientsVersion = "1.27.0"
-val googleApiServicesBigQuery = s"v2-rev20181104-$googleClientsVersion"
+val googleClientsVersion = "1.28.0"
+val googleApiServicesBigQuery = s"v2-rev20181221-$googleClientsVersion"
 val bigdataossVersion = "1.9.16"
 val gaxVersion = "1.38.0"
 val googleAuthVersion = "0.12.0"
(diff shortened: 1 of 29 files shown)


build.sbt
```diff
@@ -81,8 +81,8 @@ val caffeineVersion = "2.8.1"
 val bigtableClientVersion = "1.8.0"
 val generatedGrpcGaVersion = "1.43.0"
 val generatedGrpcBetaVersion = "0.44.0"
-val googleClientsVersion = "1.27.0"
-val googleApiServicesBigQuery = s"v2-rev20181104-$googleClientsVersion"
+val googleClientsVersion = "1.28.0"
+val googleApiServicesBigQuery = s"v2-rev20181221-$googleClientsVersion"
 val bigdataossVersion = "1.9.16"
```
ChethanUK: Why this is 1.9.16? When the latest is 2.0.0?

jto:
LGTM. 

I don't think the problems in the document are really an issue. 
As far as I understand case 1 and 2 would be so obviously broken that it's actually unlikely to go unnoticed. Case 3 is a bit trickier but the fix is not making the situation worse, merely more visible.

build.sbt
regadas: Hi @ChethanUK, we use that version to be inline with `apache/beam`.

ChethanUK: Okay, I just asked bcoz by default all other products we use (Dataproc, SparkOnK8s, Flink) use 2.0.0.

regadas: yeah I understand ... `apache/beam` sometimes just takes it's time to update deps.

project/plugins.sbt
```diff
@@ -17,6 +17,7 @@ addSbtPlugin("com.lightbend.paradox" % "sbt-paradox" % "0.6.8")
 addSbtPlugin("io.github.jonas" % "sbt-paradox-material-theme" % "0.6.0")
 addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.3.5")
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.1.1")
+addSbtPlugin("com.github.cb372" % "sbt-explicit-dependencies" % "0.2.12")
```
regadas: This change just slipped through by mistake 😢

regadas: Reverted it here https://github.com/spotify/scio/commit/4d731cab98ee13887c407e7a915ac54e2e8aac5a

PR #2635, opened 2020-01-27
Re-work generalized ApproxFilter

nevillelyh:
Decoupled Algebird `Hash128` & Guava `Funnel` from ser/de since Java serialization will likely break if impl changes. Also need to verify that we can replace the existing `hash*` ops with this new impl.


nevillelyh:
Verified that all existing transforms in `PairSCollectionFunctions` can be ported easily by replacing `Hash128[K]` with `Funnel[K]`. We can also get rid of the side input warning by hard-coding a 100MB `maxBytes`. Will do in a follow up PR.

Now I'm not sure whether we need `ABloomFilter` impl since it's really naive and inefficient.
Otherwise I'm pretty happy with this. Also re-worked the partition math to be more correct.

@anish749 PTAL.

nevillelyh:
This is outdated, moving discussion to #2651.

PR #2637, opened 2020-01-27
Add specialized coders for Elasticsearch, fix #2626

### LINKED ISSUE #2626: Add specialized coders for Elasticsearch types
For both performance reasons & avoid non-deterministic coder serialization in streaming `update`.
Related to #2488

nevillelyh:
Only covering ES5,6,7 here since ES2 support is deprecated.


nevillelyh:
Do you have the snippet for dumping graph? Laptop is rebooting I’ll check later.

nevillelyh:
Verified that all 3 branches no longer include `KryoAtomic` in the job graph.

Also for posterity the we should look out for the following warning for potential non-deterministic coders:
```
[info]  Warning: No implicit Coder found for the following type:
[info]
[info]    >> org.elasticsearch.action.index.IndexRequest
[info]
[info]  using Kryo fallback instead.
````

PR #2646, opened 2020-01-29
Add BigQuery macro Avro Schema and toAvro support

regadas:


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/BigQueryType.scala
```diff
@@ -78,9 +79,15 @@ object BigQueryType {
     /** Case class schema. */
     def schema: TableSchema
 
+    /** Case class schema. */
```
regadas:
```suggestion
    /** Case class avro schema. */
```

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -151,6 +160,100 @@ private[types] object ConverterProvider {
...
+    }
+
+    def constructor(tpe: Type, fn: TermName): Tree = {
+      val sets = tpe.erasure match {
+        case t if isCaseClass(c)(t) => getFields(c)(t).map(s => field(s, fn))
+        case _                      => c.abort(c.enclosingPosition, s"Unsupported type: $tpe")
+      }
+
+      val header =
+        q"val result = new _root_.org.apache.avro.generic.GenericRecordBuilder(${p(c, SType)}.avroSchemaOf[$tpe])"
```
nevillelyh: This means we'll have to traverse `$tpe` to create schema for every new record right? Another reason to cache schema?

regadas: hm yes it, agree it might be better to cache it. Let's follow this up.

regadas: #2648 ^^

scio-bigquery/src/test/scala/com/spotify/scio/bigquery/types/ConverterProviderSpec.scala
```diff
@@ -68,13 +68,27 @@ final class ConverterProviderSpec
...
   property("round trip optional primitive types") {
     forAll { r1: Optional =>
       val r2 = BigQueryType.fromTableRow[Optional](BigQueryType.toTableRow[Optional](r1))
       EqDerivation[Optional].eqv(r1, r2) shouldBe true
     }
   }
 
+  property("avro round trip optional primitive types") {
+    forAll { r1: Optional =>
+      val r2 = BigQueryType.fromTableRow[Optional](BigQueryType.toTableRow[Optional](r1))
```
regadas: I missed this 😢

PR #2650, opened 2020-01-31
Force GenericRecord writer in SMB Avro, fix 2649

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
```diff
@@ -93,7 +93,13 @@ public void populateDisplayData(Builder builder) {
   protected FileIO.Sink<ValueT> createSink() {
     final AvroIO.Sink<ValueT> sink =
         recordClass == null
-            ? (AvroIO.Sink<ValueT>) AvroIO.sink(getSchema())
```
nevillelyh: `AvroIO.sink()` does not set `RecordFormatter`, and the sink ends up using `ReflectWriter` for the records, which possibly does not handle `map` properly.

regadas: Yup 🤦‍♂ https://github.com/apache/beam/blob/9b30d7c9796abdf5c7e44c15b60733f1c29cbd40/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java#L1860-L1864

regadas:
@nevillelyh good to merge?

Issue #2649, opened 2020-01-31
SMB save fails on Avro map field


nevillelyh:
I can repro with the following code:

```scala
import com.spotify.scio._
import com.spotify.scio.avro._
import com.spotify.scio.smb._
import com.spotify.scio.coders.Coder
import org.apache.avro.Schema
import org.apache.avro.generic.{GenericRecord, GenericRecordBuilder}
import org.apache.beam.sdk.extensions.smb.AvroSortedBucketIO

import scala.collection.JavaConverters._
import scala.util.Random

object Debug {
  val schemaString =
    """
      |{
      |  "type": "record",
      |  "name": "Record",
      |  "namespace": "com.spotify.data",
      |  "fields": [
      |    {
      |      "name": "id",
      |      "type": "int"
      |    },
      |    {
      |      "name": "map",
      |      "type": [
      |        "null",
      |        {
      |          "type": "map",
      |          "values": "string"
      |        }
      |      ],
      |      "default": null
      |    }
      |  ]
      |}
      |""".stripMargin

  val schema = new Schema.Parser().parse(schemaString)

  def makeRecord: GenericRecord = {
    val builder = new GenericRecordBuilder(schema)
      .set("id", Random.nextInt(100))
    if (Random.nextBoolean()) {
      val m = (1 to Random.nextInt(10) + 1)
        .map(i => s"key$i" -> Random.nextInt(100).toString)
        .toMap
      builder.set("map", m.asJava)
    }
    Random.nextInt(10)
    builder.build()
  }

  def main(args: Array[String]): Unit = {
    val sc = ScioContext()
    implicit val coder = Coder.avroGenericRecordCoder(schema)
    sc.parallelize(Seq.fill(100)(makeRecord))
        .saveAsSortedBucket(AvroSortedBucketIO
          .write(classOf[Integer], "id", schema)
          .to("out")
          .withNumBuckets(2),
        )
//      .saveAsAvroFile("out", schema = schema) // this works fine
    sc.run()
  }
}
```

Stacktrace:
```
[error] Exception in thread "main" org.apache.beam.sdk.Pipeline$PipelineExecutionException: org.apache.avro.file.DataFileWriter$AppendWriteException: org.apache.avro.UnresolvedUnionException: Not in union ["null",{"type":"map","values":"string"}]: {key1=17, key2=46, key5=68, key6=16, key3=27, key4=55}
[...]

PR #2664, opened 2020-02-10
add test for #2649, Avro map field in SMB sink

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
```diff
@@ -93,6 +93,8 @@ public void populateDisplayData(Builder builder) {
   protected FileIO.Sink<ValueT> createSink() {
     final AvroIO.Sink<ValueT> sink =
         recordClass == null
+            // https://github.com/spotify/scio/pull/2649
```
ClaireMcGinty: should be `https://github.com/spotify/scio/issues/2649` ?

nevillelyh:
```suggestion
            // https://github.com/spotify/scio/issues/2649
```

scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroFileOperationsTest.java
```diff
@@ -127,4 +132,41 @@ public void testDisplayData() {
     MatcherAssert.assertThat(
         displayData, hasDisplayItem("schema", AvroGeneratedUser.SCHEMA$.getFullName()));
   }
+
+  // https://github.com/spotify/scio/pull/2649
```
nevillelyh:
```suggestion
  // https://github.com/spotify/scio/issues/2649
```

PR #2667, opened 2020-02-10
Prevent sharded-sparkey shards from having negative indices.

psobot:
https://github.com/spotify/scio/pull/2336 introduced sharded Sparkey support and assumed that the Sparkey shards would have non-negative indices (i.e.: if you specify 32 shards, you should get up to 32 files named `part-00001` through `part-00032-of-00032.sp[il]`). However, Scio 0.8.0 and 0.8.1 instead produce both negative and positive shard indices. This means that any code reading the resulting sharded datasets needs to handle the anticipated number of shards.

This PR fixes this inconsistency by wrapping each call to `MurmurHash` with `abs`. **This means that any sharded Sparkey datasets produced before this change will no longer be readable after this change, as the shard allocation will be different.** Luckily, this feature was introduced very recently and anecdotally is mostly used internally within `.asSparkeySideInput`, which produces temporary datasets that are not intended to be read outside the context of each Scio pipeline.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
@@ -34,7 +34,7 @@ import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
 import scala.util.hashing.MurmurHash3
-import scala.math.abs
+import java.lang.Math.floorMod
 
 /**
  * Main package for Sparkey side input APIs. Import all.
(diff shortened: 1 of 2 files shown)


nevillelyh:
What about `floorMod` like in #1654?

nevillelyh:
https://stackoverflow.com/questions/4412179/best-way-to-make-javas-modulus-behave-like-it-should-with-negative-numbers
More discussion here. Someone claims that `floorMod` is very slow. Not sure how big of an impact it has in the grand scheme of things though. Can maybe investigate later and be consistent for now.

regadas:
oh wow had no idea of this issue with `floorMod` :|

regadas:
briefly chatted with @nevillelyh and using `floorMod` should not be a big deal actually.

PR #2674, opened 2020-02-12
Make Coder[Set] throw NonDeterministicException

### LINKED ISSUE #2644: Fold by key not working with Set in key
When using `.foldByKey` on a key-value SCollection that has a Set in the key, the output is different than expected when the sets are created with items in different orders.
```
  it should "Fold sets correctly" in {

    val sets = Seq(
      (Set(1,2), List("hey")),
      (Set(2,1), List("hi"))
    )
    runWithContext {
      sc =>
        val setsCollection = sc.parallelize(sets)
        setsCollection.foldByKey debug()
    }
  }

// Output:
//(Set(1, 2),List(hey))
//(Set(2, 1),List(hi))
```
It works fine when the sets are created with the same ordering:
```
  it should "Fold sets correctly" in {

    val sets = Seq(
      (Set(1,2), List("hey")),
      (Set(1,2), List("hi"))
    )
    runWithContext {
      sc =>
        val setsCollection = sc.parallelize(sets)
        setsCollection.foldByKey debug()
    }
  }
// Output: (Set(1, 2),List(hey, hi))
```

jto:
fixes #2644 (sort of)

### DIFF
--- a/scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
@@ -568,6 +568,13 @@ final class CoderTest extends AnyFlatSpec with Matchers {
     Recursive(1, Option(Recursive(2, None))) coderShould roundtrip()
   }
 
+  it should "#2644 verifyDeterministic throw a NonDeterministicException exception for Set" in {
+    an[NonDeterministicException] should be thrownBy {
+      val coder = Coder[Set[Int]]
+      materialize(coder).verifyDeterministic()
+    }
+  }
+
 }
 
 object RecursiveCase {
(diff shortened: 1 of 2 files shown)


scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -568,6 +568,13 @@ final class CoderTest extends AnyFlatSpec with Matchers {
     Recursive(1, Option(Recursive(2, None))) coderShould roundtrip()
   }
 
+  it should "#2644 throw a verifyDeterministic exception for Set" in {
```
kythyra: Small detail, the description of the test is inexact: `verifyDeterministic` is the name of the function, not the name of the exception

PR #2676, opened 2020-02-12
Make files to stage list mutable

regadas:
`GcsStager` can add additional files

https://github.com/apache/beam/blob/3a79080d2a98aa3ed3bdf01a83262da86ec2f60c/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/GcsStager.java#L53-L71


jto:
Did this cause issues ?

regadas:
Yeah! I was replacing the dataflow-worker jar when i stumbled upon this!

PR #2677, opened 2020-02-14
Add BigQuery client scope for querying from Google Drive backed tables

tchagnon:
Documentation: https://cloud.google.com/bigquery/external-data-drive

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
@@ -28,6 +28,8 @@ import scala.util.Try
 
 object BigQueryConfig {
 
+  private[this] val DRIVE_SCOPE = "https://www.googleapis.com/auth/drive"
+
   /** Default cache directory. */
   private[this] val CacheDirectoryDefault: Path = Paths
     .get(CoreSysProps.TmpDir.value)
@@ -46,7 +48,10 @@ object BigQueryConfig {
 
   def location: String = DefaultLocation
 
-  def scopes: Seq[String] = DefaultScopes
+  def scopes: Seq[String] = DefaultScopes ++
+    BigQuerySysProps.DriveScope.valueOption
+      .map(x => Try(x.toBoolean).getOrElse(false))
+      .collect { case true => DRIVE_SCOPE }
 
   def isCacheEnabled: Boolean =
     BigQuerySysProps.CacheEnabled.valueOption
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
@@ -48,10 +48,11 @@ object BigQueryConfig {
 
   def location: String = DefaultLocation
 
-  def scopes: Seq[String] = DefaultScopes ++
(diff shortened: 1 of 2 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
```diff
@@ -46,7 +48,11 @@ object BigQueryConfig {
 
   def location: String = DefaultLocation
 
-  def scopes: Seq[String] = DefaultScopes
+  def scopes: Seq[String] =
+    DefaultScopes ++
+      BigQuerySysProps.DriveScope.valueOption
+        .map(x => Try(x.toBoolean).getOrElse(false))
```
regadas: use `filter(...).map(...)` instead?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
```diff
@@ -28,6 +28,8 @@ import scala.util.Try
 
 object BigQueryConfig {
 
+  private[this] val DRIVE_SCOPE = "https://www.googleapis.com/auth/drive"
```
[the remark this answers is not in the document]
tchagnon: 👍

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
tchagnon: done. lmk if that's what you were thinking

tchagnon:
Hi @regadas, reading that issue, it sounds like that would only affect default GCE compute credentials.  We're using service accounts everywhere, so I think that issue wouldn't affect us?

nevillelyh:
LGTM should we merge this?

tchagnon:
@regadas for 1. do you mean that this won't apply to calls to `ScioContext.typedBigQuery` etc?  It looked like this client gets used somehow there?

For 2.  I'm just following the existing pattern for configuration that you have.  If that should be refactored, maybe another issue/pr is appropriate.

I don't see a way to set the BigQuery client or Google credentials object that the ScioContext uses directly.  That would be helpful for this.

regadas:
@tchagnon unfortunately it won't apply to those. It will only be applied to our impl of `BigQuery` client that it's used for some ops internally and can be used outside for some extra work outside of the pipeline.

To add support for this scope, across all feature set, requires changes in `beam`.

regadas:
@tchagnon would you be ok if we close this? There's no straightforward way to achieve this at the moment unfortunately. We can however create an issue to track this and try to push changes upstream?

tchagnon:
Sure, that makes sense.

iezer:
@regadas I came across this old PR and I'm having a similar issue. I see that Beam has made changes to support setting OAuth scopes https://github.com/apache/beam/pull/23644. This is available as of Beam 2.43 which is available as of Scio 0.12.1. However it seems like we would still need some changes to similar to what's in this PR to consume the scopes set by the caller. Thoughts?

iezer:
@regadas Following up on the above, perhaps if I set the scope on my `DataflowPipelineOptions` object using that API that was added to Beam ie. `opts.setGcpOauthScopes(... "auth/drive")`, then I can use some other API other than `sc.typedBigQuery` like `sc.bigQuerySelect` that you mention above which would use the added scope.

PR #2679, opened 2020-02-14
Add java code fmt

regadas:
similar to scalafmt ... commands like `javafmtAll` et al are available


ClaireMcGinty:
+1, all the Java code in scio-smb is getting unruly

should we try to match Beam's Java style or just use https://github.com/sbt/sbt-java-formatter? It seems like they use different formatters (Beam uses the Spotless formatter and sbt-java-formatter uses the official Google Java Format guide)...

regadas:
@ClaireMcGinty Ahaha! yes! that was my first thought, to try to use spotless, but there's no integration out of the box... so I took the easy way out (lazy lol) and went with something that already works and it's also somewhat opinionated 😄

PR #2692, opened 2020-02-18
Fix BigQuery Storage default params

### LINKED ISSUE #2690: @BigQueryType.toTable annotated class can not be used as input for `typedBigQueryStorage`
```scala
  @BigQueryType.toTable
  case class Data(name: String)

  sc.typedBigQueryStorage[Data](Table.Spec(???))
```

fails with exception:
```
INFO: Exception in thread "main" java.util.NoSuchElementException: None.get
INFO: 	at scala.None$.get(Option.scala:529)
INFO: 	at scala.None$.get(Option.scala:527)
INFO: 	at com.spotify.scio.bigquery.syntax.ScioContextOps$.typedBigQueryStorage$extension1(ScioContextSyntax.scala:238)
```

Expected behaviour:
I can define case class with fields (not generated) and use it for reading BigQuery data.

Scio version 0.8.1.

regadas:
Also fixes #2690

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
@@ -249,7 +249,7 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
     self.read(BigQueryTyped.Storage[T](table))(
       BigQueryTyped.Storage.ReadParam(
         bqt.selectedFields.getOrElse(BigQueryStorage.ReadParam.DefaultSelectFields),
-        Some(rowRestriction)
+        Option(rowRestriction)
       )
     )
   }
@@ -261,7 +261,7 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
     self.read(BigQueryTyped.Storage[T](table))(
       BigQueryTyped.Storage.ReadParam(
         BigQueryType[T].selectedFields.getOrElse(BigQueryStorage.ReadParam.DefaultSelectFields),
-        Some(rowRestriction)
+        Option(rowRestriction)
       )
     )
 
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
@@ -124,12 +124,10 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
    */
   def bigQueryStorage(
     table: Table,
-    selectedFields: List[String] = Nil,
-    rowRestriction: String = null
+    selectedFields: List[String] = BigQueryStorage.ReadParam.DefaultSelectFields,
+    rowRestriction: String = BigQueryStorage.ReadParam.DefaultRowRestriction
   ): SCollection[TableRow] =
-    self.read(BigQueryStorage(table))(
-      BigQueryStorage.ReadParam(selectedFields, rowRestriction)
-    )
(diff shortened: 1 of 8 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
```diff
@@ -247,8 +248,8 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
     val table = Table.Spec(bqt.table.get)
     self.read(BigQueryTyped.Storage[T](table))(
       BigQueryTyped.Storage.ReadParam(
-        bqt.selectedFields.get,
-        rowRestriction
+        bqt.selectedFields.getOrElse(BigQueryStorage.ReadParam.DefaultSelectFields),
+        Some(rowRestriction)
```
jto: I'd rather use `Option.apply`

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
```diff
@@ -259,8 +260,8 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
   ): SCollection[T] =
     self.read(BigQueryTyped.Storage[T](table))(
       BigQueryTyped.Storage.ReadParam(
-        BigQueryType[T].selectedFields.get,
-        rowRestriction
+        BigQueryType[T].selectedFields.getOrElse(BigQueryStorage.ReadParam.DefaultSelectFields),
+        Some(rowRestriction)
```
jto: Same

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala
regadas: never hurts being on the safe side

jto: wdym ? `Option.apply` is the safer one

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/taps.scala
```diff
@@ -133,8 +132,8 @@ final case class BigQueryTaps(self: Taps) {
       () =>
         BigQueryStorage(Table.Ref(table)).tap(
           BigQueryStorage.ReadParam(
-            readOptions.getSelectedFieldsList().asScala.toList,
-            readOptions.getRowRestriction()
+            readOptions.getSelectedFieldsList.asScala.toList,
+            Some(readOptions.getRowRestriction)
```
jto: `Option` :)

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/taps.scala
```diff
@@ -151,8 +150,8 @@ final case class BigQueryTaps(self: Taps) {
         BigQueryStorage(Table.Ref(table))
           .tap(
             BigQueryStorage.ReadParam(
-              readOptions.getSelectedFieldsList().asScala.toList,
-              readOptions.getRowRestriction()
+              readOptions.getSelectedFieldsList.asScala.toList,
+              Some(readOptions.getRowRestriction)
```
jto: Option :)

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/taps.scala
regadas: this one is actually ok coz getRowRestriction returns empty String if not set

regadas: changed it coz on a second thought .. i don't wanna rely on third party behavior when i can play safe

PR #2693, opened 2020-02-18
Add SortedBucketTransform

ClaireMcGinty:
I'll add updated API bindings in a followup PR... this one is just the PTransform logic itself.

I tried to re-use functionality from `SortedBucketSink` and `SortedBucketSource` as much as possible, would appreciate any suggestions to make it clearer/more readable 👍

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
@@ -180,8 +180,6 @@ public class SortedBucketTransform<FinalKeyT, FinalValueT> extends PTransform<PB
         bucketShardsToDsts.add(KV.of(bucketShardId, dst));
       }
 
-      final CoGbkResultSchema resultSchema = BucketedInput.schemaOf(sources);
-
       final KeyGroupIterator[] iterators = sources.stream()
           .map(i -> i.createIterator(bucketId, leastNumBuckets))
           .toArray(KeyGroupIterator[]::new);
@@ -204,6 +202,7 @@ public class SortedBucketTransform<FinalKeyT, FinalValueT> extends PTransform<PB
       }
 
       final Map<TupleTag, KV<byte[], Iterator<?>>> nextKeyGroups = new HashMap<>();
+      final CoGbkResultSchema resultSchema = BucketedInput.schemaOf(sources);
       final TupleTagList tupleTags = resultSchema.getTupleTagList();
 
       while (true) {
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
@@ -18,7 +18,7 @@
 package org.apache.beam.sdk.extensions.smb;
 
 import java.io.IOException;
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -163,24 +163,24 @@ public class SortedBucketTransform<FinalKeyT, FinalValueT> extends PTransform<PB
       final int bucketId = c.element();
       final int numSources = sources.size();
 
-      final Map<BucketShardId, ResourceId> bucketShardsToDsts = new HashMap<>();
-      final Map<Integer, Writer<FinalValueT>> shardsToWriters = new HashMap<>();
+      @SuppressWarnings("unchecked")
+      final Writer<FinalValueT>[] writers = new Writer[numShards];
+      final List<KV<BucketShardId, ResourceId>> bucketShardsToDsts = new ArrayList<>();
 
       for (int shardId = 0; shardId < numShards; shardId++) {
         final BucketShardId bucketShardId = BucketShardId.of(bucketId, shardId);
         final ResourceId dst = fileAssignment.forBucket(bucketShardId, numBuckets, numShards);
 
         try {
-          shardsToWriters.put(shardId, fileOperations.createWriter(dst));
+          writers[shardId] = fileOperations.createWriter(dst);
         } catch (IOException e) {
           throw new RuntimeException(e);
         }
 
-        bucketShardsToDsts.put(bucketShardId, dst);
+        bucketShardsToDsts.add(KV.of(bucketShardId, dst));
       }
 
       final CoGbkResultSchema resultSchema = BucketedInput.schemaOf(sources);
-      final TupleTagList tupleTags = resultSchema.getTupleTagList();
 
       final KeyGroupIterator[] iterators = sources.stream()
           .map(i -> i.createIterator(bucketId, leastNumBuckets))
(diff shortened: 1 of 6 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -0,0 +1,254 @@
...
+
+  public SortedBucketTransform(
+      Class<FinalKeyT> finalKeyClass,
+      BucketMetadata<FinalKeyT, FinalValueT> bucketMetadata,
+      ResourceId outputDirectory,
+      ResourceId tempDirectory,
+      String filenameSuffix,
+      FileOperations<FinalValueT> fileOperations,
+      List<BucketedInput<?, ?>> sources,
+      SerializableFunction<KV<FinalKeyT, CoGbkResult>, Iterator<FinalValueT>> toFinalResultT) {
```
nevillelyh: There's no way to construct the output `Iterator` lazily right? The test code creates an eagerly materialized `Iterable`. Should we make it a `BiFunction<KV<...>, ProcessContext, Void>`, so one can write `(kv, ctx) -> ctx.output(...)`?

Also if we provide an more restricted wrapper of `ProcessContext` with only `void output(FinalValueT)`, we can do the `FinalKeyT` sanity check there?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -0,0 +1,273 @@
...
+        writer.close();
+      } catch (IOException e) {
+        throw new RuntimeException("Closing writer failed: ", e);
+      }
+    }
+
+    @Override
+    public void accept(ValueT t) {
+      try {
+        writer.write(t);
```
ClaireMcGinty: Is it weird to add a custom `Counter` here, or are they only supposed to be added in user code? It would be nice to have some instrumentation for users to see how many elements have been written -- as is Dataflow only reports the # of buckets, there's no visibility on actual data size. I couldn't find any examples in Beam of adding Counters to standard library `DoFn`s tho.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -0,0 +1,273 @@
...
+    this.sources = sources;
+    this.transformFn = transformFn;
+    this.bucketMetadata = bucketMetadata;
+  }
+
+  @Override
+  public final WriteResult expand(PBegin begin) {
+    final SourceSpec<FinalKeyT> sourceSpec = SortedBucketSource.getSourceSpec(finalKeyClass, sources);
+
+    Preconditions.checkArgument(
```
ClaireMcGinty: on second thought, I guess the user could specify a larger number of buckets--since at merge time we have access to the key bytes, we could just call `BucketMetadata#getBucketId(byte[] keyBytes)` using the `BucketMetadata` that's supplied for the output, then open a Writer for that bucket file... we would potentially have a lot of open file handles per DoFn though. but it could be a nicer option than offering the round robin shardingg thing.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
nevillelyh: Interesting point. This is a special case since data was read, transformed and written in the same transform, and doesn't go through any DF estimator right? Guess we can make an exception here. But the name needs to be unique?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
nevillelyh: Either way we'll have to keep all bucket/shard files open right? Could be an optimization for later?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
ClaireMcGinty: yeah, there really is no helpful information in the Dataflow UI, heh. I think the name should be okay as long as it's in the `com.spotify.scio.smb` namespace? Maybe `com.spotify.scio.smb.MergedElementsWritten` ? It could be nice to add one for number of elements/key groups read, too.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
ClaireMcGinty: Yeah, I was kind of thinking we could offer this instead of the sharding? At least for a first pass? wdyt?

nevillelyh: Yeah on second thought this should be better than sharding because it enables more parallelism downstream given that the data size is likely growing? :+1:

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
nevillelyh: Is it possible to have multiple SMB sources per workflow though? They'll need some suffix to uniquely identify each?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
ClaireMcGinty: yep! cool, I'll update the PR soon with the change.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
ClaireMcGinty: You mean multiple `SortedBucketSource`  or `SortedBucketTransform`s per workflow? Hm we could suffix with PTransform name? And maybe TupleTag of the BucketedInput if we want to measure elements read at that granularity.

How about I fix up the bucket fanout thing, merge this PR, then follow up with the API bindings/`Counter` instrumentation in separate PRs?

nevillelyh: Yeah sounds good!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
ClaireMcGinty: done! verified in unit test/dataflow. I think this PR is ready to merge

scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/TestFileOperations.java
```diff
@@ -60,8 +66,10 @@ public String readNext() throws NoSuchElementException {
 
       @Override
       public boolean hasNextElement() throws IOException {
-        next = reader.readLine();
-        return next != null;
+        reader.mark(1000);
```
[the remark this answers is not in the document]
ClaireMcGinty: i noticed it wasn't working correctly with `Iterator::forEachRemaining` -- `hasNextElement` should be more of a peek-like operation.

PR #2696, opened 2020-02-19
add ParquetExampleIO

nevillelyh:


scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/package.scala
```diff
@@ -0,0 +1,70 @@
...
+ * Main package for Parquet TensorFlow APIs. Import all.
+ *
+ * {{{
+ * import com.spotify.scio.parquet.tensorflow._
+ * }}}
+ */
+package object tensorflow {
+
+  /** Enhanced version of [[ScioContext]] with Parquet [[Example]] methods. */
+  implicit class ParquetExampleScioContext(private val self: ScioContext) extends AnyVal {
```
regadas: could we follow the same syntax structure as we have in the other modules?

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -0,0 +1,147 @@
...
+      ExampleParquetInputFormat.setFields(job, params.projection.asJava)
+    }
+    if (params.predicate != null) {
+      ParquetInputFormat.setFilterPredicate(job.getConfiguration, params.predicate)
+    }
+
+    val source = HadoopFormatIO
+      .read[JBoolean, Example]()
+      .withKeyTranslation(new SimpleFunction[Void, JBoolean]() {
+        override def apply(input: Void): JBoolean = true // workaround for NPE
```
regadas: this means that this KeyTranslation is always called?

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -0,0 +1,147 @@
...
+    tap(ParquetExampleIO.ReadParam())
+  }
+
+  override def tap(params: ReadP): Tap[Example] =
+    ParquetExapmleTap(ScioUtil.addPartSuffix(path), params)
+}
+
+object ParquetExampleIO {
+  final case class ReadParam private (
+    projection: Seq[String] = null,
```
regadas: I know we use this `null` thing a bit all over the place ... but I'm always sad to see it 😢

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
```diff
@@ -0,0 +1,61 @@
...
+import com.spotify.scio.ScioContext
+import com.spotify.scio.util.ScioUtil
+import org.apache.hadoop.mapreduce.Job
+import org.apache.hadoop.mapreduce.lib.input.FileInputFormat
+
+private[parquet] object GcsConnectorUtil {
+  def setCredentials(job: Job): Unit =
+    // These are needed since `FileInputFormat.setInputPaths` validates paths locally and
+    // requires the user's GCP credentials.
+    sys.env.get("GOOGLE_APPLICATION_CREDENTIALS") match {
```
regadas: I see that this was just moved around, but would it be worth considering using `GoogleCredential` instead? This can be done in another PR. With this it would be consistent with the other IO's

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  final case class WriteParam private (
+    schema: Schema,
+    numShards: Int = WriteParam.DefaultNumShards,
+    suffix: String = WriteParam.DefaultSuffix,
+    compression: CompressionCodecName = WriteParam.DefaultCompression
+  )
+}
+
+case class ParquetExapmleTap(path: String, params: ParquetExampleIO.ReadParam)
```
regadas: should we make it `final`?

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
nevillelyh: Not sure how you can use that. One branch requires the path to key JSON and the other just sets some hard coded key-values in Hadoop `Configuration`.

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
nevillelyh: `workaround for NPE` because Hadoop file IO always emits `<K, V>` and in this case null keys causes coder NPE.

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/package.scala
nevillelyh: That'd "require" redoing `parquet.avro` for consistency's sake? What's the add-on value? Other than binary breakage?

regadas: yeah the main value is consistency and that we should avoid having this in package objects. We can do it in another PR.

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
regadas: i would add this as a comment for the workaround NPE 😄

nevillelyh:
Yeah. It's a trivial change and not likely to pick up much usage (since there's the planned SMB+PQ work) anyway but just paving the way.

PR #2699, opened 2020-02-20
Fix: eval coder eagerly for sealead traits

regadas:
In some cases a ref to magnolia `SealedTrait` is kept. Having some hard time coming up with a concrete test example.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
@@ -25,8 +25,8 @@ object CoderMaterializer {
 
   private[scio] case class CoderOptions(nullableCoders: Boolean, kryo: KryoOptions)
   private[scio] object CoderOptions {
-    def apply(o: PipelineOptions) = {
-      val nullableCoder = o.as(classOf[com.spotify.scio.options.ScioOptions]).getNullableCoders()
+    final def apply(o: PipelineOptions): CoderOptions = {
+      val nullableCoder = o.as(classOf[com.spotify.scio.options.ScioOptions]).getNullableCoders
       new CoderOptions(nullableCoder, KryoOptions(o))
     }
   }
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -77,7 +77,6 @@ object CoderMaterializer {
         )
       case Disjunction(typeName, idCoder, id, coders) =>
         WrappedBCoder.create(
-          // `.map(identity) is really needed to make Map serializable.
```
jto: Why remove this comment ? I remember you getting bitten by the `map` and adding this comment.

scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
```diff
@@ -84,35 +83,34 @@ trait LowPriorityCoderDerivation {
...
+        Coder.disjunction[T, Int](typeName, coders) { t =>
+          sealedTrait.dispatch(t) { subtype =>
+            idx(subtype.typeName)
           }
         }
       }
-    )
+    }
+
+    Ref(typeName, coder)
```
jto: Using `Ref` is not usefull anymore since `coders` is always evaluated eagerly.

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
regadas: uhh, oops (bad friday)... somehow i thought we ere not using `.map(identity)` anymore lol

scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
regadas: ahh yes indeed! 👍

PR #2712, opened 2020-02-25
add api bindings for SortedBucketTransform

ClaireMcGinty:
+ added example usage to `SortMergeBucketExample` in scio-examples

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
@@ -178,7 +178,7 @@ object SortMergeBucketTransformExample {
         .withNumBuckets(2)
         .withNumShards(1)
         .withHashType(HashType.MURMUR3_32)
-    )({
+    ) {
       case (key, (users, accounts), outputCollector) =>
         users.foreach { user =>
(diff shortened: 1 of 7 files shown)


scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
```diff
@@ -154,3 +152,47 @@ object SortMergeBucketJoinExample {
...
+  def main(cmdLineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdLineArgs)
+    val inputL = args("inputL")
+    val inputR = args("inputR")
+    val output = args("output")
+
+    implicit val coder: Coder[GenericRecord] =
+      Coder.avroGenericRecordCoder(SortMergeBucketExample.UserDataSchema)
+
+    sc.sortMergeTransform(
```
nevillelyh: Is there a point to break/curry this into `sc.sortMergeTransform(in1, in2).with(fn).into(out)`, and reuse some bits for 3 & 4 way cases?

ClaireMcGinty: refactored the api a bit -- can you take a look? I think it's ready for merge

PR #2713, opened 2020-02-25
Add SCollection tap

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -769,18 +769,21 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
       this
     }
 
+  /**
+   * Applies f to each element of this [[SCollection]], and returns the original value.
+   *
+   * @group debug
+   */
+  def tap[U](f: T => U)(implicit coder: Coder[T]): SCollection[T] =
+    map { elem => f(elem); elem }
```
ClaireMcGinty: alternately we could implement it like

```
  def tap[U](f: T => U)(implicit coder: Coder[T]): SCollection[T] = {
    map(f)
    this
}
```

so that `map(f)` would be run in parallel with the next transform a user specifies instead of having to complete `f(elem)` for entire collection before moving on to the next step? Since the return type of `f` is discarded it seems safe to assume it can be a leaf-node in the job graph? wdyt?

also does `U` need to be parameterized? Can't `f` just be `T => Any` or `T => Unit` ?

regadas: so `tap` is actually kinda of particular since it allows `side-effects` to take place. As such you might or might not require `f` to complete before passing to the next step. I think i'm leaning towards letting the user make that decision. ¯\_(ツ)_/¯ 

by having a type param i'm basically saying almost the same thing as `T => Any` but I'm not swallowing the type, it's nicer overall (better inference). `T => Unit` is to restrictive as it would force users to actually pass a function of that type if they are not creating an in-place lambda.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -769,18 +769,21 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     out: () => PrintStream = () => Console.out,
     prefix: String = "",
     enabled: Boolean = true
-  ): SCollection[T] =
+  )(implicit coder: Coder[T]): SCollection[T] =
```
jto: Since the `SCollection` keeps the same type, the implicit param is not strictly necessary. You coud get it from `internal.getCoder`. It would avoid a change in type signature.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -769,18 +769,21 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
     } else {
       this
     }
 
+  /**
+   * Applies f to each element of this [[SCollection]], and returns the original value.
+   *
+   * @group debug
+   */
+  def tap[U](f: T => U)(implicit coder: Coder[T]): SCollection[T] =
```
jto: I fid the naming confusing as scio already have `Tap`

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: ah right! yes no need for this neither for the `tap` nice catch!

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: yeah not great but too bad, open to suggestions tbh 👍 otherwise will stick with it

nevillelyh:
Should we merge? Hopefully `tap` will catch on and default `Tap` :sob:

PR #2715, opened 2020-02-26
Remove usage of deprecated beam's AvroIO methods

### LINKED ISSUE #2262: Remove usage of deprecated org.apache.beam.sdk.io.AvroIO methods
`sinkViaGenericRecords` and `RecordFormatter` are both [deprecated](https://github.com/apache/beam/pull/8418); instead, we should be adding an additional map step to explicitly convert from `ElementT => AvroT`. This affects some of our dynamic IO utilities.

jto:
fixes #2262

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/dynamic/syntax/SCollecionSyntax.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/dynamic/syntax/SCollecionSyntax.scala
@@ -27,13 +27,13 @@ import org.apache.avro.file.CodecFactory
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecord
 import org.apache.beam.sdk.coders.StringUtf8Coder
-import org.apache.beam.sdk.io.AvroIO.RecordFormatter
 import org.apache.beam.sdk.io.{Compression, FileIO}
 import org.apache.beam.sdk.{io => beam}
 
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 import java.util.{HashMap => JHashMap}
+import java.nio.channels.WritableByteChannel
 
 object DynamicSCollectionOps {
   private[syntax] def writeDynamic[A](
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/io/dynamic/syntax/SCollecionSyntax.scala
```diff
@@ -122,12 +122,10 @@ final class DynamicGenericRecordSCollectionOps[T <: GenericRecord](private val s
       val nm = new JHashMap[String, AnyRef]()
       nm.putAll(metadata.asJava)
       val sink = beam.AvroIO
-        .sinkViaGenericRecords(schema, new RecordFormatter[T] {
-          override def formatRecord(element: T, schema: Schema): GenericRecord =
-            element
-        })
+        .sink[T](schema)
```
regadas: hm, I think this will use `ReflectDatumWriter` instead of `GenericDatumWriter` which is preferred in this case 😢 

Same thing in the case below.

jto: Good catch. Boy is AvroIO tricky!

jto: So… beam has `AvroIO.Sink`which extends `FileIO.Sink`  and `AvroSink` which extends `FileBasedSink`. `AvroIO.Sink` and `AvroSink` are completelly urealated classes. (That’s not confusing...). `AvroIO.Sink` is implemented using either `RecordFormatter` which is deprecated, or `ReflectDatumWriter` (great…). So **`AvroIO.Sink` is to be avoided.**

`AvroIO` has `AvroOI.Write` and `AvroOI.TypedWrite`.  `AvroOI.Write` is delegating to `AvroOI.TypedWrite` and only exists for compatibility reasons.

`AvroOI.TypedWrite` is implemented using `AvroSink` and does not rely on deprecated APIs or reflection (:tada:) **BUT** it does not support dynamic destination or rather it does **BUT** does not expose the API... 

`AvroIO.Write` does expose `public Write<T> to(DynamicAvroDestinations<T, ?, T> dynamicDestinations)` but `DynamicAvroDestinations` is abstract and does not have a public implementation (other than `constantDestinations`) .

I feel like Beam is trolling me.

jto:
I'm closing the PR. In its current state beam's API does not offer a better implementation. The "right" way would be to use `FileIO#writeDynamic()` but the current implementation of `AvroIO.Sink` is just not good.

PR #2719, opened 2020-02-28
make CoGbkResult Iterable slightly lazy

nevillelyh:


nevillelyh:
@ClaireMcGinty w.r.t. how to expose this in a nicer way, do you think it's possible/worth adding a flag to `SortedBucketSource` to create no-buffer `Iterable` for `CoGbkResult` instead? Once set, `LazyIterable#iterator()` can be called only once and should fail on re-iterating.

ClaireMcGinty:
@nevillelyh I'm ok with that idea! I guess the smb-scio module as whole is still considered experimental, so we can expose this extra-experimental feature with a flag :P

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -338,6 +332,9 @@ final class SortedBucketScioContext(@transient private val self: ScioContext) ex
     toR: CoGbkResult => R
   ) extends Serializable {
 
+    def withoutReiterable: SortMergeTransformReadBuilder[K, R] =
```
nevillelyh: This is the least intrusive for `sortMergeTransform` APIs and doesn't add a default arg that causes method overloading issues.

Doesn't work for `sortMergeCoGroup` though since they don't have intermediate type.
OTOH they don't offer much since co-group is a very thin layer. Maybe whoever needs this can just write Java?

nevillelyh: @ClaireMcGinty thoughts?

OTOH the single global flag via `PipelineOptions` won't work since `reiterable` must be true if it's used in a JOIN.

nevillelyh:
Just discovered another complication: when `reiterable` is turned off, mutation detector & ensure serializable logic will kick in first rendering the `Iterator` unusable in user code in a test.

nevillelyh:
Reworked in #2749

jto:
Reworked in #2749

PR #2733, opened 2020-03-03
 Customize equality in unit tests and better failure message

### LINKED ISSUE #2368: Better way to customize equality in unit tests and better failure message

jto:
Fixes #2368 

### Pretty print common types (based on [lihaoyi/PPrint](https://github.com/lihaoyi/PPrint))
Example from `AutoCompleteTest`:

#### Before:
Screenshot 2020-03-04 at 11 27 35

#### After:
Screenshot 2020-03-04 at 11 17 44

#### Before:

Screenshot 2020-03-04 at 11 43 56

#### After:

Screenshot 2020-03-04 at 14 06 08

### Pretty print `SpecificRecord` instances

Example from `AvroExampleTest`:

#### Before:
Screenshot 2020-03-03 at 14 30 18
#### After:
Screenshot 2020-03-04 at 14 03 16

## Custom `Eq`

See `SCollectionMatchersTest`:

https://github.com/spotify/scio/blob/13b21c92b8a5f723c9146dbb96726b8dca35fc5e/scio-test/src/test/scala/com/spotify/scio/testing/SCollectionMatchersTest.scala#L592-L623

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
@@ -35,12 +35,31 @@ import org.hamcrest.MatcherAssert.assertThat
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
 import com.twitter.chill.ClosureCleaner
+import org.apache.avro.generic.GenericRecord
+import org.apache.avro.specific.SpecificRecordBase
+
+private[testing] case class TestWrapper[T](value: T) {
+  override def toString: String = Pretty.printer.apply(value).render
+  override def equals(other: Any): Boolean =
+    other match {
+      case TestWrapper(o) if value.getClass.isArray && o.getClass().isArray() =>
+        value.asInstanceOf[Array[_]].sameElements(o.asInstanceOf[Array[_]])
+      case TestWrapper(o) => value.equals(o)
+      case _              => value.equals(other)
+    }
+}
+
+object TestWrapper {
+  def wrap[T: Coder](coll: SCollection[T]) =
+    coll.map(t => TestWrapper(t))
+}
 
 /**
  * Trait with ScalaTest [[org.scalatest.matchers.Matcher Matcher]]s for
  * [[com.spotify.scio.values.SCollection SCollection]]s.
  */
 trait SCollectionMatchers {
+
   sealed trait MatcherBuilder[T] {
     _: Matcher[T] =>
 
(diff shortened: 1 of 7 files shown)


jto:
Note: I did not touch `equalMapOf` because it seems that touching it would probably change its behaviour.

scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
```diff
@@ -35,12 +35,65 @@ import org.hamcrest.MatcherAssert.assertThat
...
+    coll.map(t => TestWrapper(t))
+
+  implicit def testWrapperCoder[T: Coder: Eq]: Coder[TestWrapper[T]] =
+    Coder.gen[TestWrapper[T]]
+}
+
+private object ScioMatchers {
+  private def supplierFromCoder[A: Coder, B](@transient a: A)(builder: A => B) = {
+    val options = PipelineOptionsFactory.create().as(classOf[ScioOptions])
+    options.setNullableCoders(true)
```
regadas: @jto can you add a comment on why do we need to set this?

jto: Good catch! It was not necessary. Fixed it :)

jto:
@nevillelyh wdyt ?

nevillelyh:
How would it work in CI, i.e. if it doesn't support color output? `pprint` should auto detect and turn it off?

Also this is breaking, although IDK any lib using the API. Should be part of 0.9?

jto:
@regadas Good point. I changed it

jto:
@nevillelyh 

> How would it work in CI, i.e. if it doesn't support color output? pprint should auto detect and turn it off?

I would hope that any half decent CI would either ignore ANSI codes or actually support them. I'll check.

> Does this imply that users will have to import cats Eq instances for primitives etc? What about case classes? Using our magnolify-cats?

The default behaviour is to use java's `equals` (since `SCollectionMatchers extends EqInstances`). Any other implicit in scope can override the behaviour. 

> Also this is breaking, although IDK any lib using the API. Should be part of 0.9?

In Scio, it did not break anything. For the vast majority the upgrade should be source compatible. It is however indeed a breaking change (mostly bc. the return type of the matchers changed a bit).

jto:
I think I can make it back compatible. Do not merge it yet.

jto:
I don't like the fact that `TestWrapper` is visible in the API signature but making it a pure implementation details means that it has to be` Serialiazable` (cuz in a closure), which seems impossible.

jto:
`TestWrapper` does not leak anymore 🎉

scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
```diff
@@ -29,18 +29,149 @@ import org.apache.beam.sdk.transforms.SerializableFunction
...
+}
 
 /**
  * Trait with ScalaTest [[org.scalatest.matchers.Matcher Matcher]]s for
  * [[com.spotify.scio.values.SCollection SCollection]]s.
  */
-trait SCollectionMatchers {
+trait SCollectionMatchers extends EqInstances {
+
+  import ScioMatchers.{makeFn, makeFnSingle}
```
regadas:
```suggestion
  import ScioMatchers.makeFn
```

regadas: The other one is not used ... either this or replace `makeFn` with `ScioMatchers.makeFn` up to you 😄

PR #2743, opened 2020-03-04
Add ProtobufUtil for Message<->Avro conversions

ClaireMcGinty:


scio-core/src/main/scala/com/spotify/scio/util/ProtobufUtil.scala
```diff
@@ -0,0 +1,72 @@
...
+   * A function that converts a Protobuf [[Message]] of type `T` into a [[GenericRecord]]
+   * whose [[Schema]] is a single byte array field, corresponding to the serialized bytes in `T`.
+   */
+  def toAvro[T <: Message: ClassTag]: T => GenericRecord = {
+    val protoCoder = CoderMaterializer.beamWithDefault(protoCoderOf[T])
+
+    (t: T) => AvroBytesUtil.encode(protoCoder, t)
+  }
+
+  private[scio] def protoCoderOf[T <: Message: ClassTag]: Coder[T] =
```
regadas: we can remove this i guess since:

```scala
def protoMessageCoder[T <: Message: ClassTag]: Coder[T]
```

ClaireMcGinty: you mean the implicit def in `com.spotify.scio.coders.instances.ProtobufCoders`? I can get rid of this method then 👍 
Although DynamicProtobufSCollectionOps#saveAsDynamicProtobufFile only has an `(implicit ct: ClassTag[T])` in its signature, so I either need to modify that to take`(implicit ct: ClassTag[T], coder: Coder[T])`, or within the method itself do something like `val protoCoder = implicitly[Coder[T]]`.

ClaireMcGinty: I guess another reason I prefer it not being implicit is because the Coder is used to serialize the bytes that will actually be written as Avro (`AvroBytesUtil#encode`), rather than just for intermediate serialization. So if the user defines their own overridden implicit coder, their written records may be in an incompatible format

regadas: TIL 😄

PR #2749, opened 2020-03-06
make CoGbkResult Iterable slightly lazy

nevillelyh:


nevillelyh:
I didn't add test for `SortedBucketSource` yet since the existing ones assert on the `PCollection<KV<K, CoGbkResult>>` whose `Iterable`s are vanilla ones after ser/de. So the cast to `LazyIterable` would fail.

nevillelyh:
Tested that this trick only works with `SortedBucketTransform`, not `SortedBucketSource`. For `SBS`, the `ParDo` following the source failed because `Caused by: java.lang.IllegalStateException: Iterator already started`, most likely because element sampling mechanism in the UI mutated the iterator, interfering with with our laziness...

nevillelyh:
@ClaireMcGinty at this point I'm actually regretting exposing the output as `KV<K, CoGbkResult>` just for the sake of consistency with Beam API. It'd actually be much easier if we had exposed it as a `transformFn` like in `SortedBucketTransform`.

nevillelyh:
Gonna merge this to make the next optimization (`KeyGroupIterator`) easier.

PR #2756, opened 2020-03-10
Add directory treatment and compression option to readFiles

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1126,21 +1127,39 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+
+  /**
+   * Reads each file, represented as a pattern, in this [[SCollection]].
+   *
+   * @see [[readFilesAsBytes]], [[readFilesAsString]]
+   *
+   * @param directoryTreatment Controls how to handle directories in the input.
+   * @param compression  Reads files using the given [[org.apache.beam.sdk.io.Compression]].
+   */
+  def readFiles[A: Coder](directoryTreatment: DirectoryTreatment, compression: Compression)(
```
jto: Should we have default paremeter values ? Unless I am missing smth, this PR introduces breaking changes in Scio's API.

regadas: actually this method is good `L1128` but I missed the following

```scala
  def readFiles[A: Coder](
    filesTransform: PTransform[PCollection[beam.FileIO.ReadableFile], PCollection[A]],
    directoryTreatment: DirectoryTreatment = DirectoryTreatment.SKIP,
    compression: Compression = Compression.AUTO
  )(implicit ev: T <:< String): SCollection[A] =
```

Issue #2752, opened 2020-03-09
Single file `saveAs*` functions


martinbomio:
We have some use cases where we want to save only one object to a file, usually with a fixed name, removing the sharding information stored in the file name. We do it for JSON file or a tf.Schema file. 

We end up writing implicit classes that extend the scio context (for reading) and Scollection of our type (for writing) and we thought it might be something that other people might want.

We could also implement those with beam's `FileSystem`, but this will mean we cannot test it with `JobTest`

regadas:
Hi @martinbomio, can you provide an example of what you're doing?

Also have you consider using `saveAsCustomOutput ` instead of implicit class `SCollection`?

martinbomio:
Hi @regadas these are two examples I've ended up doing in the past:
```scala
    def saveAsSchemaSnapshotProtobuf(
      path: String,
      name: String = "schema_snapshot.pb",
      compression: Compression = Compression.UNCOMPRESSED
    ): Unit =
      if (col.context.isTest) {
        col.map(_.toByteArray).saveAsBinaryFile(path + "/" + name, compression = compression)
      } else {
        val io = FileIO
          .write[Array[Byte]]
          .to(path)
          .via(new BytesSink())
          .withCompression(compression)
          .withNumShards(1)
          .withNaming(fixedFileName(name))
        col.map(_.toByteArray).internal.apply(io)
      }
```
and 
```
    def saveAsSingleTextFile(
      path: String,
      header: Option[String] = None,
      suffix: String = ".tsv",
      compression: Compression = Compression.UNCOMPRESSED
    ): Unit = {
      if (self.context.isTest) {
        self.saveAsTextFile(path, suffix = suffix, compression = compression)
      } else {
        val dest = header match {
          case Some(h) =>
            TextIO
              .write()
              .to(path)
              .withoutSharding()
              .withSuffix(suffix)
              .withHeader(h)
              .withWritableByteChannelFactory(
                FileBasedSink.CompressionType.fromCanonical(compression)
              )
          case _ =>
            TextIO
              .write()
              .to(path)
              .withoutSharding()
              .withSuffix(suffix)
              .withWritableByteChannelFactory(
                FileBasedSink.CompressionType.fromCanonical(compression)
              )
        }

        self.internal.apply(dest)
      }
```

I didn't know about `saveAsCustomOutput`, it will be nice to use it inside the implicit class, still think that having the implicit class which will contain the `PTransform` is nice.

jto:
[...]

PR #2758, opened 2020-03-10
Add Header and Footer support to TextIO

regadas:


scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
```diff
@@ -56,8 +56,8 @@ final case class TextIO(path: String) extends ScioIO[String] {
   override def tap(params: ReadP): Tap[String] =
     TextTap(ScioUtil.addPartSuffix(path))
 
-  private def textOut(path: String, params: WriteP) =
-    BTextIO
+  private def textOut(path: String, params: WriteP) = {
+    var transform = BTextIO
       .write()
       .to(pathWithShards(path))
```
martinbomio: this means that there will be a `part` added to the output path, which means the caller will not be able to fully control the single file name

regadas: hm not really, `pathWithShards` is actually a misleading name since what it does is to remove a trailing slash. :smile:

martinbomio:
@regadas so to replicate the functionality expressed in #2752 one should do:
```
saveAsTextFile(path, numShards = 1, suffix='tsv')
``` 
and the path will need to contain the file name? I believe the current implementation will add  a `part` prefix to the file name. So if `path = "gcs://buclet/path/filename"`  `saveAsTextFile` will end up writing a file named: `gcs://buclet/path/filename/part.tsv`

regadas:
@martinbomio yeah once I fix the part thing to replicate what you had you will also need to add a `header` param

martinbomio:
@regadas yep makes sense

PR #2764, opened 2020-03-11
Fix #2761: make java Enum coder deterministic

### LINKED ISSUE #2761: Java Enum coders are non deterministics
Going from Scio 0.7.4 -> 0.8.1+ breaks GBK operations where the key contains a Java Enum.

This is because the derived java enum coder is not deterministic, as it uses java serialization instead of kryo from 0.7.4. Since Java serialization is not deterministic, it fails.

This bug is related to this change:
https://github.com/spotify/scio/pull/2065/files

Here is a PR with a test case that for determinism. #2762

regadas:
Fixes #2761
Closes #2763

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
@@ -103,7 +103,10 @@ trait JavaCoders extends JavaBeanCoders {
     )
 
   implicit def coderJEnum[E <: java.lang.Enum[E]: ClassTag]: Coder[E] =
-    Coder.beam(SerializableCoder.of(ScioUtil.classOf[E]))
+    Coder.xmap(Coder[String])(
+      value => java.lang.Enum.valueOf(ScioUtil.classOf[E], value),
+      _.name
+    )
 }
 
 trait JavaBeanCoders {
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
```diff
@@ -103,7 +103,10 @@ trait JavaCoders extends JavaBeanCoders {
     )
 
   implicit def coderJEnum[E <: java.lang.Enum[E]: ClassTag]: Coder[E] =
-    Coder.beam(SerializableCoder.of(ScioUtil.classOf[E]))
+    Coder.xmap(Coder[String])(
```
jto: This *should* be strictly the same as `SerializableCoder` since `Enum` are always serialized as a string ([source](https://docs.oracle.com/javase/6/docs/platform/serialization/spec/serial-arch.html#6469)).
Can we not just use a wrapper around `SerializableCoder` that forces `verifyDeterministic` to return true ? 
It would give use the guarantee that the serialization does not change between Scio versions.

regadas: we can but I'm not sure if there's a strong reason to? maybe if we had more cases like this?

regadas: Talked with @jto and this is probably fine for now

Issue #2706, opened 2020-02-25
User-defined args show as lisp-case in --help output and error messages, but only parse as camelCase


danxmoran:
I have a pipeline with an argument class:
```scala
case class Args(
  inputPrefix: String,
  outputPrefix: String
)
```
I parse the args using `ContextAndArgs.typed`. When I run the pipeline with `--help`, I see:
```
--input-prefix  <string>
--output-prefix <string>
```
However after #1780 these are invalid. Running the pipeline with camelCase works fine, but error messages are _really_ confusing when people try to pass in the lisp-case variants:
```
sbt> run --input-prefix foo --output-prefix bar
[error] Exception in thread "main" java.lang.Exception: Required option --input-prefix not specified
[error] Required option --output-prefix not specified
```
I took a look at `ContextAndArgs` and found a call to a `hyphenizeArg` method [here](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L197). Would removing that call fix the problem? If so, I'd be happy to submit a PR.

regadas:
Hi @danxmoran thanks for the report. Unfortunately I don't think it's that easy. Gonna take a deeper look.

regadas:
closed via #2777

PR #2773, opened 2020-03-16
Fix: full camelCase typed args support

regadas:
fixes: #2706


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -194,7 +198,15 @@ object ContextAndArgs {
           case _ => true
         }
 
-      CaseApp.detailedParseWithHelp[T](customArgs.map(hyphenizeArg)) match {
+      val result =
```
regadas: There's a fix for caseapp that makes this obsolete

PR #2714, opened 2020-02-26
Cache schemas

jto:
fixes #2648

PR #2775, opened 2020-03-17
Add schema cache support in BigQueryType

regadas:
https://github.com/spotify/scio/pull/2714#pullrequestreview-370710821

### no cache
```
[info] Benchmark                         Mode  Cnt      Score       Error  Units
[info] BigQueryAvroBenchmark.convert     avgt   10  47635.637 ± 1167.784  ns/op
```

### cache
```
[info] Benchmark                         Mode  Cnt     Score     Error  Units
[info] BigQueryAvroBenchmark.convert     avgt   10  2350.097 ± 9.759  ns/op
```

```
 BigQueryType.fromAvro[FooBar](BigQueryType.toAvro[FooBar](FooBar(...)))
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -601,6 +601,8 @@ lazy val `scio-bigquery`: Project = project
   .settings(
     description := "Scio add-on for Google BigQuery",
     libraryDependencies ++= Seq(
+      //this dep seems to be required only when compilling with 2.11
+      "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % "provided",
       "org.apache.beam" % "beam-vendor-guava-26_0-jre" % beamVendorVersion,
       "com.twitter" %% "chill" % chillVersion,
       "com.google.protobuf" % "protobuf-java" % protobufVersion,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -601,6 +601,8 @@ lazy val `scio-bigquery`: Project = project
   .settings(
     description := "Scio add-on for Google BigQuery",
     libraryDependencies ++= Seq(
+      //this dep seems to be required only when compilling with 2.11
+      "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % "provided",
```
nevillelyh: This shouldn't be "provided" though? We're not expecting users to explicitly add `caffeine` dependency right?
Should we use a simple java `ConcurrentHashMap` instead?

regadas: Yeah thought about that ... but then I figured that this can probably go in `scio 0.9` and it's not needed. I probably should update the milestone

nevillelyh: Looks like `caffeine` is pretty self-contained so maybe it's fine to use it. But it shouldn't be "provided" right?

regadas: it should be ok to be "provided" JVM should only load the referenced classes and in this case is not referenced. Are you seeing issues?

nevillelyh: Oh because this is only referenced at compile time?

regadas: Yup, and it's only required by the scala 2.11 compiler ¯\_(ツ)_/¯. Should be ok, once we drop 2.11 this line will be gone.

nevillelyh: Good to merge then?

regadas: 👍

PR #2779, opened 2020-03-18
optimize KeyGroupIterator

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/KeyGroupIterator.java
```diff
@@ -17,69 +17,92 @@
...
+        iterators.stream().map(Iterators::peekingIterator).collect(Collectors.toList());
     this.keyFn = keyFn;
     this.keyComparator = keyComparator;
   }
 
   @Override
   public boolean hasNext() {
     checkState();
-    return iterator.hasNext();
+    return iterators.stream().anyMatch(it -> it.hasNext() && it.peek() != null);
```
nevillelyh: I'm not sure this 100% correct. The `!= null` condition is basically treating `null` key as not a real key but it also seems that our E2E test doesn't cover cases with multiple `null` keys that might mess up the iterator state?

nevillelyh: This might be a red herring and caused by the bug in `JsonFileOperations`.

PR #2782, opened 2020-03-18
Cogroup on Seq

psilospore:
## Description

I wanted to use cogroup on an undefined number of `SCollections[(KEY, A)]`
There are cogroup's that I can use on multiple SCollections from 2 to 22.
```
  def cogroup[KEY: Coder, A: Coder, B: Coder](a: SCollection[(KEY, A)], b: SCollection[(KEY, B)]): SCollection[(KEY, (Iterable[A], Iterable[B]))] = {…}
```
However I have a Seq[SCollection[(KEY, A)]].

Something more like this would work for me:
```
  def cogroup[KEY: Coder, A: Coder](l: Seq[SCollection[(KEY, A)]): SCollection[(KEY, Iterable[A])] = ???
```
Since the value in the SCollection is of the same type I can return an Iterable instead of a Tuple.

Let me know what you all think.

## TODO
- Handle error cases: Seq of 1
- Minor refactoring
- Tests?

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/util/MultiJoin.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/MultiJoin.scala
@@ -40,6 +40,17 @@ trait MultiJoin extends Serializable {
 
   def toOptions[T](xs: Iterator[T]): Iterator[Option[T]] = if (xs.isEmpty) Iterator(None) else xs.map(Option(_))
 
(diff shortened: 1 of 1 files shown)


nevillelyh:
Can you use `sc.unionAll(l).groupByKey` instead? Essentially the same thing, and likely more performant since it eliminates the individual lazy `Iterable[V]`s.

psilospore:
@nevillelyh thanks! That was the operation I wanted.

PR #2785, opened 2020-03-20
Initial flink runner support

regadas:


build.sbt
```diff
@@ -456,6 +469,12 @@ lazy val `scio-core`: Project = project
       "org.apache.beam" % "beam-runners-spark" % beamVersion % Provided exclude (
         "com.fasterxml.jackson.module", "jackson-module-scala_2.11"
       ),
+      "org.apache.beam" % "beam-runners-flink-1.9" % beamVersion % Provided excludeAll (
```
nevillelyh: Wait so they only publish `FlinkRunner` for 2.11?

regadas: indeed!

jto: Does it mean that it will only work with scala 2.11 ?

build.sbt
```diff
@@ -54,6 +54,7 @@ val elasticsearch5Version = "5.6.16"
 val elasticsearch6Version = "6.8.7"
 val elasticsearch7Version = "7.6.1"
 val featranVersion = "0.5.0"
+val flinkVersion = "1.9.2"
```
jto: Why not 1.10 ?

regadas: I'm actually keeping tabs with beam since the runners are `flink` "specific"

build.sbt
regadas: it works as well with 2.12 ... The job I showed you as running on a 2.12 cluster

build.sbt
ibzib: FYI Flink 1.10 will be supported in upcoming Beam releases (starting with Beam 2.21 IIRC).

Since Flink jobs are not backwards-compatible with older clusters, I suggest supporting multiple Flink versions going forward if possible, as Beam does (Beam usually tries to support the latest 3 Flink releases).

https://github.com/apache/beam/tree/master/runners/flink

Issue #2787, opened 2020-03-20
BQ Write with Schema fails with NPE


anish749:
Trying to write a `SCollection[T: Schema]` using `.saveAsBigQuery(..)` I get a NPE because Beam doesn't know the `TableSchema`

This originates from [here](https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L730)

regadas:
Hi @anish749 can you provide the full stack trace? This sounds s bit fishy since `TableSchema` should be derived from the underlying `PCollection` `Schema` which is set in the line below?

anish749:
You are right, it is not the one from where I thought it originates.
I am not having the stack trace in a parseable format to paste here, and it is in google log viewer. Let me share it privately.

RustedBones:
BQ based on beam schema has been dropped in 0.12. Closing the issue

PR #2788, opened 2020-03-20
Pass table schema for BQ write using schema

anish749:
Try out a fix for #2787


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -726,8 +726,9 @@ object BigQueryTyped {
       sc.read(underlying)
 
     override protected def write(data: SCollection[T], params: WriteP): Tap[T] = {
+      val (beamSchema, _, _) = SchemaMaterializer.materialize(Schema[T])
       val ps = BigQueryTypedTable.WriteParam(
-        null,
+        BigQueryUtils.toTableSchema(beamSchema),
```
regadas: This `TableSchema` is not used when `useBeamSchema` is set

PR #2797, opened 2020-03-24
add BucketMetadataUtil

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadataUtil.java
```diff
@@ -0,0 +1,167 @@
...
+  ////////////////////////////////////////////////////////////////////////////////
+
+  private static FileAssignment getFileAssignment(ResourceId directory, String filenameSuffix) {
+    return new SMBFilenamePolicy(directory, filenameSuffix).forDestination();
+  }
+
+  private static <K, V> Optional<BucketMetadata<K, V>> getMetadata(FileAssignment fileAssignment) {
+    final ResourceId resourceId = fileAssignment.forMetadata();
+    try {
+      InputStream inputStream = Channels.newInputStream(FileSystems.open(resourceId));
```
jto: That `InputStream` is never closed. Is it expected ?

nevillelyh: IIRC the underlying Jackson `ObjectMapper` closes the stream but let me just put a break point on it and verify. BRB once I get intellij to build.

nevillelyh: Yep verified that `ObjectMapper` will close it after reading value.

nevillelyh: Will do in a separate commit since I need to fix up `BucketMetadata.from` a bit.

nevillelyh:
@ClaireMcGinty WDYT?

PR #2803, opened 2020-03-26
Add support for default transform names

### LINKED ISSUE #2760: Allow overriding transform default names
See https://github.com/spotify/scio/pull/2735#pullrequestreview-370466469 for context.

anish749:
Fixes #2760

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -279,7 +279,7 @@ final case class BigQueryTypedTable[T: Coder](
 
   override protected def read(sc: ScioContext, params: ReadP): SCollection[T] = {
     val io = reader.from(table.ref).withCoder(CoderMaterializer.beam(sc, Coder[T]))
-    sc.wrap(sc.pipeline.apply(s"Read BQ table ${table.spec}", io))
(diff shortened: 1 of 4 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -279,7 +279,7 @@ final case class BigQueryTypedTable[T: Coder](
 
   override protected def read(sc: ScioContext, params: ReadP): SCollection[T] = {
     val io = reader.from(table.ref).withCoder(CoderMaterializer.beam(sc, Coder[T]))
-    sc.wrap(sc.pipeline.apply(s"Read BQ table ${table.spec}", io))
+    sc.wrap(sc.applyInternal(s"Read BQ table ${table.spec}", io))
```
regadas: I think eventually we will need to track down all the custom names we have and add this.

regadas:
@anish749 Thanks for this!

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
anish749: I didn't find any other usage of this. This was the only place.

regadas: oh good to know! The good thing about this change is that we can now safely add more default human readable names to transforms! which is a big plus

PR #2805, opened 2020-03-26
Add more convinience methods that support default transform names

regadas:


nevillelyh:
This will change the names of transforms without explicit name right? Can we save this for 0.9?

regadas:
It will not change the names ... but we can move it to 0.9.x

PR #2808, opened 2020-03-26
Add covary method to lift SCollection to the specified type

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -210,6 +210,9 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   // Collection operations
   // =======================================================================
 
+  /** lifts this [[SCollection]] to the specified type */
+  def covary[U]: SCollection[U] = this.asInstanceOf[SCollection[U]]
```
nevillelyh: Should it be `private`?

regadas: I'm ok with it being `public` it can be useful and it's doesn't seem harmful?

jto: In that case could it have type bounds or an implicit witness ?

regadas: On a second thought yeah this can be harmful.

@jto yeah that sounds good

PR #2809, opened 2020-03-26
Update scaladoc api mappings

regadas:
This adds a little plugin that removes the need for all the cruft around `apiMappings`

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -19,6 +19,7 @@ addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.3.5")
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.1.5")
 addSbtPlugin("com.lightbend.sbt" % "sbt-java-formatter" % "0.5.1")
 addSbtPlugin("com.github.cb372" % "sbt-explicit-dependencies" % "0.2.12")
+addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "3.0.0")
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


project/plugins.sbt
```diff
@@ -19,6 +19,7 @@ addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.3.5")
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.1.5")
 addSbtPlugin("com.lightbend.sbt" % "sbt-java-formatter" % "0.5.1")
 addSbtPlugin("com.github.cb372" % "sbt-explicit-dependencies" % "0.2.12")
+addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "3.0.0")
```
nevillelyh: Do they have hard coded mapping for Dataflow, BigQuery JavaDocs etc.?
TBH those are nice to have I don't mind not having the link & keep this simple.

regadas: The plugin resolves the javadoc.io links for them ... so when we have `PCollection` in a scaladoc ... users can actually jump to the reespective javadoc

nevillelyh: I'm just curious how does it know the URL prefix for say Beam JavaDoc?

nevillelyh: https://github.com/ThoughtWorksInc/sbt-api-mappings/tree/3.0.x/src/main/scala/com/thoughtworks/sbtApiMappings
So looks like a bunch of hard coded ones including JDK lib, spark, play, but that's about it. 🤷‍♂

regadas: for most of the deps will use the sonatype mapping rule.

PR #2817, opened 2020-03-30
Reuse ApproxFilter

anish749:
Like we have apis that allows reusing side inputs for Hash joins / hashIntersect, it would be nice to use ApproxFilters in the same way.

This allows pre-filtering of multiple SCollections before using them in `MultiJoin` steps with more than 1 collection.

Example:
```scala
val si = pairSCollRight.asApproxFilterSideInput(BloomFilter, 1000000)
val filtered1 = pairSColl1.sparseIntersectByKey(si)
val filtered2 = pairSColl2.sparseIntersectByKey(si)
```

Creating a draft for some early feedback before writing tests and documentation.

This API is simpler and aimed at smaller SCollections that fit in one ApproxFilter and doesn't need partitioning.


nevillelyh:
LGTM should we merge?

anish749:
@nevillelyh Perfect, I'll add some docs on the companion object methods, and a test cases before merging.

PR #2823, opened 2020-03-30
Fix: cancel job on waitUntilFinish timeout

### LINKED ISSUE #2813: Cancel Dataflow job on timeout?
This looks like a regression on Scio 0.8: I think Scio 0.7.x had [some logic to cancel DF jobs](https://github.com/spotify/scio/blob/v0.7.4/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala#L80) when timing out on `waitUntilDone()`/`waitUntilFinish()`.
It seems this behaviour is [no longer available on Scio 0.8.x](https://github.com/spotify/scio/blob/v0.8.3/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala).

Not sure if intentional or not.

Still, when using Beam’s PipelineResult users can do something [like this](https://github.com/spotify/dbeam/blob/346e838087e8d5b67588071491e5956d243321fc/dbeam-core/src/main/java/com/spotify/dbeam/beam/BeamHelper.java#L43) to have the same behaviour.

regadas:
Fixes #2813

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -612,10 +612,14 @@ class ScioContext private[scio] (
           case Duration.Inf => -1
           case d            => d.toMillis
         }
+        // according to PipelineResult returns null on timeout
         val state = pipelineResult.waitUntilFinish(time.Duration.millis(wait))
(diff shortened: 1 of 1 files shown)


labianchin:
Nice!

QQ: Will this be released under the 0.8.x series or will we have to wait for 0.9.x?

PR #2826, opened 2020-03-31
Fix Sharded Sparkey string hashing behaviour for strings longer than one character.

psobot:
Sharded Sparkey [was added back in Scio 0.8.0](https://github.com/spotify/scio/pull/2336) and had some [bugfixes that landed in 0.8.2](https://github.com/spotify/scio/pull/2667) - here's another one!

It turns out that `MurmurHash3`, which we're using for distributing data across the shards of a sharded Sparkey fileset, uses a different hashing algorithm if passed a `String` object vs when it's passed an `Array[Byte]`. Namely:

```scala
// if (str.length > 1)
MurmurHash3.stringHash(str, seed) != MurmurHash3.bytesHash(str.getBytes, seed) 
```

Unfortunately, Scio was using these methods interchangeably, and used `bytesHash` in one place and `stringHash` in another. This wasn't caught in tests because the tests used single-character keys, and this inconsistency only appears if the strings being hashed are longer than a single character.

This is a known issue [documented on StackOverflow](https://stackoverflow.com/a/46472986/679081):
> `MurmurHash3.bytesHash` and python's `mmh3.hash` pass characters to the hashing mixer in groups of 4, but `MurmurHash3.stringHash` mixes characters in groups of 2. This means that the two hash functions return completely different outputs.

This PR fixes the issue by:
 - using `stringHash` consistently when the input is a `String`
 - altering the tests to use multi-character keys
 - adding additional assertions to the test

Special thanks to @drubinstein for identifying this bug.

### DIFF
--- a/scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
+++ b/scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
@@ -140,6 +140,12 @@ class SparkeyTest extends PipelineSpec {
     val readers = basePaths.map(basePath => Sparkey.open(new File(basePath)))
     readers.map(_.toMap.toList.toMap).reduce(_ ++ _) shouldBe bigSideData.toMap
 
+    val shardedReader = sparkeyUri.getReader
+    shardedReader.toMap shouldBe bigSideData.toMap
+
+    bigSideData.foreach { case (expectedKey, expectedValue) =>
+      shardedReader.get(expectedKey) shouldBe Some(expectedValue) }
+
     FileUtils.deleteDirectory(new File(sparkeyUri.basePath))
   }
 
@@ -209,7 +215,7 @@ class SparkeyTest extends PipelineSpec {
 
     val input = Seq("a", "b", "a", "b")
 
-    val sparkey = sc.parallelize(sideData).asSparkey(numShards = 2)
+    val sparkey = sc.parallelize(sideData).asSparkey(numShards = 10)
     val sparkeyMaterialized = sparkey.materialize
     val si = sparkey.asSparkeySideInput
     val result = sc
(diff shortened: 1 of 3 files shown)


scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
```diff
@@ -59,7 +59,13 @@ object TestCache {
 }
 
 class SparkeyTest extends PipelineSpec {
-  val sideData = Seq(("a", "1"), ("b", "2"), ("c", "3"))
+  /* We're using keys longer than a single character here to trigger edge-case behaviour
```
regadas: this tl:dr is useful 👍

scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
```diff
@@ -140,6 +146,12 @@ class SparkeyTest extends PipelineSpec {
     val readers = basePaths.map(basePath => Sparkey.open(new File(basePath)))
     readers.map(_.toMap.toList.toMap).reduce(_ ++ _) shouldBe bigSideData.toMap
 
+    val shardedReader = sparkeyUri.getReader
+    shardedReader.toMap shouldBe bigSideData.toMap
+
+    bigSideData.foreach { case (expectedKey, expectedValue) =>
+      shardedReader.get(expectedKey) shouldBe Some(expectedValue) }
```
regadas: this will need to be formatted

psobot: D'oh! Forgot the cardinal rule of contributing to Scio - make sure everything runs through `sbt scalafmt test:scalafmt scalafmtSbt` first 🤦‍♂

PR #2844, opened 2020-04-03
rework implicits/syntax for scio-extra bigquery package

ClaireMcGinty:
now, implicits are in an `SCollectionSyntax` trait that's extended by the package object, consistent with the other implicit SCollection classes in scio. I made the `ToTableRow` and `ToTableSchema` private traits so they're harder for users to misuse -- they can only access the public methods in the `AvroConverters` object.

this *is* a small breaking change but I think use of this package is pretty low and the next release will be a major one, so we can just add it to release notes.


scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -0,0 +1,75 @@
...
+
+import scala.reflect.ClassTag
+
+trait SCollectionSyntax {
+  implicit def toAvroToBigQuerySCollection[T <: IndexedRecord: ClassTag](
+    data: SCollection[T]
+  ): AvroToBigQuerySCollection[T] = new AvroToBigQuerySCollection[T](data)
+}
+
+final class AvroToBigQuerySCollection[T <: IndexedRecord: ClassTag](
```
regadas, line 39: 🎨

```suggestion
final class AvroToBigQuerySCollectionOps[T <: IndexedRecord: ClassTag](
```

most of these are normally suffixed with `Ops`

scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/ToTableSchema.scala
```diff
@@ -17,9 +17,9 @@
 
 package com.spotify.scio.extra.bigquery
 
-import com.google.api.services.bigquery.model.{TableFieldSchema, TableSchema}
-import com.spotify.scio.annotations.experimental
-import com.spotify.scio.extra.bigquery.Implicits.AvroConversionException
+import AvroConverters.AvroConversionException
```
regadas, line 20: 🎨 fq import

```suggestion
import com.spotify.scio.extra.bigquery.AvroConverters.AvroConversionException
```

scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/syntax/SCollectionSyntax.scala
ClaireMcGinty: 👍

PR #2845, opened 2020-04-03
check jobReference.location for query location

nevillelyh:


scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryClientIT.scala
```diff
@@ -85,7 +85,7 @@ class BigQueryClientIT extends AnyFlatSpec with Matchers {
   }
 
   it should "support missing source tables" in {
-    bq.query.extractLocation("SELECT 6") shouldBe None
+    bq.query.extractLocation("SELECT 6") shouldBe Some("EU")
```
nevillelyh, line 88: So this means the query will be executed in EU despite not having any source table.
I'm not sure how it defaults to EU though, a project-wide setting?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
```diff
@@ -310,21 +310,29 @@ final private[client] class QueryOps(client: Client, tableService: TableOps, job
...
-      .map {
-        case (pId, dId) =>
-          val l = client.underlying.datasets().get(pId, dId).execute().getLocation
-          if (l != null) l else BigQueryConfig.location
-      }
-    require(locations.size <= 1, "Tables in the query must be in the same location")
-    locations.headOption
+    val job = run(QueryJobConfig(sqlQuery, dryRun = true, useLegacySql = isLegacySql(sqlQuery)))
+    job.map(_.getJobReference.getLocation) match {
+      case Success(l) if l != null => Some(l)
```
regadas, line 325: @nevillelyh if `l == null` will `job` be a `Failure`?

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/BigQueryClientIT.scala
regadas: this seems like it's probably reading your local config?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/QueryOps.scala
nevillelyh: `Success` can have `null`, also it's possible that `jobReference` exists but the `location` field is missing.
```
scala> Success("a").map(_ => null)
res1: scala.util.Try[Null] = Success(null)
```

regadas: then we are missing a `Success(l) if l == null` in that `match`

nevillelyh: Duh, fixed by matching `_` in the fallback.

Issue #2796, opened 2020-03-23
Cleanup deprecated operations in preparation for 0.9


regadas:

nevillelyh:
Can we close this now? @regadas

regadas:
No not yet. Still cleaning up a few more things.

PR #2856, opened 2020-04-08
Remove ReflectiveRecordIO

regadas:
Related to #2796

### DIFF
--- a/scio-avro/src/it/scala/com/spotify/scio/avro/types/AvroTapIT.scala
+++ b/scio-avro/src/it/scala/com/spotify/scio/avro/types/AvroTapIT.scala
@@ -19,7 +19,6 @@ package com.spotify.scio.avro.types
 import com.spotify.scio.avro.AvroTaps
 import com.spotify.scio.io.Taps
 import org.apache.avro.Schema.Parser
-import org.apache.avro.generic.GenericRecord
 import org.apache.beam.sdk.io.FileSystems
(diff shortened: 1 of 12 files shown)


scio-avro/src/it/scala/com/spotify/scio/avro/types/AvroTapIT.scala
```diff
@@ -51,7 +51,7 @@ final class AvroTapIT extends AnyFlatSpec with Matchers {
   it should "read avro file" in {
     FileSystems.setDefaultPipelineOptions(PipelineOptionsFactory.create)
 
-    val asd = AvroTaps(Taps()).avroFile[GenericRecord](
+    val asd = AvroTaps(Taps()).avroFile(
```
regadas, line 54: `asd`? 🤔 I'll rename this

nevillelyh:
I know we don't use reflective record, but are there other reasons to remove it? Since we're reducing feature parity with Beam `AvroIO` right?

regadas:
yeah I think there are a couple of reasons:

1) as you said we never used it
2) we advise people not to use it because it would have high performance implications. 
3) afaik scio we didn't actually provide the `ReflectData` support through the `AvroIO.sink(..)` we were always expecting that `T` was actually a `GenericRecord`. Having an overloaded `avroFile[T]` is actually misleading. Our types were lying.

We can add support for it but since we advertise not to use it I don't see why we should support it? wdyt?

PR #2870, opened 2020-04-13
delete unused schema in scio-avro

nevillelyh:


nevillelyh:
Oh duh that’s why it didn’t show up in grep. Will revert.

PR #2872, opened 2020-04-13
Revert "delete unused schema in scio-avro"

nevillelyh:
Reverts spotify/scio#2870

The file is used in an integration test.


regadas:
I know why this is borked ... let's move on!

PR #2888, opened 2020-04-16
remove unecessary setCoder

jto:


jto:
Yeah it's actually broken. Beam is tricky.

PR #2899, opened 2020-04-21
Don't write SMB metadata if prior steps fail (fix #2895)

### LINKED ISSUE #2895: SMB sink writes metadata file even if GBK/sort fail
looks like it's because the write metadata transform doesn't depend on sort transform, we should restructure the graph there:

Screen Shot 2020-04-20 at 1 48 33 PM

ClaireMcGinty:
I moved the metadata write to the final Rename step, before moving the buckets from their temp to their final destinations. Since metadata is just one file, I think it's still a pretty atomic operation despite no longer being written to temp storage first; if the metadata write fails, it will be cleaned up and the bucket rename won't happen.

Also simplified the graph a bit by removing redundant composite transforms

before:
Screen Shot 2020-04-21 at 1 58 50 PM

after:
Screen Shot 2020-04-21 at 2 10 42 PM


ClaireMcGinty:
@nevillelyh Added a unit test that uses a dummy implementation of `FileOperations` that throws an exception in `createSink` (the `createSink` method is called in `@ProcessElement` so it is happening during pipeline execution, rather than initialization), wdyt?

PR #2914, opened 2020-04-24
Add initial scala based PCollectionView

regadas:


regadas:
Hi @jto yeah sorry I left this in a limbo state ... update the description with some of the things we can improve with this. There are some other things we can add on top of this like reifying the `SCollection` but I'll add these in a following PR.

PR #2920, opened 2020-04-27
Add initial update alias op to ES7

6c697a61:
Hi @ClaireMcGinty and @regadas, can we add a few methods here? 

Related to #3132

### DIFF
--- a/scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala
+++ b/scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala
@@ -18,13 +18,24 @@
 package com.spotify.scio.elasticsearch
 
 import org.apache.http.HttpHost
+import org.elasticsearch.action.admin.indices.alias.IndicesAliasesRequest
+import org.elasticsearch.action.admin.indices.alias.IndicesAliasesRequest.AliasActions
+import org.elasticsearch.action.admin.indices.alias.get.GetAliasesRequest
+import org.elasticsearch.action.admin.indices.delete.DeleteIndexRequest
+import org.elasticsearch.action.support.master.AcknowledgedResponse
 import org.elasticsearch.client._
 import org.elasticsearch.client.indices.{CreateIndexRequest, CreateIndexResponse}
+import org.elasticsearch.common.unit.TimeValue
 import org.elasticsearch.common.xcontent.XContentType
+import org.slf4j.LoggerFactory
 
+import scala.collection.JavaConverters._
 import scala.util.Try
 
 object IndexAdmin {
+
+  private val Logger = LoggerFactory.getLogger(this.getClass)
+
   private def indicesClient[A](esOptions: ElasticsearchOptions)(f: IndicesClient => A): Try[A] = {
     val client = new RestHighLevelClient(RestClient.builder(esOptions.nodes: _*))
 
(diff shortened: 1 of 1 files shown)


scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala
```diff
@@ -78,4 +89,122 @@ object IndexAdmin {
...
+
+  /**
+   * Add index alias and remove the alias from all other indexes if it is already pointed to any.
+   * If index already exists or some other error occurs this results in a [[scala.util.Failure]].
+   *
+   * @param alias        to be re-assigned
+   * @param newIndexName index to point the alias to
+   * @param timeout defaults to 1 minute
+   */
+  def swapIndexAlias(
```
regadas, line 147: I think we can make this a little bit more broader.

What about renaming this to `createAlias` and have a `removePrevious: Boolean` and a `isWriteIndex: Boolean`. I think it would allow more use cases.

6c697a61: Oops, missed this comment, sorry @regadas. To make sure I understood you right:
```createAlias``` would just add a new index alias
```removePrevious``` would unlink a given alias from all indices

What does ```isWriteIndex``` do?
I assume the Boolean would be derived from ```Try(AcknowledgedResponse)```  by applying ```x.map(_.isAchnowledged)getOrElse(false)```

Would be kinda great to still add a swap method since that's what we do when we create index from scratch every time

Comment (regadas) [reply scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala:147]: Hi @6c697a61, ES alias can point to multiple indices and `isWriteIndex` will tell that a given index is the one to write to when using the alias to write. (It wouldn't actually be a `Boolean` in this case).

I think swapping can be achieved by using smth like `createAlias(alias, index, removePrevious = true)` without having specific `swap` op.

Comment (6c697a61) [reply scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala:147]: Ohh, have them as additional parameters! Got confused, sorry - thought you meant functions. 
Adding ```removePrevious``` , but still not clear of how  ```isWriteIndex``` is intended to be used. I understand what it is, just want to make sure we are on the same page about how it should be used. Do you mean we could put it as an additional parameter in ensureIndex to be able to prevent writing to an aliasedIndex @regadas?

Comment (6c697a61) [reply scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala:147]: Also, would it make sense to also add a way to obtain ElasticserchOptions from address string?
```  
def elasticSearchOptions(esAddress: String): ElasticsearchOptions = 
    ElasticsearchOptions(DnsResolver.addresses(esAddress))
  
```

File: scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala [hunk 2]
```diff
@@ -78,4 +89,134 @@ object IndexAdmin {
...
+
+    val getAliasesResponse =
+      client.getAlias(new GetAliasesRequest(alias), RequestOptions.DEFAULT)
+
+    val request = new IndicesAliasesRequest()
+      .addAliasAction(
+        new AliasActions(AliasActions.Type.ADD)
+          .index(indexName)
+          .alias(alias)
+      )
```
regadas, line 206: What I meant with `isWriteIndex: Boolean` was smth along the lines of:

```suggestion
    // indices: Iterable[(String, Boolean)] (index name, is write index)
    require(
      indices.find(_._2).size == 1,
      "Only one index per alias can be assigned to be the write index at a time"
    
    val request = indices.foldLeft(new IndicesAliasesRequest()) {
      case (request, (idx, isWriteIndex)) =>
        request.addAliasAction(
          new AliasActions(AliasActions.Type.ADD)
            .index(idx)
            .writeIndex(isWriteIndex)
            .alias(alias)
        )
    }
```

scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/IndexAdmin.scala
```diff
@@ -78,4 +89,134 @@ object IndexAdmin {
...
+  /**
+   * Add index alias and remove the alias from all other indexes if it is already pointed to any.
+   * If index already exists or some other error occurs this results in a [[scala.util.Failure]].
+   *
+   * @param alias        to be re-assigned
+   * @param indexName index to point the alias to
+   * @param timeout defaults to 1 minute
+   */
+  def createOrUpdateAlias(
+    nodes: Iterable[HttpHost],
```
regadas, line 148: I would prefer if we just have methods with `ElasticsearchOptions`

nevillelyh:
This has been ongoing for a while? What needs to be done to get it merged?

PR #2922, opened 2020-04-28
New method mapKeys for PairScollection

mkuthan:
Mapping keys seems to be as useful as mapping values. 

Example use case for stripping part of the key for next aggregations:

```
scolByCompositeKey.mapKeys { (k1, _) => k1 }
```

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -898,6 +898,14 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
     self.map(_._1)
 
   /**
+   * Pass each key in the key-value pair SCollection through a `map` function without changing
+   * the values.
+   * @group transform
+   */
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -897,6 +897,14 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
   def keys(implicit koder: Coder[K], voder: Coder[V]): SCollection[K] =
     self.map(_._1)
 
+  /**
+   * Pass each key in the key-value pair SCollection through a `map` function without changing
+   * the values.
+   * @group transform
+   */
+  def mapKeys[U: Coder](f: K => U)(implicit koder: Coder[K], voder: Coder[V]): SCollection[(U, V)] =
```
regadas, line 905: `Coder[K]` is not needed in this case

mkuthan: Good catch! Done

regadas:
Thanks!

PR #2930, opened 2020-04-30
Fix creation of local sharded Sparkey files.

psobot:
Thanks to @codewitch for discovering this - when trying to create a local sharded Sparkey file, Scio calls `LocalShardedSparkeyUri#exists`, which calls `basePathsAndCount` to get the number of shards - but in doing so, `FileSystems#match` is called on an empty local path, which throws an exception. This PR fixes this by adding `EmptyMatchTreatment.ALLOW` when this method is called via `exists`.


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -49,9 +50,12 @@ trait ShardedSparkeyUri extends SparkeyUri {
 
   val globExpression = s"$basePath/part-*"
 
-  private[sparkey] def basePathsAndCount(): (Seq[String], Short) = {
-    val matchResult: MatchResult = FileSystems.`match`(globExpression)
+  private[sparkey] def basePathsAndCount(
+    emptyMatchTreatment: EmptyMatchTreatment = DISALLOW
```
regadas, line 54: I would actually not set a default value here

PR #2953, opened 2020-05-07
implement SMB "least bucket replication" for SortedBucketSource

### LINKED ISSUE #2652: SMB "least bucket replication"
The SMB implementation currently supports joining sources with different bucket sizes, since bucket size is required to be `2^n`. However when reading such sources, the maximum parallelism is bounded by the one with the smallest bucket size.

One improvement we discussed is to replicate the smallest input to all workers. For example, say we have 3 sources of 8, 4, 2, and want to use parallelism of 8. Each worker thread may read the following buckets. This way each worker may read 2x/4x data from input 2, 3, but can filter them out by re-hashing the key.

```
worker input1 input2 input3
0           0      0      0
1           1      1      1
2           2      2      0
3           3      3      1
4           4      0      0
5           5      1      1
6           6      2      0
7           7      3      1
```

A few things we should discuss:
- A better name for this strategy?
- What's the target parallelism? It can be largest bucket size, user defined, or some heuristics based on file size?
- How configurable do we want this to be?

ClaireMcGinty:
(fix #2652) 

I haven't implemented it for SortedBucketTransform, can follow up with another PR or release Scio with just this and let users try it out.

### DIFF
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
@@ -22,8 +22,7 @@ import com.spotify.scio.annotations.experimental
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.{ClosedTap, EmptyTap}
 import com.spotify.scio.values._
-import org.apache.beam.sdk.extensions.smb.SortedBucketSource.TargetParallelism
-import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, SortedBucketTransform}
+import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, SortedBucketTransform, TargetParallelism}
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
 import org.apache.beam.sdk.transforms.join.CoGbkResult
 import org.apache.beam.sdk.transforms.{DoFn, ParDo}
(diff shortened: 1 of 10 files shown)


scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -141,9 +143,10 @@ final class SortedBucketScioContext(@transient private val self: ScioContext) ex
   def sortMergeCoGroup[K: Coder, A: Coder, B: Coder](
     keyClass: Class[K],
     a: SortedBucketIO.Read[A],
-    b: SortedBucketIO.Read[B]
+    b: SortedBucketIO.Read[B],
+    targetParallelism: TargetParallelism
```
ClaireMcGinty, line 147: not super happy with this API change (i.e not providing a default here) but scalac won't allow multiple overloaded methods with default values :/ any suggestions?

nevillelyh: Does overloading method without defaults work, or does it cause erasure issues?

```
def sMCG(k, a, b) ...
def sMCG(k, a, b, tp) ...
def sMCG(k, a, b, c) ...
def sMCG(k, a, b, c, tp) ...
```

Otherwise, what about making `TargetParallelism` an implicit argument in all methods, and have a default instance of `min` (old default behavior?). Similar to how executors are used in `scala.concurrent`. But that is if we don't add other params later.

ClaireMcGinty:
> LGTM amazing it actually worked out to be OK amount of change. I guess our abstraction paid off 😂
> 
> OTOH we should add more javadocs & comments esp. parts around `TargetParallelism`? It's getting hard to follow re-reading the code after a while.

haha yea I was also surprised at how little code change there was too!

sure, I pulled out TargetParallelism into its own file and added more javadoc, to the wiki as well 👍

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
ClaireMcGinty: oh cool that worked! 🙌

PR #2975, opened 2020-05-14
SCollection.withWindowFn with upper bound type

### LINKED ISSUE #2956: SCollection.withWindowFn does not allow upper type bound for WindowFn
Current implementation of SCollection.withWindowFn does not accept typed WindowFn.

Please look at example custom WindowFn implementation, I have to use AnyRef instead of domain type AdEvent:
https://github.com/mkuthan/beam-examples/blob/b6ba82845e1a3394154d0ba4805a29cef5d5c46e/src/main/scala/org/mkuthan/beam/examples/AdEventWindowFn.scala#L17

I would really appreciate SCollection.withWindowFn signature with upper bound type :)

Thanks,
Marcin

mkuthan:
Fixes #2956

Feel free to propose better type name than "A" :)

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -936,7 +936,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @group window
    */
   def withWindowFn[W <: BoundedWindow](
-    fn: WindowFn[_ >: T, W],
+    fn: WindowFn[_ <: Any, W],
     options: WindowOptions = WindowOptions()
   ): SCollection[T] = {
     var transform = Window.into(fn).asInstanceOf[Window[T]]
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -935,8 +935,8 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * Window values with the given function.
    * @group window
    */
-  def withWindowFn[A <: AnyRef, W <: BoundedWindow](
-    fn: WindowFn[A, W],
+  def withWindowFn[W <: BoundedWindow](
+    fn: WindowFn[_ >: T, W],
     options: WindowOptions = WindowOptions()
   ): SCollection[T] = {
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -935,8 +935,8 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * Window values with the given function.
    * @group window
    */
-  def withWindowFn[W <: BoundedWindow](
-    fn: WindowFn[AnyRef, W],
+  def withWindowFn[A <: AnyRef, W <: BoundedWindow](
+    fn: WindowFn[A, W],
```
regadas, line 939: Hi @mkuthan on a second thought, since `WindowFn` is applied on a `SCollection[T]` we should apply here a generic lower bound constraint instead. This way no need for a second type parameter.

```suggestion
    fn: WindowFn[_ >: T, W],
```

mkuthan: Indeed! Even safer!

regadas: Ahh unfortunately `2.12` doesn't like this ... I think we will need to go with smth `_ <: Any` (thinking of `AnyVal`'s)

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -936,7 +936,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @group window
    */
   def withWindowFn[W <: BoundedWindow](
-    fn: WindowFn[AnyRef, W],
+    fn: WindowFn[_ >: T, W],
```
regadas, line 939:
```suggestion
    fn: WindowFn[_ <: Any, W],
```

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: Took the liberty of going with this! Hope you don't mind @mkuthan and thanks a lot for spotting this!

regadas: Another alternative would be to have diff constraints for 2.12/2.13 🤔 not sure if I want to go that route ... but we would end up with a better 2.13.

PR #2987, opened 2020-05-19
Add SCollection reify methods

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1215,6 +1215,57 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   *  val coll: SCollection[(Int, Map[Int, Int])] = sc.parallelize(Seq(1, 2)).reify(other.asMapSideInput)
+   * }}}
+   *
+   * Reify as Multimap:
+   * {{{
+   *  val other: SCollection[(Int, Int)]  = sc.parallelize(Seq((1, 1)))
+   *  val coll: SCollection[(Int, Map[Int, Iterable[Int]])]  = sc.parallelize(Seq(1, 2)).reify(other.asMultiMapSideInput)
+   * }}}
+   */
+  def reify[U: Coder](side: SideInput[U]): SCollection[(T, U)] = {
```
nevillelyh, line 1245: This feels like a case of `def cross[U](that: SCollection[U])` where `that` is a singleton. Is there a better name to improve discovery? Something like `crossSideInput/crossSingleton(that.asSingletonSideInput)` so it's more obvious for tab-completers?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1215,6 +1215,57 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   *    sc.parallelize(Seq(1, 2)).reifyInGlobalWindow(_.asListSideInput)
+   * }}}
+   *
+   * Can be used to replace patterns like:
+   * {{{
+   *  val coll: SCollection[Iterable[Int]] = sc.parallelize(Seq(1, 2)).groupBy(_ => ())
+   * }}}
+   * where you want to actually get an empty [[Iterable]] even if no data is present.
+   */
+  def reifyInGlobalWindow[U: Coder](view: SCollection[T] => SideInput[U]): SCollection[U] =
```
nevillelyh, line 1266: The `view` part is a little hard to parse since it's applied on `this` really. And it should almost always be an `_.asListSideInput`? What about getting rid of the argument and just have singleton/list/map/set variants?

regadas: Yeah, the view part will always be applied on the current `SCollection` like. We can go with the multiple versions.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: I like the idea of having this more obvious ... not sure about the naming though. :thinking:

nevillelyh: If we rename the other as `reify{Singleton,List,Map,Set}InGlobalWindow`, can call this `reifyWithSideInput`? That also opens other `reify*` options and avoid potential overloading issues.

regadas: yeah the other one I was thinking, was smth along those lines but with `As``reifyAs{Singleton,List,Map,Set}InGlobalWindow. I think I like `reifyWithSideInput`

nevillelyh:
@regadas it's conflicting with master mind taking a look?

PR #2995, opened 2020-05-20
Use BigQuery client project consistently

### LINKED ISSUE #2928: BigQueryType.fromStorage to ingest default project_ids
### Context

At my organization, we have separate GCP projects for sandbox, QA and prod, and they all contain identical schemas of datasets & tables on BigQuery. This means that if we were to take advantage of `BigQueryType.fromStorage`'s cost-effectiveness, ideally we should pass in table specs without projectIds and have it inferred via a compile time argument.

When i try to compile with projec-id-less table specs, the macro expansion fails in the following manner:

```
/...Schemas.scala:29:6: exception during macro expansion: io.grpc.StatusRuntimeException: 

INVALID_ARGUMENT: 'x-goog-request-params' header value 'table_reference.project_id=&table_reference.dataset_id=my_dataset' is invalid.
[error]     @BigQueryType.fromStorage(
```

which seems like an empty project ID is used to build the dry run request.

### Suggested feature 

Is it possible to use a compiled argument, like `-Dbigquery.project`, somewhere within the [StorageImpl](https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala#L89) section of the type provider to fill in as a default project id?

### LINKED ISSUE #2994: Internal BigQuery client ops are not defining ProjectId consistently
On a deeper look, #2928 is not possible because of this.

regadas:
Fixes #2994
Fixes #2928, It will allow compilation without `projectId` in the spec. However, it must be defined either through `-Dbigquery.project`, env or `.config/gcloud`

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryPartitionUtil.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryPartitionUtil.scala
@@ -58,7 +58,7 @@ private[bigquery] object BigQueryPartitionUtil {
   private def getPartitions(bq: BigQuery, tableRef: TableReference): Set[String] = {
     val prefix = tableRef.getTableId.split('$')(0)
     bq.tables
-      .tableReferences(tableRef.getProjectId, tableRef.getDatasetId)
+      .tableReferences(tableRef)
       .filter(_.getTableId.startsWith(prefix))
       .map(_.getTableId.substring(prefix.length))
       .toSet
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryPartitionUtil.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryPartitionUtil.scala
@@ -58,7 +58,7 @@ private[bigquery] object BigQueryPartitionUtil {
   private def getPartitions(bq: BigQuery, tableRef: TableReference): Set[String] = {
     val prefix = tableRef.getTableId.split('$')(0)
     bq.tables
-      .tableReferences(Option(tableRef.getProjectId()), tableRef.getDatasetId())
+      .tableReferences(tableRef.getProjectId(), tableRef.getDatasetId())
       .filter(_.getTableId.startsWith(prefix))
       .map(_.getTableId.substring(prefix.length))
       .toSet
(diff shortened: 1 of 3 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala
```diff
@@ -167,6 +162,13 @@ final private[client] class TableOps(client: Client) {
     client.underlying.tables().get(p, tableRef.getDatasetId, tableRef.getTableId).execute()
   }
 
+  /** Get list of tables in a dataset. */
+  def tableReferences(tableRef: TableReference): Seq[TableReference] =
```
nevillelyh, line 166: The input `TableReference` can have `tableId`? That can be a little misleading? It's unclear from the call site that `.tableReferences(tableRef)` gets all tables from the dataset `tableRef` is in.

regadas: yeah that's true... agree it might be confusing

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryPartitionUtil.scala
```diff
@@ -58,7 +58,7 @@ private[bigquery] object BigQueryPartitionUtil {
   private def getPartitions(bq: BigQuery, tableRef: TableReference): Set[String] = {
     val prefix = tableRef.getTableId.split('$')(0)
     bq.tables
-      .tableReferences(tableRef.getProjectId, tableRef.getDatasetId)
+      .tableReferences(tableRef.getProjectId(), tableRef.getDatasetId())
```
nevillelyh, line 61:
```suggestion
      .tableReferences(tableRef.getProjectId, tableRef.getDatasetId)
```

nevillelyh: nit: this is unnecessary? IntelliJ won't add those `()`s

regadas: yeah intellij likes that :man_shrugging: ... regardless the idea is actually not to change this line ... missed that.

PR #2996, opened 2020-05-20
Add SCollection#groupMap and SCollection#groupMapReduce operations

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -510,14 +511,52 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   * Return an SCollection of grouped items. Each group consists of a key and the result of an associative
+   * reduce function. The ordering of elements within each group is not guaranteed, and may even differ each
+   * time the resulting SCollection is evaluated.
+   *
+   * The associative function is performed locally on each mapper before sending results to
+   * a reducer, similarly to a "combiner" in MapReduce
+   *
+   * @group transform
+   */
+  def groupMapReduce[K](f: T => K)(
```
nevillelyh, line 548: How does this compare to the more idiomatic `keyBy(f),reduceByKey(g)`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -510,14 +511,52 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+  /**
+   * Return an SCollection of grouped items. Each group consists of a key and a sequence of
+   * elements transformed into a value of type `U`. The ordering of elements within each group is not guaranteed,
+   * and may even differ each time the resulting SCollection is evaluated.
+   *
+   * It is equivalent to groupBy(key).mapValues(_.map(f)), but more efficient.
+   *
+   * @group transform
+   */
+  def groupMap[K, U](f: T => K)(
```
nevillelyh, line 525: How does this compare to the more idiomatic `.map(x => (f(x), g(x))).groupByKey`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: Well good question, there's a tiny bit advantage over that example, where there's one step less of conversion of `tuple` to `kv`. Also be because we wrap this in a composite transform might increase chances of fusion.

Recently also scala 2.13 added these methods to collections, so i guess this might become idiomatic?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
regadas: Same thing applies for this example one step less of conversion/serialization. However the idea behind this method is to try to mitigate other issues like:

```scala
groupBy(key).mapValues(_.map(f))
```

Users will incur in an extra copy of the list which depending on the size might be significant perf penalty.

PR #3005, opened 2020-05-22
Make SMB source a splittable BoundedSource

ClaireMcGinty:
Re-implement `SortedBucketSource` as a proper Beam `BoundedSource` splittable by the runner. "Splittable" refers to how many buckets will be processed per worker, and ranges from 1 (all buckets processed by the same worker) to the highest # of buckets among input sources.

It corresponds to a new `TargetParallelism`, `TargetParallelism.auto()`, that's now the default if the user doesn't supply a value.

I noticed that the Dataflow runner, at least, tries to call `split(long desiredBundleSizeBytes)` with too high of a target byte size than is optimal for SMB, probably because of its relatively low throughput compared to other Sources. To offset I introduced a `DESIRED_SIZE_BYTES_ADJUSTMENT_FACTOR`, currently set to 0.5, to adjust the target size by. In benchmarks, this seems to result in the most optimal parallelism*.

This is only implemented for `SortedBucketSource`, not `SortedBucketTransform`, which needs to be followed up on at some point.

**\* benchmarks:**
LHS - 33GB/256buckets, RHS - 100GB/2048 buckets

Baseline:
Screen Shot 2020-05-22 at 1 50 35 PM

BoundedSource implementation, DESIRED_SIZE_BYTES_ADJUSTMENT_FACTOR = 0.5
Screen Shot 2020-05-22 at 1 50 22 PM

BoundedSource implementation, DESIRED_SIZE_BYTES_ADJUSTMENT_FACTOR = 0.33
Screen Shot 2020-05-22 at 3 25 11 PM

^ I'm a little conflicted on what value to set `DESIRED_SIZE_BYTES_ADJUSTMENT_FACTOR` to--the higher value runs extremely fast but is a bit more resource-intensive, whereas the lower value is a few minutes slower but uses fewer resources. @nevillelyh wdyt?

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
@@ -17,6 +17,7 @@
 
 package org.apache.beam.sdk.extensions.smb;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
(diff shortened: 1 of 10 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -106,198 +113,198 @@ public SortedBucketSource(
...
+      TargetParallelism targetParallelism,
+      int bucketOffsetId,
+      int effectiveParallelism) {
     this.finalKeyClass = finalKeyClass;
     this.sources = sources;
     this.targetParallelism = targetParallelism;
+    this.bucketOffsetId = bucketOffsetId;
+    this.effectiveParallelism = effectiveParallelism;
+    this.keyGroupSize =
+        Metrics.distribution(SortedBucketSource.class, "SortedBucketSource-KeyGroupSize");
```
ClaireMcGinty, line 132: Unlike in the previous iteration where `SortedBucketSource` is a PTransform, we don't have access to a unique name for this Read so if there are multiple SortedBucketSource reads in a single pipeline the counters would overlap... any ideas on a way to disambiguate them?

ClaireMcGinty: I added an optional metricsKey argument that will default to `SortedBucketSource-{$atomicInteger}` if not supplied. If `atomicInteger == 0` it will just be `SortedBucketSource`... so for people who only have a single SMB read in their pipeline, the metrics should look exactly the same.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -272,5 +282,80 @@ public void processElement(ProcessContext c) {
             c.output(bucketShardAndDst);
           });
     }
+
+    static void merge(
```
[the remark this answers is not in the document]
ClaireMcGinty: Yea I think I can refactor once we get `SortedBucketTransform` using the same parallelism strategy as SMB Source.

PR #3006, opened 2020-05-22
Add -release 8 scalac option

### LINKED ISSUE #3000: Wrong base64 encoding used in BigQuery `scio-extra`
https://github.com/spotify/scio/blob/master/scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/ToTableRow.scala#L41

it should be `BaseEncoding.base64()`

regadas:
Fixes #3000

This flag will force scala to target 8 to avoid the mentioned issue when compiling/publishing with jdk11. It doesn't play well in some other situations (read [thread](https://github.com/scala/scala-dev/issues/139)) and in our case I had to disabled it in `scio-repl` (ClassLoader issues) and while doing `doc` generation. In both these cases it's not really important that this flag is enabled.

### DIFF
--- a/project/ScalacOptions.scala
+++ b/project/ScalacOptions.scala
@@ -20,6 +20,8 @@ import sbt._, Keys._
 object Scalac {
   // see: https://tpolecat.github.io/2017/04/25/scalac-flags.html
   val baseOptions = List(
+    "-release",
+    "8",
     "-target:jvm-1.8",
     "-deprecation", // Emit warning and location for usages of deprecated APIs.
     "-feature", // Emit warning and location for usages of features that should be imported explicitly.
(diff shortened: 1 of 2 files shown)


project/ScalacOptions.scala
```diff
@@ -57,9 +59,7 @@ object Scalac {
     // "-Ywarn-unused:privates", // Warn if a private member is unused.
     "-Ywarn-value-discard", // Warn when non-Unit expression results are unused.
     "-Xmacro-settings:show-coder-fallback=true",
-    "-Ydelambdafy:inline", // Set the strategy used for translating lambdas into JVM code to "inline"
-    "-Ybackend-parallelism",
-    "8"
```
nevillelyh, line 62: This is parallelism 8, not JDK version 8 right?

regadas: yeah that is referring to parallelism. It shouldn't be removed actually (I was experiment other stuff). Added it back.

PR #3012, opened 2020-05-26
Unify sharded & non sharded sparkey write

### LINKED ISSUE #3009: Unify sparkey IO sharded and non sharded

### LINKED ISSUE #3010: Non sharded `.asSparkey` write speed is slower compared to scio `0.7.4`

regadas:
Fixes #3009 
Fixes #3010

Depends on #3008 

before:

image

After:

image


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -224,48 +226,75 @@ package object sparkey extends SparkeyReaderInstances {
...
+                .sortWith(_.getFilename < _.getFilename)
+                .asJava
+              val dst = SparkeyUri.extensions
+                .map(ext => FileSystems.matchNewResource(s"$basePath$ext", false))
+                .asJava
+
+              FileSystems.rename(src, dst)
+
+              nonShardedUri
+            } else {
```
regadas, line 294: this is somewhat ugly but needed to keep the filename's backwards compat, important?

basically renaming `gs://example/path/part-0000-of-0001.{spi, spl}` to `gs://example/path.{spi, spl}`

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -224,48 +226,75 @@ package object sparkey extends SparkeyReaderInstances {
...
+                uri.sparkeyUriForShard(shard, numShards),
+                maxMemoryUsage,
+                xs.asJava
+              )
+          }
+
+        val shardsMap = shards.asMapSideInput
+
+        val uris = shards.context
+          .parallelize((0 until numShards).map(_.toShort))
```
psobot, line 258: This is a neat solution! I wouldn't have thought to add this step in here.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -224,48 +226,75 @@ package object sparkey extends SparkeyReaderInstances {
...
-                  writeToSparkey(
-                    uri.sparkeyUriForShard(shard, shardCount),
-                    maxMemoryUsage,
-                    xs.asJava
-                  )
-              }
-              .groupBy(_ => ())
-              .map(_ => uri: SparkeyUri)
+          .map { _ =>
+            if (numShards <= 1) {
```
psobot, line 279: Nit: we assert `numShards > 0` earlier on - should this be `numShards == 1`?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
psobot: Unfortunately I think this might be necessary, as I know of a lot of pipelines that depend on the Sparkey name being consistent.

regadas: yeah that's what I figured let's keep this for now.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
regadas, line 279:
```suggestion
            if (numShards == 1) {
```

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -39,6 +39,8 @@ import org.slf4j.LoggerFactory
 import scala.jdk.CollectionConverters._
 import scala.util.hashing.MurmurHash3
 import java.lang.Math.floorMod
+import org.apache.beam.sdk.io.FileSystems
+import org.apache.beam.sdk.io.fs.ResourceId
```
regadas, line 43:
```suggestion
import org.apache.beam.sdk.io.FileSystems
```

PR #3020, opened 2020-05-29
Add support for sharded JDBC reads from tables

stormy-ua:
## Background

The standard Beam [JdbcIO](https://beam.apache.org/releases/javadoc/2.0.0/org/apache/beam/sdk/io/jdbc/JdbcIO.html) doesn't support sharded reading from a table. As a result, reads from large SQL tables are inefficient because it isn't possible to fetch data from a table concurrently. In addition, we have observed that JdbcIO might fail with `OutOfMemory` exception when reading from a large table even if the [fetchSize](https://beam.apache.org/releases/javadoc/2.19.0/org/apache/beam/sdk/io/jdbc/JdbcIO.Read.html#withFetchSize-int-) parameter is set to a relatively small value (5-10K).

Inspired by [dbeam](https://github.com/stormy-ua/dbeam) which can help with parallelized/sharded data export from a table into GCS in Avro format. It doesn’t have an option to make a sharded read into an SCollection directly though.

## Implementation

The implementation is based on extending [BoundedReader](https://beam.apache.org/releases/javadoc/2.19.0/org/apache/beam/sdk/io/BoundedSource.BoundedReader.html) with the support for splitting.

## Constraints

SQL databases in general are not designed for massively parallel/bulk reads. Consider to create a separate DB replica optimized for reads only. It could be easily done with CloudSql - more details [here](https://cloud.google.com/sql/docs/postgres/replication/create-replica). 

## Sharding

Sharding reads could be executed only against tables/materialized views. A table/materialized view must have a column to shard by. The column must be of an integer or long type and for better performance it should be in a clustered or a secondary index. The  reader first queries for minimum and maximum values of the shard column and splits the whole range into equal smaller partitions, so the shard column should have evenly distributed values otherwise there might be no performance gains from the sharded reading.

## Usage Example

The library exposes `jdbcShardedSelect` function in the `com.spotify.scio.jdbc.sharded` package which mimics the existing Scio function `jdbcSelect`. 

Here is an example of minimum possible pipeline using the `jdbcShardedSelect` function:

```scala
package com.spotify.test

import scio.ScioContext
import scio.jdbc.{CloudSqlOptions, JdbcReadOptions}
import scio.jdbc._
import scio.jdbc.sharded._

object MyPipeline {

    val (pipelineOpts, args) = ScioContext.parseArguments[CloudSqlOptions](cmdlineArgs)
    val sc = ScioContext(pipelineOpts)

    val connOptions = JdbcConnectionOptions(
      username = "<user>",
      password = Some("<password>"),
      driverClass = classOf[org.postgresql.Driver],
      connectionUrl = s"jdbc:postgresql://google/<db>?" +
        s"cloudSqlInstance=<instance>&" +
        s"socketFactory=com.google.cloud.sql.postgres.SocketFactory&" +
        s"readOnly=true"
    )

    val shardedReadOptions = JdbcShardedReadOptions[String](
    connectionOptions = connOptions,
    tableName = "test",
    shard = Shard.range[Long],
    rowDecoder = rs => rs.getString("id"),
    shardColumn = "id",
    numShards = 8,
    fetchSize = 100000
    )

    sc.jdbcShardedSelect(shardedReadOptions)
        .saveAsTextFile("gs://test/")
  }

}    
```

## Parameters

The library defines `JdbcShardedSourceOptions` options which is the paramter for `jdbcShardedSelect`. Here is the list of currently supported parameters:

* `shardColumn`: the column to shard by. Must be of integer/long type ideally with evenly distributed values.
* `numShards`: number of shards to split the table into for reading. There is no guarantee that Beam will actually execute reads in parallel. It is up to Beam auto scaler to decide the level of parallelism to use (number of workers and threads per worker). But the behavior could be controlled with `maxNumWorkers` and `numberOfWorkerHarnessThreads` parameters (see more details about these parameters [here](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params)). Defaults to 4.
* `tableName`: name of a table or materialized view to read from
* `fetchSize`: number of records to read from the JDBC source per one call to a database. Default value is 100,000. Set to `-1` to make it unbounded.

## Performance

The sharded reader was tested by executing it against a table in the CloudSQL Postgres database. The table has 85M records and the table size as reported by Postgres `pg_size_pretty` function is 30 GB. To make things more interesting and demonstrate that the reader can actually scale the initial number of workers was set to 4 and number of threads per worker 4 as well, so the degree of the parallelism a job gets right away is 16 and scaling from 1 to 32 should prove that sharding actually works. The total elapsed time of job runs was measured. The fetch size was set to 100K.

| Number of shards | Job total elapsed time in seconds | Max Egress MB/s from the DB |
| :--------------- | :-------------------------------- | :-------------------------- |
| 1                | 860 secs                          | 8 MB/s                    |
| 2                | 530 secs                          | 13 MB/s                    |
| 4                | 423 secs                          | 18 MB/s                    |
| 8                | 323 secs                          | 30 MB/s                    |
| 16               | 273 secs                          | 37 MB/s                    |
| 32               | 250 secs                          | 45 MB/s                    |

## Testing

The function tested with a job referencing Scio built and published locally. The job was executed  using the DataFlow runner.


scio-jdbc/src/main/java/com/spotify/scio/jdbc/sharded/JdbcShardedSource.java
```diff
@@ -0,0 +1,230 @@
...
+                    ? new Range<Long>(
+                        range.min() + i * partitionSize,
+                        range.min() + i * partitionSize + partitionSize - 1)
+                    : new Range<Long>(range.min() + i * partitionSize, range.max()))
+        .collect(Collectors.toList());
+  }
+
+  private Range<Long> getShardColumnRange() throws Exception {
+    try (Connection connection = createConnection(readOptions.connectionOptions())) {
+      String query =
```
nevillelyh, line 87: A bit on the defensive side: should we also check if the column has index?

stormy-ua: I think it should be doable with [DatabaseMetaData.getIndexInfo](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getIndexInfo-java.lang.String-java.lang.String-java.lang.String-boolean-boolean-). There is one caveat though - a column could be in an index, but the index might be inefficient e.g. a composite index where the first column has low cardinality. But we can add this check just to filter out cases when the column definitely not in an index.

stormy-ua:
@nevillelyh I like the idea with adding support for `String`, `Binary` and `Numeric`. I think we can even create a separate abstraction (inner to `scio`) for this in the form of a type class e.g.:

```scala
case class Query(queryString: String)

trait Shardable[A] {
   def shard(numShards: Int, lowerBound: A, upperBound: A): Seq[Query]
}
```

The implementation for `Int/Long` would be trivial (same thing which is being done in the Source in the current implementation) and should be possible implement for strings and other naturally ordered data types.

scio-jdbc/src/main/java/com/spotify/scio/jdbc/sharded/JdbcShardedSource.java
```diff
@@ -0,0 +1,230 @@
...
+      this.source = source;
+    }
+
+    @Override
+    public boolean start() throws IOException {
+      try {
+        connection = createConnection(source.readOptions.connectionOptions());
+
+        final Statement statement = connection.createStatement();
+        if (source.readOptions.fetchSize() != JdbcShardedReadOptions.UnboundedFetchSize()) {
```
regadas, line 167: we probably should move this constants to Java land

scio-jdbc/src/main/java/com/spotify/scio/jdbc/sharded/JdbcShardedSource.java
```diff
@@ -0,0 +1,230 @@
...
+            String.format(
+                "SELECT * FROM %s WHERE %s >= %d and %s <= %d",
+                source.readOptions.tableName(),
+                source.readOptions.shardColumn(),
+                source.range.min(),
+                source.readOptions.shardColumn(),
+                source.range.max());
+
+        log.info("Running a query: [{}]", query);
+        resultSet = statement.executeQuery(query);
```
regadas, line 182: could we be using scrollable result sets (forward-only)?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/Range.scala
```diff
@@ -0,0 +1,20 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.jdbc.sharded
+
+case class Range[A](min: A, max: A)
```
regadas, line 20: This should also be moved to Java land

scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSelectTest.scala
```diff
@@ -0,0 +1,70 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.jdbc
+package sharded
```
regadas, line 19: is there a need for relative package?

scio-jdbc/src/main/java/com/spotify/scio/jdbc/sharded/JdbcShardedSource.java
```diff
@@ -0,0 +1,230 @@
...
+
+        final Statement statement = connection.createStatement();
+        if (source.readOptions.fetchSize() != JdbcShardedReadOptions.UnboundedFetchSize()) {
+          log.info("Setting a user defined fetch size: [{}]", source.readOptions.fetchSize());
+          statement.setFetchSize(source.readOptions.fetchSize());
+        }
+
+        final String query =
+            String.format(
+                "SELECT * FROM %s WHERE %s >= %d and %s <= %d",
```
regadas, line 174: we probably should move this to `static`

scio-jdbc/src/main/java/com/spotify/scio/jdbc/sharded/JdbcShardedSource.java
```diff
@@ -0,0 +1,230 @@
...
+
+      log.info("Executing query = [{}]", query);
+
+      final ResultSet rs = connection.createStatement().executeQuery(query);
+
+      if (rs.next()) {
+        return new Range<Long>(rs.getLong("min"), rs.getLong("max"));
+      } else {
+        log.warn("The table is empty. Nothing to read.");
+        return null;
```
regadas, line 100: maybe return `Optional`?

stormy-ua:
@regadas I decided to implement the `BoundedSource` in Scala so the implementation is more coherent. In addition, changed the result set to be scrollable.

@nevillelyh added `JdbcShardable` abstraction and support for range based (numeric primitive types) and prefix based (strings and binary). Here is an example of the sharded select for `BigDecimal`:

```scala
sc.jdbcShardedSelect(getShardedReadOptions(opts), ShardBy.range.of[BigDecimal])
```

and for String prefix based:

```scala
sc.jdbcShardedSelect(getShardedReadOptions(opts), ShardBy.prefix.of[String](prefixLength = 3))
```

I haven't implemented the prefix based sharding yet. It isn't trivial so might take some time, but should be doable.

scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSelectTest.scala
stormy-ua: Usually I do this to avoid redundant import package prefixes e.g.:

```scala
package com.spotify.scio
package jdbc.sharded

import io.TextIO (instead of com.spotify.scio.io.TextIO)
````

but in this case it isn't needed, so I will remove this.

Comment (stormy-ua) [reply scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSelectTest.scala:19]: In this case I bring everything in the scope of the `com.spotify.scio.jdbc` package so I don't need to import `CloudSqlOptions` and `JdbcJob`

Comment (regadas) [reply scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSelectTest.scala:19]: this is one of those things that is not really recommended for various reasons. one of them being bringing in an `io.example` dependency, suddenly you need to restructure this. Always prefer explicit import paths/package defs

**Comment by regadas** [comment]
@stormy-ua regarding the scrollable result set, I'm curious about performance, did you had time to re-run and get some numbers?

File: scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSource.scala [hunk 8]
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+  private class JdbcShardedReader(
+    private val source: JdbcShardedSource[T, S],
+    private val query: ShardQuery
+  ) extends BoundedSource.BoundedReader[T] {
+    private var connection: Connection = _
+    private var resultSet: ResultSet = _
+
+    override def start: Boolean = {
```
regadas, line 114: since this actually mutates internal state `override def start(): Boolean`

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSource.scala
```diff
@@ -0,0 +1,150 @@
...
+        new JdbcShardedReader(this, q)
+      case None =>
+        throw new UnsupportedOperationException(
+          "Not possible to create a read for a source with " +
+            "the empty query"
+        )
+    }
+
+  private class JdbcShardedReader(
+    private val source: JdbcShardedSource[T, S],
```
regadas, line 108: `private val` can be removed here?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSource.scala
```diff
@@ -0,0 +1,150 @@
...
+import scala.jdk.CollectionConverters._
+
+final private[jdbc] class JdbcShardedSource[T, S](
+  private val readOptions: JdbcShardedReadOptions[T],
+  coder: BCoder[T],
+  jdbcShardable: JdbcShardable[S],
+  private val query: Option[ShardQuery] = None
+) extends BoundedSource[T] {
+
+  private val SHARD_BOUNDS_QUERY_TEMPLATE = "SELECT min(%s) min, max(%s) max FROM %s"
```
regadas, line 40: `ShardBoundsQueryTemplate` to be coherent if other constants defined?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardable.scala
```diff
@@ -0,0 +1,163 @@
...
+      (range, numShards) => (range.upperBound - range.lowerBound) / numShards
+    )
+
+  implicit val floatRangeJdbcShardable: RangeJdbcShardable[Float] =
+    RangeJdbcShardable[Float](
+      (rs, colName) => rs.getFloat(colName),
+      (range, numShards) => (range.upperBound - range.lowerBound) / numShards
+    )
+
+  implicit val stringPrefixJdbcShardable: Int => PrefixJdbcShardable[String] = prefixLength =>
```
regadas, line 76: Since this is still not implemented, I guess it's better if we don't supply the implicit? or will you still implement it in this PR?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardable.scala
```diff
@@ -0,0 +1,163 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.jdbc.sharded
+
+import java.sql.ResultSet
+import org.slf4j.LoggerFactory
+
+trait JdbcShardable[T] extends Serializable {
```
regadas, line 23: what about calling this just `Shard`? this way we would have `RangeShard` and `PrefixShard`

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardable.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  def columnValueDecoder(resultSet: ResultSet, columnName: String): T
+
+  def partition(range: Range[T], numShards: Int): Seq[ShardQuery]
+
+}
+
+object ShardBy {
+
+  object range {
```
regadas, line 33: what about moving this under the companion object? `Shard.range[T]` / `Shard.prefix[T](p)`

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSource.scala
```diff
@@ -0,0 +1,150 @@
...
+        ShardQuery.toSelectStatement(query)(readOptions.tableName, readOptions.shardColumn)
+
+      log.info(s"Running a query: [$queryString]")
+      resultSet = statement.executeQuery(queryString)
+      resultSet.next
+    }
+
+    override def advance: Boolean = resultSet.next
+
+    override def getCurrent: T = source.readOptions.rowMapper.apply(resultSet)
```
regadas, line 136:
```suggestion
    override def getCurrent: T = source.readOptions.rowMapper(resultSet)
```

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardQuery.scala
```diff
@@ -0,0 +1,48 @@
...
+package com.spotify.scio.jdbc.sharded
+
+sealed trait ShardQuery extends Serializable
+
+case class RangeShardQuery[T](range: Range[T], upperBoundInclusive: Boolean) extends ShardQuery
+case class PrefixShardQuery[T](prefix: T) extends ShardQuery
+
+object ShardQuery {
+
+  private val RANGE_QUERY_TEMPLATE = "SELECT * FROM %s WHERE %s >= %s and %s %s %s"
```
regadas, line 27: nit: style of these as well

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/syntax/SCollectionSyntax.scala
```diff
@@ -30,6 +31,32 @@ final class JdbcScioContextOps(private val self: ScioContext) extends AnyVal {
...
+   *                    controlled with maxNumWorkers and numberOfWorkerHarnessThreads parameters
+   *                    (see more details about these parameters here). Defaults to 4.
+   *
+   *                    tableName: name of a table or materialized view to read from
+   *
+   *                    fetchSize: number of records to read from the
+   *                    JDBC source per one call to a database. Default value is 100,000. Set to
+   *                    -1 to make it unbounded.
+   */
+  def jdbcShardedSelect[T: ClassTag: Coder, S](
```
regadas, line 55: `ClassTag` not needed? outside of this PR, but I see that the previous method also defines it ... it shouldn't be needed as well.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/Range.scala
```diff
@@ -0,0 +1,20 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.jdbc.sharded
+
+case class Range[A](lowerBound: A, upperBound: A)
```
regadas, line 20:
```suggestion
final case class Range[A](lowerBound: A, upperBound: A)
```

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardQuery.scala
```diff
@@ -0,0 +1,48 @@
...
+  private val PREFIX_QUERY_TEMPLATE = "SELECT * FROM %s WHERE %s LIKE '%s%%'"
+
+  def toSelectStatement(shardQuery: ShardQuery)(tableName: String, shardColumn: String): String =
```
regadas, line 30: I don't think this needs to be curried.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardQuery.scala
```diff
@@ -0,0 +1,48 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.jdbc.sharded
+
+sealed trait ShardQuery extends Serializable
+
+case class RangeShardQuery[T](range: Range[T], upperBoundInclusive: Boolean) extends ShardQuery
```
regadas, line 22: we can mark this as `final` for now

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/package.scala
```diff
@@ -0,0 +1,3 @@
+package com.spotify.scio.jdbc
+
+package object sharded {}
```
regadas, line 3: this `package object` can be removed

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardable.scala
stormy-ua: I think sharding by an arbitrary range of strings deserves its own PR and might take some time to implement it properly, so probably would remove it from this PR.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardable.scala
stormy-ua: ok, I think it sounds better, anyways it will be in its own package so we can use shorter names (I was trying to highlight that it is JDBC related).

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/syntax/SCollectionSyntax.scala
stormy-ua: Indeed, it isn't needed. I blindly copy pasted it from the `jdbcSelect`. Actually, `jdbcSelect` doesn't need it as well. Do you think I can safely remove it from there also (don't want to break anything )?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardQuery.scala
stormy-ua: yeah, I had some idea, but it isn't needed now.

scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/JdbcShardedSelectTest.scala
stormy-ua: Agree, I had to change my habit of using this approach. It just so nice to omit imports if possible. Removed all relative package definitions.

stormy-ua:
> @stormy-ua regarding the scrollable result set, I'm curious about performance, did you had time to re-run and get some numbers?

I ran to jobs (with forward only and without one after the other) and did see only a slight minor performance improvement. Two jdbcShardedReads in the job took 21 and 15 mins as opposed to 22 and 16 mins in the old job. But this is an anecdotal evidence - might be due to some other thins fluctuating over time (network, db state etc.).

stormy-ua:
@regadas @nevillelyh looks like I addressed all comments, one thing which is still pending is the sharding implementation for string/binary fields, but even with the current implementation it is possible to extend it in a user code by implementing `PrefixShard` trait. Let me know what you think. Tested on out job by building/publishing it locally.

jto:
@nevillelyh @regadas ping :)

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/Shard.scala
```diff
@@ -0,0 +1,154 @@
...
+  def columnValueDecoder(resultSet: ResultSet, columnName: String): T
+
+  def partition(range: Range[T], numShards: Int): Seq[ShardQuery]
+
+}
+
+object ShardBy {
+
+  object range {
+    def of[T: RangeShard]: Shard[T] = implicitly[RangeShard[T]]
```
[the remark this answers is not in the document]
stormy-ua: @regadas nvm, I found your comment in the email thread. I agree, it is better to simplify - it makes the syntax clearer. Just pushed the change.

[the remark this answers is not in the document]
regadas: Awesome!

stormy-ua:
@regadas 😆  `RangeShardShardCoder` should be `RangeShardStringCoder`. I was swapping Shard and String in its name and  did a mistake. Just renamed it `RangeShardStringCoder`.

regadas:
@stormy-ua right, but i was actually thinking more about the `Coder` part ... do want users to confuse it with the `com.spotify.scio.coders.Coder`.

stormy-ua:
@regadas Ah, I c, well, I can rename it to `RangeShardStringCodec`. wdyt?

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/Shard.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  }
+
+  def apply[T: Numeric](decoder: (ResultSet, String) => T, partitionLength: (Range[T], Int) => T) =
+    new NumericRangeShard[T](decoder, partitionLength)
+
+}
+
+final class RangeStringShard[T <: ShardString](implicit
+  rangeStringShardCodec: RangeShardStringCodec[T]
```
regadas, line 158: `Codec` suits it better! 👍

PR #3022, opened 2020-06-01
Use scalaVersion semantic selector

regadas:


project/ScalacOptions.scala
```diff
@@ -16,78 +16,85 @@
...
-    // "-Ywarn-numeric-widen", // Warn when numerics are widened.
-    // "-Ywarn-unused:implicits", // Warn if an implicit parameter is unused.
-    // "-Ywarn-unused:locals", // Warn if a local definition is unused.
-    // "-Ywarn-unused:params", // Warn if a value parameter is unused.
-    // "-Ywarn-unused:patvars", // Warn if a variable bound in a pattern is unused.
-    // "-Ywarn-unused:privates", // Warn if a private member is unused.
-    "-Ywarn-value-discard", // Warn when non-Unit expression results are unused.
-    "-Xmacro-settings:show-coder-fallback=true",
-    "-Ydelambdafy:inline", // Set the strategy used for translating lambdas into JVM code to "inline"
-    "-Ybackend-parallelism",
```
regadas, line 63: On further inspection I noticed that bloop and intellij don't really like this setting when `-release 8` is added.

regadas: let's see how's the compile time without it ...

project/ScalacOptions.scala
```diff
@@ -16,78 +16,88 @@
...
+      // "-Ywarn-unused:params", // Warn if a value parameter is unused.
+      // "-Ywarn-unused:patvars", // Warn if a variable bound in a pattern is unused.
+      // "-Ywarn-unused:privates", // Warn if a private member is unused.
+      "-Ywarn-value-discard", // Warn when non-Unit expression results are unused.
+      "-Xmacro-settings:show-coder-fallback=true",
+      "-Ydelambdafy:inline" // Set the strategy used for translating lambdas into JVM code to "inline"
     )
+
+    VersionNumber(sys.props("java.version")) match {
+      case v if v.matchesSemVer(SemanticSelector(">1.8")) => base ++ List("-release", "8")
```
regadas, line 66: Since I use Java 11 I overlloked the fact that scalac issues an `error` if `-release` is applied on a java <= 1.8. This `error` somehow doesn't abort compilation.

PR #3037, opened 2020-06-11
Use a sampled subset of files to compute estimated SMB source size

ClaireMcGinty:
For sources with a very very high # of files, computing the estimated byte size of the SMB BoundedSource couldn't be done in memory as it materialized a list of every file in the Source (for the initial split(), the BoundedSource must estimate the byte size of every file in every bucket). This change introduces sampling of the total file set--it will only read the metadata of up to 1000 randomly selected files, and use that info to extrapolate the total byte size.

Tested on an internal use case with ~1.6 million unique files. `gsutil du` computes their total byte size to be ~4.57 TB. The sampling algorithm in this PR, using only 1000 files as a sample, resulted in an estimate of ~4.54 TB--so it seems fairly accurate, given the relatively small sample size. (Also, before this change, that computation ran out of memory and couldn't complete at all.)

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
@@ -442,7 +442,7 @@ public class SortedBucketSource<FinalKeyT> extends BoundedSource<KV<FinalKeyT, C
    * @param <V> the type of the values in a bucket
    */
   public static class BucketedInput<K, V> implements Serializable {
-    private static int samplingNumFilesThreshold = 1000;
+    private static int SAMPLING_NUM_FILES_THRESHOLD = 1000;
 
     private TupleTag<V> tupleTag;
     private String filenameSuffix;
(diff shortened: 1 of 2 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -419,6 +442,8 @@ public void close() throws IOException {}
    * @param <V> the type of the values in a bucket
    */
   public static class BucketedInput<K, V> implements Serializable {
+    private static int samplingNumFilesThreshold = 1000;
```
nevillelyh, line 445: Minor: should be `final int SAMPLING_NUM_FILES_THRESHOLD`?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -473,18 +498,101 @@ static CoGbkResultSchema schemaOf(List<BucketedInput<?, ?>> sources) {
       return sourceMetadata;
     }
 
-    List<ResourceId> listFilesForBucket(int bucketId, int targetParallelism) {
-      return mapBucketFiles(bucketId, targetParallelism, Function.identity());
+    long getOrSampleByteSize(int bucketId, int targetParallelism) {
+      final AtomicLong totalNumFiles = new AtomicLong(0L);
+      flatmapBucketShardIds(
```
nevillelyh, line 503: This can simply be `flatmapBucketShardIds(..., (bucketShardId, metadata) -> 1).sum()`?

nevillelyh: Alternatively, `mapBucketShardIds` into `List<ResourceId>`, do `Collections.shuffle(list)`, and `take(SAMPLING_NUM_FILES_THRESHOLD`? That'll also get rid of the random fraction trick later.

And you can probably keep `mapBucketShardIds` as is.

ClaireMcGinty: hmm, my idea here was to try to avoid materializing a list of (potentially) millions of `ResourceId`s (plus I think the GCS ResourceId/Path classes have some nontrivial initialization logic?), so the list is only ever <= 1000 ResourceIds.

PR #3039, opened 2020-06-15
Update Beam to 2.22.0

regadas:


scio-core/src/main/scala/com/spotify/scio/schemas/instances/JavaInstances.scala
```diff
@@ -83,6 +84,8 @@ trait JavaInstances {
 
       override def getArgumentType: FieldType = FieldType.STRING
     }))
+
+  implicit def jLocalDate: Type[LocalDate] = Type(CalciteUtils.DATE)
```
anish749, line 88: Thanks for adding this 🎉 !!

PR #3051, opened 2020-06-15
Use MonoidAggregator when available, fix #3050

### LINKED ISSUE #3050: Support `MonoidAggregator`
Currently `SCollection#aggregate*` takes `Aggregator[A, B, C]` and does `.map(a.prepare).sum(a.semigroup).present(c)` under the hood. This might fail for empty windows.

nevillelyh:


nevillelyh:
Yeah don't have strong opinion. I looked at Scalding and `MonoidAggregator` wasn't used anywhere in the code base. I wonder if majority of semigroups are already monoids and therefore `MonoidAggregator` should be the default?

regadas:
hm, I'm guessing they just don't leverage them 🤷 There are some `Aggregator` that are not `monoid`'s i.e `min/max` so i think supporting both is good! I lean however to have separate methods though.

nevillelyh:
Let me play around it first. `MonoidAggregator` sounds like an impl detail (although it's not) and might be confusing to users since it's not surfaced anywhere else in Algebird/Scalding docs.

nevillelyh:
Switched to overloading approach. Seems to work.

PR #3053, opened 2020-06-16
Update deprecated .toIterator with .iterator

nevillelyh:


scio-extra/src/main/scala-2.12/kantan/codecs/compat.scala
```diff
@@ -6,7 +6,7 @@ object compat {
 
   implicit class ResourceIteratorCompatOps[T](private val underlying: ResourceIterator[T])
       extends AnyVal {
-    def iterator: Iterator[T] = underlying.toIterator
+    def iterator: Iterator[T] = underlying.iterator
```
regadas, line 9: this one needs to be `toIterator`

scio-repl/src/main/scala-2.12/com/spotify/scio/repl/compat/CsvReaderOps.scala
```diff
@@ -4,6 +4,6 @@ import kantan.csv.CsvReader
 
 final class CsvReaderOps[A](val reader: CsvReader[A]) extends AnyVal {
 
-  def iterator: Iterator[A] = reader.toIterator
+  def iterator: Iterator[A] = reader.iterator
```
regadas, line 7: same here

nevillelyh: Seems to work for me locally will double check.

nevillelyh: Yeah this is fine. There was another breakage but all fixed.

regadas: ahhh cool! 😄 sorry about this I confused this with the one provided by `kantan` 🤦

PR #3058, opened 2020-06-17
Add support to allow static bq execution in traditional template

### LINKED ISSUE #3036: Using BigQuery fromQuery in a dataflow template
Hi! 

I'm trying to create a BigQuery to Bigtable (with some extra processing) dataflow template. I'm planning to trigger this dataflow job every 24 hrs. 

I'm seeing that after I create the template, the first run properly process all the data and finishes. If I run the same template again, it doesn't read any of the data. I also tried using `@BigQueryType.fromQuery()` and also `sc.bigQuerySelect(selectQuery)`, and both methods resulted in the same behavior. The select query fed into these is one that selects all rows so if anything there should be an equivalent or more number of rows than the previous run. 

I didn't think this would be a caching issue but just to be safe I did try generating the template with the `-Dbigquery.cache.enable=false` flag and that didn't work.

I do know that BigQuery temp tables are TTLed for 24 hrs (and in my attempts, my reruns have been <24 hrs apart), not sure if that has anything to do with it. 

I'm using scala `2.13.1`, scio `0.9.1`, beam `2.20.0`, and sbt `1.3.0`

I'm just trying to better understand what is going on?

regadas:
This is very much a first step. It only allows/fixes current impl to allow static bq execution using the "traditional" template scheme.

Fixes #3036

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -44,6 +44,7 @@ import org.apache.beam.sdk.transforms.SerializableFunction
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
+import org.apache.beam.sdk.options.ValueProvider
 
 private object Reads {
   private[this] val cache = new ConcurrentHashMap[ScioContext, BigQuery]()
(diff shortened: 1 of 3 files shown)


regadas (commented):
We believe that *Flex Templates* offers more flexibility and Google seems to be on the same page. For that reason we will not add *Traditional Template* support.

The solution for the referenced issue is to run the job as a *Flex Template* job

PR #3059, opened 2020-06-17
scalafmt 2.6.0

jto:


scio-avro/src/main/scala/com/spotify/scio/avro/package.scala
```diff
@@ -28,9 +28,11 @@ import com.spotify.scio.avro.syntax.Syntax
  */
 package object avro extends Syntax {
 
-  /** Typed Avro annotations and converters. */
+  /**
+   * Typed Avro annotations and converters. */
```
regadas, line 32: This are formatted in a funky way, is there a way to keep this one liners?

regadas: either that or we rewrite them to be `/* comment */` 🤷

Issue #3061, opened 2020-06-17
Support Dataflow templates


regadas:
Currently `Dataflow` has 2 types of templates:

* [ ] Traditional template
* [ ] Flex template

Related issues #696 and #2982 reference `ValueProvider`'s which is inherently related to `traditional` template model. To support this model we will need to address some changes in `IO` params.

Flex template, this model should work out of the box, however it would be good if we could have some docs around this and perhaps some `scio-flex-template.g8` to allow some quickstarts.

sfines-clgx:
Can I upvote this? I would like to upvote this. My usecase here is that I launch about 3k jobs to process a large amount of data in meaningful chunks, and driving this using other mechanisms is a real pain

bkhuong:
Any updates on the docs / quickstarts for scio flex templates?

regadas:
Hi @bkhuong sorry for late reply. This fell under my radar.

We don't have specific docs for yet (maybe you can help us out? 😄  ) but we do have support for them in our https://github.com/spotify/scio.g8

Give it a try and let us know.

regadas:
I'll close this issue as per discussion we will not add direct support for `Traditional templates` because `Google` is pushing for `Flex Templates`. Flex Templates are more versatile then `Traditional` even-though they still have some use cases where they shine. It's possible to use `Traditional Templates` by using `Beam IO` directly. 

Flex template, this model should work out of the box. I'll create a separate issue to track the addition of related docs.

PR #3069, opened 2020-06-18
Add ValueProvider support to TextIO

regadas:
Related to #3061


jto:
Do we have any status from Google about the status of `ValueProvider` ? 
I'd like to be sure they won't deprecate them soon-ish before we spend time adding support.

regadas:
It's a good question! from one of the conversations with them `Traditional Templates` were not being deprecated in favour of `Flex Templates` but simply providing a different templating approach. I think that still stands, but we should probably follow up with them.

nevillelyh:
To do this properly, we'll have to do this for all IOs right?

jto:
Can we merge this PR @regadas ?

regadas (commented):
We believe that *Flex Templates* offers more flexibility and Google seems to be on the same page. For that reason we will not add *Traditional Template* support.

PR #3079, opened 2020-06-22
add small SMB sink optimizations

ClaireMcGinty:
adds some opt-in features that may benefit certain pipelines -- both are pretty small/targeted improvements, but do work for specific pipeline cases

1) key cache -- if the SCollection is expected to contain many repeating keys, the user can configure a cache of configurable size mapping `K key --> (byte[], bucketId)`.
2) SMB write on pre-keyed collection. If key extraction is expensive/requires type conversion, and the user's pipeline already has a `map` operation before the SMB sink operation where the key is materialized, they can just modify the `map` operation to emit a `KV` tuple of (key, record):

```scala
// before
data.map(id => MyAvroRecord.newBuilder().setId(id).build()).saveAsSortedBucket(...)

// after
data.map(id => (id, MyAvroRecord.newBuilder().setId(id).build())).saveAsPreKeyedSortedBucket(...)
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1148,7 +1148,7 @@ lazy val `scio-smb`: Project = project
       "joda-time" % "joda-time" % jodaTimeVersion,
       "org.apache.beam" % "beam-vendor-guava-26_0-jre" % beamVendorVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
-      "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion
+      "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % "provided"
     ),
     beamSDKIODependencies,
     javacOptions ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -1147,7 +1147,8 @@ lazy val `scio-smb`: Project = project
       "com.google.protobuf" % "protobuf-java" % protobufVersion,
       "joda-time" % "joda-time" % jodaTimeVersion,
       "org.apache.beam" % "beam-vendor-guava-26_0-jre" % beamVendorVersion,
-      "org.slf4j" % "slf4j-api" % slf4jVersion
+      "org.slf4j" % "slf4j-api" % slf4jVersion,
+      "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -1147,7 +1147,8 @@ lazy val `scio-smb`: Project = project
       "com.google.protobuf" % "protobuf-java" % protobufVersion,
       "joda-time" % "joda-time" % jodaTimeVersion,
       "org.apache.beam" % "beam-vendor-guava-26_0-jre" % beamVendorVersion,
-      "org.slf4j" % "slf4j-api" % slf4jVersion
+      "org.slf4j" % "slf4j-api" % slf4jVersion,
+      "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion
```
nevillelyh, line 1151: This can be "provided" maybe?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -186,29 +256,117 @@ public void startBundle() {
...
+      this.extractKeyFn = extractKeyFn;
+      this.valueCoder = valueCoder;
+      this.cacheSize = cacheSize;
+      cacheHits = Metrics.counter(SortedBucketSink.class, "cacheHits");
+      cacheMisses = Metrics.counter(SortedBucketSink.class, "cacheMisses");
+    }
+
+    @Override
+    public ResourceType getResourceType() {
+      return ResourceType.PER_CLASS;
```
nevillelyh, line 323: Should be `PER_INSTANCE` in case there're multiple sinks with the same `K` type?

build.sbt
ClaireMcGinty: > Is there a way to validate the keys in `PreKeyed` sink matches metadata? Just for extra sanity check.

yeah I added a sampling transform that checks! Users can opt out of it if they're very confident/don't want to incur the cost of the extra transform: https://github.com/spotify/scio/pull/3079/files#diff-677cd9b1b47a87b82312e6df2ae18092R768 Unfortunately the Write transform doesn't depend on that check but the job will still fail.

PR #3090, opened 2020-06-25
Replace socco with socco-ng 0.1.1

### LINKED ISSUE #2186: socco examples don't compile with 2.12.9
This is dependent on socco being published for `2.12.9`

regadas:
Full cross compile version support and more updates added.

Closes #2186

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1289,7 +1289,7 @@ lazy val soccoSettings = if (sys.env.contains("SOCCO")) {
       "-P:socco:package_com.spotify.scio:https://spotify.github.io/scio/api"
     ),
     autoCompilerPlugins := true,
-    addCompilerPlugin("com.criteo.socco" %% "socco-plugin" % "0.1.9"),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -1289,7 +1289,7 @@ lazy val soccoSettings = if (sys.env.contains("SOCCO")) {
       "-P:socco:package_com.spotify.scio:https://spotify.github.io/scio/api"
     ),
     autoCompilerPlugins := true,
-    addCompilerPlugin("com.criteo.socco" %% "socco-plugin" % "0.1.9"),
+    addCompilerPlugin(("io.regadas" %% "socco-ng" % "0.1.1").cross(CrossVersion.full)),
```
jto, line 1292: Did you have to fork it ? Maybe we should consider migrating to another lib if socco is not maintained.

regadas: Yeah I did. The original one seems defunct as nobody responds to PRs. Having a different tool is a possibility, although the idea of having the code become document grew on me, and that’s why I forked it to keep it a bit alive and updated. I actually have some extra ideas to add to it.

PR #3091, opened 2020-06-25
Remove unnecessary cache of client interceptors

honnix:
Two reasons:

* `CredentialInterceptorCache` caches for default credentials already
* User should be able to provide their own credentials to get a
  corresponding interceptor instead of always using the default
  credentials


honnix:
@regadas PTAL

PR #3097, opened 2020-06-25
refactor SortedBucketTransform into a BoundedSource + reuse merge logic

ClaireMcGinty:
Refactored `SortedBucketTransform` to use a `BoundedSource` transform that shares its merging logic with `SortedBucketSource` and allows users to specify a `TargetParallelism`. Additionally I cleaned up the API a bit so it's not just partially fusing aspects of a `SortedBucketIO.Read` and `SortedBucketIO.Write`. The API now looks like:

```scala
sc.sortMergeTransform(
  classOf[KeyClass],
  AvroSortedBucketIO.read(...).from(...),
  AvroSortedBucketIO.read(...).from(...),
  TargetParallelism.auto()
).to(
  AvroSortedBucketIO
     .transformOutput(classOf[KeyClass], "key_field", classOf[ValueClass])
     .to("gs://my_output")
).via {
  case (key, (lhs, rhs), outputCollector) =>
     val v: ValueClass = ...
     outputCollector.accept(v)
}
```

The number of buckets in the output is now determined by the parallelism. I.e. if they use `TargetParallelism.auto()`, then the Dataflow runner heuristic will decide, or they can use `TargetParallelism.of(2)` to end up with 2 buckets in the output. numShards will always be 1 and the HashType will always match the source. The only aspects of the output metadata the user needs to explicitly specify is key field/key class.

I know the diff on this PR is really awful -- a lot of it is boilerplate AutoValue/Builder stuff to accommodate the API changes :/ Overall there is less duplicated logic between SMB Transform and SMB source/sink now and I think a less confusing API.

benchmarks:
not that surprisingly, there's not a huge change because the bulk of the job is not super parallelizable (writing to a single file per worker), but BoundedSource impl looks a bit better.

Existing, non-BoundedSource impl:
Screen Shot 2020-06-24 at 5 12 05 PM

New boundedSource impl:
Screen Shot 2020-06-25 at 2 42 34 PM

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
@@ -261,6 +261,35 @@ public class SortedBucketSource<FinalKeyT> extends BoundedSource<KV<FinalKeyT, C
         .collect(Collectors.toList());
   }
 
+  static int getFanout(
+      SourceSpec sourceSpec,
+      int effectiveParallelism,
+      TargetParallelism targetParallelism,
+      long estimatedSizeBytes,
+      long desiredSizeBytes) {
+    desiredSizeBytes *= DESIRED_SIZE_BYTES_ADJUSTMENT_FACTOR;
+
+    int greatestNumBuckets = sourceSpec.greatestNumBuckets;
+
+    if (effectiveParallelism == greatestNumBuckets) {
+      LOG.info("Parallelism is already maxed, can't split further.");
+      return 1;
+    }
+    if (!targetParallelism.isAuto()) {
+      return sourceSpec.getParallelism(targetParallelism);
+    } else {
+      int fanout = (int) Math.round(estimatedSizeBytes / (desiredSizeBytes * 1.0));
+
+      if (fanout <= 1) {
+        LOG.info("Desired byte size is <= total input size, can't split further.");
+        return 1;
+      }
+
+      // round up to nearest power of 2, bounded by greatest # of buckets
+      return Math.min(Integer.highestOneBit(fanout - 1) * 2, greatestNumBuckets);
+    }
+  }
+
   // `getEstimatedSizeBytes` is called frequently by Dataflow, don't recompute every time
   @Override
   public long getEstimatedSizeBytes(PipelineOptions options) throws Exception {
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
@@ -32,10 +32,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -65,7 +63,6 @@ import org.apache.beam.sdk.transforms.join.CoGbkResultSchema;
 import org.apache.beam.sdk.transforms.join.UnionCoder;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.TupleTag;
-import org.apache.beam.sdk.values.TupleTagList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.primitives.UnsignedBytes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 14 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -384,26 +397,34 @@ public boolean advance() throws IOException {
         }
 
         int keyGroupCount = 0;
+
+        // Track the canonical # buckets of each source that the key is found in.
+        // If we find it in a source with a # buckets >= the parallelism of the job,
+        // we know that it doesn't need to be re-hashed as it's already in the right bucket.
+        boolean keyFoundInSourceWithMatchingNumBuckets = false;
```
ClaireMcGinty, line 404: this is actually a generalization of the previous heuristic used to determine if a key needs re-hashing. Consider this example of a LHS with 1 bucket and a RHS with 2, where the parallelism is set to 2 (so number of buckets in output == 2).

```
Source A    | Source B               
Bucket 0    |  Bucket 0   |  Bucket 1
-------------------------------------
1           |  1          | 4
2           |  3
3                
4
```

Worker 1 is merging `SourceA-Bucket0` with `SourceB-Bucket0`. When we come across element `1`, we see it exists in Source B. Since source B has 2 buckets, which == the parallelism of the source, we know that if we performed a re-hashing of `1`, it would still be assigned to bucket 0 in a 2-bucket sink. On the other hand, when we get to element `2`, since it does not exist in Source B, we don't know where it would end up in a 2-bucket sink operation, so we must rehash.

Worker 2 is merging a copy of `SourceA-Bucket0` with `SourceB-Bucket1`. It would need to rehash 1, 2, and 3, but not 4. 

In datasets with fanout and a lot key overlap (like event data!) this should eliminate a lot of hashing operations.

@nevillelyh lmk if you can poke any holes in this idea, I tested it pretty thoroughly (including with partitioned sources where different partitions have different # buckets), but could be missing something

ClaireMcGinty:
@nevillelyh can you take a look some time this week?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
nevillelyh: Just thinking out loud: when `numBuckets` of a source >= parallelism, its pairs are read in a merge sort of 1 or more buckets, so any key in there are in the right place. We have to check keys from other sources (replicated to increase parallelism) whether they belong here. Did I get this right?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -384,26 +397,34 @@ public boolean advance() throws IOException {
...
+            entry.getValue().getValue().forEachRemaining(values::add);
+            keyGroupCount += values.size();
+
             nextKeyGroupsIt.remove();
           }
         }
 
+        final boolean acceptKeyGroup =
+            keyFoundInSourceWithMatchingNumBuckets
+                || keyGroupFilter.apply(minKeyEntry.getValue().getKey());
```
nevillelyh, line 426: This `keyGroupFilter.apply` is the hashing we're trying to save right? I'm not sure what costs more though, doing this once per source or `entry.getValue().getValue().forEachRemaining(values::add)` appending a `List<V>` even if we might filter the key group here?

I wonder if we can make `Boolean acceptKeyGroup = null` and change #410 to something like this:

```java
if (bucketsPerSource.get(tupleTag) >= parallelism || acceptKeyGroup) {
  // source # buckets >= parallelism = definitely keep
  // else, unsure, check hash
  entry.getValue().getValue().forEachRemaining(values::add);
  keyGroupCount += values.size();
} else if (keyGroupFilter.apply(minKeyEntry.getValue().getKey())) {
  acceptKeyGroup = true;
  entry.getValue().getValue().forEachRemaining(values::add);
  keyGroupCount += values.size();
} else {
  // skip key but still have to exhaust iterator
  entry.getValue().getValue().forEachRemaining(value -> {});
}
```

Basically, add items if numBuckets >= parallelism, otherwise lazily check hash, skip key if both are false. WDYT?

nevillelyh:
We can merge now, and maybe save that lazy `keyGroupFilter` as an optimization for later.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
ClaireMcGinty: Yep, exactly

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
ClaireMcGinty: that's a good idea -- should work. I'll update the PR!

ClaireMcGinty:
@nevillelyh yeah sounds good!  so merge it as-is for now and I'll put up another PR with the lazy keyGroupFilter after ?

PR #3098, opened 2020-06-25
Add support for String/ByteString TF Bucket key types

nmaraston:
`String`/`ByteString` key types are supported when writing TF `Example`s to SMB buckets (since the type need only be encodeable). These types are however not supported when reading TF `Example`s from said buckets. 

This PR adds support for `String`/`ByteString` key types when extracting SMB bucket keys.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketMetadata.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketMetadata.java
@@ -70,7 +70,7 @@ public class TensorFlowBucketMetadata<K> extends BucketMetadata<K, Example> {
     } else if (getKeyClass() == ByteString.class) {
       BytesList values = feature.getBytesList();
       Preconditions.checkState(values.getValueCount() == 1, "Number of feature in keyField != 1");
-      return (K) ByteString.copyFrom(values.getValue(0).toByteArray());
+      return (K) values.getValue(0);
     } else if (getKeyClass() == String.class) {
       BytesList values = feature.getBytesList();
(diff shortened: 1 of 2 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketMetadata.java
```diff
@@ -67,6 +67,14 @@ public K extractKey(Example value) {
       BytesList values = feature.getBytesList();
       Preconditions.checkState(values.getValueCount() == 1, "Number of feature in keyField != 1");
       return (K) values.getValue(0).toByteArray();
+    } else if (getKeyClass() == ByteString.class) {
+      BytesList values = feature.getBytesList();
+      Preconditions.checkState(values.getValueCount() == 1, "Number of feature in keyField != 1");
+      return (K) ByteString.copyFrom(values.getValue(0).toByteArray());
```
martinbomio, line 73: any reason why we copy it instead of just `values.getValue(0)`?

martinbomio:
lgtm!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketMetadata.java
nevillelyh: Good find. Shouldn't have to copy as `ByteString` is immutable.

nevillelyh:
```suggestion
      return (K) values.getValue(0);
```

Issue #2776, opened 2020-03-17
Possible improvements for reading many files


ClaireMcGinty:
Currently, our built-in file read apis (`sc.avroFile`, `sc.protobufFile`, `sc.textFile`, ...) take a single String `path` either matching a specific file, or a filepattern. The approach most users seem to take when reading many files is to either use`SCollection.unionAll` on many individual read transforms (can produce a job graph that's too huge to submit), or to specify a file pattern matching many files (not performant if it matches a huge # of files).

Although we do have the `SCollection#readFiles[A: Coder](filesTransform: PTransform[PCollection[beam.FileIO.ReadableFile], PCollection[A]])` method, it has the disadvantage that users have to work directly with Beam ReadFiles transforms.

I had two thoughts on optimizing user experience in either case:

1. Expose the `withHintMatchesManyFiles()` option in more ScioIO ReadParams. ([doc in Beam's AvroIO](https://github.com/apache/beam/blob/da9e17288e8473925674a4691d9e86252e67d7d7/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java#L672)). When this is set, the `Read` transform is converted to using a `ReadFiles` transform.

2. Offer `sc.avroFile`/`sc.protobufFile`/`sc.textFile` methods that take a `List[String]` of paths. I know this adds some clutter/maintanability issues but it does seem like a VERY common use case, and we don't have to offer it for every ScioIO.

wdyt?

nevillelyh:
Main concern re 2, it might cause method overload issues? If there's a way around it then 👍

regadas:
overloading will be an issue here, but maybe we can find some clever naming? I think we should work on this as I have seen more use cases requiring this.

regadas:
@alexclare wants to take a stab at this.

kellen:
My implicit class version of this in our repo is called `avroFiles` (plural). No collision issue.

PR #3110, opened 2020-06-27
Idiomatic APIs to support beam readFiles APIs

syodage:
Fixes: https://github.com/spotify/scio/issues/2776

Scio doesn't have a nice way to deal with beam `readFiles` (read multiple files) APIs. Scio users either have to use beam transform `PTransform[PCollection[beam.FileIO.ReadableFile], PCollection[A]]` or  use single file/file-pattern read API with  `SCollection.unionAll(..)` to read multiple filepatterns/files. Here I have extended `TextIO`, `GenericRecordIO` and `SpecificRecordIO` to support both single file/filepattern and multiple files/filepatterns. 

User have to use slightly different APIs available in ScioContext e.g `sc.textFile(path)` ,  `sc.textFiles(paths)` to read single or multiple files respectively. 

The PR introduces the following new APIs to users along with existing single file/filepattern read APIs
- `sc.textFiles(paths)`
with `import com.spotify.scio.avro._` 
- `sc.avroFiles(paths, schema)`
- `sc.avroFiles[T <: SpecificRecord: ClassTag: Coder](paths)`
- `sc.objectFiles(paths)`
- `sc.protobufFiles(paths)`
- `sc.typedAvroFiles(paths)`

Added `AvroReadFilesExample`

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -86,6 +86,7 @@ object ObjectFileIO {
   val WriteParam = AvroIO.WriteParam
 }
 
+/** Read multiple files/file-patterns with beam AvroIO readFiles API */
 final case class ObjectReadFilesIO[T: Coder](paths: Iterable[String]) extends AvroIO[T] {
   override type ReadP = Unit
   override type WriteP = Nothing
@@ -143,6 +144,7 @@ object ProtobufIO {
   val WriteParam = AvroIO.WriteParam
 }
 
+/** Read multiple files/file-patterns with beam AvroIO readFiles API */
 final case class ProtobufReadFilesIO[T <: Message: ClassTag](paths: Iterable[String])
     extends AvroIO[T] {
   override type ReadP = Unit
@@ -213,6 +215,7 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
     SpecificRecordTap[T](ScioUtil.addPartSuffix(path))
 }
 
+/** Read multiple files/file-patterns with beam AvroIO readFiles API */
 final case class SpecificRecordReadFilesIO[T <: SpecificRecord: ClassTag: Coder](
   paths: Iterable[String]
 ) extends AvroIO[T] {
@@ -339,6 +342,15 @@ final case class GenericRecordParseIO[T](path: String, parseFn: GenericRecord =>
     GenericRecordParseTap[T](ScioUtil.addPartSuffix(path), parseFn)
 }
 
+/**
+ * Given a parseFn, read [[org.apache.avro.generic.GenericRecord GenericRecord]]
+ * and apply a function mapping [[GenericRecord => T]] before producing output.
+ * This IO applies the function at the time of de-serializing Avro GenericRecords.
+ *
+ * This IO doesn't define write, and should not be used to write Avro GenericRecords.
+ *
+ * This IO read multiple files/file-patterns with beam AvroIO readFiles API
+ */
 final case class GenericRecordParseFilesIO[T](paths: Iterable[String], parseFn: GenericRecord => T)(
   implicit coder: Coder[T]
 ) extends AvroIO[T] {
@@ -447,6 +459,7 @@ object AvroTyped {
     }
   }
 
+  /** Read multiple files/file-patterns with beam AvroIO readFiles API */
   final case class AvroReadFilesIO[T <: HasAvroAnnotation: ClassTag: TypeTag: Coder](
     paths: Iterable[String]
   ) extends ScioIO[T] {
@@ -488,8 +501,10 @@ trait AvroReadFilesIO[T] extends ScioIO[T] {
   final override val tapT = TapOf[T]
 }
 
+
 object AvroReadFilesIO {
 
+  /** TestIO implementation to use with JobTests. */
   @inline final def apply[T](paths: Iterable[String]): AvroReadFilesIO[T] =
     new AvroReadFilesIO[T] with TestIO[T] {
       override def testId: String = s"AvroReadFilesIO($paths)"
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -447,3 +447,15 @@ object AvroTyped {
     }
   }
 }
+
+trait AvroReadFilesIO[T] extends ScioIO[T] {
+  final override val tapT = TapOf[T]
+}
+
+object AvroReadFilesIO {
+
+  @inline final def apply[T](paths: Iterable[String]): AvroReadFilesIO[T] =
+    new AvroReadFilesIO[T] with TestIO[T] {
+      override def testId: String = s"AvroReadFilesIO($paths)"
+    }
+}
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -446,6 +446,42 @@ object AvroTyped {
         .map(avroT.fromGenericRecord)
     }
   }
+
+  final case class AvroReadFilesIO[T <: HasAvroAnnotation: ClassTag: TypeTag: Coder](
+    paths: Iterable[String]
+  ) extends ScioIO[T] {
+    override type ReadP = Unit
+    override type WriteP = Nothing
+    final override val tapT = TapOf[T]
+
+    /**
+     * Get a typed SCollection from an Avro schema. Supports reading multiple files/file-patterns at once.
+     *
+     * Note that `T` must be annotated with
+     * [[com.spotify.scio.avro.types.AvroType AvroType.fromSchema]],
+     * [[com.spotify.scio.avro.types.AvroType AvroType.fromPath]], or
+     * [[com.spotify.scio.avro.types.AvroType AvroType.toSchema]].
+     */
+    override protected def read(sc: ScioContext, params: ReadP): SCollection[T] = {
+      val avroT = AvroType[T]
+      sc.wrap(
+        sc.applyInternal(
+          Create.of(paths.asJava)
+        ).apply(beam.FileIO.matchAll())
+          .apply(beam.FileIO.readMatches())
+          .apply(beam.AvroIO.readFilesGenericRecords(avroT.schema))
+      ).map(avroT.fromGenericRecord)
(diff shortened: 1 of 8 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -141,21 +141,32 @@ sealed trait AvroIO[T] extends ScioIO[T] {
       .withMetadata(metadata.asJava)
 }
 
-final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: String)
+final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](paths: Iterable[String])
```
jto, line 144: Could be defined this way to avoid breaking back-compat:

```suggestion
final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](paths: String*)
```

Alternatively:

```suggestion
final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: String, paths: String*)
```

which guaranties at least 1 path is passed

jto:
It is indeed a simpler API but it does not cover all the cases since `paths` needs to be known when the graph is constructed (vs when the job is actually executed). `readAll` is therefore still necessary. Still a worthy addition though. 

IMHO it should be a separate IO bc. by definition multiple files path can only be used to read data (not writing). Having that in a separate IO would also guarantee back-compatibility.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -166,46 +177,74 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
...
-    val t = beam.AvroIO
-      .readGenericRecords(schema)
-      .from(path)
-    sc.wrap(sc.applyInternal(t))
+    sc.wrap(
+      if (paths.size == 1) {
+        val t = beam.AvroIO
+          .readGenericRecords(schema)
+          .from(paths.head)
+        sc.applyInternal(t)
```
jto, line 212: I don't like that. Since multiple `paths` can only be used for reads, I think it should have its own read-only IO rather than shoehorning it into the existing IOs

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -166,46 +177,74 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
...
+}
+
+final case class GenericRecordIO(paths: Iterable[String], schema: Schema)
+    extends AvroIO[GenericRecord] {
   override type ReadP = Unit
   override type WriteP = AvroIO.WriteParam
 
-  override def testId: String = s"AvroIO($path)"
+  // TODO: support test with multiple paths
+  override def testId: String = s"AvroIO(${paths.head})"
```
jto, line 200: Does not seem super safe.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -166,46 +177,74 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
...
     data.applyInternal(
-      avroOut(data, t, path, params.numShards, params.suffix, params.codec, params.metadata)
+      avroOut(data, t, paths.head, params.numShards, params.suffix, params.codec, params.metadata)
     )
     tap(())
   }
 
   override def tap(read: ReadP): Tap[T] =
-    SpecificRecordTap[T](ScioUtil.addPartSuffix(path))
+    SpecificRecordTap[T](ScioUtil.addPartSuffix(paths.head))
```
jto, line 186: Unless I'm missing smth, `tap` should not be supported when `path.length > 1`

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
syodage: yes, we need to improve to handle multiple files, it is one of the todos in the PR. This is just to get all existing tests to pass.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
syodage: Since we have `apply` function which takes only one string argument. this won't introduce any back-compatibility issue.  But of course, this is another way to do it. the only downside with varargs approach is, it takes away passing any additional argument at the end of the constructor if we might need it in the future.

jto: Yeah sorry I did not see the added `apply`. You can disregard this comment ^^

syodage:
@jto `readAll` is deprecated in favor of `readFiles` API and as per the deprecation note, it won't get any improvement. Therefore IMO we shouldn't introduce `readAll` impls. 

> IMHO it should be a separate IOs bc. by definition multiple files path can only be used to read data (not writing). 

Agree, and I noticed most of the IOs we have are read-only or write-only, due to their own reasoning. This makes me think should we have `ScioIORead` and `ScioIOWrite` whose are subtypes of `ScioIO` and in our idiomatic APIs in `ScioContext` takes `ScioIORead` types and SCollection` takes `ScioIOWrite` types.  Which is more type-safe than existing.

This `ScioIO#write()` should have a way to get the new read params to construct the tap.  or maybe call tap 
 after write(which returns `Tappable`) e.g `scoll.write(SomeIO(...)).tap(readParam)` which returns SCollection. I think this would make it easier to understand the code. 

This of course introduces breaking changes. we can think of a way to handle that.

jto:
The existing API is already type safe and you already can declare your IO the be ro or rw simply by setting the type of `ReadP` or `WriteP` to `Nothing`. 

`tap`  only makes sense for IOs that support both `read` and `write`. If only `read` is supported, `tap` is the IO itself which is kinda useless. If only `write` is supported, you can't read the data, so `tap` can't possibly work and you'd need a separate IO.

jto:
The existing API is already type safe and you already can declare your IO the be ro or rw simply by setting the type of `ReadP` or `WriteP` to `Nothing`. 

`tap`  only makes sense for IOs that support both `read` and `write`. If only `read` is supported, `tap` is the IO itself which is kinda useless. If only `write` is supported, you can't read the data, so `tap` can't possibly work and you'd need a separate IO.

syodage:
I find this is interesting would like to discuss more https://github.com/spotify/scio/issues/3116 

I think creating read-only IO implementation for `readFiles` makes more sense than poluting exiting IOs. I will update the PR with that.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -175,6 +213,32 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
     SpecificRecordTap[T](ScioUtil.addPartSuffix(path))
 }
 
+final case class SpecificRecordReadFilesIO[T <: SpecificRecord: ClassTag: Coder](
+  paths: Iterable[String]
```
ClaireMcGinty, line 217: I'm a little hesitant about the use of the ReadAll APIs (`FileIO.matchAll()` -> `FileIO.readMatches() -> beam.AvroIO.readFiles()`) here for any `Iterable` of paths. From what I've seen in the documentation/ad hoc performance tests, ReadAll improves performance only when the # of file paths is very large (in the thousands and ten-thousands) -- or when taking the Union of many single-path Avro reads makes the job graph too large to submit to Dataflow.

I think it would be good to add this info about possible performance implications to the scaladoc for the `ScioContextSyntax` methods, and maybe logging a warning if `paths.size` is less than 10,000 that the user should switch to `SCollection.union(paths.map(sc.avroFile(_)))`... @regadas  / @jto wdyt?

[Beam documentation reference](https://github.com/apache/beam/blob/494256fbc1528c1ca7d0740188c1d96e23a93756/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java#L185-L190) (note that `Read#withHintMatchesManyFiles` converts a Read to a ReadAll transform under the hood):

```java
 * <h3>Reading a very large number of files</h3>
 *
 * <p>If it is known that the filepattern will match a very large number of files (e.g. tens of
 * thousands or more), use {@link Read#withHintMatchesManyFiles} for better performance and
 * scalability. Note that it may decrease performance if the filepattern matches only a small number
 * of files.
```

syodage: @ClaireMcGinty this is a good catch. I think logging a warning would be a good idea since small number of file patterns may match with large number of files we can't depend only on `paths.size`.

syodage: @ClaireMcGinty updated the API doc with this information, about the warning log, I am a bit concern about adding it, since it is hard to identify the number of matching file beforehand and for the valid case, this will be polluting the output. If this is something we do in other cases in the scio then I can add.

ClaireMcGinty: Yeah, agreed that it's hard to do for filepatterns. I think the documentation is 👍

ClaireMcGinty:
I feel a little conflicted that the input types for these are `paths: Iterable[String]`. In the underlying Beam `ReadAll` APIs, the input isn't a static list of file patterns/file paths, but a `PCollection` of file patterns/file paths, which are possibly dynamically created. In that case the user could use `SCollection#readFiles`, but I think it's confusing that the new IOs in this pull request wouldn't handle that case.

Just a thought, what if we had some kind of typeclass for ScioIO inputs? like:

```scala
sealed trait ReadInput
case class SinglePathInput(path: String) extends ReadInput
case class MultiplePathsInput(path: Iterable[String]) extends ReadInput
case class SCollectionPathsInput(path: SCollection[String]) extends ReadInput
```

Then the `ScioIOs` could just take an input of type `ReadInput` and case-match on which implementation is being used.

I guess the conversion logic between Strings <-> `ReadInput` types in the user-facing APIs would add a lot of complexity to the codebase in its own way though. @jto what do you think?

syodage:
@ClaireMcGinty @regadas, Yes agree with both of you, also I did check the performance of the new APIs with 5000 to 10000 files still the `SCollection.unionAll(paths.map(sc.avroFile, schema))` is the winner both in Wall time and resource consumption. 

It seems all these lines of codes give us literally nothing other than nice user API with poor performance. Considering maintain overhead and poor performance (unless use wisely), better we avoid exposing these APIs to the end-users. Let's close this one.

PR #3114, opened 2020-06-29
Add support for read only `ensureTable`

### LINKED ISSUE #3070: Support a read-only version of "ensureTables"
Currently, `TableAdmin.ensureTables()` assumes permissions such as "bigtable.tables.update" even when a table exists. This means the deploy user needs permissions we otherwise would not grant. Ideally, there would be a method like `ensureTables()` that checks for existence only, making no attempts to change anything.

In 0.7.4 we observed that `ensureTables()` effectively acted like this. That is to say, without granting additional permissions, calling `ensureTables()` worked fine as long as the tables in fact existed. If they did not exist, the check would fail (due to missing permissions) and we could separately address it with a different user.

Here's an example of current behavior when calling `ensureTables()`, when the table + CFs already exist, with a user lacking the correct permissions:

```
[main] INFO com.spotify.scio.bigtable.TableAdmin$ - Modifying or updating 9 column families for table ********************************************
Jun 02, 2020 10:03:59 PM com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation error
SEVERE: Could not complete RPC. Failure #0, got: Status{code=PERMISSION_DENIED, description=Access denied. Missing IAM permission: bigtable.tables.update., cause=null} on channel 1.
Trailers: Metadata(bigtable-channel-id=1)
Exception in thread "main" io.grpc.StatusRuntimeException: PERMISSION_DENIED: Access denied. Missing IAM permission: bigtable.tables.update.
```

Here's the current version:
https://github.com/spotify/scio/blob/master/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala#L71-L87

And the earlier version (0.7.4):
https://github.com/spotify/scio/blob/v0.7.4/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala#L51-L97

jto:
fixes #3070

### DIFF
--- a/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
+++ b/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
@@ -32,6 +32,13 @@ import scala.util.Try
 
 /** Bigtable Table Admin API helper commands. */
 object TableAdmin {
+
+  sealed trait CreateDisposition
+  object CreateDisposition {
+    case object Never extends CreateDisposition
+    case object CreateIfNeeded extends CreateDisposition
+  }
+
   private val log: Logger = LoggerFactory.getLogger(TableAdmin.getClass)
 
   private def adminClient[A](
@@ -78,12 +85,13 @@ object TableAdmin {
    */
   def ensureTables(
     bigtableOptions: BigtableOptions,
-    tablesAndColumnFamilies: Map[String, Iterable[String]]
+    tablesAndColumnFamilies: Map[String, Iterable[String]],
+    createDisposition: CreateDisposition /*= CreateDisposition.CreateIfNeeded*/
   ): Unit = {
     val tcf = tablesAndColumnFamilies.iterator.map {
       case (k, l) => k -> l.map(_ -> None)
     }.toMap
-    ensureTablesImpl(bigtableOptions, tcf).get
+    ensureTablesImpl(bigtableOptions, tcf, createDisposition).get
   }
 
   /**
@@ -101,7 +109,8 @@ object TableAdmin {
    */
   def ensureTablesWithExpiration(
     bigtableOptions: BigtableOptions,
-    tablesAndColumnFamilies: Map[String, Iterable[(String, Option[Duration])]]
+    tablesAndColumnFamilies: Map[String, Iterable[(String, Option[Duration])]],
+    createDisposition: CreateDisposition
   ): Unit = {
     // Convert Duration to GcRule
     val x = tablesAndColumnFamilies.iterator.map {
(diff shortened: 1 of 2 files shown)


scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -154,20 +166,24 @@ object TableAdmin {
...
+                  .setParent(instancePath)
+                  .setTableId(table)
+                  .build()
+              )
+            case CreateDisposition.Never =>
+              log.info("Table {} does not exist", table)
           }
 
-          ensureColumnFamilies(client, tablePath, columnFamilies)
+          ensureColumnFamilies(client, tablePath, columnFamilies, createDisposition)
```
nevillelyh, line 186: We don't need to pass `createDisposition` here right? Can just fail fast on line 183 and throw an exception? Also it's probably a desired behavior to throw exception than do nothing silently?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -154,20 +166,24 @@ object TableAdmin {
...
+              log.info("Creating table {}", table)
+              client.createTable(
+                CreateTableRequest
+                  .newBuilder()
+                  .setParent(instancePath)
+                  .setTableId(table)
+                  .build()
+              )
+            case CreateDisposition.Never =>
+              log.info("Table {} does not exist", table)
```
nevillelyh, line 183: Should probably be `log.error` and fail here?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
regadas: this is allowing the case where table exists and users want to modify CF's... they will only be able to modify it if they pass `CreateIfNeeded`

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
```diff
@@ -183,7 +199,8 @@ object TableAdmin {
   private def ensureColumnFamilies(
     client: BigtableTableAdminClient,
     tablePath: String,
-    columnFamilies: Iterable[(String, Option[GcRule])]
+    columnFamilies: Iterable[(String, Option[GcRule])],
+    createDisposition: CreateDisposition
   ): Unit = {
     val tableInfo =
```
regadas, line 205: Since we only modify if `CreateIfNeeded` can this be moved down?

scio-bigtable/src/main/scala/com/spotify/scio/bigtable/TableAdmin.scala
jto: Indeed

PR #3115, opened 2020-06-29
Update to use testId instead of toString

syodage:


syodage:
> Is this fixing a bug?

I encounter this while working on `readFiles` api improvement. 

> IIRC We identify ScioIOs by .toString which most IOs overload, and encode extra info like path, pubsub topic, etc. This change will treat all IOs in the same job with the same test ID and is most likely not what we want?

Shouldn't the same IO with different params have different `testId`?  code is inconsistent as we use the `io.testIO -> value` in the input map(see the line following the changed line) but check with `io.toString`. So far this worked because `io.testId` is equivalent to `io.toString`

nevillelyh:
Ah I remembered wrong. `ScioIO#testId` defaults to `def testId: String = this.toString`. Not the same semantics as `testId` used by `JobTest`.
https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala#L68

So it's inconsistent to have `ScioIO#toString` & `ScioIO#testId` behaving differently. Are there other uses of `.toString` we should clean up?

syodage:
yes, `testId` default to `toString` in ScioIO. 

Couldn't find any other uses of `io.toString` instead of  the `io.testId`

PR #3118, opened 2020-06-30
Cleanup scio-repl classloader

regadas:


scio-repl/src/main/scala/com/spotify/scio/repl/ReplScioContext.scala
```diff
@@ -38,11 +57,53 @@ class ReplScioContext(options: PipelineOptions, artifacts: List[String])
...
+  private def createJar(): String =
+    createJar(AbstractFile.getDirectory(Path(replOutputDir)), tempJar).getPath()
+
+  /**
+   * Creates a jar file from the classes contained in a virtual directory.
+   *
+   * @param virtualDirectory containing classes that should be added to the jar
+   */
+  private def createJar(dir: Iterable[AbstractFile], jarFile: File): File = {
+    println(dir)
```
jto, line 74: Forgot to delete that ?

PR #3121, opened 2020-06-30
Add support for PipelineOptions in ContextAndArgs

### LINKED ISSUE #3120: Support PipelineOptions in `ContextAndArgs.typed`
It seems that users sometimes rely on `PipelineOptions` as a way to parse arguments.

Currently the usage is:
```scala
trait Options extends PipelineOptions {
  ...
}
```

```scala
val (opts, remainingArgs) = ScioContext.parseArguments[Options](args, withValidation)
val sc = ScioContext(opts)
```

Maybe we could provide the same api as we provide for `case-app` use cases.

```scala
val (sc, opts): (ScioContext, Options) = ContextAndArgs.typed[Options](args)
```

regadas:
Closes #3120

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -394,6 +394,14 @@ lazy val protobufSettings = Def.settings(
   )
 )
 
+def splitTests(tests: Seq[TestDefinition], filter: Seq[String]) = {
+  val (filtered, default) = tests.partition(test => filter.contains(test.name))
+  val policy = Tests.SubProcess(ForkOptions())
+  new Tests.Group(name = "<default>", tests = default, runPolicy = policy) +: filtered.map { test =>
+    new Tests.Group(name = test.name, tests = Seq(test), runPolicy = policy)
+  }
+}
+
 lazy val root: Project = Project("scio", file("."))
   .settings(commonSettings)
   .settings(noPublishSettings)
@@ -562,7 +570,11 @@ lazy val `scio-test`: Project = project
       "com.propensive" %% "magnolia" % magnoliaVersion
     ),
     beamSDKIODependencies,
-    (Test / compileOrder) := CompileOrder.JavaThenScala
+    (Test / compileOrder) := CompileOrder.JavaThenScala,
+    Test / testGrouping := splitTests(
+      (Test / definedTests).value,
+      List("com.spotify.scio.ArgsTest")
+    )
   )
   .configs(
     IntegrationTest
@@ -1041,7 +1053,11 @@ lazy val `scio-examples`: Project = project
     },
     sources in doc in Compile := List(),
     run / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat,
-    Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
+    Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat,
+    Test / testGrouping := splitTests(
+      (Test / definedTests).value,
+      List("com.spotify.scio.examples.WordCountTest")
+    )
   )
   .dependsOn(
     `scio-core`,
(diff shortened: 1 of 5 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -239,6 +239,37 @@ object ContextAndArgs {
...
 
+  final case class PipelineOptionsParser[T <: PipelineOptions: ClassTag] private ()
+      extends ArgsParser[Try] {
+    override type ArgsType = T
+
+    override def parse(args: Array[String]): Try[Result] = Try {
+      val (opts, remainingArgs) = ScioContext.parseArguments[T](args, withValidation = true)
+      Either.cond(remainingArgs.asMap.isEmpty, (opts, opts), s"Unused $remainingArgs")
+    } match {
+      case r @ Success(Right(v)) => r
```
regadas, line 250:
```suggestion
      case r @ Success(Right(_)) => r
```

jto:
Nice. I actually think this is better thsn case-app and should be the recommended way.

build.sbt
```diff
@@ -394,6 +394,14 @@ lazy val protobufSettings = Def.settings(
   )
 )
 
+def splitTests(tests: Seq[TestDefinition], filter: Seq[String]) = {
```
regadas, line 397: Tests run within the same group which might use or not the same Classloader. One way to guarantee diff CL is to split Tests into a separate group.

nevillelyh: We're assuming that users will never run into conflicting `PipelineOptions` in the same build & having to do this right?

Although I really don't see the appeal of `trait Options extends PipelineOptions` with getters/setters over case classes. If we want to really opinionated, case app should be the way to go?

OTOH there's `ScioContext#apply(opts: PipelineOptions)`, can users just do that instead of `ContextAndArgs` instead?

regadas: yeah they will need to do smth similar to this depending on their build. Note that this change does not introduce this issue, it merely surfaced it while adding more tests.

This change is a nice to have. IMO it makes the API consistent when it comes to argument parsing. Most of our examples use `ContextAndArgs` and the ones that use `case-app` use the `ContextAndArgs.typed` variant. If users want to use `PipelineOptions` they can use what you mentioned 1) `ScioContext#apply(opts: PipelineOptions)` but can also use:
2 ) 
```
val (opts, remainingArgs) = ScioContext.parseArguments[Options](args, withValidation)
val sc = ScioContext(opts)
```

The usage is completely different and there's no reason to be. on 1) and 2) users have also the extra step of doing `sc.as(Options.class).getProperty` as opposed to direct access:

```
val (sc, {Args | CaseApp}) = ContextAndArgs
```

With this change we add:

```
val (sc, {Args | CaseApp | T <: PipelineOptions}) = ContextAndArgs
```

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -239,6 +239,37 @@ object ContextAndArgs {
...
+      case f                     => f
+    }
+  }
+
+  sealed trait TypedArgsParser[T, F[_]] {
+    def parser: ArgsParser[F]
+  }
+
+  sealed trait LowPrioTypedArgsParser {
+    implicit def caseApp[T: Parser: Help]: TypedArgsParser[T, Try] = new TypedArgsParser[T, Try] {
```
nevillelyh, line 261: Would a `T <: Product` here work as alternative to low-prio hack?

regadas: Yeah good point, i think it will.

regadas: I was thinking about case-app command support which is a sum type ... but i think it will not actually work with Parse/Help bounds

PR #3123, opened 2020-06-30
Reset PipelineOptions registry before running in JobTest

regadas:


regadas:
Actually this is not a good solution since JobTest can run in parallel ...

PR #3129, opened 2020-07-01
Compile scio-repl with -release 8

regadas:


regadas:
On a second thought we need to keep it as is because we need to keep java 11 Classloader logic.

scio-repl can be compiled/publish with java 8/11 produced artifacts will work on either versions.

PR #3140, opened 2020-07-07
AvroBucketMetadata should validate keyPath (fix #3038)

### LINKED ISSUE #3038: `AvroBucketMetadata` should validate keyPath
When extracting keys since it's stringly typed. A typo in the key would cause `null` value & a single null bucket.

ClaireMcGinty:
validates `keyPath` against the Avro schema as well as type-checking `keyClass` to the type Avro returns when calling #get on a `GenericRecord`.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
@@ -22,14 +22,21 @@ import static org.apache.beam.sdk.coders.Coder.NonDeterministicException;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Preconditions;
+import com.google.protobuf.ByteString;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Map;
-import java.util.Objects;
+import java.util.Set;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.reflect.ReflectData;
+import org.apache.avro.specific.SpecificRecordBase;
 import org.apache.beam.sdk.coders.AtomicCoder;
 import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.CannotProvideCoderException;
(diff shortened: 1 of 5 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -54,9 +58,36 @@ public AvroBucketMetadata(
...
+    this(
+        BucketMetadata.CURRENT_VERSION,
+        numBuckets,
+        numShards,
+        keyClass,
+        hashType,
+        validateKeyField(
+            keyField,
+            keyClass,
+            new ReflectData(recordClass.getClassLoader()).getSchema(recordClass)));
```
nevillelyh, line 73: What's the use case for `ReflectData`?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -108,6 +139,100 @@ public boolean isPartitionCompatible(BucketMetadata o) {
...
+          hasNull && schema.getTypes().size() == 2,
+          "A union can only be used as a key field if it contains exactly 1 null "
+              + "type and 1 key-class type. Was: "
+              + schema);
+    }
+
+    return getClassForType(schemaType);
+  }
+
+  private static Class<?> getClassForType(Schema.Type schemaType) {
```
nevillelyh, line 199: Surprised that Avro doesn't already have a mapping table like this? Or maybe it's private?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
nevillelyh: Ah looks like you're using this to get schema for `SpecificRecord` right?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
ClaireMcGinty: Sort of... it has `SpecificData#getClass`:

https://github.com/apache/avro/blob/release-1.8.2/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L152

But it doesn't support Fixed or Record; behaves strangely for union types (`"type": "string"` returns `String.class` ; whereas `"type":["string", "null"]` returns `CharSequence.class`); and returns null for Enum types unless there's a Java class of the same name in the classpath*. I guess we could check for `Record`, `Fixed`, `Enum`, and `String` types ourselves first and then delegate to `SpecificData#getClass`.

* Which actually, we could maybe make a ticket to support--since right now `extractKey` just works by casting the raw value from the generic record to type `K`: `K key = (K) node.get(keyPath[keyPath.length - 1]);`. We could have some logic that if the key schema is an enum we could use reflection on the enum class like `EnumClass.ValueOf((String) node.get(keyPath[keyPath.length - 1])`. but yeah... not sure if it's a big enough use case to matter.

nevillelyh: If enum is the only edge case right now probably not worth it. Enum doesn't have enough range and shouldn't be used as key anyway.

PR #3148, opened 2020-07-09
Clarify T <: HasAnnotation vs T: Schema methods in 0.9 migration doc

ClaireMcGinty:
related to a support question yesterday about the migration path not working for a `@BigQueryType.fromTable` read when the user switched to the `typedBigQueryTable` -- a Schema could not be inferred for the non-case-class. I wanted to add some clarification on the new BQ Read APIs since the `typedBigQueryTable[T: Schema]` may not work for every use case. 

also, Git is displaying the diff terribly! I just added alternate methods for the first two API examples, as well as a clarifying paragraph.


ClaireMcGinty:
sounds good! yeah there's actually no info on the `Schema`-based APIs in the Type-safe BigQuery docs, I'll move it there 👍

PR #3159, opened 2020-07-16
Add internal composite transfrom

regadas:


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -48,11 +45,8 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
 
   private[this] val context: ScioContext = self.context
 
-  private[this] val toKvTransform =
-    ParDo.of(Functions.mapFn((kv: (K, V)) => KV.of(kv._1, kv._2)))
-
   private[scio] def toKV(implicit koder: Coder[K], voder: Coder[V]): SCollection[KV[K, V]] =
-    self.applyTransform(toKvTransform).setCoder(CoderMaterializer.kvCoder[K, V](context))
+    self.map(kv => KV.of(kv._1, kv._2)).setCoder(CoderMaterializer.kvCoder[K, V](context))
```
jto, line 49: `kvCoder` should already be set by `map` so it's either unnecessary or need to be passed to `map` directly to avoid setting a `Coder` twice.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -422,9 +422,10 @@ final case class TableRowJsonIO(path: String) extends ScioIO[TableRow] {
       .map(e => ScioUtil.jsonFactory.fromString(e, classOf[TableRow]))
 
   override protected def write(data: SCollection[TableRow], params: WriteP): Tap[TableRow] = {
-    data
-      .map(e => ScioUtil.jsonFactory.toString(e))
-      .applyInternal(data.textOut(path, ".json", params.numShards, params.compression))
+    data.transform_("BigQuery write") {
```
jto, line 425: Could that break streaming jobs when they try to update ?

scio-core/src/main/scala/com/spotify/scio/io/PubsubIO.scala
```diff
@@ -344,22 +344,17 @@ final private case class PubsubIOWithAttributes[T: ClassTag: Coder](
...
-
+    data.transform_ { coll =>
+      coll
+        .withName("Encode Pubsub message and attributes")
+        .map { kv =>
+          val payload = CoderUtils.encodeToByteArray(coder, kv._1)
+          val attributes = kv._2.asJava
+          new beam.PubsubMessage(payload, attributes)
+        }
+        .setCoder(PubsubMessageWithAttributesCoder.of())
```
jto, line 355: Same here, the Coder is already set by `map`

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
regadas: yeah I noticed this one but it cant be changed right now. There are some ops that depend on this method and requires the coder to be explicitly `KvCoder`. If we set it through `map` it will still end up as `NullableCoder(KvCoder(NullableCoder(StringUtf8Coder),NullableCoder(StringUtf8Coder)))` which doesn't fly with `GroupByKey`.

I plan to look at this in a separate PR.

regadas: Hi again @jto! I was thinking, could this fit the bill? #3164 ... there are cases like this where we want the beam coder only and not even wrapped in the `NullCoder` (if the case) coz it doesn't make sense.

PR #3163, opened 2020-07-16
Avoid circular dep between SCollection and PCollectionWrapper

regadas:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -184,6 +184,39 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     ensureSerializable(coder).fold(throw _, pApply(name, transform).setCoder)
   }
 
+  private[scio] def pApply[U](
```
regadas, line 187: I would like also to rename this method and be inline with the `applyTransform`

nevillelyh: IIRC `applyTransform` calls `pApply` and the only extra thing it does is setting output coder right? Something like `apply[Transform]Internal`?

regadas: yeah pretty much. Regarding the name yeah smth like that, but let me think about it ... i also don't wanna mix the renaming with this.

PR #3164, opened 2020-07-20
Add raw beam coder materializer

regadas:
The idea here is to allow us to set the raw beam coder (native or not) without being wrapped in  `NullableCoder` and `WrappedBCoder`. This is useful in situations where 1) the underlying transform checks for a given coder instance and 2) where wrapping it in a `NullableCoder` does not make sense.


syodage:
t]Jus to understand, What's the difference between, this (`RawCoder[T]`) and `Beam[T]`?

regadas:
@syodage updated PR descp.

syodage:
@regadas thanks for the description.

PR #3166, opened 2020-07-20
Remove unused param of internal partitionFn

regadas:


scio-core/src/main/scala/com/spotify/scio/util/Functions.scala
```diff
@@ -263,7 +263,7 @@ private[scio] object Functions {
       c.output(g(c.element()))
   }
 
-  def partitionFn[T](numPartitions: Int, f: T => Int): PartitionFn[T] =
+  def partitionFn[T](f: T => Int): PartitionFn[T] =
     new NamedPartitionFn[T] {
       private[this] val g = ClosureCleaner.clean(f) // defeat closure
       override def partitionFor(elem: T, numPartitions: Int): Int = g(elem)
```
nevillelyh, line 269: Should we use do a `require(g(elem) < numPartitions` here or does Beam already perform this sanity check?

regadas: Hi @nevillelyh! this fn is meant to be used with `PartitionDoFn` which takes care of both scenarios. Don't believe it's worth replicate it.

PR #3171, opened 2020-07-22
Use raw coder in SCollection#applyKvTransform

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -239,7 +239,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     name: String,
     transform: PTransform[_ >: PCollection[T], PCollection[KV[K, V]]]
   )(implicit koder: Coder[K], voder: Coder[V]): SCollection[KV[K, V]] =
-    applyTransform(name, transform)(Coder.beam(CoderMaterializer.kvCoder[K, V](context)))
+    applyTransform(name, transform)(Coder.raw(CoderMaterializer.kvCoder[K, V](context)))
```
nevillelyh, line 242: It should be `raw` to be consistent with this change right?
https://github.com/spotify/scio/blob/e70dbf99c85b736fb6bbbc38aeb4a7d1a33d6ca4/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala#L55

PR #3172, opened 2020-07-22
Add SparkeyIO

kellen:
Adds `SparkeyIO`, makes sparkey endpoints testable in `JobTest`. See #3173

Open questions:
* Should `SparkeyIO` be used on the write path and if so how?
* Is there a better way to pass inputs from JobTests than `SparkeyIO.testUri(...)`?


scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
```diff
@@ -58,6 +60,44 @@ object TestCache {
...
+    val uris = items.map(SparkeyUri(_, () => ???))
+    testJobTestInput(uris)(path => SparkeyIO(path)) { case (sc, path) => sc.sparkeyUri(path) }
+  }
+
+  it should "work in a job test" in {
+    val lookup = items.map(i => i -> i)
+    val expected = items.map(i => s"$i -> $i")
+    JobTest[SparkeyJobTest.type]
+      .args("--input=input.txt", "--sparkey=mySparkey", "--output=out.txt")
+      .input(SparkeyIO("mySparkey"), List(SparkeyIO.testUri(lookup)))
```
nevillelyh, line 93: So this is really to allow mock data as Sparkey `SideInput` in `JobTest`. It's not quite an `IO` since `sc.read()` behaves differently than other `IO`s and `write` is unimplememnted.

How does one test Sparkey side input without this right now? And I wonder if there're somewhere in between that we can make it easier without "abusing" the `IO` mechanics?

kellen: Correct. You can't currently test them in a `JobTest` so far as I can tell.

To test now, I've rewritten my pipelines such that there is a method taking the `SideInput[SparkeyReader]` as an input, rewritten the `JobTest` to instead use `runWithContext` and then passed a side input using a mock `SparkeyUri` like you see in `SparkeyIO.testUri`.

nevillelyh: I like the idea, maybe we can rename a few things to make it less misleading. I'll comment in the code.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,56 @@
...
+import java.util
+
+import com.spotify.scio.ScioContext
+import com.spotify.scio.io.{EmptyTap, EmptyTapOf, ScioIO, Tap, TapOf}
+import com.spotify.scio.util.RemoteFileUtil
+import com.spotify.scio.values.SCollection
+import com.spotify.sparkey.{IndexHeader, LogHeader, SparkeyReader}
+import org.apache.beam.sdk.options.PipelineOptions
+
+final case class SparkeyUriTap(private val singleSparkeyUri: SparkeyUri) extends Tap[SparkeyUri] {
```
nevillelyh, line 12: We probably don't need the tap since the IO is read-only. No writes = no tap.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,56 @@
...
+          override def duplicate(): SparkeyReader = ???
+          override def close(): Unit = ???
+          override def iterator(): util.Iterator[SparkeyReader.Entry] = ???
+        }
+      }
+    }
+  }
+}
+
+final case class SparkeyIO(path: String) extends ScioIO[SparkeyUri] {
```
nevillelyh, line 44: Rename this to `SparkeySideInput`/`SparkeyUriSideInput`, or something that differentiates it from a regular IO?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,56 @@
...
+  override def open(sc: ScioContext): SCollection[SparkeyUri] = sc.parallelize(value.toIterable)
+}
+
+object SparkeyIO {
+  object ReadP {
+    def apply(options: PipelineOptions): ReadP = ReadP(() => RemoteFileUtil.create(options))
+  }
+  case class ReadP(rfuCreateFn: () => RemoteFileUtil)
+
+  def testUri(in: List[(String, String)]): SparkeyUri = {
```
nevillelyh, line 23: `in` can just be a `Map`? Also `SparkeySideInput.forTest(data)` if we rename the class & the method?
Also we should overload this to support `Array[Byte] => Array[Byte]` too, but you'll have to use something like `ByteString` internally to make map lookups work.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -84,11 +83,11 @@ trait ShardedSparkeyUri extends SparkeyUri {
 }
 
 private[sparkey] object ShardedSparkeyUri {
-  def apply(basePath: String, options: PipelineOptions): ShardedSparkeyUri =
+  def apply(basePath: String, rfuCreateFn: () => RemoteFileUtil): ShardedSparkeyUri =
```
nevillelyh, line 86: Why do we need a `fn` here? Can't seem to find a place where it's not `RemoteFileUtil.create()`.

kellen: The original code deferred creation of `RemoteFileUtil` until it knew it was dealing with a remote uri, so this is replicating that behavior. We could also eagerly create `RemoteFileUtil` but it would be unused when reading locally.

nevillelyh:
@psobot thoughts? I like the idea. Naming of `SparkeyIO` etc. can be tweaked a bit but overall this could be useful?

scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
kellen: Integrated your comments. 

Can you explain why is a "normal" SparkeyIO doesn't make sense? I don't think I have enough context

nevillelyh: A normal `ScioIO` is intended to support read/write in a pipeline, like text or Avro, e.g. `sc.sparkeyFile("gs://...").map(...).saveAsSparkeyFile()`. And `T` should be something like `(String, String)`.

In this case we're not supporting reading Sparkey files as `SCollection[(String, String)]` (even though it's technically doable) but "abusing" the `ScioIO` wiring for testing. That's why I'm saying it's not a typical IO.

Also we can save the `SparkeyIO` name in case we want to support reading/writing sharded Sparkey files as `SCollection[(String, String)]`.

kellen: Couldn't you write out the `SparkeyUri`s if you wanted to? The contents of the sparkey is just a side-effect of Uris passing through the pipeline, sorta.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUriSideInput.scala
```diff
@@ -0,0 +1,66 @@
...
+    override def getAsByteArray(key: Array[Byte]): Array[Byte] = ???
+    override def getAsEntry(key: Array[Byte]): SparkeyReader.Entry = ???
+    override def getIndexHeader: IndexHeader = ???
+    override def getLogHeader: LogHeader = ???
+    override def duplicate(): SparkeyReader = ???
+    override def close(): Unit = ???
+    override def iterator(): util.Iterator[SparkeyReader.Entry] = ???
+  }
+
+  def forTest[T](items: Map[String, String]): SparkeyUri = new SparkeyUri {
```
nevillelyh, line 31: `T` is unused?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUriSideInput.scala
```diff
@@ -0,0 +1,66 @@
...
+    override val basePath: String = "i should be ignored"
+    override private[sparkey] def exists = true
+    override def getReader: SparkeyReader = new TestSparkeyReader {
+      override def getAsString(key: String): String = items.get(key).orNull
+    }
+  }
+
+  def forTest(items: Map[Array[Byte], Array[Byte]])(implicit d: DummyImplicit): SparkeyUri = new SparkeyUri {
```
nevillelyh, line 39: `Array[Byte]` shouldn't be used as map key. I doubt there'll be any collision but still bad practice.

Is the `DummyImplicit` to avoid method overloading issues? Maybe we have to use `Seq[(Array[Byte], Array[Byte])]` after all and that'll solve both issues but a bit inconsistent.

Alternatively maybe rename them to `fromStrings/fromBytes`?

kellen: Ack, good point. Using `Seq` resolves both issues.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUriSideInput.scala
```diff
@@ -0,0 +1,66 @@
...
+    override def getAsEntry(key: Array[Byte]): SparkeyReader.Entry = ???
+    override def getIndexHeader: IndexHeader = ???
+    override def getLogHeader: LogHeader = ???
+    override def duplicate(): SparkeyReader = ???
+    override def close(): Unit = ???
+    override def iterator(): util.Iterator[SparkeyReader.Entry] = ???
+  }
+
+  def forTest(items: Map[String, String]): SparkeyUri = new SparkeyUri {
+    override val basePath: String = "i should be ignored"
```
regadas, line 32: I think it's preferable to make a change in the `trait` to `def`

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUriSideInput.scala
```diff
@@ -0,0 +1,66 @@
...
+  override def testId: String = s"SparkeyUriSideInput($path)"
+
+  override type ReadP = SparkeyUriSideInput.ReadP
+  override type WriteP = Nothing // ReadOnly
+
+  final override val tapT = EmptyTapOf[SparkeyUri]
+  override def tap(read: ReadP): Tap[tapT.T] = EmptyTap
+
+  override protected def read(sc: ScioContext, params: ReadP): SCollection[SparkeyUri] = {
+    sc.parallelize(Seq(SparkeyUri(path, params.rfuCreateFn)))
```
regadas, line 61: I think this is missing the `ShardedSparkeyUri`?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUriSideInput.scala
```diff
@@ -0,0 +1,66 @@
...
+    override private[sparkey] def exists = true
+    override def getReader: SparkeyReader = new TestSparkeyReader {
+      override def getAsByteArray(key: Array[Byte]): Array[Byte] = {
+        byteStrItems.get(ByteString.copyFrom(key)).orNull
+      }
+    }
+  }
+}
+
+final case class SparkeyUriSideInput(path: String) extends ScioIO[SparkeyUri] {
```
regadas, line 51: I find this naming confusing 😬 also having users supplying test data through `forTest` does not conform with other IO's. This leads me to believe that this might not be 1) exactly what we want 2) some more deeper changes are needed to support these use cases. In any case imo this requires a bit more thought.

nevillelyh:
So what should we do next? It is a bit hacky but also solves a practical need. Also it's in `scio-extra` so I think we can live with it as long as we avoid confusing naming. @regadas?

regadas:
I would like to give this some thought. I agree, it's practical but in `scio-extra` this is probably the most used io/sideInput and for that reason I'm a little bit hesitant in adding things that we might discard in a near future.

kellen:
@regadas any ideas on how we might restructure this to fit better w/ the rest of the architecture?

kellen:
Closing in favor of a different approach

PR #3179, opened 2020-07-23
Add scalafix ExplicitReturnType and ProcedureSyntax rules

regadas:
🍿

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,4 +1,10 @@
 rules = [
   RemoveUnused,
   LeakingImplicitClassVal
+  ProcedureSyntax
+  ExplicitResultTypes
 ]
+
+ExplicitResultTypes.memberKind = [Def, Val, Var]
+ExplicitResultTypes.memberVisibility = [Public]
+ExplicitResultTypes.skipSimpleDefinition = false
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/types/MacroUtil.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/types/MacroUtil.scala
@@ -49,7 +49,7 @@ private[types] object MacroUtil {
 
   val ScioAvro = "_root_.com.spotify.scio.avro"
   val ApacheAvro = "_root_.org.apache.avro"
-  val ScioAvroType = s"$ScioAvro.types.AvroType"
+  val ScioAvroType: String = s"$ScioAvro.types.AvroType"
 
   def p(c: blackbox.Context, code: String): c.Tree = c.parse(code)
 }
--- a/build.sbt
+++ b/build.sbt
@@ -151,6 +151,8 @@ lazy val formatSettings = Seq(
   javafmtOnCompile := false
 )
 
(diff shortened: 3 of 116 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/types/MacroUtil.scala
```diff
@@ -47,9 +47,9 @@ private[types] object MacroUtil {
 
   // Namespace helpers
 
-  val ScioAvro = "_root_.com.spotify.scio.avro"
-  val ApacheAvro = "_root_.org.apache.avro"
-  val ScioAvroType = s"$ScioAvro.types.AvroType"
+  val ScioAvro: String = "_root_.com.spotify.scio.avro"
```
jto, line 50: Requiring explicit type ascription for basic `val` that are easily inferred seems excessive. IIRC this was not required when we had scalastyle doing the validation.

.scalafix.conf
```diff
@@ -1,4 +1,10 @@
 rules = [
   RemoveUnused,
   LeakingImplicitClassVal
+  ProcedureSyntax
+  ExplicitResultTypes
 ]
+
+ExplicitResultTypes.memberKind = [Def, Val, Var]
```
jto, line 8:
```suggestion
ExplicitResultTypes.memberKind = [Def, Var]
```

😁

regadas: ahah 😄 I don't want to turn it off for `Val` there are cases where sometimes it might not be immediately obvious i.e `val schema: Schema = Schema.createRecord("GenericTestRecord", null, null, false)` and all the `TapT[T]` (which we should probably revise at some point).

I'm still looking into this because I get the feeling that `ExplicitResultTypes.skipSimpleDefinition` is not working as i expected ... I was unable to remove those `val s: String = "foo"` but i have the feeling that it should be possible

regadas: turns out that if we simply set `ExplicitResultTypes.skipSimpleDefinition = false` it behaves a lot more as we wanted it to.

scio-avro/src/main/scala/com/spotify/scio/avro/types/MacroUtil.scala
```diff
@@ -49,7 +49,7 @@ private[types] object MacroUtil {
 
   val ScioAvro = "_root_.com.spotify.scio.avro"
   val ApacheAvro = "_root_.org.apache.avro"
-  val ScioAvroType = s"$ScioAvro.types.AvroType"
+  val ScioAvroType: String = s"$ScioAvro.types.AvroType"
```
jto, line 52: Does `ScioAvroType` require a type annotation but not `ApacheAvro`?

regadas: couldn't find yet a solution for cases where string interpolation is used.

regadas: I can leave with that for now though.

PR #3187, opened 2020-07-26
Allow union types in non leaf field for key

### LINKED ISSUE #3184: Validate key path in AvroBucketMetadata disallows keying by field in record
It looks like the `validateKeyField` method in AvroBucketMetadata disallows specifying a key which is in an avro record [here](https://github.com/spotify/scio/blob/master/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java#L152).

For example, if I use as a key `message.foo` where the message is an avro `Record` type and foo is a `String` type, the line mentioned before will throw an exception.

martinbomio:
Fixes #3184.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
@@ -149,10 +149,9 @@ public class AvroBucketMetadata<K, V extends GenericRecord> extends BucketMetada
           field, String.format("Key path %s does not exist in schema %s", keyPath[i], currSchema));
 
       currSchema = field.schema();
-      Schema.Type schemaTypeForUnionType = getSchemaTypeForUnionType(currSchema);
       Preconditions.checkArgument(
           currSchema.getType() == Schema.Type.RECORD
-              || schemaTypeForUnionType == Schema.Type.RECORD,
+              || getSchemaOrInnerUnionType(currSchema) == Schema.Type.RECORD,
           "Non-leaf key field " + keyPath[i] + " is not a Record type");
     }
 
@@ -175,12 +174,12 @@ public class AvroBucketMetadata<K, V extends GenericRecord> extends BucketMetada
   }
 
   private static Class<?> getKeyClassFromSchema(Schema schema) {
-    Schema.Type schemaType = getSchemaTypeForUnionType(schema);
+    Schema.Type schemaType = getSchemaOrInnerUnionType(schema);
 
     return getClassForType(schemaType);
   }
 
-  private static Schema.Type getSchemaTypeForUnionType(Schema schema) {
+  private static Schema.Type getSchemaOrInnerUnionType(Schema schema) {
     Schema.Type schemaType = schema.getType();
 
     if (schemaType == Schema.Type.UNION) {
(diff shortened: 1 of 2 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -173,6 +175,12 @@ private static String validateKeyField(String keyField, Class keyClass, Schema s
   }
 
   private static Class<?> getKeyClassFromSchema(Schema schema) {
+    Schema.Type schemaType = getSchemaTypeForUnionType(schema);
+
+    return getClassForType(schemaType);
+  }
+
+  private static Schema.Type getSchemaTypeForUnionType(Schema schema) {
```
ClaireMcGinty, line 183:
```suggestion
  private static Schema.Type getSchemaTypeForUnionType(Schema schema) {
  private static Schema.Type getSchemaOrInnerUnionType(Schema schema) {
```
I think this method name is a bit clearer, wdyt?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -149,8 +149,10 @@ private static String validateKeyField(String keyField, Class keyClass, Schema s
           field, String.format("Key path %s does not exist in schema %s", keyPath[i], currSchema));
 
       currSchema = field.schema();
+      Schema.Type schemaTypeForUnionType = getSchemaTypeForUnionType(currSchema);
```
ClaireMcGinty, line 152:
```suggestion
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -149,8 +149,10 @@ private static String validateKeyField(String keyField, Class keyClass, Schema s
...
       Preconditions.checkArgument(
-          currSchema.getType() == Schema.Type.RECORD,
+          currSchema.getType() == Schema.Type.RECORD
```
ClaireMcGinty, line 154:
```suggestion
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -149,8 +149,10 @@ private static String validateKeyField(String keyField, Class keyClass, Schema s
...
+              || schemaTypeForUnionType == Schema.Type.RECORD,
```
ClaireMcGinty, line 155:
```suggestion
          getSchemaOrInnerUnionType(currSchema) == Schema.Type.RECORD,
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
ClaireMcGinty: we can omit this check because the getSchema method below will return the result of Schema#getType for non-union records :)

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -173,6 +175,12 @@ private static String validateKeyField(String keyField, Class keyClass, Schema s
   }
 
   private static Class<?> getKeyClassFromSchema(Schema schema) {
+    Schema.Type schemaType = getSchemaTypeForUnionType(schema);
```
ClaireMcGinty, line 178:
```suggestion
    final Schema.Type schemaType = getSchemaOrInnerUnionType(schema);
```

PR #3193, opened 2020-07-27
Fix issue with union type as non-leaf field of smb key

martinbomio:
Found another issue while running a job that uses the following schema:
```
{
  "type": "record",
  "name": "Bar",
  "namespace": "some.namespace",
  "fields": [
    {
    "name": "message",
    "type": [
      "null",
      {
       "name": "foo"
       "type": ["null", "string"],
       ....
      } 
    ]
    },
    ......
  ]
}
```

and we want to key by `messafe.foo`. 

This line https://github.com/spotify/scio/blob/master/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java#L158 fails when currSchema is of union type.

This time I tested it running a job locally and it works.

### DIFF
--- a/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadataTest.java
+++ b/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadataTest.java
@@ -51,12 +51,15 @@ public class AvroBucketMetadataTest {
           false,
           Lists.newArrayList(
               new Schema.Field("countryId", Schema.create(Type.BYTES), "", ""),
-              new Schema.Field("postalCode", Schema.createUnion(Schema.create(Type.NULL), Schema.create(Type.BYTES)), "", ""),
+              new Schema.Field("postalCode", Schema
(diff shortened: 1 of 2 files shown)


scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadataTest.java
```diff
@@ -51,13 +51,16 @@
...
-              new Schema.Field("postalCode", Schema.createUnion(Schema.create(Type.NULL), Schema.create(Type.BYTES)), "", ""),
+              new Schema.Field("postalCode", Schema
+                  .createUnion(Schema.create(Type.NULL), Schema.create(Type.BYTES)), "", ""),
               new Schema.Field(
                   "prevCountries",
                   Schema.createArray(Schema.create(Schema.Type.STRING)),
                   "",
                   Collections.<String>emptyList())));
 
+  private static final Schema LOCATION_UNION_SCHEMA = Schema.createUnion(Schema.create(Type.NULL), LOCATION_SCHEMA);
```
martinbomio, line 62: we want to test the two conditions, when the non-leaf field is a union type and when it is not

PR #3209, opened 2020-08-03
Fix JdbcShardedReadOptions.fetchSize description

viktorjonsson:
Description of JdbcShardedReadOptions.fetchSize is incorrect.


scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/JdbcShardedReadOptions.scala
```diff
@@ -28,7 +28,7 @@ import com.spotify.scio.jdbc.JdbcConnectionOptions
  * @param shardColumn Column to shard by. Must be of integer/long type ideally with evenly
  *                    distributed values
  * @param rowMapper Function to map from a SQL [[java.sql.ResultSet]] to `T`
- * @param fetchSize Name of a table or materialized view to read from. Default value is 100000.
+ * @param fetchSize Amount of rows fetched per java.sql.ResultSet. Default value is 100000.
```
regadas, line 31:
```suggestion
 * @param fetchSize Amount of rows fetched per [[java.sql.ResultSet]]. Default value is 100000.
```

PR #3215, opened 2020-08-03
allow user-supplied filename prefix for smb writes/reads

ClaireMcGinty:
I.e. `"part"` instead of `"bucket"` so that the output file naming is in parity with non-smb writes.

This is a relatively small change (it's just an optional field added to `BucketMetadata`) but unfortunately requires a **ton** of small code changes to Java boilerplate 😭 

I tested that it's backwards compat--if the `metadata.json` file lacks a `filenamePrefix` field, it will default to `"bucket"`.


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -70,7 +71,8 @@ public AvroBucketMetadata(
         validateKeyField(
             keyField,
             keyClass,
-            new ReflectData(recordClass.getClassLoader()).getSchema(recordClass)));
+            new ReflectData(recordClass.getClassLoader()).getSchema(recordClass)),
+        filenamePrefix);
```
nevillelyh, line 75: indentation?

```suggestion
            filenamePrefix);
```

PR #3240, opened 2020-08-07
Avoid scala.jdk.CollectionConverters implicit import in Bigquery macro

### LINKED ISSUE #3222: @BigQueryType.fromSchema macro expansion fails in scala 2.13 if scala.collection.JavaConverters._ is imported in scope
```
[error] type mismatch;
[error]  found   : java.util.List[AnyRef]
[error]  required: ?{def asScala: ?}
[error] Note that implicit conversions are not applicable because they are ambiguous:
[error]  both method asScalaBufferConverter in object JavaConverters of type [A](l: java.util.List[A]): collection.JavaConverters.AsScala[scala.collection.mutable.Buffer[A]]
[error]  and method ListHasAsScala in trait AsScalaExtensions of type [A](l: java.util.List[A]): jdk.CollectionConverters.ListHasAsScala[A]
[error]  are possible conversion functions from java.util.List[AnyRef] to ?{def asScala: ?}
[error]   @BigQueryType.fromSchema(
```

regadas:
Fixes #3222

### DIFF
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
@@ -121,7 +121,7 @@ private[types] object ConverterProvider {
 
     def list(tree: Tree, tpe: Type): Tree = {
       val jl = tq"_root_.java.util.List[AnyRef]"
-      q"$tree.asInstanceOf[$jl].asScala.iterator.map(x => ${cast(q"x", tpe)}).toList"
+      q"asScala($tree.asInstanceOf[$jl]).iterator.map(x => ${cast(q"x", tpe)}).toList"
     }
 
     def field(symbol: Symbol, fn: TermName): Tree = {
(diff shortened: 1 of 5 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -153,7 +153,7 @@ private[types] object ConverterProvider {
 
     val tn = TermName("r")
     q"""(r: _root_.org.apache.avro.generic.GenericRecord) => {
-          import _root_.scala.jdk.CollectionConverters._
```
nevillelyh, line 156: Wait we were using 2.13 converters this whole time? That should've broken 2.12 tests?

nevillelyh: Never mind. Looks like these are from `scala-collection-compat`. We should fix it for `scio-avro` macros too right?

regadas: the tests were also using `_root_.scala.jdk.CollectionConverters._` so no problem there ... and by luck most of ppl where not importing `_root_.scala.jdk.CollectionConverters._`

PR #3242, opened 2020-08-07
Remove coders deprecation warns

regadas:


scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -504,8 +505,8 @@ object Coder
   implicit def iterableCoder[T: Coder]: Coder[Iterable[T]] = ScalaCoders.iterableCoder
   implicit def throwableCoder[T <: Throwable: ClassTag]: Coder[T] = ScalaCoders.throwableCoder
   implicit def listCoder[T: Coder]: Coder[List[T]] = ScalaCoders.listCoder
-  implicit def traversableOnceCoder[T: Coder]: Coder[TraversableOnce[T]] =
```
nevillelyh, line 507: Will this break people who actually use `TraversableOnce` as surface type? Not that they should.
But on second thought, it'll break mutability check so probably a non-issue?

regadas: on 2.12/2.13 IterableOnce/TraversableOnce are type aliases so this shouldn't:tm: be an issue.

PR #3246, opened 2020-08-10
Use ## and support consistent Array hashCode

### LINKED ISSUE #2450: Non primitive keys result in incorrect sparse transforms for more than 125M sparse keys

regadas:
Closes #2450

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -30,6 +30,7 @@ import com.twitter.algebird.{Aggregator, Monoid, MonoidAggregator, Semigroup}
 import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.values.{KV, PCollection, PCollectionView}
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -173,20 +174,21 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
    */
-  def hashPartitionByKey(numPartitions: Int): Seq[SCollection[(K, V)]] =
-    self.partition(
-      numPartitions,
-      {
-        case (key, _) =>
-          Math.floorMod(key.hashCode(), numPartitions)
-      }
-    )
+
```
nevillelyh, line 184:
```suggestion
```

PR #3247, opened 2020-08-10
Change default SMB codec to Deflate to match Scio

ClaireMcGinty:
The default value now matches Scio AvroIO's default value, so users can transition more easily

### DIFF
--- a/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroFileOperationsTest.java
+++ b/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroFileOperationsTest.java
@@ -141,7 +141,7 @@ public class AvroFileOperationsTest {
     MatcherAssert.assertThat(
         displayData, hasDisplayItem("compression", Compression.UNCOMPRESSED.toString()));
     MatcherAssert.assertThat(
-        displayData, hasDisplayItem("codecFactory", CodecFactory.snappyCodec().getClass()));
+        displayData, hasDisplayItem("codecFactory", CodecFactory.deflateCodec(6).getClass()));
     MatcherAssert.assertThat(
         displayData, hasDisplayItem("schema", AvroGeneratedUser.SCHEMA$.getFullName()));
   }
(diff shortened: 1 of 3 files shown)


scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroFileOperationsTest.java
```diff
@@ -141,7 +141,7 @@ public void testDisplayData() {
     MatcherAssert.assertThat(
         displayData, hasDisplayItem("compression", Compression.UNCOMPRESSED.toString()));
```
regadas, line 142: This change LGTM, However I just noticed that this display data will look a bit sketchy/confusing perhaps.

nevillelyh: Huh what do you mean?

regadas: it will show `compression: uncompressed` when in reality there's probably compression applied through the `CodecFactory`. I'm guessing this is due to the fact that we inherit from `FileOperations`.

nevillelyh: Ah yeah it's uncompressed from the File IO stream point of view, not Avro internal.

nevillelyh: We can override parent `FileOperations` by adding a line here? @ClaireMcGinty WDYT?
https://github.com/spotify/scio/blob/246382227f3592465738163709877a9da2f366e8/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java#L97

ClaireMcGinty: ahh good catch @regadas ! @nevillelyh can you expand on that change? Since we do already have the codec in the `AvroFileOperations` `DisplayData`. I feel like we could either:

- remove the `super()` call here so it's not inheriting from FileOperations: https://github.com/spotify/scio/blob/246382227f3592465738163709877a9da2f366e8/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java#L95

- Modify `FileOperations` to only include `compression` in the `DisplayData` if `compression != Compression.UNCOMPRESSED`

- Change the display names from "compression" -> "FileCompression" and "codecFactory" -> "AvroCompression"

PR #3253, opened 2020-08-12
improve SMB error handling

ClaireMcGinty:
Adds some more specific messaging for common errors.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadataUtil.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadataUtil.java
@@ -97,12 +97,13 @@ public class BucketMetadataUtil {
               .parallelStream()
               .map(
                   dir ->
-                      BucketMetadataUtil.getMetadata(dir)
-                          .map(metadata -> (BucketMetadata<K, V>) metadata)
-                          .orElseThrow(
-                              () ->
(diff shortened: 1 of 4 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadataUtil.java
```diff
@@ -95,19 +92,21 @@ public static PartitionMetadata create(
...
               .parallelStream()
-              .map(BucketMetadataUtil::<K, V>getMetadata)
+              .map(
+                  dir ->
+                      BucketMetadataUtil.getMetadata(dir)
+                          .map(metadata -> (BucketMetadata<K, V>) metadata)
+                          .orElseThrow(
+                              () ->
+                                  new RuntimeException(
+                                      "Could not find SMB metadata for source directory " + dir)))
```
regadas, line 105: avoid extra allocation
```suggestion
                      (BucketMetadata<K, V>)
                          BucketMetadataUtil.getMetadata(dir)
                              .orElseThrow(
                                  () ->
                                      new RuntimeException(
                                          "Could not find SMB metadata for source directory "
                                              + dir)))
```

ClaireMcGinty: good idea! thanks 👍

Issue #3260, opened 2020-08-17
External sorter improvements


nevillelyh:
Right now the sorter has a superficial 2048MB limit on sorter memory due to Hadoop. This is no longer true for `InMemorySorter` or `NativeFileSorter`, since both uses `ArrayList<KV<byte[], byte[]>>` internally and is only limited by JVM heap, not a single `byte[]`'s 2GB limit.

https://github.com/apache/beam/blob/master/sdks/java/extensions/sorter/src/main/java/org/apache/beam/sdk/extensions/sorter/BufferedExternalSorter.java#L73

We could also experiment with fast compression (maybe snappy or low setting deflate) of temp files in `NativeFilteSorter`, and see if the reduced disk IO is worth the additional CPU.

nevillelyh:
Tried deflate (level 1, 5, 9) & snappy compression in `NativeFileSorter` with bare metal CPU & SSD, got spacing saving varying from 40-60% with a real data set, but 20-130% slower. Probably not worth the effort unless worker disk is really really slow.

nevillelyh:
Another possible improvement: delete external temp files once the merge sort iterator is exhausted.

PR #3269, opened 2020-08-19
workaround for sorter memory limit #3260

nevillelyh:


nevillelyh:
WIP, need to test if both in-memory & native sorter can handle the large memory usage.

PR #3274, opened 2020-08-25
[SMB] delete file early in NativeFileSorter

nevillelyh:
Upstream: https://github.com/apache/beam/pull/12681
Delete `dataFile` as soon as we're done sort in batch from it.


regadas (approved):
LGTM

PR #3285, opened 2020-09-01
sparkey improvements

nevillelyh:


nevillelyh:
Pushed to master since I want to keep the separate commits.

PR #3291, opened 2020-09-02
Add SCollection#filterNot

### LINKED ISSUE #3286: Add SCollection.filterNot function
It would be nice to have the `SCollection.filter` function counterpart `Scollection.filterNot` which returns elements which do not satisfy a predicate. Sometimes this helps with a transformation code readability. `filterNot` is available for regular Scala collections.

regadas:
Fix #3286


stormy-ua:
Awesome! Thanks!

PR #3295, opened 2020-09-02
support sparkey compression, fix #3210

### LINKED ISSUE #3210: Support Sparkey compressions
As of Sparkey 3.1.0 it supports snappy and zstd compressions. We should look into supporting these.

nevillelyh:


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -215,13 +217,21 @@ package object sparkey extends SparkeyReaderInstances {
     def asSparkey(
       path: String = null,
       maxMemoryUsage: Long = -1,
-      numShards: Short = DefaultNumShards
+      numShards: Short = DefaultNumShards,
+      compressionType: CompressionType = CompressionType.NONE,
```
regadas, line 221: 🎨 Since these defaults are used in multiple methods might be worth moving them to a constant.

regadas (commented):
Minor comment, otherwise LGTM.

regadas (approved):

PR #3303, opened 2020-09-04
Use beam-runners-flink-1.10 when using BEAM_USERS

regadas:


nevillelyh:
2.24.0 will be released soon with 1.11 runner, should we wait?

regadas:
this is making it consistent with what we have https://github.com/spotify/scio/blob/3503b0c3d2db718c474f9457ad40ec5003ffe7b9/build.sbt#L487-L492

I thought it would be only be on `2.25.0` if it's early awesome

regadas:
actually let me change this since it might happen in the future

regadas (commented):
@nevillelyh gonna merge this, just because we already have `1.10` on scio-core

PR #3307, opened 2020-09-09
Add support for JDBC sharding by UUID encoded as string

stormy-ua:
There was a request from a team for JDBC sharding by a column with UUIDs encoded as strings e.g. "a2c9cba1-eaa5-4c3d-b099-896730eb6302". In their case the column is the primary key and sharding gives a significant improvement in the performance. The job running time decreased from 1 hr down to 20 mins with fewer workers. 

Even though we were able to produce a temporary solution w/o changing `scio-jdbc` it would be great to add a cleaner solution to scio directly so that it could be used by others for similar use cases.

The change adds two new `ShardString` subclasses, namely `UuidLowerString` and `UuidUpperString`. The former represents a UUID encoded as a string in the lower case e.g. "a2c9cba1-eaa5-4c3d-b099-896730eb6310" and the latter a UUID encoded as a string in the upper case "A2C9CBA1-EAA5-4C3D-B099-896730EB6310". The implementation projects UUIDs onto a space of `BigInt`s by removing dashes `-` first and then converting from the hex representation to `BigInt` exactly the same were it is done for `HexUpperString` and `HexLowerString`.

 If UUIDs have more or less uniform distribution the sharding approach might produce relatively even shards by using the approach described above.

### DIFF
--- a/scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/ShardedQueryTests.scala
+++ b/scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/ShardedQueryTests.scala
@@ -52,7 +52,7 @@ class ShardedQueryTests extends AnyFlatSpec {
   "toSelectStatement" must "produce the correct statement for hex string in upper case" in {
 
     val shardQuery = RangeShardQuery[HexUpperString](
-      Range(HexUpperString("a"), HexUpperString("f")),
+      Range(HexUpperString("A"), HexUpperString("F")),
       upperBoundInclusive = true,
       quoteValues = true
     )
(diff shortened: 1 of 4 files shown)


scio-jdbc/src/test/scala/com/spotify/scio/jdbc/sharded/ShardedQueryTests.scala
```diff
@@ -52,7 +52,7 @@ class ShardedQueryTests extends AnyFlatSpec {
   "toSelectStatement" must "produce the correct statement for hex string in upper case" in {
 
     val shardQuery = RangeShardQuery[HexUpperString](
-      Range(HexUpperString("a"), HexUpperString("f")),
+      Range(HexUpperString("A"), HexUpperString("F")),
```
nevillelyh, line 55: Looks like `HexUpperString` etc. does to upper/lower and can handle such cases. Should they also validate that prefix is valid hex & fail fast? We can do that as a separate PR.

stormy-ua: That was one my changes. I removed that logic from `HexUpperString` etc. completely and moved it to `RangeShardStringCodec`s to keep it one place. It was confusing to have in both `HesUpper/LowerStrings` and `ShardCodec`s. I think adding the validation logic is a good idea even though it isn't going to give 100% guarantee that there are no bad strings in the table. We would apply validation only to min and max strings and there might be some badly formatted strings in between.

PR #3322, opened 2020-09-16
Filter out potentially included env directories

regadas:
* Less jars being staged
* example of excluded directories: .sbt / .sdkman
* When using SBT to launch jobs SBT jars are included in the staged files. This will avoid their inclusion and aliviate the need to use `run / fork := true`. Staged files will be the same with or without this SBT setting.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -136,14 +136,11 @@ private object RunnerContext {
     pipelineOptions: PipelineOptions,
     classLoader: ClassLoader
   ): Iterable[String] = {
-    // exclude jars from JAVA_HOME and files from current directory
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -136,14 +136,11 @@ private object RunnerContext {
     pipelineOptions: PipelineOptions,
     classLoader: ClassLoader
   ): Iterable[String] = {
-    // exclude jars from JAVA_HOME and files from current directory
-    val javaHome = new File(CoreSysProps.Home.value).getCanonicalPath
-    val userDir = new File(CoreSysProps.UserDir.value).getCanonicalPath
```
regadas, line 141: This is outdated and pre `GlassGraph` usage. By default system jars are not included

ClaireMcGinty (approved):

Issue #3279, opened 2020-08-31
Mutable `ScalableBloomFilter`


nevillelyh:
This came up last week during an internal training. It's one of the oldest Scala code at Spotify and used heavily by some large-ish pipelines (via copy/paste). The original PR here:
https://github.com/twitter/algebird/pull/494

Since it's not a Monoid we're better off using Guava BF here. Also seems the current impl ser/de the entire `List[BF]` to append to the first (largest) one. So in practice we can just pretend a new BF if `bfs.head` is full or only ser/de the first BF to add items. As a bonus we'll also have better/more predictable coder behavior.

nevillelyh:
I'm thinking something down the line of this, a good Java impl with light Scala wrapper & coder
https://github.com/spotify/scio/pull/2651

kellen:
CLosed by #3339

nevillelyh:
Did a more thorough benchmark, old Algebird based SBF vs new MSBF.
1. Create new filter with fpp = 0.01 and initial capacity of 128, insert 100 elements
2. Encode, decode, add 2x more elements (200, 400, ...)
3. Repeat 2 until 102400 elements.

New implementation seems to perform ~40% better pretty consistently.

https://docs.google.com/spreadsheets/d/1DgtFSGm9vkDiIbHZjXR_01fr6zlJycV3U5IUayI73xo/edit#gid=0

PR #3339, opened 2020-09-23
Mutable ScalableBloomFilter

kellen:
First pass at a mutable scalable bloom filter per #3279

Wasn't quite sure how to address this part of the issue:
> Also seems the current impl ser/de the entire List[BF] to append to the first (largest) one. So in practice we can just pretend a new BF if bfs.head is full or only ser/de the first BF to add items. As a bonus we'll also have better/more predictable coder behavior.

From what I saw the serialization seemed to just be via kryo? 

For tests, I retained the relevant ones from the referenced algebird PR.


scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
```diff
@@ -0,0 +1,21 @@
+package com.spotify.scio.coders.instances
+
+import java.io.{InputStream, OutputStream}
+
+import com.google.common.hash.BloomFilter
+import com.spotify.scio.coders.Coder
+import com.google.common.{hash => g}
+import org.apache.beam.sdk.coders.AtomicCoder
+
+class GuavaBloomFilterCoder[T](implicit val funnel: g.Funnel[T])
```
nevillelyh, line 10: https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala#L337

We have this already but private, maybe move it here?

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
```diff
@@ -0,0 +1,85 @@
...
+  growthRate: Int,
+  tighteningRatio: Double,
+  var filters: List[g.BloomFilter[T]],
+  // storing a count of items in the head avoids calling the relatively expensive `approximateElementCount` after each insert
+  var headCount: Long
+)(implicit funnel: g.Funnel[T])
+    extends Serializable {
+  def contains(item: T): Boolean = filters.exists(f => f.mightContain(item))
+  def approximateElementCount: Long = filters.foldLeft(0L) { case (sum, filter) =>
+    sum + filter.approximateElementCount()
```
nevillelyh, line 58: Minor but is `filters.iterator.map(_.approximateElementCount).sum` more readable?

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
```diff
@@ -0,0 +1,85 @@
...
+      if (filters != Nil) {
+        fpProb = fpProb * tighteningRatio
+        headCapacity = growthRate * headCapacity
+      }
+      headCount = 0
+      filters = g.BloomFilter.create[T](funnel, headCapacity, fpProb) :: filters
+    }
+  }
+
+  def +(item: T): MutableScalableBloomFilter[T] = {
```
nevillelyh, line 74: Should probably be `+=` to make it clear that it's mutating `this`.

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
```diff
@@ -0,0 +1,85 @@
...
+    scale()
+    val changed = filters.head.put(item)
+    if (changed) headCount = headCount + 1
+    this
+  }
+
+  def ++(items: Iterable[T]): MutableScalableBloomFilter[T] = {
```
nevillelyh, line 81: Likewise, `++=`

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
```diff
@@ -0,0 +1,85 @@
...
+ * @param headCapacity      The capacity of the filter at the head of `filters`
+ * @param growthRate        The growth rate of each subsequent filter added to `filters`
+ * @param tighteningRatio   The tightening ratio applied to the current `fpProb` to maintain the false positive probability over the sequence of filters
+ * @param filters           The underlying 'plain' bloom filters
+ * @param headCount         The number of items currently in the filter at the head of `filters`
+ * @param funnel            The funnel to turn `T`s into bytes
+ * @tparam T                The type of objects inserted into the filter
+ */
+case class MutableScalableBloomFilter[T](
+  var fpProb: Double,
```
nevillelyh, line 47: What about making the `var`s private? We only need to expose `fpProb`, everything else is impl detail right?

nevillelyh (commented):
Re: yeah ser/de should be derivable `Coder[MSBF]` (not kryo) since `MSBF` is a case class. Need to double check if you make some fields `private` though.

Re: mutating just the head BF, I was thinking, say we encode an MSBF like this:

`| fpProb bytes | headCapacity bytes | headCount bytes | growthRate bytes | ... | head BF bytes | BF1 bytes | ... | BFn bytes |`

And persistent it as a giant byte array on disk. When adding items, instead of decoding all BFs, we do this:
- Decode `fpProb`, `headCapacity`, `headCount`, `growthRate`, `tighteningRatio`
- If `headCount` >= `headCapacity`, create new head BF(s) & add elements
- Else decode head BF & add elements, or even (need to reverse engineer) mutate original bytes in place
- Encode params, new head BF(s) if any, then original tail BFs bytes without decoding them.

The API could be in the form of `def add(sbfBytes: Array[Byte], elems: Iterable[T]): Array[Byte]`, that allows us to add elements to a giant MSBF on disk without decoding every member BF, all but the head won't change anyway. In the very likely case that head BF is not full, we might even mutate the bits in place without allocating new byte arrays.

For the typical use case of one MSBF per user, several MBs x 1 billion, the saving from wasteful decode/encode & memory allocation could be huge. The down side is it's pretty hacky, and require duplicating code in the guava BF :shrug:

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
kellen: I guess they could all be private? Do you ever actually want to know what fpProb is after creation?

scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
kellen: Realized I could `xmap` to the ApproxFilter wrapper, so updated.

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
kellen: Made `fpProb` a val and added a private `headFPProb` to track the current fpp.

I made the fields `private[hash]` so the tests would still have access, but looks kinda ugly. Is there some more elegant way?

nevillelyh: Yeah `fpProb`(the original desired fpp) should be exposed since it's a "configuration", in case one is reading SBF produced by someone else.

Where are you accessing them in tests? I don't see any.

kellen: Mostly inspecting the size of filters, e.g. [here](https://github.com/spotify/scio/pull/3339/files#diff-a62c7d77f3404a228897cae20f78774aR19)

nevillelyh: Can just add a `def numBFs: Int = filter.size`? Most other `private[hash]` can just be `private`?

nevillelyh:
Another nice to have: make it extend `ApproxFilter[T]` so all the syntactic sugar works out of box (plus consistency)? Should be fairly trivial. The main difference vs existing BF is we don't need `expectedInsertions` (maybe warn if it's set to >0) when constructing SBF.

https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala#L34

Back to the lazy add idea, maybe change the impl a bit to have `filters: List[BF] = Nil` and `tail: Array[Byte]` when reading from persisted bytes, and only decode the entire `List[BF]` & throw away `tail` when `mightContain` is called? That way we retain the consistent API, and quietly optimize the use case of read + append only case.

kellen:
Did a pass to make it lazy. Not simple any more, but neither is it terribly complicated. 

I also wrote an `ApproxFilter` wrapper but it doesn't seem to have any advantage over the single BF ApproxFilter.

scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala
```diff
@@ -355,3 +347,41 @@ object BloomFilter extends ApproxFilterCompanion {
...
+class ScalableBloomFilter[T: g.Funnel] private (private val impl: MutableScalableBloomFilter[T])
+    extends ApproxFilter[T] {
+
+  /** Return an estimate for the total number of distinct elements that have been added to this [[ApproxFilter]]. */
+  override def approxElementCount: Long = impl.approximateElementCount
+  override def mightContain(elem: T): Boolean = impl.mightContain(elem)
+  override val expectedFpp: Double = impl.fpProb
+}
+
+// TODO this has questionable value; the API has already materialized `elems`, so you can with less overhead just create a single normal BF
```
nevillelyh, line 360: That's true, maybe it's indeed overkill to fit `ApproxFilter` here, plus the mutable nature doesn't really fit `AF` 's immutable design. Should we undo that?

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
```diff
@@ -0,0 +1,215 @@
...
+case class MutableScalableBloomFilter[T](
+  fpProb: Double,
+  private var headCapacity: Long,
+  private val growthRate: Int,
+  private val tighteningRatio: Double,
+  private var headFPProb: Double,
+  // storing a count of items in the head avoids calling the relatively expensive `approximateElementCount` after each insert
+  private var headCount: Long,
+  private var head: Option[g.BloomFilter[T]],
+  private var tail: List[Either[g.BloomFilter[T], SerializedBloomFilters]]
```
nevillelyh, line 139: If it's not serialized yet, `tail` should be `List(Right(SerializedBloomFilters(...)))` i.e. always `List` of a single element? Can this type be `Either[List[g.BF[T]], SerializedBF]]` so it's clearer?

kellen: Let's say you've pulled an existing SBF off disk and you're adding new items, then you saturate the `head` and need to add another BF, pushing the current head to `tail`.At that point, if we have `Either[List[g.BF[T]], SerializedBF]]` we'll be forced to serialize the old `head` which seemed like a costly operation to have on insert, so here it's deferred until you actually want to encode to bytes.

scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala
kellen: Yeah will do.

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
```diff
@@ -0,0 +1,215 @@
...
+    val dis = new DataInputStream(bais)
+
+    val fpProb = dis.readDouble()
+    val headCapacity = dis.readLong()
+    val growthRate = dis.readInt()
+    val tighteningRatio = dis.readDouble()
+    val headFPProb = dis.readDouble()
+    val headCount = dis.readLong()
+    val numFilters = dis.readInt()
+    val head = if (numFilters > 0) Some(g.BloomFilter.readFrom[T](dis, funnel)) else None
```
nevillelyh, line 82: There's another tiny opt we can do here, if `headCount >= headCapacity` i.e. head is already full, make it `None` and leave it as part of `tail` instead?

syodage (approved):
LGTM! 👍

scio-core/src/main/scala/com/spotify/scio/hash/MutableScalableBloomFilter.scala
kellen: This feels very edge-casey. I gave it a shot but it would introduce a bug where the next BF after append would have the same size as the previous one since we're using `None` as the flag for initial-BF-creation where the parameters don't get adjusted.

The other option would be to create the next head at deserialization time but that then requires exposing some of the internals in an unaesthetic way :/

nevillelyh: Yeah maybe not worth it. The chance of that happening is `1/headCapacity` and increasing slim as we add more.

PR #3340, opened 2020-09-23
Add `scio-google-cloud-platform`

### LINKED ISSUE #3237: `scio-google-cloud-platform`
Issues #3202 #3231 gave second thoughts. One of the end goals of moving things out of `scio-core` is to really keep it with core types, minimal dependencies and allow users to not pull extra cruft. That's what led to the separation of `bigtable`, `bigquery` and `spanner`. Now with the separation of `datastore` and `pubsub` we will end up with a lot of modules (possibly more in the future).

* `scio-bigtable`
* `scio-bigquery`
* `scio-spanner`
* `scio-pubsub`
* `scio-datastore`

In reality cruft separation doesn't really happen:

```bash 
cs resolve com.spotify::scio-bigtable:0.9.3 --scala 2.13 | grep bigquery

com.google.api.grpc:proto-google-cloud-bigquerystorage-v1:0.90.0:default
com.google.api.grpc:proto-google-cloud-bigquerystorage-v1alpha2:0.90.0:default
com.google.api.grpc:proto-google-cloud-bigquerystorage-v1beta1:0.90.0:default
com.google.api.grpc:proto-google-cloud-bigquerystorage-v1beta2:0.90.0:default
com.google.apis:google-api-services-bigquery:v2-rev20191211-1.30.9:default
com.google.cloud:google-cloud-bigquery:1.106.0:default
com.google.cloud:google-cloud-bigquerystorage:0.125.0-beta:default
````

```bash
cs resolve com.spotify::scio-bigquery:0.9.3 --scala 2.13 | grep bigtable

com.google.api.grpc:proto-google-cloud-bigtable-v2:1.9.1:default
```

Dependency wise, all share a lot of the same underlying dependencies i.e `grpc`, `google-cloud`, `beam-sdks-java-io-google-cloud-platform` and it`s hard to actively make transitive dependencies `provided` 

On a side note: Coders for these modules have been leaving within `scio-core` and they should ideally be in their respective submodule meaning that `BigQuery` `TableRow` coders should be under `scio-bigquery` requiring users to:

```scala
import com.spotify.scio.coders._
import com.spotify.scio.bigquery.coders._

// if we had scio-google-cloud-platform
import com.spotify.scio.gcp.coders._
```

Because of this I'm entertaining the idea of bringing `GCP` IO's together; one potential side-effect is that it would allow us to solve dep conflict better when used together.

wdyt?

regadas:
Closes #3237

This PR is moving all GCP related IO's, extension methods and coders to a different project.

Every IO follows the following package convention:

```scala
// brings extension methods into scope
import com.spotify.scio.< IO >._

// brings coders into scope.
import com.spotify.scio.< IO >.coders._
```

**Note** that some packages don't require extra coders so `com.spotify.scio.< IO >.coders` might not be available.

A few breaking changes are introduced:

* Users of `Bigquery`, `Bigtable`, `Spanner` will need to add a new dependency to theis build
  ```scala
  libraryDependencies += "com.spotify" %% "scio-google-cloud-platform" % scioVersion
  ```
  
  `Bigquery` users might require `import com.spotify.scio.bigquery.coders._` to avoid fallback. I might actually remove this one since `TableRow` coder can be supplied by `genericJsonCoder`.

* `Pubsub` and `Datastore` users will require adding a new package:
  ```scala
  import com.spotify.scio.pubsub._
  import com.spotify.scio.pubsub.coders._
  import com.spotify.scio.datastore._
  ```


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryCoders.scala
```diff
@@ -0,0 +1,30 @@
...
+package com.spotify.scio.bigquery
+
+import com.spotify.scio.coders.Coder
+
+import com.google.api.services.bigquery.model.TableRow
+import org.apache.beam.sdk.io.gcp.bigquery.TableRowJsonCoder
+
+trait BigQueryCoders {
+
+  implicit def tableRowCoder: Coder[TableRow] = Coder.beam(TableRowJsonCoder.of())
```
regadas, line 26: I will see if can just use `BeamTypeCoders#genericJsonCoder`

nevillelyh (commented):
So `bigquery/bigtable/spanner` into a single sub-project? No conflict among them?

regadas:
@nevillelyh right! for me this is a big plus. By having unit/IT tests playing nice together give us some degree of assurance that deps play nice together.

PR #3343, opened 2020-09-24
Remove BeamCoders unchecked warning

regadas:


syodage (commented):
What is the warning we suppressed here? I don't see any related warning when I build the `scio-core` 🤷

PR #3345, opened 2020-09-24
Deprecate ScioContext pubsubIO methods

### LINKED ISSUE #2535: Review PubSub context methods
Old `apply` method of `PusubIO` is deprecated in favor of specific format records. I guess we need to review these as well:

https://github.com/spotify/scio/blob/b42cf6a217929f7bad5382ebb96329901cb91a13/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L722-L752

Also we need to review corresponding tests.

jto:
Fixes #2535 

I don't think it makes much sense to add a method for each of the supported types in `ScioContext` since it would add quite a lot of methods:

- `pubsubSubscriptionAsAvro`
- `pubsubSubscriptionAsProtobuf`
- `pubsubSubscriptionAsPubSubMessage`
- `pubsubSubscriptionAsCoder`
- `pubsubTopicAsAvro`
- `pubsubTopicAsProtobuf`
- `pubsubTopicAsPubSubMessage`
- `pubsubTopicAsCoder`

Adding all those methods seems quite ridiculous so I'd rather encourage peeps to just use `sc.read(...)`

I also deprecated and renamed the `readX` methods on `PubsubIO` since they can be used both for reads and writes

wdyt ?


jto:
Can I merge this ?

syodage (approved):
LGTM! 👍 

I was thinking, we should move all `scio-core` GCP IOs to separate module, e.g `scio-gcp-io` or maybe part of this module https://github.com/spotify/scio/issues/3237 . Definitely not something address with these changes.

PR #3350, opened 2020-09-24
Add specialized TupleCoders

regadas:
I recently noticed that Tuple3 / Tuple4 are not that uncommon so decided to specialize them to save on serialization. Users should rarely rely on above then that but this adds for the other ones as well (bonus).

We should at some point ditch these python scripts and use either paiges or scalafix.

### DIFF
--- a/scio-core/src/main/scala-2.12/com/spotify/scio/coders/instances/TupleCoders.scala
+++ b/scio-core/src/main/scala-2.12/com/spotify/scio/coders/instances/TupleCoders.scala
@@ -24,6 +24,7 @@ package com.spotify.scio.coders.instances
 
 import java.io.{InputStream, OutputStream}
 
+import shapeless.Strict
 import com.spotify.scio.coders.{Coder, CoderStackTrace}
 import org.apache.beam.sdk.coders.Coder.NonDeterministicException
 import org.apache.beam.sdk.coders.{Coder => BCoder, _}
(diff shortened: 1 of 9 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -499,8 +499,7 @@ object Coder
   implicit val noneCoder: Coder[None.type] = ScalaCoders.noneCoder
   implicit val bitSetCoder: Coder[BitSet] = ScalaCoders.bitSetCoder
   implicit def seqCoder[T: Coder]: Coder[Seq[T]] = ScalaCoders.seqCoder
-  import shapeless.Strict
-  implicit def pairCoder[A, B](implicit CA: Strict[Coder[A]], CB: Strict[Coder[B]]): Coder[(A, B)] =
+  implicit def pairCoder[A, B](implicit CA: Coder[A], CB: Coder[B]): Coder[(A, B)] =
```
regadas, line 502: Hmm, I think we don't actually need the PairCoder here anymore we can have it as part of `TupleCoders` 🤔

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -527,10 +527,9 @@ trait ScalaCoders {
   implicit def seqCoder[T: Coder]: Coder[Seq[T]] =
     Coder.transform(Coder[T])(bc => Coder.beam(new SeqCoder[T](bc)))
 
-  import shapeless.Strict
```
regadas, line 530: hmm, I guess 2.12 will still require the use of `shapeless.Strict` ...

scio-core/src/main/scala-2.12/com/spotify/scio/coders/instances/TupleCoders.scala
```diff
@@ -0,0 +1,4715 @@
...
+  implicit def tuple3Coder[A, B, C](implicit
+    CA: Strict[Coder[A]],
+    CB: Strict[Coder[B]],
+    CC: Strict[Coder[C]]
+  ): Coder[(A, B, C)] =
+    Coder.transform(CA.value) { ac =>
+      Coder.transform(CB.value) { bc =>
+        Coder.transform(CC.value)(cc => Coder.beam(new Tuple3Coder[A, B, C](ac, bc, cc)))
+      }
+    }
```
jto, line 213: Nesting `Coder.transform` will be slower than the code generated using magnolia which is why the old version was using `Coder.gen`.

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -391,19 +393,18 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
       intercept[NonDeterministicException] {
-        val coder = Coder[(Double, Double, Double)]
+        val coder = Coder[NonDeterministic]
 
         materialize(coder).verifyDeterministic()
       }
 
     val expectedMsg =
-      "RecordCoder[scala.Tuple3](_1 -> DoubleCoder, _2 -> DoubleCoder, _3 -> DoubleCoder)" +
+      "RecordCoder[com.spotify.scio.coders.NonDeterministic](a -> DoubleCoder, b -> DoubleCoder)" +
```
jto, line 402: This seems like a worse error message. Why change it ?

jto (commented):
I'm pretty sure this will actually be slower than the current version. Did you run benchmarks ?

regadas (commented):
Hi @jto! good points I should have made them clear in the PR description.

To emphasise the main point of this PR is to address serde time. From the quick benchmarks below we can see improvements between **20%** to **50%**. This can translate to somewhat big savings when using `cogroup` ops.

### master
```
[info] Benchmark                               Mode  Cnt    Score    Error  Units
[info] CoderBenchmark.tuple3Decode             avgt    5   53.833 ±  0.326  ns/op
[info] CoderBenchmark.tuple3Encode             avgt    5   86.729 ± 10.622  ns/op
[info] CoderBenchmark.tuple4Decode             avgt    5   69.513 ±  1.937  ns/op
[info] CoderBenchmark.tuple4Encode             avgt    5  106.868 ±  1.345  ns/op
```
### PR
```
[info] Benchmark                               Mode  Cnt   Score   Error  Units
[info] CoderBenchmark.tuple3Decode             avgt    5  26.200 ± 0.136  ns/op
[info] CoderBenchmark.tuple3Encode             avgt    5  65.883 ± 2.037  ns/op
[info] CoderBenchmark.tuple4Decode             avgt    5  40.348 ± 0.209  ns/op
[info] CoderBenchmark.tuple4Encode             avgt    5  72.218 ± 0.524  ns/op
```

~Regarding **compile** time I'm aware that there's in fact some degradation (not really sure how much in user code base but there's some in scio.) but as I mention that's not the main focus here and perhaps this can be alleviated with PR #3170~

~We can try however a few approaches:~
~1) Apply this to tuples up to 4/5. Above that we can resort to `Coder.gen`.~
~2) If possible try see if we can optimise nested `Coder.transform` it in a separate PR.~

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
regadas: This is actually test `case class` with non deterministic fields. `tuple` is being tested above.

jto:
I was also talking about serde time. I think I remember fiddling with nested `Coder.transform` which was performing worse than the implementation of `LowPriorityCoderDerivation` because of the closure nesting being slower that iterating on an array. I'll need to have another look at that.

regadas:
So I was looking at what impacted **compile time** by **2.8%** and found out that it was the removal of `pairCoder` from `Coder` companion object. Not being under object scope makes implicit search take a little bit longer. We could bring it back plus the new ones but that would mean to also bring the `shapeless.Strict` into the scene to make it work with `2.12` but I want to avoid it so I'm gonna let be as is.

jto (approved):

PR #3351, opened 2020-09-24
add type param T for withWindowFn, fix #2980

### LINKED ISSUE #2980: Introduce a second type parameter in `withWindowFn` for better type safety.
This is not done right now so we don't break backwards compat.

https://github.com/spotify/scio/pull/2975#discussion_r425094916

nevillelyh:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1099,7 +1102,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
       transform = transform.every(period)
     }
     transform = transform.withOffset(offset)
-    this.withWindowFn(transform, options)
+    this.withWindowFn(transform.asInstanceOf[WindowFn[T, IntervalWindow]], options)
```
nevillelyh, line 1105: It's kinda ugly though since most Beam `WindowFn` instances just hard code `Object` for `T` :shrug:

regadas: yeah! hmm this is tricky the right fix for this would be https://github.com/spotify/scio/pull/2975#discussion_r425094916 but this only works on 2.13.

I don't really want users to start using `.asInstanceOf` 😢 One way I was thinking was to have 2 versions of this fn. 2.12 would be kept as is and 2.13 would be stricter. 🤷 

OTOH we can always just leave it as is? `WindowFn[_ <: Any, W]`

nevillelyh: I'm fine keeping it as is. Since `T` should just be the current `SCollection[T]`, the `_ <: Any` part is more of an artifact of Java type inference defects? Basically it's not more unsafe leaving it as is.

regadas: 👍 keeping it as is!

Issue #2824, opened 2020-03-31
Dynamic Destination for Parquet Files on GCS


jayadeep-jayaraman:
Is there an example of FileIO.WriteDynamic for Parquet files. I see that there is saveAsDynamicTextFile and saveAsDynamicAvroFile. It would be great if we can have a similar saveAsDynamicParquetFile API.

regadas:
Hi @cejj at the moment we don't have such example. If you come up with one, could you make a PR? 😄

cejj:
Sure let me work on it.

cejj:
@regadas - I have created an example. Can you let me know the process for raising a PR and submitting the example ?

regadas:
Hi @cejj sorry for the late reply! There's no process really, just make a PR against `master` 😄

PR #3356, opened 2020-09-26
scio-parquet to support dynamic destinations for windowed scollections

RyanBerti:
This is a draft PR to support https://github.com/spotify/scio/issues/2824; it provides an API for defining a custom filename based on the window associated with the SCollection (see https://spotify-foss.slack.com/archives/C015189TFDH/p1600776553015800). This differs from other dynamic file support implementations, which provide an API for defining the filename based on the SCollection object directly; it's unclear which functionality the original issue was requesting, but this PR addresses the functionality requested in the slack thread.

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -43,6 +43,7 @@ import org.apache.parquet.hadoop.metadata.CompressionCodecName
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
 import com.spotify.scio.io.TapT
+import com.spotify.scio.parquet.avro.ParquetAvroIO.WriteParam
 import org.apache.beam.sdk.io.FileBasedSink.FilenamePolicy
 import org.apache.beam.sdk.io.fs.ResolveOptions.StandardResolveOptions
 import org.apache.beam.sdk.io.fs.ResourceId
@@ -92,7 +93,10 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
       FileBasedSink.convertToFileResourceIfPossible(ScioUtil.pathWithShards(path))
     val prefix = StaticValueProvider.of(resource)
     val fileNamePolicy =
-      if (params.windowedFilenameFunction.isDefined || params.filenameFunction.isDefined) {
+      if (
+        params.windowedFilenameFunction != WriteParam.DefaultWindowedFilenameFunction ||
+        params.filenameFunction != WriteParam.DefaultFilenameFunction
+      ) {
         createFilenamePolicy(
           resource,
           params.windowedFilenameFunction,
@@ -125,8 +129,8 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
 
   def createFilenamePolicy(
     baseFileName: ResourceId,
-    windowedFilenameFunction: Option[(Int, Int, BoundedWindow, PaneInfo) => String],
-    filenameFunction: Option[(Int, Int) => String],
+    windowedFilenameFunction: (Int, Int, BoundedWindow, PaneInfo) => String,
+    filenameFunction: (Int, Int) => String,
     filenameSuffix: String
   ): FilenamePolicy = {
 
@@ -139,13 +143,11 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
         paneInfo: PaneInfo,
         outputFileHints: FileBasedSink.OutputFileHints
       ): ResourceId = {
-        windowedFilenameFunction
-          .map(f => f(shardNumber, numShards, window, paneInfo))
-          .map(fileName =>
-            baseFileName.getCurrentDirectory
-              .resolve(fileName + filenameSuffix, StandardResolveOptions.RESOLVE_FILE)
-          )
-          .getOrElse(???)
+        val filename = windowedFilenameFunction(shardNumber, numShards, window, paneInfo)
+        baseFileName.getCurrentDirectory.resolve(
+          filename + filenameSuffix,
+          StandardResolveOptions.RESOLVE_FILE
+        )
       }
 
       override def unwindowedFilename(
@@ -153,13 +155,11 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
         numShards: Int,
         outputFileHints: FileBasedSink.OutputFileHints
       ): ResourceId = {
-        filenameFunction
-          .map(f => f(shardNumber, numShards))
-          .map(fileName =>
-            baseFileName.getCurrentDirectory
-              .resolve(fileName + filenameSuffix, StandardResolveOptions.RESOLVE_FILE)
-          )
-          .getOrElse(???)
+        val filename = filenameFunction(shardNumber, numShards)
+        baseFileName.getCurrentDirectory.resolve(
+          filename + filenameSuffix,
+          StandardResolveOptions.RESOLVE_FILE
+        )
       }
     }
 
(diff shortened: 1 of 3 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -87,23 +91,80 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
...
     data.applyInternal(t)
     tap(ParquetAvroIO.ReadParam[T, T](writerSchema, null, identity))
   }
 
   override def tap(params: ReadP): Tap[T] =
     ParquetAvroTap(ScioUtil.addPartSuffix(path), params)
+
+  def createFilenamePolicy(
+    baseFileName: ResourceId,
+    windowedFilenameFunction: Option[(Int, Int, BoundedWindow, PaneInfo) => String],
```
nevillelyh, line 128: Should we require at least one or both `Option` to be defined? That'll remove the `.getOrElse(???)` code smell?
Is there a case where both should be defined?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -87,23 +91,80 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
...
+    new FileBasedSink.FilenamePolicy {
+
+      override def windowedFilename(
+        shardNumber: Int,
+        numShards: Int,
+        window: BoundedWindow,
+        paneInfo: PaneInfo,
+        outputFileHints: FileBasedSink.OutputFileHints
+      ): ResourceId = {
+        windowedFilenameFunction
```
nevillelyh, line 142: Minor: see if we can make `Option[F]` `F` instead to get rid of the 2 `.map` & object allocation? If either one of the functions can be absent, we can provide a fallback e.g. `(Int, Int) => throw new IllegalArgumentException()` with a meaningful message?

nevillelyh (commented):
Thanks, overall LGTM. If we can make the parameter expectation more explicit.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
RyanBerti: I was wondering how to do this - I don't think there's different SCollection syntax if you're using the default global window or have explicitly provided a window. Because of this, it's sort of up to the user to understand which file naming function they need to define. I could explicitly check to make sure that correct filenaming function is defined depending on whether the global window is in use or not, and then conditionally create Filenaming instances with the correct method implemented depending on which of the filenaming functions are defined. I think I started going down that path initially but decided against it as deferring to the unimplemented exception felt like a lot less code. Happy to update the implementation to be more strict, just let me know.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
RyanBerti: So I'm currently giving the new WriteParams a default of `None`, but I think what you're proposing (change the default to a function that throws an exception instead) would get rid of all of the Options and allow us to get rid of the `getOrElse(???)` as well. I'll do this, as it'll clean things up a little more.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
RyanBerti: I think I've resolved this with my latest commit, ty!

RyanBerti:
Circleci seems to keep timing out after running the tests successfully - known issue?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -87,23 +92,79 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
     val resource =
       FileBasedSink.convertToFileResourceIfPossible(ScioUtil.pathWithShards(path))
     val prefix = StaticValueProvider.of(resource)
-    val usedFilenamePolicy =
-      DefaultFilenamePolicy.fromStandardParameters(prefix, null, params.suffix, false)
-    val destinations = DynamicFileDestinations.constant[T](usedFilenamePolicy)
+    val fileNamePolicy =
+      if (
+        params.windowedFilenameFunction != WriteParam.DefaultWindowedFilenameFunction ||
```
nevillelyh, line 97: I'd avoid comparing functions. FWIW they could behave the same and still fail equality. We should rename the 2 `Default*Function`s `Fallback*Function`s? Also see comments below, can we use a `Option[Either[WFn, Fn]]` here?

Also, if `windowedFilenameFunction` is required for windowed writes, we should check `data.internal.getWindowingStrategy != WindowingStrategy.globalDefault()` like you did on line 119 right?

So basically something like:

```
if (optFn.isDefined) {
  check collection is windowed
  windowedFilenameFn = optFn.get.left.orElse(fallbackWindowedFn)
  filenameFn = optFn.get.right.orElse(fallback...)
  createFilenamePolicy
} else {
  DefaultFilenamePolicy...
}
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -138,5 +139,47 @@ package object avro {
...
+     *                         (the shard number,
+     *                         the total number of shards)
+     */
+    def saveAsDynamicParquetAvroFile(
+      path: String,
+      schema: Schema = WriteParam.DefaultSchema,
+      numShards: Int = WriteParam.DefaultNumShards,
+      suffix: String = WriteParam.DefaultSuffix,
+      compression: CompressionCodecName = WriteParam.DefaultCompression,
+      windowFilenameFunction: (Int, Int, BoundedWindow, PaneInfo) => String =
```
nevillelyh, line 166: Looking at `FilenamePolicy`, these 2 functions are mutually exclusive? Like the write is either windowed if underlying `PCollection` is, or not, but never the same time?

If that's the case, we can use a `Either[WindowFilenameFunction, FilenameFunction]`? Basically requiring the user to always provide 1, but not both.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -147,13 +208,23 @@ object ParquetAvroIO {
...
+      throw new NotImplementedError("saveAsDynamicParquetAvroFile requires a filename function")
   }
 
   final case class WriteParam private (
     schema: Schema = WriteParam.DefaultSchema,
     numShards: Int = WriteParam.DefaultNumShards,
     suffix: String = WriteParam.DefaultSuffix,
-    compression: CompressionCodecName = WriteParam.DefaultCompression
+    compression: CompressionCodecName = WriteParam.DefaultCompression,
+    windowedFilenameFunction: (Int, Int, BoundedWindow, PaneInfo) => String =
```
nevillelyh, line 225: If these 2 functions are indeed mutually exclusive as discussed below, this should be an `Option[Either[WindowedFilenameFunction, FilenameFunction]]`?

The semantics being: the `Option[T]` must be defined for dynamic writes, absent for regular writes. And for dynamic writes, either one of the 2 functions must be provided.

RyanBerti: I like this idea - I'll update implementation to use an Option[Either[]] approach.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
RyanBerti: I agree re: comparing functions; the Option[Either[]] approach makes things a lot easier. WRT the check that the user provides a windowed filename function when the collection is windowed, I'd prefer to defer to the exception thrown by the default function as that seems more concise?

nevillelyh: But that exception will only be throw after job submission and when runner starts writing? Checking early would prevent job from being submitted.

RyanBerti: That's a good point; I'll update the explicit check.

nevillelyh (approved):
Much cleaner!

nevillelyh:
LGTM thanks!

PR #3357, opened 2020-09-28
Support $LATEST replacement for Query

liufuyang:
So to support Bigtable queries containing `$LATEST` in the table name.

### DIFF
--- a/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
+++ b/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
@@ -316,6 +316,7 @@ object StorageIT {
   @BigQueryType.fromStorage("data-integration-test:partition_a.table_%s", List("$LATEST"))
   class StorageLatest
 
+  // Require setting default project to allow this work locally. Use command `gcloud config set project data-integration-test`.
   @BigQueryType.fromStorage("partition_a.table_%s", List("$LATEST"))
   class StorageEnvProject
 
--- a/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
+++ b/scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
@@ -316,7 +316,6 @@ object StorageIT {
   @BigQueryType.fromStorage("data-integration-test:partition_a.table_%s", List("$LATEST"))
   class StorageLatest
 
-  // Require setting default project to allow this work locally. Use command `gcloud config set project data-integration-test`.
   @BigQueryType.fromStorage("partition_a.table_%s", List("$LATEST"))
   class StorageEnvProject
 
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
@@ -39,7 +39,7 @@ sealed trait Source
 final case class Query(underlying: String) extends Source {
 
   /**
-   *  A helper method to replace the "$LATEST" laceholder in query to the latest common partition.
+   *  A helper method to replace the "$LATEST" placeholder in query to the latest common partition.
    *  For example:
    *  {{{
    *  @BigQueryType.fromQuery("SELECT ... FROM `project.data.foo_%s`", "$LATEST")
--- a/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
+++ b/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
@@ -19,23 +19,26 @@ package com.spotify.scio.bigquery
 import java.math.MathContext
 import java.nio.ByteBuffer
 
-import com.spotify.scio.ScioContext
-import com.spotify.scio.values.SCollection
 import com.google.api.services.bigquery.model.{
+  TableReference => GTableReference,
   TableRow => GTableRow,
-  TimePartitioning => GTimePartitioning,
-  TableReference => GTableReference
+  TimePartitioning => GTimePartitioning
 }
-import org.apache.beam.sdk.io.gcp.bigquery.{BigQueryHelpers, BigQueryInsertError, WriteResult}
+import com.spotify.scio.ScioContext
+import com.spotify.scio.bigquery.client.BigQuery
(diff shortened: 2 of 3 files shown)


scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -19,23 +19,26 @@ package com.spotify.scio.bigquery
...
+import org.joda.time._
 import org.joda.time.format.{DateTimeFormat, DateTimeFormatterBuilder}
-import org.joda.time.DateTimeZone
-import org.joda.time.{Instant, LocalDate, LocalDateTime, LocalTime}
 
 sealed trait Source
 
-final case class Query(underlying: String) extends Source
+final case class Query(underlying: String) extends Source {
+  def latest(bq: BigQuery): Query =
```
regadas, line 39: Thank you for this! Would you mind add a small scaladoc here?

regadas: I actually got some more context around this so the issue here is that `ScioContext#bigQuerySelect` is not respecting `Query` `$LATEST`.

Instead of exposing `Query#latest` I would probably explore first if we can have `BigQuerySelect` respect `$LATEST` without having the need to use `Query#latest`. 

This should be possible by formatting the query here https://github.com/spotify/scio/blob/16756a3e0e768baa4188c7135a1a2f00eb51460c/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L65

regadas (changes requested):
Gotta more context around this change. Let me know what you think

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
liufuyang: Thanks for the info. I am totally new to Scio. Before I try to format it query at line 65 on `BigQueryIO.scala`, I would like to ask:

Do you know that there is a formatting operation happens on 
https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala#L127

So if we move the format latest to `bqReadQuery()` in `BigQueryIO.scala` then perhaps we don't need to do that on the `TypeProvider` line 127 or...? 🤔  

And this format of `$LATEST` then basically will become some default and fixed logic (and hopefully no one will really name their table sending with $LATEST?)

regadas: > So if we move the format latest to bqReadQuery() in BigQueryIO.scala then perhaps we don't need to do that on the TypeProvider line 127 or...?

That will still be needed to resolve the table schema

> And this format of $LATEST then basically will become some default and fixed logic (and hopefully no one will really name their table sending with $LATEST?)

I'm ok with that since tables are only allowed to contain letters, numbers and underscores.

If we support the above, users will be able to do the following:
```scala
@BigQueryType.fromQuery("select ... from foo_$LATEST")
object Foo
```

```scala
scioContext.bigQuerySelect(Foo.queryAsSource())
```

regadas: @liufuyang As I was thinking a little bit more what I proposed I realised I don't really like that it would be pushing this logic to `BigQueryIO`'s as it would bring more inconsistencies. 

Let's keep what you had before! add scaladoc and we can add an additional method that uses `BigQuery.defaultInstance` so we would have `Query#latest()`

In reality the above example would be:
```scala
@BigQueryType.fromQuery("select ... from foo_%s", "$LATEST")
object Foo

scioContext.bigQuerySelect(Foo.queryAsSource("$LATEST").latest())
```

liufuyang: @regadas Thanks for the input. I updated the code. How does it look now?

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -19,23 +19,48 @@ package com.spotify.scio.bigquery
...
+   *  Or, if your query string is a dynamic value which uses a "$LATEST" placeholder,
+   *  {{{
+   *  val dynamicSQLStr = "SELECT ... FROM some_table_$LATEST"
+   *  scioContext.bigQuerySelect(Query(dynamicSQLStr).latest(bq))
+   *  }}}
+   *
+   * @param bq [[BigQuery]] client
+   * @return [[Query]] with "$LATEST" replaced
+   */
+  def latest(bq: BigQuery): Query =
```
regadas, line 61: @liufuyang can you also add?

```suggestion
  def latest(): Query = latest(BigQuery.defaultInstance)
```

scio-bigquery/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
```diff
@@ -316,6 +316,7 @@ object StorageIT {
   @BigQueryType.fromStorage("data-integration-test:partition_a.table_%s", List("$LATEST"))
   class StorageLatest
 
+  // Require setting default project to allow this work locally. Use command `gcloud config set project data-integration-test`.
```
regadas, line 319:
```suggestion
  project data-integration-test`.
```

I think this comment is better added in the docs. I'll purpose smth soon.

scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -19,23 +19,50 @@ package com.spotify.scio.bigquery
...
-import org.joda.time.{Instant, LocalDate, LocalDateTime, LocalTime}
 
 sealed trait Source
 
-final case class Query(underlying: String) extends Source
+/** A wrapper type [[Query]] which wraps a SQL String. */
+final case class Query(underlying: String) extends Source {
+
+  /**
+   *  A helper method to replace the "$LATEST" laceholder in query to the latest common partition.
```
regadas, line 42:
```suggestion
   *  A helper method to replace the "$LATEST" placeholder in query to the latest common partition.
```

regadas (approved):

regadas:
Thanks @liufuyang

PR #3359, opened 2020-09-28
suport null-key records in smb writes

ClaireMcGinty:
Records with null keys are diverted to a separate file, `bucket-null-keys.[avro|tfrecord|json]`. this completes a feature we started implementing in an earlier version and didn't finish.

I got rid of the sharding aspect of the null-key bucket -- I'm envisioning this would just be a small spillover of records, @nevillelyh wdyt? We could also add an optimization so that the null-key bucket skips the SortDoFn step but I don't think it matters too much.

this file is ignored in SMB reads/transforms, it's designed for regular file reads.

(the failing CircleCI test is from a transient bigtable integration test failure)

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketShardId.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketShardId.java
@@ -47,8 +47,8 @@ public abstract class BucketShardId {
     return new AutoValue_BucketShardId(bucketId, shardId);
   }
 
-  public static BucketShardId ofNullKey(int shardId) {
-    return new AutoValue_BucketShardId(NULL_KEYS_BUCKET_ID, shardId);
+  public static BucketShardId ofNullKey() {
+    return new AutoValue_BucketShardId(NULL_KEYS_BUCKET_ID, 0);
   }
 
   public boolean isNullKeyBucket() {
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SMBFilenamePolicy.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SMBFilenamePolicy.java
@@ -145,7 +145,7 @@ public final class SMBFilenamePolicy implements Serializable {
       return directory.resolve(timestamp + METADATA_FILENAME, StandardResolveOptions.RESOLVE_FILE);
     }
 
-    /** Returns a ResourceId matching the */
+    /** Returns a ResourceId matching the null keys. */
     public ResourceId forNullKeys() {
       return directory.resolve(
           NULL_KEYS_BUCKET_TEMPLATE + "*" + filenameSuffix, StandardResolveOptions.RESOLVE_FILE);
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SMBFilenamePolicy.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SMBFilenamePolicy.java
@@ -18,6 +18,7 @@
 package org.apache.beam.sdk.extensions.smb;
 
 import java.io.Serializable;
+import java.util.List;
 import java.util.UUID;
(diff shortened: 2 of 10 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SMBFilenamePolicy.java
```diff
@@ -144,6 +145,12 @@ public ResourceId forMetadata() {
       return directory.resolve(timestamp + METADATA_FILENAME, StandardResolveOptions.RESOLVE_FILE);
     }
 
+    /** Returns a ResourceId matching the */
```
nevillelyh, line 148:
```suggestion
    /** Returns a ResourceId matching the null keys. */
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketShardId.java
```diff
@@ -47,8 +47,8 @@ public static BucketShardId of(int bucketId, int shardId) {
     return new AutoValue_BucketShardId(bucketId, shardId);
   }
 
-  public static BucketShardId ofNullKey(int shardId) {
-    return new AutoValue_BucketShardId(NULL_KEYS_BUCKET_ID, shardId);
```
nevillelyh, line 51: Does this mean we used to create `shardId` null key files instead just 1 for the entire sink?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -343,21 +339,26 @@ public void processElement(ProcessContext c) {
       final V record = c.element();
       final K key = extractKeyFn.apply(record);
 
-      final KV<BucketShardId, byte[]> bucketIdAndSortBytes =
-          Optional.ofNullable(getResource().getIfPresent(key))
```
nevillelyh, line 347: The `key` param to `getIfPresent` is marked `@NonNull`, how does that work before when `key == null`?  I guess the entire `Optional` becomes `None` and it hits `orElseGet` then `processKey`, which should still handle the null key?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -790,9 +791,12 @@ public final WriteResult expand(PCollection<KV<K, V>> input) {
...
                       @ProcessElement
                       public void processElement(ProcessContext c) {
-                        if (!bucketMetadata
-                            .extractKey(c.element().getValue())
-                            .equals(c.element().getKey())) {
+                        final K key = bucketMetadata.extractKey(c.element().getValue());
+                        final boolean keysMatch =
+                            key == null
+                                ? c.element().getKey() == null
+                                : key.equals(c.element().getKey());
```
nevillelyh, line 798: This reminds me, for Avro, we're relying on `exrtraKey` returns `String` instead of `CharSequence`? Doesn't seem the case looking at `AvroBucketMetadata`. Do we want to compare encoded bytes or deep compare?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketShardId.java
ClaireMcGinty: yep, we used to.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
ClaireMcGinty: I think it was a bug before... we just didn't have a test case for it.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
ClaireMcGinty: ah yeah that's a good point. I can update to compare bytes, that might be simplest.

Issue #3030, opened 2020-06-09
Use HLL+ for approximate distinct counts over vanilla HLL


hansencc:
Beam offers approximate distinct count implementations using HLL+ that has accuracy and performance benefits over HLL.

https://github.com/apache/beam/blob/master/sdks/java/extensions/sketching/src/main/java/org/apache/beam/sdk/extensions/sketching/ApproximateDistinct.java

https://github.com/apache/beam/blob/master/sdks/java/extensions/zetasketch/src/main/java/org/apache/beam/sdk/extensions/zetasketch/HllCount.java 

It would be nice to allow these implementations to be leveraged through Scio.

nevillelyh:
I'm thinking something down the line of this, a good Java impl with light Scala wrapper & coder
https://github.com/spotify/scio/pull/2651

PR #3361, opened 2020-09-29
Add ApproxDistinctCount trait and two HLL++ beam extension implementations.

syodage:
Fixes https://github.com/spotify/scio/issues/3030

- Added `ApproxDistincCounter[T]` trait and following implementations
   - `ApproximateUniqueCounter` - wraps  [ApproximateUnique](https://beam.apache.org/documentation/transforms/java/aggregation/approximateunique/) with sample size.
   - `ApproximateUniqueCounterByError` - wraps  [ApproximateUnique](https://beam.apache.org/documentation/transforms/java/aggregation/approximateunique/) with maximum estimation error.
   - `SketchingHyperLogLogPlusPlus[T]` - wraps [ApproximateDistinct](https://beam.apache.org/releases/javadoc/2.24.0/org/apache/beam/sdk/extensions/sketching/ApproximateDistinct.html) which uses [addthis's stream-lib](https://github.com/addthis/stream-lib)
   - `ZetasketchHll{Int|Long|String|ByteArray}Counter` - wraps [HllCount](https://beam.apache.org/documentation/transforms/java/aggregation/hllcount/), which uses [google's zetasketch HLL++ implementation](https://github.com/google/zetasketch)
 
- Added ` def countApproxDistinct(estimator: ApproxDistinctCounter[T]): SCollection[Long] = {...}`
   - updated both `def countApproxDistinct(sampleSize: Int)` and `def countApproxDistinct(maximumEstimationError: Double = 0.02)` to call the new method with correct estimator.
   - ~Added deprecate notes to deprecate from the `SCollection` in favor of the new method.~

- Added ` def countApproxDistinctByKey(estimator: ApproxDistinctCounter[V])(implicit
    koder: Coder[K],
    voder: Coder[V]
  ): SCollection[(K, Long)]` to the `PairSCollectionFunction`
    - updated both ` def countApproxDistinctByKey(sampleSize: Int)` and  `def countApproxDistinctByKey(maximumEstimationError: Double = 0.02)` to call above method with correct estimator.
    - ~Added deprecate notes to deprecate from the `PairSCollectionFunction` in favor of the new method.~

- Add tests.

Have added both beam HLL++ implementations to `scio-extra` since those are in `beam-exention` packages.


ClaireMcGinty (commented):
i haven't looked through it all yet, but can you add the [Spotify copyright template](https://spotify.github.io/scio/dev/Style-Guide.html#intellij-idea) to your IDE and reformat the new files? We used to have `scalastyle` to check for the presence of copyright header and it seems there's no equivalent in `scalafmt` 🤔

scio-extra/src/test/scala/com/spotify/scio/extra/hll/HLLSpec.scala
```diff
@@ -0,0 +1,30 @@
+package com.spotify.scio.extra.hll
+
+import com.spotify.scio.testing.PipelineSpec
+
+trait HLLSpec extends PipelineSpec {
+
+  def checkWithErrorRate(actual: Seq[Long], expected: Seq[Long], errorRate: Double): Unit = {
```
ClaireMcGinty, line 24: hm... do you think this would be valuable to publish as part of a Scio artifact for consumers to use in their own tests? Since it's there aren't any HLL-specific imports, maybe it could be moved into `scio-test/src/main` and refactored into a `ProbabilisticSpec` or something like that?

syodage: Yes, good idea, even though having only these two functions are not that helpful in most of the probabilistic test, better we start with these and add more helper function as it requires.

syodage:
some of the build configurations keep failing with the following exception 

```
Exception in thread "Thread-47" java.io.EOFException
	at java.base/java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:3141)
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1619)
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:482)
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:440)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:839)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:828)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

and with the following error. 

```
[error] Error: Total 25, Failed 0, Errors 0, Passed 25
[error] Error during tests:
[error] 	Running java with options -classpath .... sbt.ForkMain 43545 failed with exit code 137
```

This is related to https://github.com/sbt/sbt/issues/653

scio-test/src/main/scala/com/spotify/scio/testing/ApproxPipelineSpec.scala
```diff
@@ -0,0 +1,89 @@
+/*
```
syodage, line 1: This adds a couple of helper methods for testing with error rates, there are other ways to do this, 
- add these methods as matchers so we can call `actual shouldBeInRage (expected, errorRate)` 
- Add these as methods of implicits wrapper

I find adding these methods in an implicit class is a bit uglier and adding these in SCollectionMatcher also will pollute it, as these are specific to approximate algorithms. IMO, start with the current approach and later change it once we have a lot of approximation tests would be a good option. But let me know what you all think.

scio-core/src/main/scala/com/spotify/scio/estimators/ApproxDistinctCounter.scala
```diff
@@ -0,0 +1,93 @@
...
+   */
+  def estimateDistinctCount(in: SCollection[T]): SCollection[Long]
+
+  /**
+   * Approximate distinct element per each key in the given key value SCollection.
+   * This will output estimated distinct count per each unique key.
+   */
+  def estimateDistinctCountPerKey[K](
+    in: SCollection[(K, T)]
+  )(implicit koder: Coder[K], voder: Coder[T]): SCollection[(K, Long)]
```
nevillelyh, line 46: We shouldn't need implicit coders here as they can be infered from `in` for 0.10.x

scio-core/src/main/scala/com/spotify/scio/estimators/ApproxDistinctCounter.scala
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * ApproxDistinctCounter impl for [[org.apache.beam.sdk.transforms.ApproximateUnique]] with sample size.
+ *
+ * Count approximate number of distinct values for each key in the SCollection.
+ * @param sampleSize the number of entries in the statistical sample; the higher this number, the
+ * more accurate the estimate will be; should be `>= 16`.
+ */
+case class ApproximateUniqueCounter[T](sampleSize: Int) extends ApproxDistinctCounter[T] {
```
nevillelyh, line 56: These 2 impls are the same as `SCollection#countApproxDistinct` & `PairSCollectionFunctions#countApproxDistinctByKey`?

EDIT: ah I see you deprecated those in favor of these "pluggable" ones. We don't really need to deprecate those though? They still function perfectly fine and has idiomatic names similar to Spark, etc. The new ones with `estimator: ApproxDistinctCounter[V]` can be optional for those who want to use HLL++.

nevillelyh:
For this to be more useful than another way of approximate distinct count, we should expose the raw `HyperLogLogPlus` data structure under the hood? People might want to persist it with the included coder (we might need to wrap it in an Scio `Coder`?) and aggregate distinct counts across pipelines/partitions.
https://beam.apache.org/releases/javadoc/2.4.0/org/apache/beam/sdk/extensions/sketching/ApproximateDistinct.ApproximateDistinctFn.html

BQ supports storing HLL++ as `BYTES`, so we should also guaranteed that the one produced by Scio is compatible with them.
https://cloud.google.com/bigquery/docs/reference/standard-sql/hll_functions

scio-core/src/main/scala/com/spotify/scio/estimators/ApproxDistinctCounter.scala
syodage: I didn't see much value of keeping those APIs, but if those APIs align with other frameworks, don't' see a reason to not support it. I will remove the deprecation note.

syodage:
@nevillelyh, Idea was to have this first and support distributed HLL++ next (as a separate PR). I wasn't thinking to expose the underline data structure(sketch) to users before, will take that to account when building the new API.

One of the assumptions I have is; not all approximate district counts algorithms are distributed friendly(mergeable), this means we can have another trait that extends the `ApproxDistinctCounter` with a set of new APIs which expose the underline data structure + merging capability.

scio-core/src/main/scala/com/spotify/scio/estimators/ApproxDistinctCounter.scala
syodage: tried to build the coder like this `Coder.raw(CoderMaterializer.kvCoder[K, java.lang.Long](in.context))` but it complains `Cannot find an implicit Coder instance for type:   >> K` , is it because this is outside the SColleciton?

nevillelyh: You have to "localize" the implicits like this: https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala#L50

scio-core/src/main/scala/com/spotify/scio/estimators/ApproxDistinctCounter.scala
nevillelyh: Well it's best to not remove things unless necessary. These are supplemented by the 2 new methods, not replaced nor do they cause and confusion with the new ones.

.jvmopts
```diff
@@ -6,6 +6,7 @@
 -XX:ReservedCodeCacheSize=512m
 -XX:+TieredCompilation
 -XX:+UseParallelGC
+-XX:+CMSClassUnloadingEnabled
```
regadas, line 9: Why is this needed?

scio-test/src/main/scala/com/spotify/scio/testing/ApproxPipelineSpec.scala
regadas: I think I'm leaning towards having these as matches; that makes this consistent with the rest.

in regards to:

> adding these in SCollectionMatcher 

We can supply them as we supply `CoderAssertions`

https://github.com/spotify/scio/tree/master/scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala

.jvmopts
syodage: This was added with `-XX:+UseConcMarkSweepGC`  to fix https://github.com/spotify/scio/pull/3361#issuecomment-702776811, but `UseConcMarkSweepGC` is obsoleted and the issue vanished after rebased with the new master. This is leftover from the cleaning, I will remove this.

scio-extra/src/main/scala/com/spotify/scio/extra/hll/sketching/package.scala
```diff
@@ -0,0 +1,73 @@
...
+   * @param p Precision,
+   *          Controls the accuracy of the estimation. The precision value will have an impact on the number of
+   *          buckets used  to store information about the distinct elements. In general one can expect a relative
+   *          error of about 1.1 / sqrt(2&#94;p). The value should be of at least 4 to guarantee a minimal accuracy.
+   *
+   * @param sp Sparse Precision,
+   *           Uses to create a sparse representation in order to optimize memory and improve accuracy at small
+   *           cardinalities. The value of sp should be greater than p(precision), but lower than 32.
+   */
+  case class SketchingHyperLogLogPlusPlus[T](p: Int, sp: Int) extends ApproxDistinctCounter[T] {
```
nevillelyh, line 46: Should we add range check on `p` and `sp`?

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/package.scala
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * The HyperLogLog++ (HLL++) algorithm estimates the number of distinct values in a data stream.
+   * HLL++ is based on HyperLogLog; HLL++ more accurately estimates the number of distinct values in very large and
+   * small data streams.
+   *
+   * @param p Precision, controls the accuracy of the estimation. The precision value will have an impact on the number of buckets
+   *          used to store information about the distinct elements.
+   *          should be in the range `[10, 24]`, default precision value is `15`.
+   */
+  case class ZetasketchHllLongCounter(p: Int = HllCount.DEFAULT_PRECISION)
```
nevillelyh, line 80: You can probably dedup these for specializations with a type class like `ZetaSketchable[T]` and implicit impls for `Int/Long/String/Array[Byte]`?

scio-extra/src/test/scala/com/spotify/scio/extra/hll/zetasketch/ZetasketchHllIntCounterTest.scala
```diff
@@ -0,0 +1,78 @@
...
+class ZetasketchHllIntCounterTest extends PipelineSpec {
+
+  "ZetasketchHLL++" should "estimate int distinct count" in {
+    val estimator = ZetasketchHllIntCounter()
+    val input = for (i <- 0 to 1000000) yield (i % 20)
+    val output = runWithData(input) { scl =>
+      scl
+        .countApproxDistinct(estimator)
+    }
+    output shouldApproximate withErrorRate(Seq(20L), 0.6d)
```
nevillelyh, line 32: Minor: we usually do `0.6f` for floats but leave out `d` for double. Is there a particular reason for this style?

scio-test/src/main/scala/com/spotify/scio/testing/ApproximationAssertions.scala
```diff
@@ -0,0 +1,114 @@
...
+   * @param errorRate - how much percentage off from expected value is acceptable.
+   */
+
+  def withErrorRate(
+    expected: Iterable[Long],
+    errorRate: Double
+  ): ApproximationAssertion[Iterable[Long]] = { (actual: Iterable[Long]) =>
+    actual.size shouldBe expected.size
+    (actual zip expected).foreach { case (act, expt) =>
+      val error = ((expt / 100) * errorRate).toLong
```
nevillelyh, line 82: Is `errorRate` a percentage? So `0.5` actually means 0.0005? I thought the convention is usually absolute error (epsilon), without scaling by 100.
Also `expt` is `Long` so `expt / 100` will truncate. But won't matter if you remove `/ 100`.

Seems Algebird calls it `errorPct` but uses it like `epsilon` without scaling by 100.
https://github.com/twitter/algebird/blob/develop/algebird-test/src/test/scala/com/twitter/algebird/HyperLogLogSeriesTest.scala#L53

Should we just use `epsilon` to be precise?

scio-test/src/main/scala/com/spotify/scio/testing/ApproximationAssertions.scala
```diff
@@ -0,0 +1,114 @@
...
+   * @param errorRate - how much percentage off from expected value is acceptable.
+   */
+  def withErrorRatePerKey[K](
+    expected: Iterable[(K, Long)],
+    errorRate: Double
+  ): ApproximationAssertion[Iterable[(K, Long)]] = { (actual: Iterable[(K, Long)]) =>
+    actual.size shouldBe expected.size
+    (actual zip expected)
+    val ex = expected.toMap
+    actual.toMap.foreach { case (k, act) =>
```
nevillelyh, line 105: `.toMap` seems redundant here?

nevillelyh (commented):
One last thing, we already have a `hash` package. Should we move this there, since HLL is also a hash-based algorithm. The name `estimators` might cause confusion with statistical one.
https://en.wikipedia.org/wiki/Estimator

scio-extra/src/main/scala/com/spotify/scio/extra/hll/sketching/package.scala
syodage: yeap, good idea 👍

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/package.scala
syodage: do you have an example where we do this type of abstraction, I am struggling to see the complete picture here and how to use implicit impls. 🤔

scio-extra/src/test/scala/com/spotify/scio/extra/hll/zetasketch/ZetasketchHllIntCounterTest.scala
syodage: nope, just explicitly adding. will remove it.

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/package.scala
nevillelyh: You can unify the 4 case cases like this, basically abstract away the `HllCount.Init.for*()` method calls, because everything else is the same. This also allows us to support other types without adding a new case class.

```
case class ZetaSketchHLLCounter[T](...)(implicit zs: ZetaSketchable[T]) {
  def estimateDistinctCount(in: ...) = in
    .sth
    .applyTransform(zs.init)
    .sth
}

sealed trait ZetaSketchable[T] {
  def init: ...
}

object ZetaSketchable {
  implicit val intZetaSketchable = new ZetaSketchable[T] {
    def init: ... = HllCount.Init.forInts()
  }
}
```

And later in the follow up PR we can add:
```
sealed trait ZetaSketchable[T] {
  def decode(ba: Array[Byte]): ZetaHLL[T]
  def encode(zh: ZetaHLL[T]): Array[Byte]
}

sealed trait ZetaHLL[T] {
  def estimateSize: Long = ...
  def merge(that: ZetaHLL[T]): ZetaHLL[T]
}
```

That allows as to have methods like `SCollection[T]#asZetaHLL: SCollection[ZetaHLL[T]]` similar to the `ApproxFilter` abstractions.

scio-test/src/main/scala/com/spotify/scio/testing/ApproximationAssertions.scala
syodage: yes, it is a percentage here (that's why named it `errorRate`), ZetaSketch talks about it as a relative error (percentage) , https://github.com/google/zetasketch , that's what made me use a percentage instead of absolute error.  BigQuerry hll also talk in relative error with CI https://cloud.google.com/bigquery/docs/reference/standard-sql/hll_functions

because of those reasons, I would say let's use it as a percentage? (in Algebird case, maybe it is a bug, since they named it errorPct?)

nevillelyh: Yeah more curious about the inconsistency really. Just name it `errorPct` and update doc to reflect it's %? Clearer than `rate`.

syodage: Thinking this again, since this should be generic to all (approximation) cases not just for this case. In this approximation domain, what is the common thing, percentage, or epsilon?

PR #3367, opened 2020-10-01
Add custom GenericJson pretty print

### LINKED ISSUE #3355: containsInAnyOrder matcher failing with bigquery.TableRow
I'm facing a weird issue with the containsInAnyOrder matcher and not sure how to fix it. 

Here's a screenshot of the error output when I run my tests:
Screen Shot 2020-09-25 at 1 32 14 PM

This is the test 
```
      .output(CustomIO[STableRow]("Write Event To BigQuery"))( events => {
        events.map(f => f.sDataType) should containInAnyOrder (Seq[STableRow](OUTPUT_VALID_EVENTS).map(f => f.sDataType))
        events.map(f => f.location) should containInAnyOrder (Seq[STableRow](OUTPUT_VALID_EVENTS).map(f => f.location))
        events.map(f => f.tableRow) should containInAnyOrder (Seq[STableRow](OUTPUT_VALID_EVENTS).map(f => f.tableRow))
      })
```

for some reason, the line comparing the tableRows is failing despite output that shows they are the same. The tableRow field is of type `com.spotify.scio.bigquery.TableRow`

scio version 0.9.0
scala version 2.12.12

regadas:
Since we use a custom Printer in our `Matchers` we can add a custom handler for `GenericJson` where instead of the default call to `toString` we can force it to create a proper JSON string.

Fixes #3355

```
[info]   java.lang.AssertionError: apply@{Matchers.scala:6770}:2/ParMultiDo(Anonymous).output:
[info] Expected: Expected: iterable with items [
[info]          {"booleanField":"true","intField":1,"stringField":"someString","longField":1,"doubleField":1.0,"floa
tField":1.0}
[info] ]
[info]      but: not matched: <{"booleanField":true,"intField":1,"stringField":"someString","longField":1,"doubleFie
ld":1.0,"floatField":1.0}>
```

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/Pretty.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/Pretty.scala
@@ -22,6 +22,8 @@ import scala.jdk.CollectionConverters._
 import com.spotify.scio.{registerSysProps, SysProp}
 import scala.util.Try
 import pprint.PPrinter
+import com.google.api.client.json.GenericJson
+import org.apache.beam.sdk.extensions.gcp.util.Transport
 
 @registerSysProps
 object PrettySysProps {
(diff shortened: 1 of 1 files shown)


nevillelyh (approved):
So it'll use JSON strings for equality? There might be types with same string repr that can fail in runtime but guess there's nothing we can do in tests? :shrug:

PR #3371, opened 2020-10-02
Use TupleCoder instead of PairCoder

regadas:


nevillelyh:
Since we're talking about merging 0.10 branch back to master, should we just keep this and close the other PR?

PR #3375, opened 2020-10-05
Make BQ inherit labels from Beam/Dataflow job

### LINKED ISSUE #3333: Scio BigQuery client could inherit labels from Beam/Dataflow job
We are using labels on each and every GCP resource for better accountability. So every Dataflow job (and every resource consumed by the job) is labeled with job name. Together with costs BigQuery export we will get nice and detailed costs report for every Dataflow job.

One missing piece of the puzzle is a BigQuery jobs spawned by Scio. What do you think about labels inheritance, e.g. labels form the job will be passed to the com.google.api.services.bigquery.model.JobConfiguration created by internal Scio BigQuery client?

BTW. I need to fill similar change for Beam BigQueryIO and its internal BigQuery client :)

jto:
fixes #3333

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -52,6 +52,7 @@ import scala.io.Source
 import scala.reflect.ClassTag
 import scala.util.control.NoStackTrace
 import scala.util.{Failure, Success, Try}
+import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions
 
 /** Runner specific context. */
 trait RunnerContext {
@@ -509,6 +510,13 @@ class ScioContext private[scio] (
     o.setScalaVersion(BuildInfo.scalaVersion)
     o.setScioVersion(BuildInfo.version)
   }
+
+  private[scio] def labels: Map[String, String] =
+    (for {
+      // Check if class is in classpath
+      _ <- Try(Class.forName("org.apache.beam.runners.dataflow.options.DataflowPipelineOptions"))
+      o <- Try(optionsAs[DataflowPipelineOptions])
+    } yield o.getLabels.asScala.toMap).getOrElse(Map.empty)
 
   private[scio] val testId: Option[String] =
     Try(optionsAs[ApplicationNameOptions]).toOption.map(_.getAppName).filter(TestUtil.isTestId)
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -68,28 +68,14 @@ private object Reads {
     flattenResults: Boolean = false
   ): SCollection[T] = {
     val bigQueryClient = client(sc)
-    if (bigQueryClient.isCacheEnabled) {
-      val read = bigQueryClient.query
-        .newQueryJob(sqlQuery, flattenResults)
-        .map { job =>
-          sc.onClose(_ => bigQueryClient.waitForJobs(job))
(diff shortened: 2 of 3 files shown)


jto:
It is not possible to write an IT test bc. the Java BQ client does not have an API to read labels. I just tested in the scala console:

```scala
import com.spotify.scio.bigquery.client.BigQuery
val bq = BigQuery.defaultInstance()
val sqlQuery = "SELECT word, word_count FROM `bigquery-public-data.samples.shakespeare` LIMIT 10"
bq.query.run(sqlQuery, labels = Map("test-label" -> "yup"))

val res6: com.google.api.services.bigquery.model.TableReference = GenericData{classInfo=[datasetId, projectId, tableId], {datasetId=scio_bigquery_staging_us, projectId=scio-playground, tableId=scio_query_20201007102841_88209221}}
```

```sh
❯ bq show -j --format=pretty scio-playground:US.scio-playground-ba02d67f-8c17-4136-8b1f-f9353af5bc16
Job scio-playground:scio-playground-ba02d67f-8c17-4136-8b1f-f9353af5bc16

+----------+---------+-----------------+----------------+---------------------+-----------------+--------------+--------------+----------------+
| Job Type |  State  |   Start Time    |    Duration    |     User Email      | Bytes Processed | Bytes Billed | Billing Tier |     Labels     |
+----------+---------+-----------------+----------------+---------------------+-----------------+--------------+--------------+----------------+
| query    | SUCCESS | 07 Oct 12:28:41 | 0:00:00.819000 | julient@spotify.com | 2650191         | 10485760     | 1            | test-label:yup |
+----------+---------+-----------------+----------------+---------------------+-----------------+--------------+--------------+----------------+
```

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -510,6 +511,13 @@ class ScioContext private[scio] (
     o.setScioVersion(BuildInfo.version)
   }
 
+  private[scio] def labels: Map[String, String] =
+    (for {
+      // Check if class is in classpath
+      _ <- Try(Class.forName("org.apache.beam.runners.dataflow.options.DataflowPipelineOptions"))
+      o <- Try(optionsAs[DataflowPipelineOptions])
```
nevillelyh, line 518: This line should be an no-op if the one above fails right? So it shouldn't cause a `ClassNotFoundException` if DF runner deps are absent?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -68,30 +68,18 @@ private object Reads {
...
-      val baseQuery = if (!flattenResults) {
-        typedRead.fromQuery(sqlQuery).withoutResultFlattening()
-      } else {
-        typedRead.fromQuery(sqlQuery)
-      }
-      val query = if (bigQueryClient.query.isLegacySql(sqlQuery, flattenResults)) {
-        baseQuery
-      } else {
-        baseQuery.usingStandardSql()
+    val labels = sc.labels
```
nevillelyh, line 71: Are we removing the cache flag here? The `else` branch is in case if the user disables cache.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
jto: Yes that's exactly what it does :)

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
jto: The BQ client already supports it.

Issue #3208, opened 2020-08-03
Add more documentation to BigQuery API


lindahamp:
Would it be possible to add some documentation to [these](https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala) types? They are exposed but lack documentation, so it's a bit hard to understand how/when to use them? Specifically thinking about `Query` and `Table`. 

Also, adding some documentation [here](https://github.com/spotify/scio/blob/master/scio-bigquery/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala#L107-L135) could be nice as well since it's currently lacking.

liufuyang:
question links are move to 
https://github.com/spotify/scio/blob/master/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala

And 
https://github.com/spotify/scio/blob/master/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala#L107-L135

(Though I am not sure which lines exactly for now)

jasmineytchen:
I would like to pick this up if it is still open.

liufuyang:
I added a few things via this PR
https://github.com/spotify/scio/pull/3376 

Do you think there is more to add? You can continue with adding extra doc if you feel there is more things can be added 👍 

I will not add more doc than the ones I already have in my PR above :)

PR #3376, opened 2020-10-05
Support $LATEST replace for BQ Table type

### LINKED ISSUE #3363: Add method to resolve latest `Table` partition
Similar to #3357

liufuyang:
closes #3363

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
@@ -65,7 +65,7 @@ final case class Query(underlying: String) extends Source {
 }
 
 /**
- * A wrapper type [[Table]] which wraps a table spec String and a table [[GTableReference]]
+ * [[Table]] abstracts the multiple ways of referencing Bigquery tables. Tables can be referenced by a table spec `String` or by a table reference [[GTableReference]]. 
  *
  * Example:
  * {{{
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
@@ -107,14 +107,14 @@ object Table {
     override lazy val spec: String = BigQueryHelpers.toTableSpec(ref)
     def latest(bq: BigQuery): Ref =
       Ref(Spec(spec).latest(bq).ref)
-    def latest() = latest(BigQuery.defaultInstance())
+    def latest(): Ref = latest(BigQuery.defaultInstance())
 
   }
   final case class Spec(spec: String) extends Table {
     override val ref: GTableReference = BigQueryHelpers.parseTableSpec(spec)
     def latest(bq: BigQuery): Spec =
       Spec(BigQueryPartitionUtil.latestTable(bq, spec))
-    def latest() = latest(BigQuery.defaultInstance())
+    def latest(): Spec = latest(BigQuery.defaultInstance())
   }
 }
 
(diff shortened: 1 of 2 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -68,14 +68,23 @@ sealed trait Table extends Source {
   def spec: String
 
   def ref: GTableReference
+
+  def latest(bg: BigQuery): Table
```
regadas, line 72: can we add `def latest()` as well?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -68,14 +68,23 @@ sealed trait Table extends Source {
...
     override lazy val spec: String = BigQueryHelpers.toTableSpec(ref)
+    def latest(bq: BigQuery): Ref =
+      Ref(Spec(spec).latest(bq).ref)
+
   }
   final case class Spec(spec: String) extends Table {
     override val ref: GTableReference = BigQueryHelpers.parseTableSpec(spec)
+    def latest(bq: BigQuery): Spec = {
+      val specWithLatestReplaced = BigQueryPartitionUtil.latestTable(bq, spec)
+      Spec(specWithLatestReplaced)
```
regadas, line 86:
```suggestion
      Spec(BigQueryPartitionUtil.latestTable(bq, spec))
```

regadas: it's ok to be one liner

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
regadas: Also some scaladoc as we did for `Query`?

liufuyang: Indeed, sorry I forgot that one :)

Do you know where I can see some examples how how to use this `Table` type or what is the most common way to use it? Then I can perhaps write a little scala doc here about it as I saw that
https://github.com/spotify/scio/issues/3208

regadas: You can have a look at `scio-examples` as we have some things there like https://github.com/spotify/scio/blob/4ca965e0c1d82f8aafb2628eaa92edb5b5f64337/scio-examples/src/main/scala/com/spotify/scio/examples/cookbook/BigQueryTornadoes.scala#L34

The migration guide also gives you an overview of where it's used https://spotify.github.io/scio/migrations/v0.9.0-Migration-Guide.html#bigquery.

Although for this PR is enough just to add doc around the latest.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -64,18 +64,56 @@ final case class Query(underlying: String) extends Source {
   def latest(): Query = latest(BigQuery.defaultInstance())
 }
 
+/**
+ * A wrapper type [[Table]] which wraps a table spec String and a table [[GTableReference]]
```
regadas, line 68:
```suggestion
 * [[Table]] abstracts the multiple ways of referencing Bigquery tables. Tables can be referenced by a table spec `String` or by a table reference [[GTableReference]]. 
```

regadas: @liufuyang I think this suits it better.

liufuyang: Indeed. I had the concept but not exactly sure how to phrase it. Thank you very much for the suggestion 👍

liufuyang:
Looks like there is more code unrelated to my PR having format issues. 

Isn't there some command one can run to reformat everything on master branch and push in to solve this or?

regadas (approved):
Just missing a rebase an we are good to go!

PR #3377, opened 2020-10-05
Fix master build check failure

syodage:


syodage:
now `scalafix` check is failing

nevillelyh:
Fixed both on master.

Issue #3331, opened 2020-09-21
Add Redis IO


regadas:
`Redis` is widely used for a lot of use cases. It would be nice if we could support it more seamlessly.

PR #3391, opened 2020-10-08
Add RedisMutation support for RedisIO (#3331)

stormy-ua:
This is the next PR in a series of PRs adding support for Redis.

There are several main additions/changes:

* The `RedisMutation` abstraction was added. It represents a generic mutation to be applied to Redis. `RedisWrite` IO was changed to support `RedisMutation`s directly.
* Support for both byte-array and string keys and values for commands supporting this.
* Fine grained TTLs.  It is possible to specify TTL on a mutation level as opposed to one general TTL for all mutations as it was implemented in Beam's out-of-the-box RedisIO.
* Beam's Write IO is completely replaced with Scio specific write IO supporting some new features. Both transform and doFn are implemented in Scala to simplify its integration with the rest of Scala scio Redis code.
* All Redis examples changed to use mutations rather than raw string values.

That's an example of how the new mutation based API works from a user point of view:

```scala
    sc.parallelize(
      Iterable(
        Append("key1", "1"),
        Append("key2", "3"),
        Append("key3", "3"),
        Set("key4", "4")
      )
    ).saveAsRedis(connectionOptions)
```

### DIFF
--- a/scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
+++ b/scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2020 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.redis.write
+
+import org.joda.time.Duration
+import redis.clients.jedis.Pipeline
+
+/** Represents an abstract Redis command. */
+sealed trait RedisMutation[T] extends Serializable {
+  val key: T
+  val ttl: Option[Duration]
+}
+
+// This is needed to overcome the issue with type erasure
+sealed trait StringKeyMutation extends RedisMutation[String]
+
+// This is needed to overcome the issue with type erasure
+sealed trait ByteArrayKeyMutation extends RedisMutation[Array[Byte]]
+
+/** See Redis commands documentation for the description of commands: https://redis.io/commands */
+object RedisMutation {
+
+  // Subspace of mutations having string key
+  object String {
+
+    // String mutations
+    case class Append(key: String, value: String, ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    case class Set(key: String, value: String, ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    case class IncrBy(key: String, value: Long, ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    case class DecrBy(key: String, value: Long, ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    // Set mutations
+    case class SAdd(key: String, value: Seq[String], ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    // List mutations
+    case class RPush(key: String, value: Seq[String], ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    case class LPush(key: String, value: Seq[String], ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+    // HyperLogLog mutations
+    case class PFAdd(key: String, value: Seq[String], ttl: Option[Duration] = None)
+        extends StringKeyMutation
+
+  }
+
+  // Subspace of mutations having byte-array key
+  object ByteArray {
+
+    // String mutations
+    case class Append(key: Array[Byte], value: Array[Byte], ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    case class Set(key: Array[Byte], value: Array[Byte], ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    case class IncrBy(key: Array[Byte], value: Long, ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    case class DecrBy(key: Array[Byte], value: Long, ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    // Set mutations
+    case class SAdd(key: Array[Byte], value: Seq[Array[Byte]], ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    // List mutations
+    case class RPush(key: Array[Byte], value: Seq[Array[Byte]], ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    case class LPush(key: Array[Byte], value: Seq[Array[Byte]], ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+    // HyperLogLog mutations
+    case class PFAdd(key: Array[Byte], value: Seq[Array[Byte]], ttl: Option[Duration] = None)
+        extends ByteArrayKeyMutation
+
+  }
+
+}
+
+object RedisMutator {
+
+  import RedisMutation._
+
+  /**
+   * Applies a mutation to the Redis pipeline.
+   * Sets ttl if needed.
+   */
+  def apply(mutation: RedisMutation[_], pipeline: Pipeline): Unit = {
+    mutation match {
+      // Append
+      case String.Append(key, value, _)    => pipeline.append(key, value)
+      case ByteArray.Append(key, value, _) => pipeline.append(key, value)
+      // Set
+      case String.Set(key, value, _)    => pipeline.set(key, value)
+      case ByteArray.Set(key, value, _) => pipeline.set(key, value)
+      // IncrBy
+      case String.IncrBy(key, value, _)    => pipeline.incrBy(key, value)
+      case ByteArray.IncrBy(key, value, _) => pipeline.incrBy(key, value)
+      // DecrBy
+      case String.DecrBy(key, value, _)    => pipeline.decrBy(key, value)
+      case ByteArray.DecrBy(key, value, _) => pipeline.decrBy(key, value)
+      // LPush
+      case String.LPush(key, values, _)    => pipeline.lpush(key, values: _*)
+      case ByteArray.LPush(key, values, _) => pipeline.lpush(key, values: _*)
+      // RPush
+      case String.RPush(key, values, _)    => pipeline.rpush(key, values: _*)
+      case ByteArray.RPush(key, values, _) => pipeline.rpush(key, values: _*)
+      // SAdd
+      case String.SAdd(key, value, _)    => pipeline.sadd(key, value: _*)
+      case ByteArray.SAdd(key, value, _) => pipeline.sadd(key, value: _*)
+      // PFAdd
+      case String.PFAdd(key, value, _)    => pipeline.pfadd(key, value: _*)
+      case ByteArray.PFAdd(key, value, _) => pipeline.pfadd(key, value: _*)
+    }
+
+    mutation match {
+      case mutation: StringKeyMutation =>
+        mutation.ttl.foreach(expireTime => pipeline.pexpire(mutation.key, expireTime.getMillis))
+      case mutation: ByteArrayKeyMutation =>
+        mutation.ttl.foreach(expireTime => pipeline.pexpire(mutation.key, expireTime.getMillis))
+    }
+
+  }
+
+}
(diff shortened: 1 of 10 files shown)


scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
```diff
@@ -0,0 +1,169 @@
...
+
+package com.spotify.scio.redis.write
+
+import redis.clients.jedis.Pipeline
+
+/** Represents an abstract Redis command. */
+sealed trait RedisMutation extends Product with Serializable
+
+/** See Redis commands documentation for the description of commands: https://redis.io/commands */
+object RedisMutation {
```
regadas, line 26: hm, not really sure about this 🤔 

I believe smth like this works a little bit better? wdyt?

```scala
sealed abstract class RedisType[T]

object RedisType {
  implicit case object StringRedisType extends RedisType[String]
  implicit case object ByteArrayRedisType extends RedisType[Array[Byte]]
}

sealed abstract class RedisMutation extends Serializable
final case class Append[T: RedisType](key: T, value: T, ttl: Option[Long] = None)
    extends RedisMutation
final case class SAdd[T: RedisType](key: T, value: Seq[T], ttl: Option[Long] = None)
    extends RedisMutation

sealed abstract class RedisMutator[T] {
  def mutate(client: Pipeline, mutation: T): Unit
}

object RedisMutator {

  implicit val appendString: RedisMutator[Append[String]] = new RedisMutator[Append[String]] {
    override def mutate(client: Pipeline, mutation: Append[String]): Unit = {
      client.append(mutation.key, mutation.value)
      mutation.ttl.foreach(client.pexpire(mutation.key, _))
    }
  }

  implicit val appendByteArray: RedisMutator[Append[Array[Byte]]] =
    new RedisMutator[Append[Array[Byte]]] {
      override def mutate(client: Pipeline, mutation: Append[Array[Byte]]): Unit = {
        client.append(mutation.key, mutation.value)
        mutation.ttl.foreach(client.pexpire(mutation.key, _))
      }
    }

  implicit val saddByteArray: RedisMutator[SAdd[String]] =
    new RedisMutator[SAdd[String]] {
      override def mutate(client: Pipeline, mutation: SAdd[String]): Unit = {
        client.sadd(mutation.key, mutation.value: _*)
        mutation.ttl.foreach(client.pexpire(mutation.key, _))
      }
    }

  def mutate[T: RedisMutator](client: Pipeline)(value: T): Unit =
    implicitly[RedisMutator[T]].mutate(client, value)
}
```

scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
```diff
@@ -0,0 +1,169 @@
...
+
+  // Subspace of mutations having string key
+  object String {
```
regadas, line 29: I don't think this is necessary.

scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
regadas: Also, it would be nice if `ttl` is expressed as `Duration` as well.

stormy-ua: Yeah, that's what I was trying to do at first. It would work well if not for the nested `Seq[T]` inside mutation case classes e.g. 

```scala
final case class SAdd[T: RedisType](key: T, value: Seq[T], ttl: Option[Long] = None)
    extends RedisMutation
```

inside a function which is actually processing the mutation we would need to pattern match on it:
```scala
mutation match {
case SAdd(key: String, value: Seq[String], ttl) =>
    //process the command
}
```
This is where things break. Scala compiler can't match on `value: Seq[String]` because of the type erasure. Effectively, the compiler is seeing only Seq[Any]. `ClassTag` doesn't help in this case, because `T` is nested into the `Seq`. In theory, `TypeTag` might help, but there might be some performance issue with this.

stormy-ua: I think initially I had ttl defined as `Option[Duration]`, but looks like there is no default `Coder` for the `Duration` type, so I replaced it with `Long`. I think it is still possible to make it `Duration` by defining a custom `Coder`. I can try to do this.

stormy-ua: Actually, I think you are proposing a slightly different approach from I tried at first. Let me check if it will work.

stormy-ua: Looks like with your approach there is a problem with stitching `ObjectMutator.mutate[T: RedisMutator]` and `RedisWriteFn extends DoFn[RedisMutation, Void]`. The latter doesn't know anything about the nested type of a mutation. We can parameterize `RedisWriteFn` with a type of mutation e.g.g `RedisWriteFn[A, T <: RedisMutation[A] : RedisMutator]` or just require a `RedisMutator` for every element in a collection e.g. `RedisWriteFn[T <: RedisMutator]`. It won't be possible to mix mutations with such approach, because `RedisMutator` is specific to a mutation type and it wouldn't work for a mixed SCollection. I will try to play with it more and report back.

stormy-ua:
@regadas I have made several changed based on the review feedback:

1. I settled on a hybrid approach for `RedisMutation`/`RedisMutator`. `RedisMutation` is type parameterized by a type of a key (string or byte array) and `RedisMutator` is implemented for `RedisMutation[String]` and `RedisMutation[Array[Byte]]` only. All specific mutations handling is pushed down into these mutator implementations and pattern matching is used to match on a specific mutation. `RedisMutator` is also contravariant so mutations could be mixed in a single SCollection:

```scala
sealed trait RedisMutation[T] extends Product with Serializable {
  val key: T
  val ttl: Option[Duration]
}

final case class Append[T: RedisType](key: T, value: T, ttl: Option[Duration] = None)
    extends RedisMutation[T]
final case class SAdd[T: RedisType](key: T, value: Seq[T], ttl: Option[Duration] = None)
    extends RedisMutation[T]

object RedisMutator {

  implicit val stringRedisMutator: RedisMutator[RedisMutation[String]] =
    new RedisMutator[RedisMutation[String]] {
      override def mutate(client: Pipeline, mutation: RedisMutation[String]): Unit = {
        mutation match {
          case Append(key, value, _) => client.append(key, value)
          case SAdd(key, value, _)   => client.sadd(key, value: _*)
        }

        mutation.ttl.foreach(expireTime => client.pexpire(mutation.key, expireTime.getMillis))
      }
    }

  implicit val byteArrayRedisMutator: RedisMutator[RedisMutation[Array[Byte]]] =
    new RedisMutator[RedisMutation[Array[Byte]]] {
      override def mutate(client: Pipeline, mutation: RedisMutation[Array[Byte]]): Unit = {
        mutation match {
          case Append(key, value, _) => client.append(key, value)
          case SAdd(key, value, _)   => client.sadd(key, value: _*)
        }

        mutation.ttl.foreach(expireTime => client.pexpire(mutation.key, expireTime.getMillis))
      }
    }

  def mutate[T: RedisMutator](client: Pipeline)(value: T): Unit =
    implicitly[RedisMutator[T]].mutate(client, value)

}
```

2. TTL type changed to Joda Duration. I had to add a new `Coder` to `JavaCoders`. Its implementation is trivial and usese `Coder.xmap`:

```scala
trait JodaCoders {
  implicit def jodaDurationCoder: Coder[Duration] =
    Coder.xmap(Coder[Long])(Duration.millis, _.getMillis)
}
```

Let me know what you think! it is definitely got better in compare to the previous revision.

scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
regadas: Ah I see. My previous example missed the ability to mix `RedisMutation` in the same `SCollection`. However I think we can get there with a few tweaks:

```scala

// let's make this change to `RedisMutation`
sealed abstract class RedisMutation extends Serializable {
  def rt: RedisType[_]
}
// this is just syntax sugar for the implicit added below
object RedisMutation {
  def unapply[T <: RedisMutation](mt: T): Option[(T, RedisType[_])] = Some(mt -> mt.rt)
}

// let's exposed the resolved implicit `RedisType`
final case class Append[T](key: T, value: T, ttl: Option[Long] = None)(implicit
  val rt: RedisType[T]
) extends RedisMutation

// add this new implicit under `RedisMutator`
implicit def redisMutation[T <: RedisMutation]: RedisMutator[T] = new RedisMutator[T] {
  override def mutate(client: Pipeline, mutation: T): Unit = {
    mutation match {
      case RedisMutation(mt: Append[String @unchecked], RedisType.StringRedisType) =>
        RedisMutator.mutate(client)(mt)
      case RedisMutation(mt: Append[Array[Byte] @unchecked], RedisType.ByteArrayRedisType) =>
        RedisMutator.mutate(client)(mt)
      case RedisMutation(mt: SAdd[Array[Byte] @unchecked], RedisType.ByteArrayRedisType) =>
        RedisMutator.mutate(client)(mt)
    }
  }
}

// and make `T <: RedisMutation`
def mutate[T <: RedisMutation: RedisMutator](client: Pipeline)(value: T): Unit =
  implicitly[RedisMutator[T]].mutate(client, value)

// We also need to make
// RedisWriteTransform[T <: RedisMutation: RedisMutator]
// RedisWriteFn[T <: RedisMutation: RedisMutator]
// SCollectionRedisOps[T <: RedisMutation: RedisMutator]
```

The main reason to bubble up `RedisMutator` is performance. If users just have `SCollection[Append[String]]` the resolved mutator will be the most specific (Per my previous example would be `RedisMutator.appendString`) and will avoid the pattern match. If users use `SCollection[RedisMutation]` the generalised one will be used.

regadas (commented):
Yes this one looks a lot better! 

Left this comment https://github.com/spotify/scio/pull/3391#discussion_r504264121 to try to address the mix of `RedisMutation`

scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
stormy-ua: @regadas Well, looks like this worked out and now it pulls specific mutators for homogeneous collections or falls back to the generic mutator for mixed collections. There is still one limitation though. It isn't possible to mix string and byte array mutations, because the Coder can't be inferred e.g.:

```scala
    sc.parallelize(
      Iterable(
        Append("key1", "1"),
        Append("key2".getBytes, "2".getBytes)
      )
    ).saveAsRedis(connectionOptions)
``` 

is failing because Coder couldn't be generated for `com.spotify.scio.redis.write.Append[_ >: Array[Byte] with String <: java.io.Serializable]`. I am not sure whether it would be easy to find a workaround for this issue.

regadas: @stormy-ua for this we need to come up with our own coder:

```scala
trait CoderInstances {
  implicit def appendCoder[T: Coder: RedisType] = Coder.gen[Append[T]]
  implicit def saddCoder[T: Coder: RedisType] = Coder.gen[SAdd[T]]

  private[this] def coders: Map[Int, Coder[_]] = Map(
    1 -> appendCoder[String],
    2 -> appendCoder[Array[Byte]],
    3 -> saddCoder[String]
  )

  implicit def redisMutationCoder[T <: RedisMutation] =
    Coder.disjunction[T, Int]("RedisMutation", coders.asInstanceOf[Map[Int, Coder[T]]]) {
      case RedisMutation(_: Append[String @unchecked], RedisType.StringRedisType)         => 1
      case RedisMutation(_: Append[Array[Byte] @unchecked], RedisType.ByteArrayRedisType) => 2
      case RedisMutation(_: SAdd[String @unchecked], RedisType.StringRedisType)           => 3
    }
}
```
This coders can be added following this [approach].(https://github.com/spotify/scio/blob/4466eb228ba838cf2d9acaed17bc030e1bd6a13f/scio-google-cloud-platform/src/main/scala/com/spotify/scio/pubsub/coders/package.scala#L22)

`import com.spotify.scio.redis.coders._` is now recommended.

Also we can just have `sealed abstract class RedisMutation` and drop the type param.

regadas (commented):
Hi @stormy-ua I think we are able to address the coder issue as well with https://github.com/spotify/scio/pull/3391#discussion_r504658861

scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
stormy-ua: Got it, will try to add the custom coder.

stormy-ua: @regadas The custom coder solved the issue with mixed collections. Just pushed the commit. I am wondering a disjunction coder creation could be automated. It seems redundant to manually assign IDs to coders :) Also dropped type param from `RedisMutation.`

regadas: Right now deriving sum types with type params is limited so we need to stick with this for now.

regadas (approved):
@stormy-ua Thank you for this! 🎉

PR #3392, opened 2020-10-09
Integrate docs in CI

### LINKED ISSUE #3151: Have CircleCI run sbt mdoc + paradox when anything in "/site" folder is changed in a PR
Changes to any of the markdown files in `/site` could cause `sbt site/mdoc` to fail if a code snippet is incorrect, or `sbt site/paradox` to fail if a `@ref` link is broken or something. Maybe we could the CircleCI "when" clause: https://circleci.com/docs/2.0/configuration-reference/#the-when-step-requires-version-21 to execute these commands if any site files are in the git diff?

It would also be nice to have CircleCI push the site to GH pages when the PR is merged, but the credentialing might be an issue

RustedBones:
Fix #3151


site/src/main/paradox/FAQ.md
```diff
@@ -253,6 +253,7 @@ This can be done by using fixed windows and using the window bounds to infer dat
 
 ```scala mdoc:reset:silent
 import com.spotify.scio._
+import com.spotify.scio.pubsub._
```
RustedBones, line 256: Caught by [build](https://app.circleci.com/pipelines/github/spotify/scio/5198/workflows/2a4d2618-d996-4f40-949b-ec85143e2330/jobs/51399)

regadas (commented):
@RustedBones Thanks for this!

When we talked I thought we were going to start using GHA for building the `site`.

Also what's the reason to change to `Compile` scope? `Paradox` scope works just fine.

.circleci/config.yml
```diff
@@ -106,14 +106,9 @@ jobs:
     steps:
       - build:
           build-steps:
-            - run:
```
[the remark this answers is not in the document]
RustedBones: Just making all run definition consistent. Some were defined with `command` some weren't

regadas:
So what I was thinking was:

1) start using GHA to build the website; action would be triggered on PR and master.
2) (Follow up PR) on tag publish we deploy the website

Sorry if I wasn't clear eralier 😞 Let me know wdyt.

RustedBones (commented):
I dropped the `Paradox` config because of the update to [latest](https://www.scala-sbt.org/sbt-site/migration-guide.html) `sbt-site`. 

The usage of `Paradox` config was a mistake on their side. Now it integrates way better with the defaults of `paradox`.

regadas:
Ahh ok I missed that update! Then 👍 There are a lof of other duplicated settings though

.circleci/config.yml
```diff
@@ -127,13 +122,12 @@ jobs:
...
-                  sbt -v $SBT_OPTS \
-                  coverage \
-                  "++$SCALA_VERSION test" \
-                  "++$SCALA_VERSION coverageAggregate"
+            - run: sbt -v $SBT_OPTS
+                coverage
+                "++$SCALA_VERSION test"
+                "++$SCALA_VERSION coverageAggregate"
             - run: ./scripts/circleci_repl.sh
+            - run: ./scripts/circleci_docs.sh <<pipeline.git.base_revision>> <<pipeline.git.revision>>
```
regadas, line 130: We would like to build it against 2.13 as well

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,47 @@
+name: Deploy
+
+on:
+  push:
+    tags: [ v.* ]
+
+jobs:
+  build:
+    name: Publish documentation
+      runs-on: ubuntu-latest
```
regadas, line 10: 🎨 bad ident

regadas (commented):
Can't wait to have this in 😄

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  pull_request:
+    branches: [ master ]
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: Build documentation
+    if: ${{ secrets.ENCRYPTION_KEY }}
```
regadas, line 12: Ah it seems that we can't do conditionals on secrets... One way to do this is to move the the `env` L32 to the build workflow and then check the `env` var

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    strategy:
+      matrix:
+        scala: [ 2.13.3, 2.12.12 ]
+        java: [ openjdk@1.8, openjdk@1.11 ]
```
regadas, line 16: since this is just docs we can just run against the default scalaVersion and just use jdk 11

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,39 @@
+name: Deploy
+
+on:
+  push:
+    tags: [ v.* ]
+
+jobs:
+  build:
+    name: Publish documentation
+    if: ${{ secrets.ENCRYPTION_KEY }}
```
regadas, line 10: since this is push to a tag secrets will be available so no need to check.

regadas (approved):
@RustedBones Thank you for this! Gonna give it a spin let's see what happens!

regadas:
LMAO well some hiccup I'll give it a nudge

RustedBones:
Thanks for fixing the last bits :)

PR #3401, opened 2020-10-15
Propagate coder and simplify Bigquery Format type

### LINKED ISSUE #3399: Bigquery avro `saveAsBigQueryTable` fails if table doesn't yet exist
This is due to the change introduced in #3385.

https://github.com/spotify/scio/blob/d41513925f5a18e9dd9d686db3e07e62bc3faf88/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L263-L266

regadas:
Right now this is only needed for `Format.GenericRecord`. 

For performance reasons users are recommended to supply their own `Coder` through:

```scala
val schema: org.apache.avro.Schema = ???
implicit val coder = Coder.avroGenericRecordCoder(schema)
```

Fixes #3399

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
@@ -202,16 +202,10 @@ object BigQuerySelect {
 object BigQueryTypedTable {
 
   /** Defines the format in which BigQuery can be read and written to. */
-  sealed abstract class Format {
-    type F
-  }
+  sealed abstract class Format[F]
   object Format {
-    case object GenericRecord extends Format {
-      override type F = org.apache.avro.generic.GenericRecord
-    }
-    case object TableRow extends Format {
-      override type F = com.google.api.services.bigquery.model.TableRow
-    }
+    case object GenericRecord extends Format[GenericRecord]
+    case object TableRow extends Format[TableRow]
   }
 
   trait WriteParam {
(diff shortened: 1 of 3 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -275,10 +271,11 @@ object BigQueryTypedTable {
    * Writting: Supports LogicalTypes only for DATE and TIME.
    *           DATETIME is not yet supported. https://issuetracker.google.com/issues/140681683
    */
-  def apply(table: Table, format: Format): BigQueryTypedTable[format.F] = format match {
-    case Format.GenericRecord => genericRecord(table).asInstanceOf[BigQueryTypedTable[format.F]]
-    case Format.TableRow      => tableRow(table).asInstanceOf[BigQueryTypedTable[format.F]]
-  }
+  def apply[F: Coder](table: Table, format: Format[F]): BigQueryTypedTable[F] =
```
nevillelyh, line 274: Is `format` a required argument now? Do we want to make it an implicit and provide a default, e.g. `TableRow` and later switch to Avro?

regadas: It's only required if users choose to use the IO directly coz otherwise:
https://github.com/spotify/scio/blob/b78dd27ea8caf196cdc7d03dea4862fc4ce92bf9/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala#L73-L87

imo I don't think we would gain by having it implicitly declared.

nevillelyh (approved):
Good to merge?

regadas:
I'm having second thoughts around the BQ API. I think we can generalise it a bit more. I'll propose something soon. I'll merge this in the meantime since it's inline with `0.9.5` and it fixes an issue.

PR #3402, opened 2020-10-15
Add Predicate for SMB source, fix #3398

### LINKED ISSUE #3398: Predicate pushdown for SMB sources
In the form of a `(List[T], T) => Boolean` so that we can stop adding elements once a key group reached certain limit (e.g. hot key) or bad elements.

nevillelyh:
WIP need to update Scala binding once we're happy with the approach


nevillelyh:
@ClaireMcGinty thoughts? Wanna have another pair of :eyes: before proceeding.

nevillelyh:
Turns out we don't need Scala binding since `SortedBucketSource.Read` Java API is used directly in Scala and implicit conversion from `(T, U) => R` to `BiFunction[T, U, R]` works.

@ClaireMcGinty PTAL

PR #3406, opened 2020-10-16
Sanitize staged file path string

### LINKED ISSUE #3393: Scio DataflowRunner cannot be launched from Windows platform
The CL linked at the bottom does not account for the \ characters found in Windows user home directories. On my machine, this leads to the following regex pattern C:\Users\info/\..+/.+ which is an error due to \U and \i.

https://github.com/spotify/scio/pull/3322/commits/af1395c7c9b7e47396ddd5e64e9da8b1265b4275

regadas:
Fixes #3393

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -136,11 +136,13 @@ private object RunnerContext {
     pipelineOptions: PipelineOptions,
     classLoader: ClassLoader
   ): Iterable[String] = {
-    val matchesEnvDir: String => Boolean = _.matches(s"${sys.props("user.home")}/\\..+/.+")
+    val sanitize: String => String = _.replace("\\", "/")
+    val matchesEnvDir: String => Boolean =
+      _.matches(s"${sanitize(sys.props("user.home"))}/\\..+/.+")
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -136,11 +136,13 @@ private object RunnerContext {
     pipelineOptions: PipelineOptions,
     classLoader: ClassLoader
   ): Iterable[String] = {
-    val matchesEnvDir: String => Boolean = _.matches(s"${sys.props("user.home")}/\\..+/.+")
+    val sanitize: String => String = _.replace("\\", "/")
+    val matchesEnvDir: String => Boolean =
+      _.matches(s"${sanitize(sys.props("user.home"))}/\\..+/.+")
```
syodage, line 141: since L145 sanitize before call matcheEvnDir, is sanitize redundant here?

nevillelyh: Whoops fat fingered & merged. Feel free to amend later.

regadas: No, that it's not redundant. That is sanitising the `sys.prop`

PR #3416, opened 2020-10-20
Redis lookup Do Fn #3331

stormy-ua:
Adding Lookup Do Fn for Redis. It just simply extends `GuavaAsyncLookupDoFn`. The client in this case is an instance of [Jedis client](https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/Jedis.java). 

There is one issue with the client which complicates its integration with the `BaseAsyncLookupDoFn`: Jedis client isn't thread safe. It has to be be created and used in the same thread. `BaseAsyncLookupDoFn` allocates clients per lookup fn instance and even though it isn't possible for multiple threads to access the client simultaneously, it is possible that different threads will access it on after another because Beam/Dataflow might re-cycle a thread after a bundle is processed (this is based on my observations). I have done some testing both on my local and on DataFlow and was able to re-produce the issue. To mitigate the issue I am wrapping the client into `ThreadLocal` so that every thread gets its own copy of the Jedis client. Jedis client allocation should be relatively lightweight operation. My main concern is that it isn't possible to explicitly close the client after a bundle is processed or/and after the thread is getting re-cycled. A better solution would be to add support for per-bundle clients to the `BaseAsyncLookupDoFn`. This would allow to properly create and dispose clients on each bundle and I believe Beam/DataFlow uses a single thread to process a single bundle (but I am not sure here).

I tested this implementation with a batch and a streaming job and seems to work fine. We can use this as a P0 implementation, but re-visit it if there will be resource management issues.

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/RedisExamples.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/RedisExamples.scala
@@ -18,6 +18,7 @@
 
 package com.spotify.scio.examples.extra
 
+import com.google.common.util.concurrent.{Futures, ListenableFuture}
 import com.spotify.scio.{ContextAndArgs, ScioContext}
 import com.spotify.scio.redis._
 import org.apache.beam.examples.common.ExampleUtils
@@ -25,6 +26,8 @@ import org.apache.beam.sdk.options.{PipelineOptions, StreamingOptions}
 import com.spotify.scio.pubsub._
 import com.spotify.scio.redis.write._
 import com.spotify.scio.redis.coders._
+import com.spotify.scio.redis.lookup.RedisLookupDoFn
+import redis.clients.jedis.Jedis
 
 // ## Redis Read Strings example
 // Read strings from Redis by a key pattern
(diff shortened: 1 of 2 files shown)


scio-examples/src/main/scala/com/spotify/scio/examples/extra/RedisExamples.scala
```diff
@@ -57,6 +60,43 @@ object RedisReadStringsExample {
...
+// `sbt "runMain com.spotify.scio.examples.extra.RedisLookUpStringsExample
+// --project=[PROJECT] --runner=DataflowRunner --zone=[ZONE]
+// --redisHost=[REDIS_HOST]
+// --redisPort=[REDIS_PORT]
+object RedisLookUpStringsExample {
+
+  private def lookupFn(connectionOptions: RedisConnectionOptions) =
+    new RedisLookupDoFn[String, Option[String]](connectionOptions) {
+      override def asyncLookup(
+        client: ThreadLocal[Jedis],
```
regadas, line 76: Hi @stormy-ua what about using `JedisPool`? It seems that it's intended for multi-threaded envs.

Also stumbled upon this https://github.com/lettuce-io/lettuce-core the other day and it seems that it would be a better fit here since it supports async calls.

We can start with `Jedis` and at some point switch to `lettuce`

stormy-ua: I can look into the `JedisPool`. The issue with the `JedisPool` is that a `JedisClient` must be explicitly returned to the pool after it is no longer needed. It isn't possible to implement this with the `BaseAsyncLookupDoFn`. I can look into how the `BaseAsyncLookupDoFn` could be changed or extended to support an explicit resource cleanup handling. 

As for `lettuce`, it looks very interesting and I think this might be a better way to go forward. I sticked to `jedis`, because this is what the original Beam Redis IO used, but we can easily replace it.

regadas: Hi @stormy-ua I was thinking about this a little bit and I think we can have an alternative take on this. We don't really need to make this an `async` lookup... we can still leverage `Jedis` and it's `Pipeline` for batch requests. The idea here is to generalize a little bit more `RedisWriteFn` into a `RedisFn` that we can use for lookups as well. I'll put up smth soon so we can see how it would look like and discuss.

stormy-ua: @regadas Sounds good! This would be the best option if we can just use `Jedis`/`Pipeline`.

regadas:
@stormy-ua I think we can close this one in favor of #3461

stormy-ua:
@regadas Yes, let's kill it.

PR #3432, opened 2020-10-26
add scalafix rules for 0.10

nevillelyh:


nevillelyh:
@regadas there are no API/import changes for `scio-google-cloud-platform` right? Only `build.sbt` changes? Those can't be fixed with rules?

regadas (approved):
There are mode changes I believe and yeah there are some import changes. We can start with there ones though.

PR #3435, opened 2020-10-27
Clarify extra.csv.CsvIO docs

pgoggijr:
- Clarify Encoder to use when reading headerless file (`HeaderEncoder.ordered` doesn't exist, `RowDecoder.ordered` is used in the linked docs)
- Add explicit configuration parameter that ignores the header.  If not included, kantan.csv will silently skip the first row of the csv file

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
@@ -61,8 +61,11 @@ import org.apache.beam.sdk.values.PCollection
  * [[https://nrinaudo.github.io/kantan.csv/rows_as_arbitrary_types.html Kantan docs]]
  * {{{
  *  case class User(name: String, age: Int)
- *  implicit val decoder = HeaderDecoder.ordered { (name: String, age: Int) => User(name, age) }
- *  val users: SCollection[User] = scioContext.csvFile(path)
+ *  implicit val decoder = RowDecoder.ordered { (name: String, age: Int) => User(name, age) }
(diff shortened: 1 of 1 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
```diff
@@ -61,8 +61,11 @@ import org.apache.beam.sdk.values.PCollection
  * [[https://nrinaudo.github.io/kantan.csv/rows_as_arbitrary_types.html Kantan docs]]
  * {{{
  *  case class User(name: String, age: Int)
- *  implicit val decoder = HeaderDecoder.ordered { (name: String, age: Int) => User(name, age) }
- *  val users: SCollection[User] = scioContext.csvFile(path)
+ *  implicit val decoder = RowDecoder.ordered { (name: String, age: Int) => User(name, age) }
+ *  val csvConfiguration = CsvIO.DefaultReadParams.copy(
+ *    csvConfiguration = CsvIO.DefaultCsvConfig.copy(header = CsvConfiguration.Header.None)
+ *  )
```
regadas, line 67:
```suggestion
 *  CsvIO.ReadParam(csvConfiguration = CsvIO.DefaultCsvConfig.withoutHeader)
```

regadas: @pgoggijr Thanks for spotting this. Just added a small suggestion.

pgoggijr: np! suggestion works locally, will commit 👍

regadas (approved):

Issue #1938, opened 2019-05-27
Avoiding saveAsElasticsearch() reshuffle


mdvorsky:
saveAsElasticsearch() currently reshuffles all data into a fixed number of shards. What do you think about adding an option to avoid AssignToShard transform? For our usecase, reshuffle is adding unnecessary overhead. (We're limiting the concurrent access to ES by --maxNumWorkers which is a lot more efficient of limiting parallelism than shuffling a lot of data into small number of very hot keys).

nevillelyh:
@regadas WDYT? @mdvorsky if you think it's a trivial change, mind submitting a PR and we can discuss there?

regadas:
Yeah we can definitely disable sharding! This will require adding config to allow config of batch size.

regadas:
@alexclare will take a stab on this one. Thanks!

RustedBones:
@regadas isn't that resolved too ?

kellen:
Calling this done in #3566

Issue #1829, opened 2019-04-10
Split elasticsearch bulk requests based on its size and document size


regadas:
ES doesn't deal well when bulk requests become too big. We should be able to split the request base on its size and document size.

KieranCTaylor:
Has this issue been looked at / resolved yet? I would be interested in tackling.

regadas:
Hi @KieranCTaylor go for it!

KieranCTaylor:
Great I will jump on this week and see what I can do

KieranCTaylor:
Hey @regadas is this not satisfied by the ES write param maxBulkRequestSize?

RustedBones:
@regadas isn't this closed since we now have `maxBulkRequestBytes` setting ?

RustedBones:
On `scio-elasticsearch8`
- `maxBulkRequestOperations` for the max number of operations in a bulk request
- `maxBulkRequestBytes` for the max total size of operations in a bulk request

PR #3437, opened 2020-10-28
Add byte size limit option to ElasticsearchIO writer (#1829)

alexclare:
Adds the ability to split up bulk Elasticsearch writes by request byte size in addition to request count.

Sort of a prerequisite for #1938. Not strictly required, but that issue involves larger Java changes so I figured it would be better to get this in beforehand.


scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -152,6 +160,7 @@
     public static class Bound<T> extends PTransform<PCollection<T>, PDone> {
 
       private static final int CHUNK_SIZE = 3000;
+      private static final long CHUNK_BYTES = 5L * 1024L * 1024L;
```
regadas, line 163: Can you add a small doc line about this value?

scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -415,32 +456,47 @@ public void setup() throws Exception {
...
         final Stream<DocWriteRequest> docWriteRequests =
             StreamSupport.stream(values.spliterator(), false)
                 .map(toDocWriteRequests::apply)
                 .flatMap(ar -> StreamSupport.stream(ar.spliterator(), false));
 
-        final Iterable<List<DocWriteRequest>> chunks =
-            Iterables.partition(docWriteRequests::iterator, maxBulkRequestSize);
+        int currentSize = 0;
+        long currentBytes = 0L;
+        ImmutableList.Builder<DocWriteRequest> chunk = ImmutableList.builder();
```
regadas, line 466: I think `ImmutableList` is a bit unnecessary here.

scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -415,32 +456,47 @@ public void setup() throws Exception {
       public void processElement(ProcessContext c) throws Exception {
         final Iterable<T> values = c.element().getValue();
 
-        // Elasticsearch throws ActionRequestValidationException if bulk request is empty,
-        // so do nothing if number of actions is zero.
-        if (!values.iterator().hasNext()) {
```
regadas, line 420: we can still do an early get out here and avoid the allocations bellow, right?

regadas (commented):
Thank you for this @alexclare! just some minor comments!

scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
alexclare: Yeah, that's true. Should I leave this block in then?

regadas: yeah let's keep it.

scio-elasticsearch/es6/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
alexclare: I ended up just constructing the `BulkRequest` directly,

regadas (approved):
@alexclare Thanks for pushing this through!

PR #3440, opened 2020-10-28
Distributed HyperLogLog++ support with ZetaSketch

syodage:


scio-extra/src/test/scala/com/spotify/scio/extra/hll/zetasketch/ZetasketchHllPlusPlusTest.scala
```diff
@@ -75,4 +75,50 @@ class ZetaSketchHllPlusPlusTestTest extends PipelineSpec {
     }
     output shouldApproximate withErrorRatePerKey(expt, 0.5)
   }
+
+  "ZetaSketchHLL" should "approximate distinct count" in {
+    val input = for (i <- 0 to 10000) yield (i % 20)
+    val output = runWithData(input) { scl =>
+      import com.spotify.scio.extra.hll.zetasketch._
```
regadas, line 82:
```suggestion
```

It's not needed and that's what causing 2.12 to complain

regadas (commented):
Overall this looks good. Although there's one thing that I would like to see changed ... All of this should be moved out of the package object and that will make it aligned as well with the rest of the codebase.

regadas:
One thing interesting about this is that they are shading `ByteString` in `HyperLogLogPlusPlus` wondering why 🤔

syodage:
@regadas updated the PR with review feedback and improved scala docs with examples. 

One thing I didn't do is explicitly define `zetasketch` dependency, instead depends on it transitively from the beam's `extensions-zetasketch` dep. Is it ok or should I explicitly add it?

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/distributed/package.scala
```diff
@@ -0,0 +1,267 @@
...
+import com.google.protobuf.ByteString
+import com.google.zetasketch.HyperLogLogPlusPlus
+import com.spotify.scio.coders.Coder
+import com.spotify.scio.values.SCollection
+import com.twitter.algebird.{Monoid, MonoidAggregator}
+
+package object distributed {
+
+  // Distributed HyperLogLogPlusPlus impl support //
+  sealed trait HllPlus[T] extends Serializable {
```
regadas, line 31: @syodage Sorry if I wasn't clear in my previous comment about the structure. It's ok for `HllPlus` to be under `com.spotify.scio.extra.hll.zetasketch` only but on a different file `HllPlus.scala` and all `syntax` extensions to `SCollection` under a `com.spotify.scio.extra.hll.zetasketch.syntax` package. Check for example the `csv` package under `scio-extra`.

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/distributed/package.scala
```diff
@@ -0,0 +1,267 @@
...
+      override def convert(t: ByteString): ByteString = t
+    }
+  }
+
+  /**
+   * This is a wrapper for internal HyperLogLogPlusPlus implementation.
+   * @param arrOpt - serialized byte array to construct the initial HyperLogLogPlusPlus instance.
+   * @param elemOpt - first element to add to the constructed new HyperLogLogPlusPlus instance.
+   */
+  final class ZetaSketchHLL[T](arrOpt: Option[Array[Byte]], elemOpt: Option[T] = None)(implicit
```
regadas, line 92: Same think here .. separate file. Basically we want to avoid type definitions inside `package object`

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/distributed/package.scala
```diff
@@ -0,0 +1,267 @@
...
+      extends MonoidAggregator[T, ZetaSketchHLL[T], Long] {
+    override def monoid: Monoid[ZetaSketchHLL[T]] = ZetaSketchHLLMonoid()
+
+    override def prepare(input: T): ZetaSketchHLL[T] = ZetaSketchHLL.create[T](input)
+
+    override def present(reduction: ZetaSketchHLL[T]): Long = reduction.estimateSize()
+  }
+
+  // Syntax
+  implicit class ZetaSCollection[T](private val scol: SCollection[T]) extends AnyVal {
```
regadas, line 177: we want to move this out to `syntax` package

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/package.scala
```diff
@@ -30,7 +30,7 @@ import org.apache.beam.sdk.values.{KV, PCollection}
 
 package object zetasketch {
 
-  trait ZetaSketchable[T] {
```
regadas, line 33: Same thing should be applied to all type definitions here as well

regadas (commented):
Sorry if I wasn't clear on my previous comment.

syodage:
@regadas thank you for reviewing this again, sorry I misunderstood you before. Hope new changes to address the code consistency.

regadas (approved):

PR #3442, opened 2020-10-30
Add multiple key range read for BigTableIO

### LINKED ISSUE #3447: read multiple keyranges from one read in BigTableIO
Currently, BigTableIO only supports single keyRange lookup, we would like to request multiple key ranges lookup in the same read instead of having to do multiple reads. Something like below

def bigtable(
    bigtableOptions: BigtableOptions,
    tableId: String,
    keyRange: Seq[ByteKeyRange],
    rowFilter: RowFilter
  )

Can you add this to the BigTableIO? Thanks!

syodage:
Closes #3447

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigtable/BigTableIO.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigtable/BigTableIO.scala
@@ -65,8 +65,8 @@ final case class BigtableRead(bigtableOptions: BigtableOptions, tableId: String)
             opts.toBuilder
         }
       )
-    if (!params.keyRange.isEmpty) {
-      read = read.withKeyRanges(params.keyRange.asJava)
+    if (!params.keyRanges.isEmpty) {
+      read = read.withKeyRanges(params.keyRanges.asJava)
     }
     if (params.rowFilter != null) {
       read = read.withRowFilter(params.rowFilter)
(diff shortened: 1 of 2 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigtable/BigTableIO.scala
```diff
@@ -85,12 +85,12 @@ final case class BigtableRead(bigtableOptions: BigtableOptions, tableId: String)
 
 object BigtableRead {
   object ReadParam {
-    private[bigtable] val DefaultKeyRange: ByteKeyRange = null
+    private[bigtable] val DefaultKeyRange: Seq[ByteKeyRange] = Seq()
```
regadas, line 88:
```suggestion
    private[bigtable] val DefaultKeyRange: Seq[ByteKeyRange] = Seq.empty[ByteKeyRange]
```

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigtable/BigTableIO.scala
```diff
@@ -85,12 +85,12 @@ final case class BigtableRead(bigtableOptions: BigtableOptions, tableId: String)
...
     private[bigtable] val DefaultRowFilter: RowFilter = null
   }
 
   final case class ReadParam private (
-    keyRange: ByteKeyRange = ReadParam.DefaultKeyRange,
+    keyRange: Seq[ByteKeyRange] = ReadParam.DefaultKeyRange,
```
regadas, line 93: I think we can make `ReadParam` at least source compat righ?

regadas (approved):

PR #3448, opened 2020-11-02
Added documentation for typedBigQuery

### LINKED ISSUE #3208: Add more documentation to BigQuery API

jasmineytchen:
closes #3208 

This PR adds documentation to an existing BigQuery API (_typedBigQuery_). Documentations for `Table` and `Query` type have been added as part of #3357.

**Question** 
Is this level of documentation adequate?


nevillelyh (approved):
LGTM, thanks!

PR #3454, opened 2020-11-04
Rework `com.spotify.scio.transforms` package

regadas:
There are transforms available under the package object which could be made available i.e `safeFlatMap`


regadas:
talked with @nevillelyh and we will not make them automatically available as it may cause confusion to the users. 

I'll update this PR to reflect that.

regadas (commented):
Gonna go ahead and merge this as it's simply cosmetic 💅

PR #3456, opened 2020-11-04
Add RedisType implicit error message

regadas:


scio-redis/src/main/scala/com/spotify/scio/redis/write/RedisMutation.scala
```diff
@@ -32,6 +33,7 @@ object RedisMutation {
   def unapply[T <: RedisMutation](mt: T): Option[(T, RedisType[_])] = Some(mt -> mt.rt)
 }
 
+@implicitNotFound("${T} is not a supported type. Redis only supports String or Array[Byte]")
```
stormy-ua, line 36: Thanks for adding this!

PR #3461, opened 2020-11-06
Add RedisDoFn

regadas:
A generalized `redis` `DoFn` that can be used for writting and reading.

Screenshot 2020-11-12 at 14 41 09


scio-redis/src/main/scala/com/spotify/scio/redis/RedisIO.scala
```diff
@@ -41,17 +49,12 @@ object RedisConnectionOptions {
...
+      .withSSL(StaticValueProvider.of(connectionOptions.useSSL))
 
-    if (connectionOptions.useSSL) {
-      config = config.enableSSL()
-    }
-
-    connectionOptions.auth.foreach(a => config = config.withAuth(a))
-
-    config
+    connectionOptions.auth.fold(config)(config.withAuth(_))
```
stormy-ua, line 57: I guess `(_)` could be dropped here?

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
```diff
@@ -0,0 +1,137 @@
...
+  private val results: ConcurrentLinkedQueue[Future[Result]] = new ConcurrentLinkedQueue()
+  private val requests: ConcurrentLinkedQueue[(List[Response[_]], Promise[List[_]])] =
+    new ConcurrentLinkedQueue()
+  private var batchCount = 0
+  private val client = new Client()
+
+  private case class Result(input: I, output: O, ts: Instant, w: BoundedWindow)
+
+  abstract class Request {
+    def create(pipeline: Pipeline): List[Response[_]]
```
stormy-ua, line 53: info: could this be replaced with a pure Scala function e.g.:
```scala
type Request = Pipeline => List[Response[_]]
```

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
```diff
@@ -0,0 +1,137 @@
...
+    pipeline.exec
+    pipeline.sync()
+
+    val iter = requests.iterator()
+    while (iter.hasNext()) {
+      val (rsp, promise) = iter.next()
+      promise.success(rsp.flatMap(r => Option(r.get())))
+    }
+
+    val future = Future.sequence(results.asScala).andThen {
```
stormy-ua, line 77: I am wondering if we want to drop all results even if a single request failed. In general, I think Redis ops should be idempotent, but with the existing design it is possible to add support for mutations adding something to the end of a list and executing this twice would lead to a duplicate in a list. I guess this might be out of scope for now.

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
```diff
@@ -0,0 +1,137 @@
...
+  @StartBundle
+  def startBundle(): Unit = {
+    pipeline = jedis.pipelined
+    pipeline.multi
+    batchCount = 0
+  }
+
+  @ProcessElement
+  def processElement(c: ProcessContext, window: BoundedWindow): Unit = {
+    val result = request(c.element(), client).map { r =>
```
stormy-ua, line 107: info: so technically all Redis requests will be put into the Jedis pipeline in the current thread. The only place where it forks is in `RedisWriter.Writer` in the `map(_ => ())`. `pipeline.exec/sync` sends all requests and reads everything back in the current thread as well. The only concurrency we have here is after all Redis requests are completed. I am wondering what we gain in return by adding future/promises here. Is the main idea to have concurrency in the part processing Redis results? (Might be a dumb question - might be missing the main point here)

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
regadas: Yeah had it before like that but then thought it would be easier for users to reason about it

scio-redis/src/main/scala/com/spotify/scio/redis/RedisIO.scala
regadas: it's ok not to do eta expansion as it will be removed in the future

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
regadas: correct, I'm planing further changes to address this but they are out of scope for this PR

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
regadas: The reason to use `Future`/`Promise` is merely to convey to the user that the result from the `Request` function is not computed right away as it is included in the pipeline bundle. Without it users would be able to try getting a value out of a response that was not yet obtained.

stormy-ua: Got it, this makes sense. Looks like clients still can try to access result before it was processed:

```scala
new RedisDoFn[T, Unit](connectionConfig, batchSize) {
        override def request(value: T, client: Client): Future[Unit] =
          client
            .request { pipeline =>
                val resps = RedisMutator.mutate(pipeline)(value)
                // do something with resps...
                resps
             }
            .map(_ => ())
      }
```

But probably it is difficult to put a guard against this.

scio-redis/src/main/scala/com/spotify/scio/redis/RedisIO.scala
stormy-ua: 😢

scio-redis/src/main/scala/com/spotify/scio/redis/RedisDoFn.scala
regadas: Yeah they can, but that's fine as they will not be able to do much.

PR #3465, opened 2020-11-10
Moving BQ Partitioned Table IO ops to BigQueryIO

### LINKED ISSUE #3372: Simplify "stream to date-partitioned BQ table" usecase
When a user wants to stream data to a date-partitioned BQ table, the way to do this is:

```scala
//noinspection ScalaStyle
class DayPartitionFunction()
  extends SerializableFunction[ValueInSingleWindow[TableRow], TableDestination] {
  override def apply(input: ValueInSingleWindow[TableRow]): TableDestination = {
    val partition = DateTimeFormat.forPattern("yyyyMMdd").withZone(DateTimeZone.UTC)
      .print(input.getWindow.asInstanceOf[IntervalWindow].start())
    new TableDestination("myproject:mydataset.mytable_" + partition, "")
  }
}

object MyJob {
    ...
    .saveAsCustomOutput(
       "myOutputName",
        BigQueryIO.writeTableRows().to(
          new DayPartitionFunction())
          .withSchema(JsonSchemaHelpers.tableSchema)
          .withWriteDisposition(WriteDisposition.WRITE_APPEND)
          .withCreateDisposition(CreateDisposition.CREATE_IF_NEEDED))
}
```

which is a lot of code for a (seemingly) simple operation.

If this is a common usecase (I'm not sure), maybe we could make this a bit simpler. E.g. removing the need for the `DayPartitionFunction` class would be nice.

jasmineytchen:
PR Summary

Closes #3372. Support BQ partitioned table IO use case in BigQueryIO. Currently the usecase is being supported in a less visible module `dynamic/`. 

I am unsure what's the best way to indicate deprecations so I have just marked them as @deprecated as with other deprecated functions in the code base. Also note that `BigQueryPartitionedTableIOIT` is a new test file, not a copy or a move from `DynamicBigQueryIT` - I am unsure why it was presented as such in git diff.


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -495,6 +500,103 @@ object TableRowJsonIO {
   )
 }
 
+final case class BigQueryPartitionedTable[T: Coder](writer: beam.BigQueryIO.Write[T])(
```
regadas, line 503: I wonder if we could just tweak `BigQueryTypedTable` a little bit.
Currently it looks like:

https://github.com/spotify/scio/blob/d387eeb901babd091b83ac5630eb793f9be2416d/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L324-L329

and:

https://github.com/spotify/scio/blob/d387eeb901babd091b83ac5630eb793f9be2416d/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L341

I think we could just not assume that `Table` is always defined and move this logic to the companion object. This way we can have an extra method that accepts a `DynamicDestination` or a `ValueInSingleWindow[T] => TableDestination`.

We would avoid some duplication and reuse a lot of the existing logic.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -100,6 +100,7 @@ final class DynamicBigQueryOps[T](private val self: SCollection[T]) extends AnyV
   }
 }
 
+@deprecated("use SCollectionTableRowOps instead")
```
regadas, line 103: I think we can add the `@deprecated` to the method itself and point to the new method. Also we should add `since = "0.10.0"`

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -124,7 +125,7 @@ final class DynamicTableRowBigQueryOps[T <: TableRow](private val self: SCollect
       createDisposition
     )
 }
-
+@deprecated("use SCollectionTypedOps instead")
```
regadas, line 128: same thing here.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -84,6 +88,22 @@ final class SCollectionTableRowOps[T <: TableRow](private val self: SCollection[
     val param = TableRowJsonWriteParam(numShards, compression)
     self.covary[TableRow].write(TableRowJsonIO(path))(param)
   }
+
+  /**
+   * Save this SCollection to partitioned BigQuery tables using the specified table function.
+   * Note that elements must be of type
+   * [[com.google.api.services.bigquery.model.TableRow TableRow]].
+   */
+  def saveAsPartitionedTable(
```
regadas, line 97: I think we could just use the same naming scheme here and have a `saveAsBigQueryTable` with the curried fn.

We probably could have one method that accepts `DynamicDestination`

We also would want to take into account the `Format[T]` because we will want this to work with `GenericRecord`.

regadas (commented):
@jasmineytchen Thanks for taking this!

Left a few comments. wdyt?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
jasmineytchen: Hmm, I see your point but `BigQueryTypedTable` also supports `read` and can be used in [test context](
https://github.com/spotify/scio/blob/e5708e62111f30dd34fff39d851ea97cb42ff017/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/dynamic/syntax/SCollectionSyntax.scala#L74:L77). These weren't supported in a `dynamic/syntax/SCollectionSyntaz`. 

Should a partitioned table support these as well?

regadas: right now we don't have a really good way to test it so we can keep it as is. However, the idea of tweaking `BigQueryTypeTable` was so you could use it as underlying impl ([example](https://github.com/spotify/scio/tree/master/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala)) for the `BigQueryPartionedTable`. On a side note we should also rename it to smth like `BigQueryDynamicTable`.

jasmineytchen: 👍  Addressed this in the latest commits

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -495,6 +500,103 @@ object TableRowJsonIO {
...
+  }
+
+  object WriteParam extends Writes.WriteParamDefauls {
+    @inline final def apply(
+      schema: TableSchema,
+      wd: WriteDisposition,
+      cd: CreateDisposition,
+      ei: ExtendedErrorInfo
+    )(it: SCollection[ei.Info] => Unit): WriteParam = new WriteParam {
+      val schema: TableSchema = schema
```
regadas, line 561: this seems odd; ref itself?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -84,6 +88,22 @@ final class SCollectionTableRowOps[T <: TableRow](private val self: SCollection[
     val param = TableRowJsonWriteParam(numShards, compression)
     self.covary[TableRow].write(TableRowJsonIO(path))(param)
   }
+
+  /**
+   * Save this SCollection to partitioned BigQuery tables using the specified table function.
+   * Note that elements must be of type
+   * [[com.google.api.services.bigquery.model.TableRow TableRow]].
+   */
+  def saveAsDynamicBigQueryTable(
```
regadas, line 97: Sorry if I wasn't clear before I think these should be:
```suggestion
  def saveAsBigQueryTable(
```

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -186,6 +205,23 @@ final class SCollectionTypedOps[T <: HasAnnotation](private val self: SCollectio
     val param = TableWriteParam(writeDisposition, createDisposition, timePartitioning)
     self.write(BigQueryTyped.Table[T](table))(param)
   }
+
+  /**
+   * Save this SCollection to partitioned BigQuery tables using the specified table function.
+   * Note that element type `T` must be annotated with
+   * [[com.spotify.scio.bigquery.types.BigQueryType BigQueryType]].
+   */
+  def saveAsDynamicTypedBigQueryTable(
```
regadas, line 214:
```suggestion
  def saveAsTypedBigQueryTable(
```

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -495,6 +500,103 @@ object TableRowJsonIO {
...
+    val wFn = ClosureCleaner.clean(writerFn)
+    val writer = beam.BigQueryIO
+      .write[T]()
+      .withFormatFunction(Functions.serializableFn(wFn))
+
+    BigQueryDynamicTable(writer)(tableFn)
+  }
+
+  def apply[T <: TableRow: Coder](
+    schema: TableSchema,
```
regadas, line 589: `schema` is not being used and it will be needed down the line to build the `DynamicDestinations`

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -495,6 +500,103 @@ object TableRowJsonIO {
   )
 }
 
+final case class BigQueryDynamicTable[T: Coder](writer: beam.BigQueryIO.Write[T])(
```
regadas, line 503: Here since we are keeping this IO for now, we can habe `DynamicDestinations` as the only argument here.

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryDynamicTableIOIT.scala
```diff
@@ -44,28 +46,30 @@ object DynamicBigQueryIT {
...
 
-  private val bq = BigQuery.defaultInstance()
+  import BigQueryDynamicTableIOIT._
+  import ItUtils.project
 
-  private val options = PipelineOptionsFactory
-    .fromArgs(s"--project=$projectId", "--tempLocation=gs://data-integration-test-eu/temp")
+  private val bq = BigQuery.defaultInstance()
+  private val options: PipelineOptions = PipelineOptionsFactory
+    .fromArgs(s"--project=$project", s"--tempLocation=$tempLocation")
```
regadas, line 56: `projectId`?

```suggestion
    .fromArgs(s"--project=$projectId", s"--tempLocation=$tempLocation")
```

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryDynamicTableIOIT.scala
```diff
@@ -44,28 +46,30 @@ object DynamicBigQueryIT {
...
     .create()
 
-  "Dynamic BigQuery" should "support typed output" in {
+  it should "support typed output" in {
     val prefix = UUID.randomUUID().toString.replaceAll("-", "")
     val sc = ScioContext(options)
-    // Assigning the value to coll to workaround a bug in scalac...
-    sc.parallelize(1 to 10)
+
+    sc.parallelize(1 to 3)
```
regadas, line 63: why decrease to 3?

regadas (commented):
Some minor comments but then I think we are good for a first iteration.

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryDynamicTableIOIT.scala
jasmineytchen: the results for 10 / 3 are the same, I reduced it as I was running tests locally - bumped this back to 10.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
jasmineytchen: this follows the same pattern as the other WriteParam constructors in BigQueryIO. See below:

https://github.com/spotify/scio/blob/master/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L221:L237

regadas (commented):
Hi @jasmineytchen I was giving this some more thought I really think that we should express this using `BigQueryTypedTable` instead of creating a new `IO`.

I started some exploration based on this PR to articulate a little bit better what I mean. https://github.com/spotify/scio/compare/master...regadas:rework_bq

It needs further work but it's coming together. Let me know wdyt!

jasmineytchen:
> Hi @jasmineytchen I was giving this some more thought I really think that we should express this using `BigQueryTypedTable` instead of creating a new `IO`.
> 
> I started some exploration based on this PR to articulate a little bit better what I mean. [master...regadas:rework_bq](https://github.com/spotify/scio/compare/master...regadas:rework_bq)
> 
> It needs further work but it's coming together. Let me know wdyt!

Hey, I did think that expressing it within `BigQueryTypedTable` would have been more ideal but was hesitant to make a big change in the source file to accommodate for the WriteOnly and test context scenarios. I like this though one concern I have with this approach is that the pattern would diverge from the other IOs. Wdyt? 

Another question, would we want to keep this change in the same PR?

regadas:
> Hey, I did think that expressing it within BigQueryTypedTable would have been more ideal but was hesitant to make a big change in the source file to accommodate for the WriteOnly and test context scenarios. I like this though one concern I have with this approach is that the pattern would diverge from the other IOs. Wdyt?

Diverges from the BQ IO's but it's a common pattern on other IO's, so it's fine. With that said there's a need to revise the other IO's as well.

This work will need to be splitted and doesn't really fit in this PR anymore. I guess we will need to punt on this for a little bit.

jasmineytchen:
> > Hey, I did think that expressing it within BigQueryTypedTable would have been more ideal but was hesitant to make a big change in the source file to accommodate for the WriteOnly and test context scenarios. I like this though one concern I have with this approach is that the pattern would diverge from the other IOs. Wdyt?
> 
> Diverges from the BQ IO's but it's a common pattern on other IO's, so it's fine. With that said there's a need to revise the other IO's as well.
> 
> This work will need to be splitted and doesn't really fit in this PR anymore. I guess we will need to punt on this for a little bit. By this I mean the deprecation of the dynamic ops as we have some foundation work to do first.

Hmm I think [this](https://github.com/spotify/scio/blob/master/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala) is an example of an IO that has a similar pattern as the BQ IO in its current state.

Alright, that makes sense. Do you want to open a separate issue (tagged enhancement maybe) for this? And can I close this PR for now?

regadas:
> Hmm I think this is an example of an IO that has a similar pattern as the BQ IO in its current state.

Those IO's don't need to have diff read/write ... if you are looking for examples these ones are good [BigTableIO](https://github.com/spotify/scio/blob/master/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigtable/BigTableIO.scala) or [RedisIO](https://github.com/spotify/scio/blob/master/scio-redis/src/main/scala/com/spotify/scio/redis/RedisIO.scala).


Yeah let's close this for now and open separate issues.

PR #3468, opened 2020-11-11
Fix README Scio logo path

viktorjonsson:


regadas (approved):
Nice catch @viktorjonsson! Thanks

PR #3475, opened 2020-11-13
Fix logging in ApproxFilter

martinbomio:
Fixing the logging of the partition setting in ApproxFilter since it logged the sequence in the first placeholder.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala
@@ -276,9 +276,10 @@ sealed trait ApproxFilterCompanion {
     } else {
       val settings = partitionSettings(expectedInsertions, fpp, 100 * 1024 * 1024)
       logger.info(
-        s"Partition settings for approximate filter side input of $expectedInsertions keys: " +
-          s"partitions=${settings.partitions}, expectedInsertions=${settings.expectedInsertions}," +
-          "sizeBytes=$settings.sizeBytes"
+        s"""Partition settings for approximate filter side input of $expectedInsertions keys:
+           |partitions=${settings.partitions}
+           |expectedInsertions=${settings.expectedInsertions}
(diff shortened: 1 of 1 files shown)


regadas (approved):

scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala
```diff
@@ -278,12 +278,10 @@ sealed trait ApproxFilterCompanion {
       logger.info(
         "Partition settings for approximate filter side input of {} keys: " +
```
regadas, line 279: So looks like that this doesnt fly with 2.12. It's probly better to use string interpolation here. Also don't forget to apply `scalafmt`

regadas (changes requested):
Needs some tweak

regadas:
@martinbomio nice catch btw. Thanks! 👏

scio-core/src/main/scala/com/spotify/scio/hash/ApproxFilter.scala
```diff
@@ -276,14 +276,9 @@ sealed trait ApproxFilterCompanion {
...
-        "Partition settings for approximate filter side input of {} keys: " +
-          "partitions={}, expectedInsertions={}, sizeBytes={}",
-        Seq(
-          expectedInsertions,
-          settings.partitions,
-          settings.expectedInsertions,
-          settings.sizeBytes
-        )
+        s"Partition settings for approximate filter side input of $expectedInsertions keys: " +
+          "partitions=$settings.partitions, expectedInsertions=$settings.expectedInsertions," +
```
regadas, line 280: This will need to be smth like:
```suggestion
          s"partitions=${settings.partitions}, expectedInsertions=${settings.expectedInsertions}," +
```

Although I think it's better if you do a multiline string here

regadas: It's still borked on the next line. I'll fix it really quick.

regadas (approved):

PR #3488, opened 2020-11-18
Add env variable to disable version check

regadas:


regadas:
I realised i was doing this the wrong way.

PR #3489, opened 2020-11-18
Fix scio version warning in tests

regadas:


scio-core/src/main/scala/com/spotify/scio/VersionUtil.scala
```diff
@@ -80,7 +88,10 @@ private[scio] object VersionUtil {
   }
 
   private[scio] def ignoreVersionCheck: Boolean =
-    Option(System.getProperty("scio.ignoreVersionWarning")).exists(_.trim == "true")
+    sys.props
+      .get("scio.ignoreVersionWarning")
+      .orElse(sys.env.get("SCIO_IGNORE_VERSION_WARNING"))
```
regadas, line 93: So apparently even though we set `scio.ignoreVersionWarning` in `.jvmopts` sys props don't get propagated to tests.

I don't really like the usage of an env var here.

regadas: ok, so apparently this is due to the test splitting. We fork the tests and javaOptions is not propagated properly.

PR #3490, opened 2020-11-18
Fix getRecord should not throw on a missing field

sfines-clgx:
Use getValue[T] inside of getRecord a la getRepeated to avoid a NullPointerException if a non-existent field was requested.


regadas (approved):
Thanks @sfines-clgx

PR #3493, opened 2020-11-20
Remove deprecated BigQueryTable

regadas:
Users should now be using `BigQueryTypedTable` if they use the IO directly


nevillelyh (approved):
This is mostly internal right? No user code change unless they use `.{read,write}` directly?

regadas:
@nevillelyh correct!

PR #3505, opened 2020-12-01
Port scio-macros and scio-core to Scala 3

jto:
## TODO 

- [x] IsJavaBean
- [x] Coder derivation
- [x] Schema derivation
- [x] Coder fallback warning macro
- [x] `AvroCodersMacros`
- [ ] `MacroSettings` ?

## Blocked

- [ ] `KryoRegistrar` (macro annotation)
- [ ] `registerSysProps` (macro annotation)
- [ ] `ArrayCoder` (blocked by https://github.com/lampepfl/dotty/issues/10599)
- [ ] `To.safe`. Is there an equivalent to `c.eval(expr)` ???

## Things to be careful about

- [ ] Closure Cleaner
- [ ]  ⚠️  Serializing derived typeclasses (Coder, Schema) ⚠️ 
- [ ] How do I run test (need to also port `scio-test` ?)


scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -188,21 +188,22 @@ private class VectorCoder[T](bc: BCoder[T]) extends SeqLikeCoder[Vector, T](bc)
...
+//     val arr = new Array[T](size)
+//     var i = 0
+//     while (i < size) {
+//       arr(i) = bc.decode(inStream)
+//       i += 1
+//     }
+//     arr
+//   }
+//   override def consistentWithEquals(): Boolean = false
+// }
```
jto, line 206: see https://github.com/lampepfl/dotty/issues/10599

scio-core/src/main/scala-3/com/spotify/scio/coders/instances/AvroCoderMacros.scala
```diff
@@ -0,0 +1,57 @@
...
+private[instances] trait AvroCodersMacros {
+  // XXX: scala3 - this should not need to be transparent for for some reason
+  // it throws the following exception:
+  //  [error] java.lang.AssertionError: assertion failed: asTerm called on not-a-Term val <none>
+  //  [error] scala.runtime.Scala3RunTime$.assertFailed(Scala3RunTime.scala:8)
+  //  [error] dotty.tools.dotc.core.Symbols$Symbol.asTerm(Symbols.scala:156)
+  //  [error] dotty.tools.dotc.sbt.ExtractAPICollector.apiDefinition(ExtractAPI.scala:338)
+  //  [error] dotty.tools.dotc.sbt.ExtractAPICollector.apiDefinitions$$anonfun$1(ExtractAPI.scala:323)
+  //  [error] scala.collection.immutable.List.map(List.scala:250)
+  transparent inline implicit def genAvro[T <: SpecificRecordBase]: Any =
```
jto, line 55: The return type of the macro should be `Coder[T]` and it should not need to be `transparent`. This might be a bug in the compiler

PR #3512, opened 2020-12-02
TestInput is not thread safe and can result in flaky tests.

rteabeault:
TestInput's apply method calls add on a mutable Set as each
input is read. This call is not thread safe and can result in
some adds getting lost (probably when the underlying data structure
is resized). This results in flaky tests. This commit changes
the Sets used in TestDataManager to be thread safe.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
@@ -22,9 +22,9 @@ import com.spotify.scio.io.ScioIO
 import com.spotify.scio.values.SCollection
 import com.spotify.scio.{ScioContext, ScioResult}
 import org.apache.beam.sdk.testing.TestStream
-
 import scala.collection.concurrent.TrieMap
 import scala.collection.mutable.{Set => MSet}
+import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success, Try}
 
 /* Inputs are Scala Iterables to be parallelized for TestPipeline, or PTransforms to be applied */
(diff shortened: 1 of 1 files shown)


rteabeault:
Please see this example code that shows `add` is not thread safe.

```import java.util.concurrent.{ExecutorService, Executors, TimeUnit}
import scala.collection.mutable.{Set => MSet}

object SetThreadTest extends App {
  // Run with this set and you will get more or less than 1000 integers added to the set.
  val s: MSet[Int] = MSet.empty

  // Run with this set instead and you will get exactly 1000 integers added to the set.
  //  val s: MSet[Int] =
  //    java.util.concurrent.ConcurrentHashMap.newKeySet[Int]().asScala

  val pool: ExecutorService = Executors.newFixedThreadPool(10)

  (1 to 1000).foreach { x =>
    pool.execute(() => s.add(x))
  }

  pool.shutdown();
  pool.awaitTermination(30, TimeUnit.SECONDS);

  println(s.size)
}```

nevillelyh (approved):

nevillelyh:
Thanks. Just curious how were you building the tests? IIRC we always construct job/tests single threaded so this never came up.

rteabeault:
Thank for the merge @nevillelyh. I don't believe we are doing anything special with our tests. We extend `PipelineSpec`, build a JobTest with inputs/outputs and call run. How exactly are you creating your tests to be single threaded?

nevillelyh:
I mean `Test{Input,Output,DistCache}` abstracts IOs in the same job and job construction in `main` and IO assertions in `JobTest` are usually single threaded.

`TestDataManager` tracks all jobs/tests via `testId` but it uses a `TrieMap` so concurrently tests should be safe as long as the job/test code themselves are single-threaded. Any case doesn't hurt to make sure. :)

PR #3532, opened 2020-12-16
Upgrade Sparkey to work around hash collision bug.

psobot:
From [sparkey-java#3.2.1](https://github.com/spotify/sparkey-java/blob/master/CHANGELOG.md#321):

> Fixed bug where creating hash files would erroneously lose some keys. The bug only applies to cases where: construction mode was SORTING and there existed hash collisions (so typically only when hash mode is 32 bits and the number of keys is more than 100000). The bug was introduced along with SORTING in 2.3.0.

Scio users may have been impacted by this bug if:

- writing any Sparkey files from their pipelines (`.asSparkey`, `.asSparkeySideInput`, etc)
- writing between 1M and 8,388,608 elements per individual Sparkey file or shard

Users can work around this behaviour in Scio versions prior to this patch by manually setting `maxMemoryUsage` on any Sparkey-related calls to `Long.MaxValue` to force `sparkey-java` to avoid the buggy code path.


kellen (approved):
👍

spkrka (approved):

regadas (approved):
Thank you all, I've just caught up on the thread nice finding @spkrka @psobot @kellen 👍 

Waiting now for sonatype to propagate

regadas:
Users on previous versions can also just override the Sparkey dependency like so:

```scala
 dependencyOverrides += "com.spotify.sparkey" % "sparkey" % "2.3.1"
```

PR #3543, opened 2020-12-31
Rework RedisMutator

regadas:
Closes #3543


stormy-ua (approved):
👍  So you pushed the mutation logic branching by type inside mutators. I guess this makes sense, because it reduces the number of implicit instances and it is easier to read and comprehend the implementation.

PR #3570, opened 2021-01-20
set Parquet writer row group size, fix #3569

### LINKED ISSUE #3569: Make Parquet sink row group configurable?
Hadoop configuration `parquet.block.size` seems to affect HDFS block size only (not applicable to GCS or other network storage) but not row group size, at least not with the `AvroParquetWriter.builder` we're using for writing, which has its own `withRowGroupSize(int rowGroupSize)` setter.

We should probably set this if `parquet.block.size` is set since that seems to be the convention.

nevillelyh:


regadas (approved):

scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroSink.java
```diff
@@ -106,11 +107,14 @@ public ParquetAvroWriter(
     @Override
     protected void prepareWrite(WritableByteChannel channel) throws Exception {
       BeamOutputFile outputFile = BeamOutputFile.of(channel);
+      // https://github.com/apache/parquet-mr/tree/master/parquet-hadoop#class-parquetoutputformat
+      int rowGroupSize = conf.get().getInt(ParquetOutputFormat.BLOCK_SIZE, 128 * 1024 * 1024);
```
regadas, line 111: @nevillelyh Just noticed that we have a `ParquetWriter.DEFAULT_BLOCK_SIZE` , same value, but probably better instead of keeping track of this ourselves?

nevillelyh: Yeah I'll use that. Parquet's constants are all over the place 🤷

Issue #2229, opened 2019-09-12
Add auth support for ElasticsearchIO


regadas:

likeanowl:
Hi @regadas !
I would like to help with this issue, if you could provide more details. :)

kellen:
Fixed by #3572

PR #3572, opened 2021-01-22
Support username password auth for Elasticsearch 7.x

mrkm4ntr:
Resolve part of https://github.com/spotify/scio/issues/2229

### DIFF
--- a/scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
+++ b/scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
@@ -63,7 +63,7 @@ final case class ElasticsearchIO[T](esOptions: ElasticsearchOptions) extends Sci
       .withError((t: BulkExecutionException) => params.errorFn(t))
 
     data.applyInternal(
-      params.usernameAndPassword
+      esOptions.usernameAndPassword
         .map { case (username, password) =>
           write.withCredentials(new UsernamePasswordCredentials(username, password))
         }
@@ -100,8 +100,7 @@ object ElasticsearchIO {
     numOfShards: Long = WriteParam.DefaultNumShards,
     maxBulkRequestSize: Int = WriteParam.DefaultMaxBulkRequestSize,
     maxBulkRequestBytes: Long = WriteParam.DefaultMaxBulkRequestBytes,
-    retry: RetryConfig = WriteParam.DefaultRetryConfig,
-    usernameAndPassword: Option[(String, String)] = None
+    retry: RetryConfig = WriteParam.DefaultRetryConfig
   )
 
   final case class RetryConfig(maxRetries: Int, retryPause: Duration)
--- a/scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
+++ b/scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
@@ -30,6 +30,7 @@ import org.joda.time.Duration
 
 import scala.jdk.CollectionConverters._
 import com.spotify.scio.io.TapT
+import org.apache.http.auth.UsernamePasswordCredentials
 
 final case class ElasticsearchIO[T](esOptions: ElasticsearchOptions) extends ScioIO[T] {
   override type ReadP = Nothing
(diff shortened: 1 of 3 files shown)


scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
```diff
@@ -96,7 +100,8 @@ object ElasticsearchIO {
     numOfShards: Long = WriteParam.DefaultNumShards,
     maxBulkRequestSize: Int = WriteParam.DefaultMaxBulkRequestSize,
     maxBulkRequestBytes: Long = WriteParam.DefaultMaxBulkRequestBytes,
-    retry: RetryConfig = WriteParam.DefaultRetryConfig
+    retry: RetryConfig = WriteParam.DefaultRetryConfig,
+    usernameAndPassword: Option[(String, String)] = None
```
regadas, line 104: I was thinking that we could add these in `ElasticsearchOptions`. It kinda is like a `ConnectionOptions`. (which is a far better name)

regadas (changes requested):
@mrkm4ntr Thanks for taking this one! do you think it will be possible to have for 6.x as well?

scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
regadas: Created and issue around the naming #3573

mrkm4ntr:
> do you think it will be possible to have for 6.x as well?

Yes, I will do it.

scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
mrkm4ntr: Is it allowed to introduce breaking changes?

mrkm4ntr:
> do you think it will be possible to have for 6.x as well?

I think we don't need to maintain modules for Elasticsearch 6 and 7. I created the issue.
https://github.com/spotify/scio/issues/3575

scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
regadas: I would rather not introduce breaking changes at least with this PR. From a brief look i think it's possible to easily deprecate that type.

regadas:
@mrkm4ntr on a second though per my comment #3575 let's not due the changes on ES6.

For this PR I guess we just need to address this comment and we should be good. https://github.com/spotify/scio/pull/3572#discussion_r562513995

scio-elasticsearch/es7/src/main/scala/com/spotify/scio/elasticsearch/ElasticsearchIO.scala
mrkm4ntr: I'd like to add this overloaded method like below not to introduce breaking changes.
```scala
def saveAsElasticsearch(
      connectionOptions: ConnectionOptions)(
      flushInterval: Duration = WriteParam.DefaultFlushInterval,
      numOfShards: Long = WriteParam.DefaultNumShards,
      maxBulkRequestSize: Int = WriteParam.DefaultMaxBulkRequestSize,
      maxBulkRequestBytes: Long = WriteParam.DefaultMaxBulkRequestBytes,
      errorFn: BulkExecutionException => Unit = WriteParam.DefaultErrorFn,
      retry: RetryConfig = WriteParam.DefaultRetryConfig
    )(f: T => Iterable[DocWriteRequest[_]]): ClosedTap[Nothing]
```

But it can't be compiled due to `in class ElasticsearchSCollection, multiple overloaded alternatives of method saveAsElasticsearch define default arguments`. Is there any reasonable method not to introduce breaking changes? 

And if we change the option name to `ConnectionOptions`, users might think it's natural that ConnectionOptions has RetryConfig and maxBulkRequestXXX. For me, keeping name as ElasticsearchOptions is not so bad.

regadas: >But it can't be compiled due to in class ElasticsearchSCollection, multiple overloaded alternatives of method saveAsElasticsearch define default arguments. Is there any reasonable method not to introduce breaking changes?

That's the downside of using default arguments, unfortunately. We would have to add a new method.

Let's do the following let's explore the ElasticsearchOptions vs ConnectionOptions on a follow-up PR and just wrap this up with moving username/password to ConnectionOptions. wdyt?

mrkm4ntr: > Let's do the following let's explore the ElasticsearchOptions vs ConnectionOptions on a follow-up PR and just wrap this up with moving username/password to ConnectionOptions. wdyt?

Sounds good!

regadas (approved):

regadas:
Thanks @mrkm4ntr

PR #3588, opened 2021-02-02
add serializeToByteArray

wayoung:
Adding a function `serializeToByteArray` to `ZetasketchHLL` to allow for storage of the `ByteArray` representation of the HLL.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/ZetaSketchHLL.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/ZetaSketchHLL.scala
@@ -67,6 +67,9 @@ final class ZetaSketchHll[T](arrOpt: Option[Array[Byte]], elemOpt: Option[T] = N
     new ZetaSketchHll[T](Option(nhll.serializeToByteArray()))
   }
 
+  /** @return the byte array representation of the hll */
+  def serializeToByteArray(): Array[Byte] = hll.serializeToByteArray()
+
   /** @return the estimated distinct count */
   def estimateSize(): Long = hll.result()
 


scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/ZetaSketchHLL.scala
```diff
@@ -67,6 +67,9 @@ final class ZetaSketchHll[T](arrOpt: Option[Array[Byte]], elemOpt: Option[T] = N
     new ZetaSketchHll[T](Option(nhll.serializeToByteArray()))
   }
 
+  /** @return the byte array representation of the hll */
+  def serializeToByteArray(): Array[Byte] = hll.serializeToByteArray()
```
nevillelyh, line 71: Make it `def internal = hll` so you get access to other methods also? More future proof.

wayoung: `hll` is already a `@transient private lazy val` on line 34

nevillelyh: I know. Hence a `def` just exposes it.

nevillelyh: @wayoung opinion? Want this merged?

wayoung: no longer needed since hll is public now.

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/ZetaSketchHLL.scala
```diff
@@ -67,6 +67,9 @@ final class ZetaSketchHll[T](arrOpt: Option[Array[Byte]], elemOpt: Option[T] = N
     new ZetaSketchHll[T](Option(nhll.serializeToByteArray()))
   }
 
+  /** @return the byte array representation of the hll */
```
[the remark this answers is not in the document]
wayoung: no longer needed since hll is public now.

scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/ZetaSketchHLL.scala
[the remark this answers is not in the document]
wayoung: no longer needed since hll is public now.

nevillelyh:
Style check is failing. Can you run `sbt scalafmtAll` to fix?

nevillelyh:
Looks like we broke something else, mind fixing?
```
[error] /home/runner/work/scio/scio/scio-extra/src/main/scala/com/spotify/scio/extra/hll/zetasketch/ZetaSketchHLL.scala:34:33: private value hp escapes its defining scope as part of type com.google.zetasketch.HyperLogLogPlusPlus[ZetaSketchHll.this.hp.In]
[error]   @transient lazy val internal: HyperLogLogPlusPlus[hp.In] = (arrOpt, elemOpt) match {
[error]                            
```

wayoung:
I'm reverting the changes to expose the hll publically. I think the way it is set up internally doesn't want it to be exposed at all, and that's what's causing the build errors. It should be enough for me to just have access to this one method.

PR #3622, opened 2021-02-18
add ParquetBucketMetadata and ParquetTypeFileOperations

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroUtils.java
```diff
@@ -0,0 +1,191 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.beam.sdk.extensions.smb;
+
+import com.google.common.base.Preconditions;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.beam.sdk.coders.*;
```
ClaireMcGinty, line 23: i don't think you need the wildcard - we're only importing `Coder` itself it looks like

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+ */
+package org.apache.beam.sdk.extensions.smb;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.reflect.ReflectData;
+import org.apache.beam.sdk.coders.*;
```
ClaireMcGinty, line 25: same comment as above

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+    switch (getRecordType(recordClass)) {
+      case AVRO:
+        return AvroUtils.validateKeyField(
+            keyField,
+            keyClass,
+            new ReflectData(recordClass.getClassLoader()).getSchema(recordClass));
+      case SCALA:
+        return validateScalaKeyField(keyField, keyClass, recordClass);
+    }
+    throw new IllegalStateException("Should never reach here");
```
ClaireMcGinty, line 211: shouldn't this go in the `default` case of the switch statement?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
```diff
@@ -0,0 +1,266 @@
...
+  // Logic for dealing with Avro records vs Scala case classes
+  ////////////////////////////////////////////////////////////////////////////////
+
+  private static RecordType getRecordType(Class<?> recordClass) {
+    if (GenericRecord.class.isAssignableFrom(recordClass)) {
+      return RecordType.AVRO;
+    } else if (scala.Product.class.isAssignableFrom(recordClass)) {
+      return RecordType.SCALA;
+    } else {
+      throw new IllegalArgumentException("Unsupported record class " + recordClass.getName());
```
ClaireMcGinty, line 197:
```suggestion
      throw new IllegalArgumentException("Unsupported record class " + recordClass.getName() +". Must be an Avro record or a Scala case class.");
```
added a suggestion to make the error message a little clearer (for example, if someone tries to use a Scala non-case-class, they might be confused by the error msg).

ClaireMcGinty (approved):

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
nevillelyh: Brain fart, Pavlovian response IntelliJ squiggly line at the bottom 🐕

Issue #3630, opened 2021-02-20
Expose `Configuration` in all Parquet read/write APIs


nevillelyh:
Expose this for all variants, Avro, typed, and TF, vanilla + SMB. Since both reader/writer might use it for extra configuration, e.g. `parquet.avro.data.supplier` when reading.
https://github.com/apache/parquet-mr/tree/master/parquet-avro

regadas:
Closed via #3635

PR #3635, opened 2021-02-25
Expose Hadoop Configuration in non-SMB Parquet IOs #3630

nevillelyh:


ClaireMcGinty (approved):
Looks good! It's probably time to update the GH site docs for Parquet, and maybe add some explanation/examples of relevant Hadoop configuration options?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -169,10 +152,16 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
...
+    private[avro] val DefaultConfiguration = new Configuration()
+  }
   final case class ReadParam[A: ClassTag, T: ClassTag] private (
-    projection: Schema,
-    predicate: FilterPredicate,
-    projectionFn: A => T
+    projectionFn: A => T,
+    projection: Schema = ReadParam.DefaultProjection,
+    predicate: FilterPredicate = ReadParam.DefaultPredicate,
+    conf: Configuration = ReadParam.DefaultConfiguration
```
regadas, line 164: Not sure how many users are really using this module but this will a breaking change. I believe we have a few more PR's like this and it will hard to track down for v0.10.0. maybe we should try to mark these PR's with a `breaking-change` label?  

for this case we could have an `apply` method with the old order and `@deprecated`.

nevillelyh:
@ClaireMcGinty yeah the docs need a revamp, maybe not so much on `Configuration` (that one just in case we need to fine tune something) but more like when to use what how, e.g. Avro vs typed IO, schema evolution, field type mapping ,etc.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
nevillelyh: Yeah this is only to accommodate `projectionFn` being the only "required" arg now. I can revert and work around it some other way to keep source compat.

nevillelyh: Actually this constructor is `private` so probably fine?

regadas: right, to create it is through the companion default `apply` `ReadParam(...)`. That's why I suggested adding just another one with the old order and all is good in the hood 😄

nevillelyh: Which one? `object ReadParam` is new and doesn't have an `apply` method. `case class ReadParam`'s constructor has always been private.

regadas (approved):

PR #3637, opened 2021-02-26
expose Configuration in SMB Parquet reads

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperations.java
```diff
@@ -83,8 +83,13 @@ private ParquetAvroFileOperations(
 
   public static <V extends GenericRecord> ParquetAvroFileOperations<V> of(
       Schema schema, FilterPredicate predicate) {
+    return of(schema, predicate, new Configuration());
```
regadas, line 86: The same way we have a protected `DEFAULT_COMPRESSION` could we have one for the config? We can use it every making things consistent.

scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeFileOperations.scala
```diff
@@ -46,12 +48,16 @@ object ParquetTypeFileOperations {
   ): ParquetTypeFileOperations[T] =
     ParquetTypeFileOperations(compression, new SerializableConfiguration(conf), null)
 
+  def apply[T: Coder: ParquetType](predicate: FilterPredicate): ParquetTypeFileOperations[T] =
+    apply(predicate, new Configuration())
```
regadas, line 52: I reckon we can define here a `DefaultConfiguration` as well and use it in `ParquetTypeSortedBucketIO`

scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeSortedBucketIO.scala
```diff
@@ -48,7 +48,8 @@ object ParquetTypeSortedBucketIO {
     inputDirectories: Seq[ResourceId] = Nil,
     filenameSuffix: String = DefaultSuffix,
     filterPredicate: FilterPredicate = null,
-    predicate: Predicate[T] = null
+    predicate: Predicate[T] = null,
+    configuration: Configuration = new Configuration()
```
regadas, line 52:
```suggestion
    configuration: Configuration = ParquetTypeFileOperations.DefaultConfiguration
```

regadas (approved):

PR #3638, opened 2021-02-26
add Parquet SMB E2E test

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
```diff
@@ -146,8 +146,13 @@ private K extractAvroKey(V value) {
     for (int i = 0; i < keyPath.length - 1; i++) {
       node = (GenericRecord) node.get(keyPath[i]);
     }
+    Object keyObj = node.get(keyPath[keyPath.length - 1]);
+    // Always convert CharSequence to String, in case reader and writer disagree
+    if (getKeyClass() == CharSequence.class || getKeyClass() == String.class) {
```
nevillelyh, line 151: @ClaireMcGinty this might be useful for regular Avro IO? That is if reader and writer schema are compiled differently to Java classes with `String` vs `CharSequence` for the string field.

PR #3640, opened 2021-02-27
Fix BigQuery implicit coder

### LINKED ISSUE #3606: Bigquery `TableRow` default coder
With the move to `scio-google-cloud-platform` the default coder for `TableRow` became the `GenericJson` coder. This coder differs from `TableRowCoder` (backed by Jackson ObjectMapper) slightly in value conversion. 

`JsonObjectParser` converts every (int, float, double) to `java.math.BigDecimal`. Users updating to 0.10.0 if casting values will face smth like:

```
org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.ClassCastException: java.math.BigDecimal cannot be cast to java.lang.Integer
```

1) We could introduce back `Coder[TableRow]`. Users migrating will need to:

```
import com.spotify.bigquery.coders._
```

**or**

2) we can change the `Coder[T <: GenericJson]` to use jackson `ObjectMapper` instead. This would make things smooth for those using this module but it would break things for those relying on this coder.

I think I'm going with option 1) since users will need to migrate to that new module which already introduces some other changes.

regadas:
fixes #3606

In a follow-up PR, I'll update the migration notes for the new `scio-google-platform` module. We need users to be aware that some of the coders are not resolved implicitly anymore due to the dependency decoupling.

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/instances/CoderInstances.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/instances/CoderInstances.scala
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2021 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigquery.instances
+
+import com.spotify.scio.coders.Coder
+import com.google.api.services.bigquery.model.TableRow
+import org.apache.beam.sdk.io.gcp.bigquery.TableRowJsonCoder
+
+trait CoderInstances {
+  implicit def tableRowCoder: Coder[TableRow] = Coder.beam(TableRowJsonCoder.of())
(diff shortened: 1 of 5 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/instances/CoderInstances.scala
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.spotify.scio.bigquery.instances
+
+import com.spotify.scio.coders.Coder
+import com.google.api.services.bigquery.model.TableRow
+import org.apache.beam.sdk.io.gcp.bigquery.TableRowJsonCoder
+
+trait CoderInstances {
+  implicit def tableRowCoder: Coder[TableRow] = Coder.beam(TableRowJsonCoder.of())
```
nevillelyh, line 25: Maybe be more source-compat if we make this available via `import com.spotify.scio.bigquery._`?

regadas: Hi @nevillelyh right, tbh I thought about it; I avoided that because it deviates a bit from the overall way to add coders; in most modules now (including this one) `Coder`'s are brought into scope through `import com.spotify.scio.<smth>.coders._`. 

Since we also set explicitly the coder for the `TableRow` based IO's should cover most of the cases. wdyt? should we still add it?

On a side note, I'm thinking of removing the `GenericJson` coder from scope. Json encode/decode tends to be different and specific to IO's.

nevillelyh: Yeah should probably be enough.
Maybe we can document all the available coders/artifacts/namespaces in a table?

regadas: That's a good idea!

PR #3647, opened 2021-03-02
SortedBucketTransform lazily evaluates the inputs to transformFn

ClaireMcGinty:
`CoGbkResult`s are backed by `Iterable<Object>`s , which until now we've always fully materialized as lists. However, for the SortedBucketTransform, we can actually lazily read those records as we're feeding them into the user's `transformFn`, so they have the choice to fully materialize it, or not. In the case of extremely large key groups this could help the user avoid an OOM as long as they reduce it properly (i.e. not calling `toList()`, `size()` etc).

If the user supplies a `Predicate` though, we have to materialize, since the Predicate operates on a `List` of buffered results.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
@@ -463,9 +463,11 @@ public class SortedBucketSource<FinalKeyT> extends BoundedSource<KV<FinalKeyT, C
                         ((List<Object>) values).add(v);
                       }
                     });
+                keyGroupSize += ((List<Object>) values).size();
               } else {
                 // Can't use predicate here as it acts on a List<Object>
                 values = () -> (Iterator<Object>) entry.getValue().getValue();
+                keyGroupSize = -1; // We can augment this in the transformFn maybe?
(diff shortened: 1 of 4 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -426,49 +436,57 @@ public boolean advance() throws IOException {
...
+            final Predicate<Object> predicate =
+                predicates[index] != null ? predicates[index] : (xs, x) -> true;
+
+            final Iterator<Object> keyGroupIterator =
+                (Iterator<Object>) entry.getValue().getValue();
+
+            if (emitKeyGroup && !materialize) {
+              valueMap.set(
+                  index,
+                  () ->
```
nevillelyh, line 461: What happens if the user re-iterates this `Iterable`? We want it to fail fast instead of emitting nothing right?

nevillelyh: Also, what if user code doesn't exhaust the iterator? IIRC I tried adding something like that (the "lazy iterable" attempt) but couldn't get it to work.

PR #3651, opened 2021-03-03
Only allow accesor methods for BigQuery fields in case classes

jordiolivares:
This commit fixes an issue we detected with a case class that contained a method in it where the code wouldn't compile if it had the `@BigQueryType` annotation.

If methods were defined, the macro would treat them as fields in the table and attempt to construct the case class with it, leading to extra arguments in the constructor


regadas (commented):
Thank you @jordiolivares! Can you add a test case for this?

jordiolivares:
Sure! Let me whip one up real quick

jordiolivares:
I added a test case for schema and round trip serialization from TableRow.

Additionally I identified that I wasn't really getting the list of construction parameters from the case class. But a bigger subset than that. I've modified the `isAccessor` method to `isCaseAccessor` which is the intended case for this.

Feel free to re-review it!

regadas (approved):
LGTM Thank you @jordiolivares !

PR #3665, opened 2021-03-08
Update Bigtable proto dependencies to 1.14.0

mrkm4ntr:
I use scio v0.10.0.
Due to generatedGrpcBetaVersion, proto-google-cloud-bigtable-admin-v1:1.9.1 is selected in our project.
```
 * com.google.api.grpc:proto-google-cloud-bigtable-admin-v2:1.9.1 is selected over 1.14.0
[info]      +- com.spotify:scio-google-cloud-platform_2.12:0.10.0 (depends on 1.9.1)
[info]      +- com.google.api.grpc:grpc-google-cloud-bigtable-admin-v2:1.14.0 (depends on 1.14.0)
[info]      +- com.google.cloud.bigtable:bigtable-client-core:1.16.0 (depends on 1.14.0)
[info]      +- com.google.cloud:google-cloud-bigtable:1.14.0      (depends on 1.14.0)
```

But bigtable-client-core:1.16.0 requires 1.14.0 and this causes following error.
```
Cause: org.apache.beam.sdk.util.UserCodeException: java.lang.NoClassDefFoundError: com/google/bigtable/admin/v2/ListBackupsRequest
```

I think there is no use to specify 1.9.1 in this project.


regadas (changes requested):
Hi @mrkm4ntr thanks for reporting this. We actually use this as a direct dep. It's however outdated and not kept in sync with bigtable-core.

If you are able to I suggest using `bigtableClientVersion` for these deps as well.

mrkm4ntr:
I think the vserion of `bigtable-client-core` is not same as `proto-google-cloud-bigtable-admin-v2`.  The version of `proto-google-cloud-bigtable-admin-v2` should be same as `grpc-google-cloud-bigtable-admin-v2`. `bigtable-client-core`: 1.16.0 depends `proto-google-cloud-bigtable-admin-v2`: 1.14.0 and `grpc-google-cloud-bigtable-admin-v2`: 1.14.0.
I will change `generatedGrpcBetaVersion` to 1.14.0. Is it OK?

regadas:
Ahh sorry I was looking at the wrong place. Yes, please update this to 1.14.0. Thank you!

regadas (approved):
Thank you @mrkm4ntr

PR #3677, opened 2021-03-11
Add Sparkey-backed .asLargeMapSideInput with coder support.

psobot:
This PR improves our Sparkey side-input functionality by adding automatic support for Sparkeys that store _any_ type that has a `Coder[T]` defined. At the moment, `.asSparkeySideInput` only supports `(String, String)` and `(Array[Byte], Array[Byte])` types. As Scio doesn't guarantee stability of `Coder[T]` across versions, this PR only adds this automatic ser/de functionality when used as a SideInput, but not for saving to long-term storage.

Why do this? Well, if the size of a `MapSideInput` grows beyond a certain threshold (somewhere **around 50MB**), the default implementation hits GCS (six times!) whenever `.get()` is called. This is absurdly slow - reading 1,000 values from a 2GB `MapSideInput` takes _over 7 hours_, whereas reading the same data with the code in this pull request takes under two minutes, proving **a speed up of over 200x**:

image

image

While this may be a bit of an outlier of a use case, a future pull request will extend this functionality to `hashJoin` and its related variants, which should allow orders-of-magnitude speedups there as well. Anecdotally, this code has been tested with certain large internal Spotify pipelines, and has resulted in **a 2x reduction** in both overall runtime and shuffle data used.

To do this, this PR:
 - adds a new `SparkeyMap[K, V]` type that wraps `SparkeyReader` with `Coder[K]` and `Coder[V]` to do automatic ser/de.
 - adds a new helper method, `.asLargeMapSideInput` (naming up for debate) that is a drop-in replacement for `MapSideInput`, but uses Sparkey under the hood
 - adds a new `. asLargeMultiMapSideInput` that offers the same drop-in replacement for `MultiMapSideInput`
 - includes tests for the above

TODO (for myself, @regadas, and @nevillelyh) - what should this be called? I named this new helper `LargeMapSideInput` to hint that it's useful if the SI's data is "large," for some definition of "large." We could call it something more descriptive (i.e.: `DiskMapSideInput`) or something more specific to the underlying technology (`SparkeySideInputWithCoders`), but I wanted to make this as accessible and discoverable as possible to encourage its adoption.


nevillelyh (commented):
LGTM. I realized some potential issues w.r.t. `null.asInstanceOf[T]` and 2.12/2.13 duplicate code. Will fix up & build on top of this branch.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -362,10 +354,9 @@ package object sparkey extends SparkeyReaderInstances {
      * [[Cache]] will be used to cache reads from the resulting [[SparkeyReader]].
      */
     @experimental
+    @deprecated("Use asLargeMapSideInput if no cache is required.")
```
regadas, line 357: just noticed that we should include `since` here

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -200,9 +200,17 @@ package object sparkey extends SparkeyReaderInstances {
...
   }
 
-  /** Enhanced version of [[com.spotify.scio.values.SCollection SCollection]] with Sparkey methods. */
-  implicit class SparkeyPairSCollection[K, V](@transient private val self: SCollection[(K, V)]) {
+  private object SparkeyPairSCollection {
     private val logger = LoggerFactory.getLogger(this.getClass)
+  }
+
+  /** Enhanced version of [[com.spotify.scio.values.SCollection SCollection]] with Sparkey methods. */
+  implicit class SparkeyPairSCollection[K, V](@transient private val self: SCollection[(K, V)])
```
regadas, line 208: @nevillelyh since you are making changes here 😄 can you remove `@transient`? we don't really need it here

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/TypedSparkeyReader.scala
```diff
@@ -1,17 +1,18 @@
 /*
- * Copyright 2020 Spotify AB
+ * Copyright 2019 Spotify AB.
```
regadas, line 2: 2021?

regadas (approved):
Nice stuff! @psobot and @nevillelyh

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/SparkeyCoderUtils.scala
```diff
@@ -0,0 +1,36 @@
...
+object SparkeyCoderUtils {
+  private[sparkey] def encode[T](value: T, coder: sdk.coders.Coder[T]): Array[Byte] = {
+    val output = new ByteArrayOutputStream()
+    coder.encode(value, output)
+    output.toByteArray
+  }
+
+  private[sparkey] def decode[T](value: Array[Byte], coder: sdk.coders.Coder[T]): T =
+    if (value == null) {
+      null.asInstanceOf[T]
```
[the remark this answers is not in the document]
nevillelyh: Turns out to be a false alarm. `extends Map[String, T]` defeats the primitive specialization of `T` so it ends up boxed.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -397,6 +384,62 @@ package object sparkey extends SparkeyReaderInstances {
...
+    }
+
+    /**
+     * Convert this SCollection to a SideInput, mapping key-value pairs of each window to a
+     * `Map[key, Iterable[value]]`, to be used with [[SCollection.withSideInputs]]. In contrast to
+     * [[asLargeMapSideInput]], it is not required that the keys in the input collection be
+     * unique. The resulting map is required to fit on disk on each worker. This is strongly
+     * recommended over a regular MultiMapSideInput if the data in the side input exceeds 100MB.
+     */
+    def asLargeMultiMapSideInput: SideInput[SparkeyMap[K, Iterable[V]]] =
```
regadas, line 440: :art: missing `@experimental`

PR #3678, opened 2021-03-11
Add Sparkey-backed .asLargeSetSideInput.

psobot:
This PR follows https://github.com/spotify/scio/pull/3677 and adds support for `SparkeySet[T]`, a Sparkey-backed set object that uses our existing `Coder` infrastructure to provide dramatically faster performance than `SetSideInput` for large (>1GB) side inputs. Similar to https://github.com/spotify/scio/pull/3677, this is exposed as `.asLargeSetSideInput`. A given `SCollection[T]` will be serialized as a Sparkey fileset (sharded by default) with `T` as the keys and an empty `Array[Byte]` as the values.


nevillelyh (approved):
LGTM applied the same clean up as #3677, implicit coders, 2.12/2.13 divergence etc.

Issue #3695, opened 2021-03-19
Deprecate and remove scio-sql


nevillelyh:
We should consider deprecate them and remove in a future release, for the following reasons:

- Very few people use them
- Value proposition of mixing SQL in a typed language is unclear
- Macros hard to maintain and significantly slow down build
- Scala 3 breakage

PR #3706, opened 2021-03-23
deprecate Beam SQL #3695

nevillelyh:


nevillelyh:
@regadas what do we want do with `com.spotify.scio.schemas`? Deprecate & remove also?

regadas:
Hi @nevillelyh well to be honest I believe we could deprecated it yes. There 3 usages of it right iirc:

1. scio-sql which is being deprecated with this
2. SCollection#to
3. BigQueryIO#BeamSchema

I believe that their usages are pretty low to none so yea we could start deprecating them and make that clear in the next release notes

PR #3715, opened 2021-03-26
Fix header/footer in TextIO.textOut

### LINKED ISSUE #3714: Bug in saveAsTextFile when using header/footer
https://github.com/spotify/scio/blob/c3feec3414b74d25ef11cf5ce89d6bf30e8fce14/scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala#L70

Using parameter `header` puts the value as both header and footer.
Using parameter `footer` does not put the value at all.

`sc.parallelize(Seq("One Line")).saveAsTextFile("gs://dummy-bucket/output_with_header.txt", numShards = 1, header = Some("One Header"), suffix = "", shardNameTemplate = "")`

> One Header
> One Line
> One Header

`sc.parallelize(Seq("One Line")).saveAsTextFile("gs://dummy-bucket/output_with_footer.txt", numShards = 1, footer = Some("One Footer"), suffix = "", shardNameTemplate = "")`

> One Line

shomolomo:
Fixes #3714

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
@@ -67,7 +67,7 @@ final case class TextIO(path: String) extends ScioIO[String] {
       )
 
     transform = params.header.fold(transform)(transform.withHeader)
-    transform = params.header.fold(transform)(transform.withFooter)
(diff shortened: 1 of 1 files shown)


regadas (approved):
Thank you for spotting this! @shomolomo

PR #3744, opened 2021-04-08
Update how-to-release.md with 0.10.2 version

stormy-ua:


nevillelyh:
Should we go back to the auto-incremental naming scheme, i.e. something that starts with C?

noppadolr665:
site/src/main/paradox/dev/How-to-Release.md

PR #3750, opened 2021-04-09
Extract avro macros in their own submodule

### LINKED ISSUE #3710: Move `AvroType` to separate sub-projects
To make ways for Scala 3 migration.

jto:
Fixes #3710


build.sbt
```diff
@@ -577,10 +568,32 @@ lazy val `scio-avro`: Project = project
     )
   )
   .dependsOn(
-    `scio-core` % "compile;it->it"
+    `scio-core` % "compile;it->it",
+    `scio-avro-base`
   )
   .configs(IntegrationTest)
 
+lazy val `scio-avro-base`: Project = project
```
regadas, line 576: Looks good to me. The only suggestion here would be to keep this named as `scio-avro`. I reckon that users not using the typed API will not need to update the dep.

jto: If we do that it becomes a breaking change since people will need to update their dependencies. If we're okay with that I can make to suggested change. @nevillelyh wdyt ?

jto:
Don't merge this. We'll just have different build path for different Scala versions. See internal slack discussion.

PR #3758, opened 2021-04-13
Add some SMB read tuning options

nevillelyh:


regadas (approved):
lgtm!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -667,14 +671,26 @@ long getOrSampleByteSize() {
...
-                  return fileOperations.iterator(file);
+                  if (backoff) {
+                    // we hit 10,000ms limit when partition = 23, 32,442ms cumulatively
+                    // 26.71min at partition = 180
+                    long millis = (long) Math.min(10000, Math.pow(1.5, partition));
+                    Thread.sleep(millis);
+                  }
+
+                  Iterator<V> iterator = fileOperations.iterator(file);
+                  return bufferSize > 0 ? new BufferedIterator<>(iterator, bufferSize) : iterator;
```
ClaireMcGinty, line 693: hm since the default `bufferSize` is 10000, unless users explicitly set it to 0 when they upgrade to Scio 0.10.3 they'll get this change in behavior... did you have a chance to perf test it on some Dataflow jobs as well as DirectRunner?

the PR looks good otherwise! 👍

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BufferedIterator.java
```diff
@@ -0,0 +1,63 @@
...
+  private final int bufferSize;
+  private final Object[] buffer;
+  private int idx;
+  private int size;
+
+  public BufferedIterator(Iterator<T> internal, int bufferSize) {
+    this.internal = internal;
+    this.bufferSize = bufferSize;
+    buffer = new Object[bufferSize];
+    refill();
```
stormy-ua, line 35: I wonder if it would help to refill on attempt to get the first element with next. Could this spread out the load even more?

nevillelyh: It's essentially the same thing, since the merge sort algorithm will `peek()` each iterator in a loop which triggers a `next()` anyway.

nevillelyh:
Do not merge yet. Just realized I should prefix the option names to prevent collision.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -667,14 +671,26 @@ long getOrSampleByteSize() {
...
               bucketId,
               targetParallelism,
-              file -> {
+              (partition, file) -> {
                 try {
-                  return fileOperations.iterator(file);
+                  if (backoff) {
+                    // we hit 10,000ms limit when partition = 23, 32,442ms cumulatively
+                    // 26.71min at partition = 180
+                    long millis = (long) Math.min(10000, Math.pow(1.5, partition));
```
stormy-ua, line 688: info: so basically the delay would be proportional to the partition index which growth exponentially with the partition index. Did you check if a uniform/linear (i.e. proportional to the partition index) delay would result in a similar to exponential delay performance?

nevillelyh: Yeah they don't differ much since the 10000ms cap kicks in after 23 partitions. The only reason for an exponential increase is to not penalize jobs that reads reasonable number of partitions. Also according to the following doc, if we can spread such request spike over 20+ minutes that should give load balancer the time to ramp up.

> As a bucket approaches its IO capacity limit, Cloud Storage typically takes on the order of minutes to detect and accordingly redistribute the load across more servers.

> If your request rate is less than 1000 write requests per second or 5000 read requests per second, then no ramp-up is needed. If your request rate is expected to go over these thresholds, you should start with a request rate below or near the thresholds and then double the request rate no faster than every 20 minutes.

https://cloud.google.com/storage/docs/request-rate#redistribution

Issue #3541, opened 2020-12-30
[FeatureRequest] updateNumberOfBigtableNodes can take specific cluster as a parameter


sisidra:
Currently `com.spotify.scio.bigtable.syntax.ScioContextOps#updateNumberOfBigtableNodes` updates all clusters for the instance.
That is not desired behaviour if each cluster is dedicated for specific task - one for batch jobs, another for service, etc.
Desired API is to allow change node count for only specific cluster (dedicated for batch jobs).

Proposed API change:
```
- def updateNumberOfBigtableNodes(
-    projectId: String,
-    instanceId: String,
-    numberOfNodes: Int,
-    sleepDuration: Duration = DefaultSleepDuration
-  )
+ // Empty list means *all* clusters, for backwards compatibility.
+ private val DefaultClusters: List[String] = List.empty()
+ def updateNumberOfBigtableNodes(
+    projectId: String,
+    instanceId: String,
+    numberOfNodes: Int,
+    clusterId: List[String] = DefaultClusters,
+    sleepDuration: Duration = DefaultSleepDuration
+  )

- def updateNumberOfBigtableNodes(
-    bigtableOptions: BigtableOptions,
-    numberOfNodes: Int,
-    sleepDuration: Duration
-  )
+ def updateNumberOfBigtableNodes(
+    bigtableOptions: BigtableOptions,
+    numberOfNodes: Int,
+    clusterId: List[String] = DefaultClusters,
+    sleepDuration: Duration
+  )
```

sisidra:
Maybe `getBigtableClusterSizes` should also be included in change.

sisidra:
Alternative is to introduce functions with different name that take single instance as parameter, e.g. `updateNumberOfBigtableClusterNodes`, `getBigtableClusterSize`.

regadas:
Hi @sisidra! updating only selected clusters makes sense to me!

Changes would need to be added here https://github.com/deacondesperado/scio/tree/update_beam_226/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java and then propagated to the Scala API.

Adding an extra argument should be enough here.

syodage:
@sisidra are you still working on this? if so could you give an update?

PR #3760, opened 2021-04-13
update nodes by cluster name

syodage:
This allows users to update node counts per cluster or set of clusters of a given instance. 

Fixes: https://github.com/spotify/scio/issues/3541

### DIFF
--- a/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
+++ b/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
@@ -36,13 +36,17 @@ import java.security.GeneralSecurityException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-/** Utilities to deal with Bigtable. */
+/**
+ * Utilities to deal with Bigtable.
+ */
 public final class BigtableUtil {
 
-  private BigtableUtil() {}
+  private BigtableUtil() {
+  }
 
   private static final Logger LOG = LoggerFactory.getLogger(BigtableUtil.class);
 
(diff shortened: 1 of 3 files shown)


scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
```diff
@@ -85,8 +89,15 @@ public static void updateNumberOfBigtableNodes(
       final ListClustersResponse clustersResponse =
           bigtableInstanceClient.listCluster(clustersRequest);
 
+      final List<Cluster> clusters = clustersResponse
+          .getClustersList()
+          .stream()
+          // if no clusters specified then apply to all clusters in the instance.
+          .filter(c -> clusterNames.isEmpty() || clusterNames.contains(shorterName(c.getName())))
```
regadas, line 96: Since we are passing a set of cluster names the expectation should be either:
* update if they exist
* or none if otherwise. We should also return an error i.e `IllegalArgumentExcpetion` if that's the case.

It would be nice to keep the previous method as well.

regadas (changes requested):

scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
syodage: Refactored the code; made clusterNames an optional type and throw exceptions if it is an empty set.

scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
```diff
@@ -62,15 +68,24 @@ private BigtableUtil() {}
...
+   * @throws IOException          If setting up channel pool fails
    * @throws InterruptedException If sleep fails
    */
   public static void updateNumberOfBigtableNodes(
-      final BigtableOptions bigtableOptions, final int numberOfNodes, final Duration sleepDuration)
-      throws IOException, InterruptedException {
+      final BigtableOptions bigtableOptions,
+      final int numberOfNodes,
+      final Duration sleepDuration,
+      final Optional<Set<String>> clusterNamesOpt) throws IOException, InterruptedException {
```
regadas, line 81: Instead of adding and `Optional` here we can simply keep as a `Set` and then add 

```java
  public static void updateNumberOfBigtableNodes(
       final BigtableOptions bigtableOptions, final int numberOfNodes, final Duration sleepDuration) {

List<Cluster> clusters = clustersResponse
           .getClustersList();

return updateNumberOfBigtableNodes(bigtableOptions, numberOfNodes, sleepDuration, clusters);
}

  public static void updateNumberOfBigtableNodes(
       final BigtableOptions bigtableOptions, final int numberOfNodes, final Duration sleepDuration, Set<String> clusterNamesOpt) {

List<Cluster> clusters =  clusters = clusters
             .stream()
             .filter(cluster -> clustersToScale.contains(shorterName(cluster.getName())))
             .collect(Collectors.toList());

return updateNumberOfBigtableNodes(bigtableOptions, numberOfNodes, sleepDuration, clusters);
}
```

scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableUtil.java
```diff
@@ -62,15 +68,24 @@ private BigtableUtil() {}
    * end to lower costs yet still get high throughput during bulk ingests/dumps.
    *
    * @param bigtableOptions Bigtable Options
-   * @param numberOfNodes New number of nodes in the cluster
-   * @param sleepDuration How long to sleep after updating the number of nodes. Google recommends at
-   *     least 20 minutes before the new nodes are fully functional
-   * @throws IOException If setting up channel pool fails
+   * @param numberOfNodes   New number of nodes in the cluster
+   * @param sleepDuration   How long to sleep after updating the number of nodes. Google recommends at
+   *                        least 20 minutes before the new nodes are fully functional
```
nevillelyh, line 73: Should document `clusterNamesOpt` here too

PR #3767, opened 2021-04-19
To.safe

vincenzobaz:
This PR proposes a solution to the To.safe blocker in Scala 3 on which @MaximeKjaer, @liufengyun and I worked.
It took us some time to get familiar with the API and we tested different strategies before landing on this code, but we think it shows the strengths of Scala 3 macros and reflection by combining the two APIs in a relatively concise and idiomatic function that does not need `eval`

You can see a smaller diff [here](https://github.com/scalacenter/scio/pull/10)

The critical section is the file `scio-core/src/main/scala-3/com/spotify/scio/schemas/To.scala` where we use Scala 3 macros to pattern match on the type argument `T` of `Expr[Schema[T]]` to derive a `Schema[T]` at compile time which can then be used to invoke `To.checkCompatibility`. To deal with case class and compile time derived `Schema`s we rely on the reflection API which allows us to access the names and types of the fields of the case class.
You can see this the implementation [here](https://github.com/scalacenter/scio/blob/tosafe/scio-core/src/main/scala-3/com/spotify/scio/schemas/To.scala#L30-L118)

This draft pr brings along many changes, in particular to build and syntax, that we adopted to use RC2, sbt 1.5, new syntax and to reduce the number of warnings.

We would like to have your opinion on this approach.

A few steps might be needed to finalize this pr:

 - Polish and refine
 - Extensive testing
 - Strengthen case class detection and safety: I am not sure that non-empty `caseFields` and `<: Product` is a strong-enough condition
 - Rebase on `main` as it progresses.


scio-core/src/main/scala-3/com/spotify/scio/schemas/To.scala
```diff
@@ -0,0 +1,118 @@
...
+          .fold(message => report.throwError(message), identity)
+    }
+  }
+
+  private def sequence[T](ls: List[Option[T]]): Option[List[T]] =
+    if ls.exists(_.isEmpty) then None
+    else Some(ls.collect { case Some(x) => x })
+
+  private def interpret[T: scala.quoted.Type](using Quotes): Option[Schema[T]] = 
+    Type.of[T] match {
```
jto, line 65: (copying my comment here since I originally made it on your fork)

Unless I'm missing something this means that the possible set of supported schema types is effectively limited to those already defined in scio. 

If a user adds a `given` instance of `Schema` for type that isn't supported (say for example a Java class), the derivation will simply ignore it. This could be a problem for aliaswa support in `Schema` but probably something we can live with. We don't really expect users to define their own instances of `Schema`.

I'd be curious to see how the implementation looks like for Java beans and Avro's `SpecificRecord`.

PR #3770, opened 2021-04-19
fix sparse transforms split on the wrong side

nevillelyh:
regression from https://github.com/spotify/scio/commit/a824db3bf0#diff-e957bccad978e20a8db993e00bf89c894f82521d980de57684402176351fd5b6L492-R435

I don't think it affect correctness. By using the potentially much larger LHS keys by mistake, we're saturating the BFs so that they'll return a lot of false positives, but they'll be filtered out in the exact join of "hot" sides. The only problem is it's not keeping enough "chill" keys from shuffling (BF return false, i.e. key not in RHS).


nevillelyh:
@anish749 PTAL since you're probably most familiar with this part.

anish749 (approved):
This looks right. Not sure how we had missed it earlier.

nevillelyh:
@anish749 it was my copy/pasta error. Test didn't catch it since the result is always correct and data size was too small to reveal any issue.

PR #3771, opened 2021-04-20
Use the same hash fn in hashPartition and hashPartitionByKey

anish749:
Just noticed while reviewing the code that the hash functions used for partitioning the lhs and the rhs can have slightly different behaviors in transforms like sparseIntersectByKey. Not sure if there is a practical implication of this.


scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -86,4 +86,9 @@ private[scio] object ScioUtil {
 
   def pathWithShards(path: String): String =
     path.replaceAll("\\/+$", "") + "/part"
+
+  def hashCodeFn[K]: K => Int = {
```
regadas, line 90: we should make this ` def hashCodeFn[K](k :K): Int`. I feel like we should change the name as well, perhaps smth like `consistentHashCode`?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -406,7 +406,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   def hashPartition(numPartitions: Int): Seq[SCollection[T]] =
     self.partition(
       numPartitions,
-      t => Math.floorMod(t.hashCode(), numPartitions)
+      t => Math.floorMod(ScioUtil.hashCodeFn[K](t), numPartitions)
```
regadas, line 409: should be `T` right?

anish749: yes absolutely, I am now convinced that my coding skills are bad without a compiler!!!

regadas (commented):
@anish749 👍 can you run scalafix? there's an unused import now.

regadas (approved):
Thanks @anish749 !

PR #3805, opened 2021-05-10
add SMB local disk buffer

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/FileOperations.java
```diff
@@ -64,6 +90,35 @@ protected FileOperations(Compression compression, String mimeType) {
...
     final Reader<V> reader = createReader();
+
+    Long bytes = diskBufferBytes.get();
+    if (bytes != null && bytes > 0) {
+      final long fileSize = readableFile.getMetadata().sizeBytes();
+      final long prevSize = diskBufferBytes.getAndUpdate(prev -> prev > 0 ? prev - fileSize : prev);
+
+      // Buffer available, an update was made
+      if (prevSize > 0) {
+        LOG.info("Buffering SMB source file {}, size = {}B", resourceId, fileSize);
```
ClaireMcGinty, line 101: is this log going to be really noisy for the millions-of-files use case? maybe the counter is enough of a signal?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/FileOperations.java
```diff
@@ -83,6 +138,13 @@ public void populateDisplayData(Builder builder) {
 
   /** Per-element file reader. */
   public abstract static class Reader<V> implements Serializable {
+    private transient Path path = null;
+
+    public void prepareRead(Path path) throws IOException {
```
ClaireMcGinty, line 143: this method naming makes it a little confusing when to call this `prepareRead` vs the other `prepareRead(ReadableByteChannel channel)`... maybe instead of this method we could rename it like:

```java
private transient Path deletablePath = null;

public void markForCleanup(Path deletablePath) throws IOException {
  this.deletablePath = deletablePath;
}
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/FileOperations.java
```diff
@@ -64,6 +90,35 @@ protected FileOperations(Compression compression, String mimeType) {
     final ReadableFile readableFile = toReadableFile(resourceId);
 
     final Reader<V> reader = createReader();
+
+    Long bytes = diskBufferBytes.get();
+    if (bytes != null && bytes > 0) {
```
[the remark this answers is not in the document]
nevillelyh: Because the value could go negative after `prev - fileSize` in `getAndUpdate`. Otherwise we'll have to do a `Math.max(prev - fileSize, 0)` there.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/FileOperations.java
nevillelyh: Yeah probably redundant after I added counters. Will change to `debug`.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/FileOperations.java
```diff
@@ -64,6 +91,40 @@ protected FileOperations(Compression compression, String mimeType) {
...
+        } while (copied < fileSize);
+        dst.close();
+        src.close();
+
+        bytesBuffered.inc(fileSize);
+        filesBuffered.inc();
+        reader.whenDone(
+            () -> {
+              path.toFile().delete();
+              return diskBufferBytes.getAndUpdate(prev -> prev + fileSize);
```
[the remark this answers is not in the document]
ClaireMcGinty: yeah makes a lot of sense! good call

PR #3820, opened 2021-05-14
Fix removed dependency

kndarp:
Documentation contains `scio-bigquery` instead of `scio-spanner`


nevillelyh:
Good catch, thanks!

PR #3824, opened 2021-05-19
Document ordering of selectedFields

iezer:
I've found that the order of the fields specified to `selectedFields` doesn't seem to have any effect.

Say I have a Big Query table with 2 columns in the order `first_column, last_column`, and a single row `first_column: "foo", last_column: "bar"`, and I have the following code:

```scala
@BigQueryType.fromStorage("my-table", selectedFields = List("last_column", "first_column")
class MyClass

sc.typedBigQueryStorage[MyClass](Table.Spec(MyClass.table.format))
.map(r => println(s"first_column: %s, last_column: %s", r.first_column, r.last_column)
```

This prints out `first_column: bar, last_column: foo`

This PR updates Scio's documentation. I haven't been able to find any documentation about this upstream in Google's API Docs.


nevillelyh:
Thanks!

PR #3828, opened 2021-05-23
Add support for tearing down the resources automatically if the resource is AutoCloseable

jordiolivares:
Hi, I was looking at some of our pipelines and discovered that the resources weren't being released. This normally isn't a problem as the teardown implies the machine is shutting down, but in our use-case it led to an ever-increasing amount of threads being created, which polluted the logs with warnings.

Instead of doing so from our codebase I thought it would be better if the DoFnWithResource automatically closed them based on whether the resource is AutoCloseable or not, liberating the users from that mental overhead.

The algorithm is adapted from the TensorflowDoFn test as it already had to solve this issue.

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
@@ -20,18 +20,20 @@ package com.spotify.scio.transforms;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.transforms.display.DisplayData;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /** A {@link DoFn} that manages an external resource. */
 public abstract class DoFnWithResource<InputT, OutputT, ResourceT> extends DoFn<InputT, OutputT> {
 
   private static final Logger LOG = LoggerFactory.getLogger(DoFnWithResource.class);
-  private static final ConcurrentMap<String, Object> resources = new ConcurrentHashMap<>();
+  private static final ConcurrentMap<String, Pair<AtomicInteger, Object>> resources = new ConcurrentHashMap<>();
 
   private final String instanceId;
   private String resourceId = null;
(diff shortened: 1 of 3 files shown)


scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
```diff
@@ -117,14 +119,43 @@ public void setup() {
...
-          LOG.debug("Creating resource {}", resourceId);
-          return createResource();
+        (key, value) -> {
+          if (value == null) {
+            LOG.debug("Creating resource {}", resourceId);
+            AtomicInteger resourceUsersCounter = new AtomicInteger(1);
+            return Pair.of(resourceUsersCounter, createResource());
+          }
+          else {
+            int newCurrentUsers = value.getLeft().incrementAndGet();
```
nevillelyh, line 131: This happens per element right? So this `AtomicInteger` could be up to total number of elements the `DoFn` has processed, and could overflow? OTOH decrement only happens in `teardown`, so by default once per CPU core. That's not enough to get the counter back to 0.

It seems that for this to work, we need to add a `releaseResource()` method to decrement the counter, and rely on the user to match each `getResource` with a call `releaseResource`. That doesn't seem reliable.

OTOH we can increment the counter in `setup()`? That's usually called once per CPU or `targetParallelism`. Same for `teardown()`. So the counter will be back to 0 when all instances of the `DoFn` are torn down.

scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
```diff
@@ -117,14 +119,43 @@ public void setup() {
...
+            LOG.debug("Incrementing resource {} users to {}", resourceId, newCurrentUsers);
+            return value;
+          }
         });
   }
 
+  @Teardown
+  public void teardown() {
+    try {
+      Pair<AtomicInteger, Object> resourcePair = resources.get(resourceId);
```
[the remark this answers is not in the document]
nevillelyh: OTOH it won't be null if we create the resource & increment the counter in `setup()`, since that's guaranteed to be called at least once.

scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
jordiolivares: This happens at the setup of the DoFn, not per element processed. The algorithm is actually as you are describing, increasing at the setup and decrementing at the teardown.

scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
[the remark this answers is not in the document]
jordiolivares: As I mention in the other comment, this is essentially what you're proposing. As the resource is created in the setup, we'll never have it be null here as there's always a guarantee of it existing

scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java
nevillelyh: Oh duh right! For some reason I remembered the resource was computed lazily in `getResource()`.

nevillelyh:
Thanks!

PR #3841, opened 2021-05-26
Fix for the non serializable Redis IO DoFn summoned execution context

stormy-ua:
`RedisDoFn` implicitly summons global execution context which isn't serializable. The following exception is raised on attempt to save to Redis: 

```
Exception in thread "main" java.lang.IllegalArgumentException: unable to serialize DoFnWithExecutionInformation{doFn=com.spotify.scio.redis.RedisWrite$Writer$$anon$1@144ab54, mainOutputTag=Tag<output>, sideInputMapping={}, schemaInformation=DoFnSchemaInformation{elementConverters=[]}}
	at org.apache.beam.sdk.util.SerializableUtils.serializeToByteArray(SerializableUtils.java:59)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.translateDoFn(ParDoTranslation.java:695)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation$1.translateDoFn(ParDoTranslation.java:253)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.payloadForParDoLike(ParDoTranslation.java:817)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.translateParDo(ParDoTranslation.java:249)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.translateParDo(ParDoTranslation.java:210)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation$ParDoTranslator.translate(ParDoTranslation.java:176)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.PTransformTranslation.toProto(PTransformTranslation.java:248)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.getParDoPayload(ParDoTranslation.java:746)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.isSplittable(ParDoTranslation.java:761)
	at org.apache.beam.repackaged.direct_java.runners.core.construction.PTransformMatchers$6.matches(PTransformMatchers.java:274)
	at org.apache.beam.sdk.Pipeline$2.visitPrimitiveTransform(Pipeline.java:289)
	at org.apache.beam.sdk.runners.TransformHierarchy$Node.visit(TransformHierarchy.java:587)
[...]

PR #3842, opened 2021-05-27
Fix RedisDoFn ExecutionContext

regadas:
This is an alternative approach to #3841 while still allowing users to provide their own `ExecutionContex`.

@stormy-ua please take a look, wdyt?


stormy-ua (approved):
LGTM! Going to close my PR in favor of this one.

martinbomio (approved):
👍

PR #3860, opened 2021-06-03
Remove java8 steps

regadas:
Tbh we don't really need them. why? Build is being done with `-release 8` and `-source 8`, `-target 8` which targets scala/java to java8. Adding java11 specific features will fail.

Because we now have less steps ... this should:tm: reduce some stress on GHA, which in the majority of cases seems to be the reason for failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,6 @@ jobs:
     strategy:
       matrix:
         java:
-          - graalvm-ce-java8@21.0.0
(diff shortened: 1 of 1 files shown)


nevillelyh:
It won't prevent cases where we use Java 11+ stdlib API by mistake though right? But I guess that's super rare.

regadas:
@nevillelyh if java11+ specific stdlib is used it will fail as it will not be possible to gen bytecode compat with java 8. I know a case with `ByteBuffer` that i can double check.

Issue #3809, opened 2021-05-11
`covary` etc. should reset coder


nevillelyh:
Right now they only cast underlying `SCollection`, which might not work for mixing `SCollection`s of ADTs.
https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L294

https://spotify-foss.slack.com/archives/C015189TFDH/p1620769273059800

Patil2099:
I would love to help in solving this issue @nevillelyh. Can you guide me a little?

PR #3868, opened 2021-06-07
covary etc. should reset coder #3809

stormy-ua:
The PR fixes the issue with `covary`/`contravary` not resetting the coder to the target underlying type. More details about the issue [here](https://github.com/spotify/scio/issues/3809).

The fix introduces the new function `unsafeCastElement` which casts the underlying type and sets the coder to the type `covary`/`contravary` is changing it to. 

Also added a unit test which mimics the original issue by concatenating two SCollections of type B and A and covarying the first to the second. It fails w/o the fix.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -298,19 +298,19 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    * @tparam U The result underlying type.
    * @return The result SCollection with the changed underlying type.
    */
-  private def unsafeChangeUnderlyingType[U: Coder]: SCollection[U] = {
+  private def unsafeCastElement[U: Coder]: SCollection[U] = {
     val coder = CoderMaterializer.beam(context, Coder[U])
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -290,14 +290,27 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
 
+  /**
+   * Changes the underlying type of the SCollection from T to U and resets the coder
+   * to match the result underlying type.
+   * The function is unsafe and should be used internally only to support covary/contravary
+   * implementations.
+   * @tparam U The result underlying type.
+   * @return The result SCollection with the changed underlying type.
+   */
+  private def unsafeChangeUnderlyingType[U: Coder]: SCollection[U] = {
```
nevillelyh, line 301: What about `unsafeCastElement`? Feels more accurate?

nevillelyh (approved):

stormy-ua:
I had to add internal to scio unsafe versions of covary/contravary to preserve backward compatibility in several places (mostly PubsubIO) and minimize the pain of migrating to the new versions of covary/contravary resetting the underlying coder.

PR #3882, opened 2021-06-18
Fix memory leak in AsyncDoFns due to race-condition and large bundles

jordiolivares:
Hi, I was observing an error occurring in an AsyncLookupDoFn due to a rather large bundle of work causing the ConcurrentHashMap to balloon in size and cause a worker memory exhaustion.

To explain the error we were encountering the exact race condition explained in the comment repeatedly with a rather large payload. In particular, I added the following lines to BaseAsyncLookupDoFn in the `addCallback` method:

```java
cacheSupplier.put(instanceId, input, output);
results.add(new Result(input, success(output), c.timestamp(), window));
F tmp = futures.remove(uuid);
if (tmp == null) {
  LOG.error("Failed to remove future {}", uuid.toString());
}
LOG.info("Current number of futures in map is {}", futures.size());
```

Running it with a `maxPendingRequests` value of 30 yielded the following after just some seconds of processing:

Screenshot 2021-06-18 at 17 48 07

As we can see this is not ideal and indicates that the race condition is happening much more often than expected. As such, the ConcurrentHashMap keeps growing in size if on a large bundle of elements to process until the worker suddenly dies due to running out of memory.

To address this we've moved the future removal from the HashMap into the `flush()` method and just for good measure used `computeIfAbsent` to make the race condition harder to hit by reducing the time between the future starting to process and when it's in the Map. This yields a stable amount of elements (stable at around 30 in our case) in the ConcurrentHashMap, which should hopefully solve the issue by GCing the dead elements

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
@@ -71,24 +71,20 @@ public abstract class BaseAsyncDoFn<InputT, OutputT, ResourceT, FutureT>
     flush(c);
 
     final UUID uuid = UUID.randomUUID();
-    FutureT future =
-        addCallback(
-            processElement(c.element()),
-            r -> {
-              results.add(new Result(r, c.timestamp(), window));
-              futures.remove(uuid);
-              return null;
-            },
-            t -> {
-              errors.add(t);
-              futures.remove(uuid);
-              return null;
-            });
-    // This `put` may happen after `remove` in the callbacks but it's OK since either the result
-    // or the error would've already been pushed to the corresponding queues and we are not losing
-    // data. `waitForFutures` in `finishBundle` blocks until all pending futures, including ones
-    // that may have already completed, and `startBundle` clears everything.
-    futures.put(uuid, future);
+    futures.computeIfAbsent(
+      uuid,
+      key -> addCallback(
+        processElement(c.element()),
+        r -> {
+          results.add(new Result(r, key, c.timestamp(), window));
+          return null;
+        },
+        t -> {
+          errors.add(t);
+          return null;
+        }
+      )
+    );
   }
 
   private void flush(ProcessContext c) {
@@ -104,6 +100,7 @@ public abstract class BaseAsyncDoFn<InputT, OutputT, ResourceT, FutureT>
     Result r = results.poll();
     while (r != null) {
       c.output(r.output);
+      futures.remove(r.futureUuid);
       r = results.poll();
     }
   }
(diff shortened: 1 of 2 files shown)


nevillelyh:
Good find! It's been a while since I touched this code, but since we clear `futures` in `startBundle`, the concurrent map size ballooning depends on the bundle size, and in your case it's heavy enough to cause problem?

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
```diff
@@ -71,24 +71,20 @@ public void processElement(ProcessContext c, BoundedWindow window) {
...
+    futures.computeIfAbsent(
+      uuid,
+      key -> addCallback(
+        processElement(c.element()),
+        r -> {
+          results.add(new Result(r, key, c.timestamp(), window));
+          return null;
+        },
+        t -> {
+          errors.add(t);
```
nevillelyh, line 83: Do we need to `futures.remove(key)` here since if it fails, it'll never be added to `results` and leaked inside the map?

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -165,16 +151,21 @@ public void processElement(ProcessContext c, BoundedWindow window) {
             },
             throwable -> {
               semaphore.release();
-              results.add(new Result(input, failure(throwable), c.timestamp(), window));
-              futures.remove(uuid);
+              results.add(new Result(input, failure(throwable), key, c.timestamp(), window));
```
nevillelyh, line 154: Should `futures.remove(key)` here too?

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncDoFn.java
nevillelyh: Although I guess since errors will cause the DoFn to fail and GCed?

jordiolivares: Correct, this will cause the DoFn to fail, so it'll be automatically GCed I believe

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
jordiolivares: In the LookupDoFn family errors are actually passed through to the consumer by wrapping them in a `Try`, so this doesn't indicate failure of the DoFn.

In this case the normal `flush()` calls will remove the futures as if everything went correctly

jordiolivares:
> Good find! It's been a while since I touched this code, but since we clear `futures` in `startBundle`, the concurrent map size ballooning depends on the bundle size, and in your case it's heavy enough to cause problem?

Yeah, it's actually pretty easy to hit it though. We found this by just consuming BigQuery via the Storage API, the bundle sizes grow to million of elements in our case with the DataflowRunner

nevillelyh (approved):

PR #3884, opened 2021-06-21
Add support for additional read sources in smb cogroup

bkalra23:
Adding support for reading 5 and 6 sources with smb cogroup. 

See [issue](https://github.com/spotify/scio/issues/3555). The original issue was filed for 8 sources but we have identified that we only need support for 6 sources for the time being.

### DIFF
--- a/scio-smb/src/it/scala/com/spotify/scio/smb/SortMergeBucketParityIT.scala
+++ b/scio-smb/src/it/scala/com/spotify/scio/smb/SortMergeBucketParityIT.scala
@@ -18,10 +18,10 @@
 package com.spotify.scio.smb
 
 import java.nio.file.{Files, Path}
-
 import com.spotify.scio.ScioContext
 import com.spotify.scio.avro._
 import com.spotify.scio.coders.Coder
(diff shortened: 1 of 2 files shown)


nevillelyh:
For conventional joins, we [generate code](https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/util/MultiJoin.scala) for 5-22 way with a [python script](https://github.com/spotify/scio/blob/main/scripts/multijoin.py). It's more scalable plus we can also [exclude](https://github.com/spotify/scio/blob/main/build.sbt#L167) it from test coverage, etc. WDYT?

bkalra23:
@nevillelyh So you recommend creating a similar script for smb joins and then generating code for 5-22 way cogroup for that?

nevillelyh:
Yeah exactly!

bkalra23:
Sure, I can look into that!

bkalra23:
Merged another [PR](https://github.com/spotify/scio/pull/3904) based on the suggestions in this one. Hence, closing this one.

PR #3888, opened 2021-06-29
Fix Elasticsearch authentication

mrkm4ntr:
There is a bug in https://github.com/spotify/scio/pull/3572. I fixed this and confirmed this change with my Elasticsearch cluster actually.


regadas (approved):
Thanks @mrkm4ntr

PR #3890, opened 2021-06-30
Create a jar from the current REPL session when reading a remote file

jmcconnell:
Beam attempts to stage the jar for the current session as it is on the classpath, but it has not been created at this point. We need to create it as we do with run().


kellen:
Naively I don't think this is valid; closing. If you can provide steps to reproduce, please reopen

PR #3891, opened 2021-06-30
Update the REPL documentation to use vals in place of defs.

jmcconnell:


site/src/main/paradox/Scio-REPL.md
```diff
@@ -82,22 +82,23 @@ Projects generated from [scio-template.g8](https://github.com/spotify/scio-templ
 Let's start with simple local-mode word count example:
 
 ```scala mdoc:invisible
+// The REPL loads the following for you.
 import com.spotify.scio._
 
 def sc: ScioContext = ???
 ```
 
 ```scala mdoc
```
regadas, line 91: `val` makes total sense here but when we added this, mdoc didn't support compile only so that's why the `def`s ... otherwise this will be eval on doc generation.

Looking again at mdoc we can do now `scala mdoc:compile-only` for these blocks. Can you adjust them?

regadas (changes requested):
Thanks @jmcconnell. Left a comment.

site/src/main/paradox/Scio-REPL.md
jmcconnell: Apologies for the delay, just noticed this comment. I've added the `compile-only` qualifiers.

PR #3894, opened 2021-07-02
Rename Sparkey hash* methods to avoid shadowing regular methods

devictr:
The `hashJoin`, `hashLeftOuterJoin`, `hashFullOuterJoin`, `hashIntersectByKey`, `hashSubtractByKey` methods in `com.spotify.scio.extra.sparkey.PairLargeHashSCollectionFunctions` were shadowing the corresponding methods in `com.spotify.scio.values.PairHashSCollectionFunctions`.

As an example, the following test doesn't compile:
```
  it should "not shadow the regular hashJoin method" in {
    import com.spotify.scio.extra.sparkey._

    val sc = ScioContext()

    val lhsInput = Seq((1, "a"), (2, "c"), (3, "e"), (4, "g"))
    val rhsInput = Seq((1, "b"), (2, "d"), (3, "f"))

    val rhs = sc.parallelize(rhsInput)
    val lhs = sc.parallelize(lhsInput)

    val result = lhs
      .hashJoin(rhs)
      .materialize

    val scioResult = sc.run().waitUntilFinish()
    val expectedOutput = List((1,("a", "b")), (2,("c","d")), (3,("e","f")))

    scioResult.tap(result).value.toList should contain theSameElementsAs expectedOutput
}
```

This PR contains a breaking change that fixes this issue. I think renaming those (relatively new) methods to `largeHash*` makes more sense, since they all take an input like `SideInput[SparkeySet[K]]` or `SideInput[SparkeyMap[K]]`

Please let me know if this isn't an approach you want to consider and if you'd like to fix this another way.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/LargeHashSCollectionFunctions.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/LargeHashSCollectionFunctions.scala
@@ -28,7 +28,7 @@ class LargeHashSCollectionFunctions[T](private val self: SCollection[T]) {
    *
    * @group transform
    */
-  def hashFilter(sideInput: SideInput[SparkeySet[T]]): SCollection[T] = {
+  def largeHashFilter(sideInput: SideInput[SparkeySet[T]]): SCollection[T] = {
     implicit val coder = self.coder
     self.map((_, ())).largeHashIntersectByKey(sideInput).keys
   }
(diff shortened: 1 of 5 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/LargeHashSCollectionFunctions.scala
```diff
@@ -30,6 +30,6 @@ class LargeHashSCollectionFunctions[T](private val self: SCollection[T]) {
    */
   def hashFilter(sideInput: SideInput[SparkeySet[T]]): SCollection[T] = {
```
psobot, line 31: Should this also change to `largeHashFilter` while we're at it? (There is a regular `hashFilter` as well, it seems.)

psobot (approved):
LGTM with one comment - although I'm curious how @nevillelyh and/or @regadas feel about breaking API compatibility. (This is a brand new feature without much usage yet, so I'm personally okay with it here.)

regadas (approved):
👍 we will release this under `0.11.0`

Issue #3851, opened 2021-06-01
Add sample Parquet job to scio-examples


clairemcginty:
ideally this would demonstrate projections, predicates, & tuning options like block size (using the `Configuration` parameter)

AkshatJindal1:
I would like to take up this task. Can you provide a background on the requirement

regadas:
Closed via #3903

PR #3903, opened 2021-07-07
Adds a Parquet example

live-wire:
This serves as an example for Reading and Writing `parquet` files.
- Followed [this documentation](https://spotify.github.io/scio/io/Parquet.html)
- Corresponding issue: #3851

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
@@ -74,7 +74,7 @@ object ParquetExample {
 
       // The result Account records are not complete Avro objects. Only the projected columns are present while the rest are null.
       // These objects may fail serialization and it’s recommended that you map them out to tuples or case classes right after reading.
-      .map(x => AccountOutput(x.getId(), x.getName().toString()))
+      .map(x => AccountOutput(x.getId(), x.getName.toString))
 
       // This case class can now be saved as a parquet file
       // numShards should be explicitly set so that the size of each output file is smaller than
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2019 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Example: Read and Write specific and generic Avro records
+// Usage:
+
+// `sbt "runMain com.spotify.scio.examples.extra.ParquetExample
+// --project=[PROJECT] --runner=DataflowRunner --zone=[ZONE]
+// --input=[INPUT].parquet --output=[OUTPUT].parquet --method=[METHOD]"`
+package com.spotify.scio.examples.extra
+
+import com.spotify.scio._
+import com.spotify.scio.parquet.avro._
+import com.spotify.scio.parquet.types._
+import com.spotify.scio.avro.Account
+import com.spotify.scio.io.ClosedTap
+import org.apache.hadoop.conf.Configuration
+import org.apache.avro.generic.GenericRecord
+
+object ParquetExample {
+
+  case class AccountInput(id: Int, `type`: String, name: String, amount: Double)
+
+  case class AccountOutput(id: Int, name: String)
+
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+
+    val m = args("method")
+    m match {
+      // Read Parquet files as Specific Avro Records
+      case "avroSpecific" => avroSpecific(sc, args)
+
+      // Read Parquet files as Generic Avro Records
+      case "avroGeneric" => avroGeneric(sc, args)
+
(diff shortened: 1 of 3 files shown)


scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,125 @@
...
+          .setName(x.name)
+          .setAmount(x.amount)
+          .build()
+      )
+      .saveAsParquetAvroFile(args("output"))
+
+  }
+
+  // See this link for parquet writer tuning https://spotify.github.io/scio/io/Parquet.html#performance-tuning
+  private def fineTunedParquetWriterConfig(): Configuration = {
```
ClaireMcGinty, line 119: I think this could be converted to a `lazy val` !

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,125 @@
...
+
+import com.spotify.scio._
+import com.spotify.scio.parquet.avro._
+import com.spotify.scio.parquet.types._
+import com.spotify.scio.avro.Account
+import com.spotify.scio.io.ClosedTap
+import org.apache.hadoop.conf.Configuration
+import org.apache.avro.generic.GenericRecord
+
+object ParquetExample {
```
ClaireMcGinty, line 34: looks pretty good! I think it would be good to have another job configuration that only writes Parquet data. (for example, `AvroExample` lets the user specify whether they want to read or write; `SortMergeBucketExample` has different jobs for writing, and reading, SMB data.) the idea is that the user could actually directly run those jobs on the runner of their choice, not just in a unit test! since the jobs in your ParquetExample all read Parquet input and write Parquet output, there’s not a clear way for a user to test it out themselves unless they have pre-written Parquet data lying around :slightly_smiling_face:

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright 2019 Spotify AB.
```
ClaireMcGinty, line 2: copyright date needs to be updated! you can set it dynamically in your IDE like this: https://spotify.github.io/scio/dev/Style-Guide.html#intellij-idea

live-wire:
@ClaireMcGinty Incorporated your feedback :) 
Please have a look.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,138 @@
...
+      Projection[Account](_.getId, _.getName, _.getAmount) // Only these columns will be projected
+    val predicate = Predicate[Account](x =>
+      x.getAmount > 0
+    ) // Will skip row groups where this column value check is not satisfied
+
+    sc.parquetAvroFile[Account](args("input"), projection, predicate)
+      // The result Account records are not complete Avro objects. Only the projected columns are present while the rest are null.
+      // These objects may fail serialization and it’s recommended that you map them out to tuples or case classes right after reading.
+      .map(x => AccountOutput(x.getId, x.getName.toString))
+      .map(_.toString)
```
ClaireMcGinty, line 89:
```suggestion
```

this conversion is done automatically in `saveAsTextFile` 👍

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private def avroGenericIn(sc: ScioContext, args: Args): ClosedTap[String] =
+    // Now the fields in Account's schema act as our projection
+    sc.parquetAvroFile[GenericRecord](args("input"), Account.getClassSchema)
+
+      // GenericRecord objects are also not Serializable.
+      // Map out projected fields into something type safe
+      .map(r => AccountOutput(r.get("id").asInstanceOf[Int], r.get("name").toString))
+      .map(_.toString)
```
ClaireMcGinty, line 100:
```suggestion
```

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,138 @@
...
+      .saveAsTextFile(args("output"))
+
+  private def typedIn(sc: ScioContext, args: Args): ClosedTap[String] =
+    // All fields in the case class definition act as our projection
+    sc.typedParquetFile[AccountInput](args("input"))
+      .map(_.toString)
```
ClaireMcGinty, line 106:
```suggestion
```

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
```diff
@@ -0,0 +1,138 @@
...
+      .map(x => AccountOutput(x.getId, x.getName.toString))
+      .map(_.toString)
+      .saveAsTextFile(args("output"))
+  }
+
+  private def avroGenericIn(sc: ScioContext, args: Args): ClosedTap[String] =
+    // Now the fields in Account's schema act as our projection
+    sc.parquetAvroFile[GenericRecord](args("input"), Account.getClassSchema)
+
+      // GenericRecord objects are also not Serializable.
```
ClaireMcGinty, line 97: this statement isn't true--`GenericRecord`s as a class _are_ serializable it can just be inefficient (because the `schema` [needs to be ser/de-ed as well](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala).

ClaireMcGinty (approved):

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
live-wire: 🆒

scio-examples/src/main/scala/com/spotify/scio/examples/extra/ParquetExample.scala
live-wire: True, My bad. I confused it with Avro's `Schema` class that is not serializable (unless converted to a String).
I will get rid of this line 👍

PR #3904, opened 2021-07-08
Add a script and methods for smb cogroup with additional sources

bkalra23:
- Adding support for reading up to 22 sources with smb cogroup.
- Added a python script to generate the Scala code.
- See [issue](https://github.com/spotify/scio/issues/3555)


bkalra23:
@nevillelyh Added a python script as recommended in [this](https://github.com/spotify/scio/pull/3884) PR.

I wasn't sure about adding the Scala code under scio-core/src/main/scala/com/spotify/scio/util as that required `org.apache.beam.sdk.extensions.smb` to be imported to scio-core (from scio-smb?). I ended up creating a util in scio-smb/src/main/scala/com/spotify/scio/smb/util. Let me know if this should be added somewhere else instead.

scio-smb/src/test/scala/com/spotify/scio/smb/util/SMBMultiJoinTest.scala
```diff
@@ -0,0 +1,129 @@
...
+      new Field("key", Schema.create(Schema.Type.INT), "", -1),
+      new Field("value", Schema.create(Schema.Type.STRING), "", "")
+    ).asJava
+  )
+
+  implicit private val coder: Coder[GenericRecord] = Coder.avroGenericRecordCoder(schema)
+
+  private val keyFn: GenericRecord => Integer = _.get("key").toString.toInt
+
+  "SMBMultiJoin" should "have parity with a 5-way CoGroup" in withNumSources(5) { inputs =>
```
clairemcginty, line 39: I think it would be simpler to just move this test case into `SortMergeBucketParityIT` :)

clairemcginty (changes requested):
looks great, thanks for adding this! I added a comment about moving the unit test case into an existing class, after that change we can merge this.

bkalra23:
Thanks for the review @clairemcginty! Moved the test to the suggested file.

scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
```diff
@@ -0,0 +1,3354 @@
...
+package com.spotify.scio.smb.util
+
+import com.spotify.scio.ScioContext
+import com.spotify.scio.coders.Coder
+import com.spotify.scio.values._
+import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, TargetParallelism}
+
+import scala.jdk.CollectionConverters._
+
+final class SMBMultiJoin(@transient private val self: ScioContext) extends Serializable {
```
regadas, line 29: There's no need for this class to be Serializable; we can also remove the `@transient` and make it extend `AnyVal`

scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
```diff
@@ -0,0 +1,3354 @@
...
+      .and(d)
+      .and(e)
+      .withTargetParallelism(targetParallelism)
+    val (tupleTagA, tupleTagB, tupleTagC, tupleTagD, tupleTagE) =
+      (a.getTupleTag, b.getTupleTag, c.getTupleTag, d.getTupleTag, e.getTupleTag)
+
+    val tfName = self.tfName
+
+    self
+      .wrap(self.pipeline.apply(s"SMB CoGroupByKey@$tfName", input))
```
regadas, line 54: :art: here we can replace it with `applyTransform`

scio-smb/src/it/scala/com/spotify/scio/smb/SortMergeBucketParityIT.scala
```diff
@@ -170,6 +170,38 @@ class SortMergeBucketParityIT extends AnyFlatSpec with Matchers {
     }
   }
 
+  "SMBMultiJoin" should "have parity with a 5-way CoGroup" in withNumSources(5) { inputs =>
+    compareResults(
+      com.spotify.scio.smb.util
+        .SMBMultiJoin(_)
```
regadas, line 176:
```suggestion
      SMBMultiJoin(_)
```

regadas (commented):
Thanks @bkalra23 I just noticed some minor tweaks

scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
bkalra23: Extending AnyVal is giving the following compile time error:
` implementation restriction: nested class is not allowed in value class
[error] This restriction is planned to be removed in subsequent releases.
[error]       .map { kv =>`
This would be line 56.
I am still debugging this but please let me know if you have any suggestions for this.

regadas: Ahh this is in fact an issue with scala 2.12; let's not make it a value class for now; I need to dig if there's a simple workaround.

regadas (approved):

clairemcginty (approved):

regadas:
Thanks @bkalra23!

Issue #3314, opened 2020-09-14
Side inputs for SMB transforms


kellen:
I have a few use-cases where I am using SMB transforms and want to merge in some (small) external data, usually from some other SCollection. The way I can achieve this at present is to write out a sparkey or a distcache in some other job then load that manually in the `via` block.

Ideally I'd be able to merge in normal side inputs to a SMB transform without this additional job & code overhead & while still avoiding a shuffle penalty.

Something like:

```
    val mySideInput: SideInput[Map[String, String]] = ???
    sc
      .sortMergeTransform[String, MyEntity]( ... )
      .to( ... )
      .withSideInputs(mySideInput)
      .via { case (key, myEntity, outputCollector, ctx) =>
          val fromSI = ctx(mySideInput).get(myEntity.someField)
          // ...
      }
```

ClaireMcGinty:
Initially I thought this would be a really easy fix -- allow the user-supplied [`TransformFn`](https://github.com/spotify/scio/blob/master/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java#L132)  to optionally take in a `ProcessContext` argument to retrieve a side input from. Unfortunately, since we rewrote this as a `BoundedSource` implementation instead of a plain `PTransform`, we no longer have access to `ProcessContext` and can't surface that. 🤦‍♀️ 

I wonder if Dataflow would still fuse transforms if we re-structured SortedBucketTransform to be a chained `SortedBucketSource` followed by a `MapElements` transform that applies the transformFn (with access to side inputs) and writes to temp file.... that could eliminate some code re-use if so, too. I can check.

ClaireMcGinty:
Ok, I did some investigation into this today. I couldn't find any way to extract the transformation logic into a PTransform without materializing the entire bucket contents into memory, which defeats the point of the SortedBucketTransform optimization. I was looking into how we could get side inputs into our `BoundedSource` impl, and we could wrap it in something like this:

```diff
begin
  .getPipeline()
-  .apply("SortedBucketSource", Read.from(boundedSource))
+  .apply("SortedBucketSource", new PTransform<PBegin, PCollection<MergedBucket>>() {
+    Iterable<PCollectionView<?>> sideInputs = ... // user-supplied side inputs
+
+     @Override
+     public PCollection<MergedBucket> expand(PBegin input) {
+        return input.apply(Read.from(boundedSource));
+     }
+     @Override
[...]

PR #3915, opened 2021-07-13
Side inputs for SMB transforms

kellen:
Adds support for side inputs in SMB transforms. See #3314

This changes when a SMB read occurs. In `main` the read occurs in a `BoundedReader`, directly after a `PBegin`. Here, we instead output bucket metadata from a reader, and do the SMB read in a `DoFn`. This allows the side input to be joined with the incoming bucket metadata. Performance looks comparable to that on `main`.

Refactors SMB source read into `MultiSourceKeyGroupReader`, removing the `BoundedReader` interface.

Scala api is similar to using side inputs on an `SCollection`:
```
val si: SideInput[Seq[Int]] = ???
sc
  .sortMergeTransform[String, In](...)
  .to[Out](...)
  .withSideInputs(si)
  .via { case (key, ins, ctx, outputCollector) =>
    val s: Seq[Int] = ctx(si)
    // ...
  }
```


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/MultiSourceKeyGroupReader.java
```diff
@@ -0,0 +1,234 @@
...
+public class MultiSourceKeyGroupReader<FinalKeyT> {
+  private static final Logger LOG = LoggerFactory.getLogger(MultiSourceKeyGroupReader.class);
+
+  private enum AcceptKeyGroup {
+    ACCEPT,
+    REJECT,
+    UNSET
+  }
+
+  private Optional<KV<FinalKeyT, CoGbkResult>> head = null;
```
clairemcginty, line 35: overall, for Java code we've been preferring the plain null-check style rather than `Optional`s... it's just a bit less overhead and more streamlined.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/MultiSourceKeyGroupReader.java
```diff
@@ -0,0 +1,234 @@
...
+                  KV.of(keyCoder.decode(new ByteArrayInputStream(next.getKey())), next.getValue()));
+          break;
+        } catch (Exception e) {
+          throw new RuntimeException("Failed to decode key group", e);
+        }
+      }
+    }
+  }
+
+  private static class BucketedInputSource<K, V> {
```
clairemcginty, line 189: I'm a bit torn on whether it's clearer to have this as a static class or just inline the method code in `MultiSourceKeyGroupReader`. Since we already have `BucketedInput`, `SortedBucketSource`, & `KeyGroupIterator` I'm worried about added complexity/name overlap :) wdyt?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -72,56 +72,72 @@
...
             newBucketMetadataFn,
             filenamePolicy.forDestination(),
             sourceSpec.hashType);
+
+    final FileAssignment fileAssignment = filenamePolicy.forTempFiles(tempDirectory);
+    final Distribution dist = Metrics.distribution(getName(), getName() + "-KeyGroupSize");
+    if (transformFn != null) {
+      this.doFn =
+          ParDo.of(
+              new NoSides<>(
```
clairemcginty, line 117: for the sake of a clearer job graph, maybe rename to `TransformWithNoSides`/`TransformWithSides` ?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -342,104 +347,237 @@ public int hashCode() {
...
+      int bucketId = e.bucketOffsetId;
+      int effectiveParallelism = e.effectiveParallelism;
+
+      ResourceId dst =
+          fileAssignment.forBucket(BucketShardId.of(bucketId, 0), effectiveParallelism, 1);
+      OutputCollector<FinalValueT> outputCollector;
+      try {
+        outputCollector = new OutputCollector<>(fileOperations.createWriter(dst));
+      } catch (IOException err) {
+        throw new RuntimeException(err);
```
clairemcginty, line 422:
```suggestion
        throw new RuntimeException("Failed to create file writer for transformed output", err);
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -342,104 +347,237 @@ public int hashCode() {
...
+        final BoundedWindow window) {
+      transformFn.writeTransform(mergedKeyGroup, context, outputCollector, window);
+    }
+  }
+
+  private static class BucketReader extends BoundedSource.BoundedReader<BucketItem> {
+    private final BoundedSource<BucketItem> currentSource;
+    private boolean started;
+    // there is only ever a single item in this source
+    @Nullable private BucketItem next;
```
clairemcginty, line 517: Maybe not worth having `BucketItem` be its own class since it's only used once

clairemcginty (commented):
Overall this is really impressive you managed to refactor such a huge piece of code! Users are going to be really happy with this change. I have a few comments regarding code style - unfortunately sometimes we prefer to have simpler/more inlined code over the more elegant, "technically correct" java :) Since it's such a big change, I think I'll take another look tomorrow having slept on it and thought about it, and then approve it 👍

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/MultiSourceKeyGroupReader.java
kellen: I refactored this code out of its original location so the code in the `MultiSourceKeyGroupReader` would be more intelligible. It's private so hopefully not polluting the namespace. Definitely open to suggestions for a better name!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
kellen: Sounds reasonable!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
kellen: IIRC I did this to get the types to be clear e.g. here: 
https://github.com/spotify/scio/pull/3915/files#diff-750c44e50ebc665919ed39fdf777fe6a3607e6ff25386bbfd874eed26956921eR536

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -222,7 +246,7 @@ public void accept(ValueT t) {
     }
   }
 
-  private static class MergedBucket implements Serializable {
+  public static class MergedBucket implements Serializable {
```
clairemcginty, line 249:
```suggestion
static class MergedBucket implements Serializable {
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -72,56 +71,72 @@
...
       TransformFn<FinalKeyT, FinalValueT> transformFn,
+      TransformFnWithSideInputContext<FinalKeyT, FinalValueT> sideInputTransformFn,
       ResourceId outputDirectory,
       ResourceId tempDirectory,
+      Iterable<PCollectionView<?>> sides,
       NewBucketMetadataFn<FinalKeyT, FinalValueT> newBucketMetadataFn,
       FileOperations<FinalValueT> fileOperations,
       String filenameSuffix,
       String filenamePrefix) {
+    assert !((transformFn == null) && (sideInputTransformFn == null)); // at least one defined
```
clairemcginty, line 96: switch to Preconditions like in SortedBucketSource for continuity?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
```diff
@@ -187,7 +202,17 @@ ResourceId getTempDirectoryOrDefault(Pipeline pipeline) {
     @Override
     public WriteResult expand(PBegin input) {
       Preconditions.checkNotNull(outputDirectory, "outputDirectory is not set");
-      Preconditions.checkNotNull(toFinalResultT, "TransformFn<K, V> via() is not set");
+      Preconditions.checkState(
+          toFinalResultT != null || toFinalResultTWithSides != null,
+          "One of TransformFn<K, V> or TransformFnWithSideInputContext<K, V> must be set by via()");
+      Preconditions.checkState(
+          toFinalResultT == null || toFinalResultTWithSides == null,
```
clairemcginty, line 209: small error with this statement since it's an inclusive or!

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -22,12 +22,15 @@ import com.spotify.scio.annotations.experimental
...
+import org.apache.beam.sdk.extensions.smb.SortedBucketTransform.{BucketItem, MergedBucket}
 import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, SortedBucketTransform, TargetParallelism}
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
 import org.apache.beam.sdk.transforms.join.CoGbkResult
+import org.apache.beam.sdk.transforms.windowing.BoundedWindow
 import org.apache.beam.sdk.transforms.{DoFn, ParDo}
-import org.apache.beam.sdk.values.KV
+import org.apache.beam.sdk.values.{KV, PCollectionView}
 
+import java.lang
```
clairemcginty, line 33: is this needed?

clairemcginty (approved):

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
kellen: This needs to be accessible in `SortMergeBucketScioContextSyntax`

clairemcginty (approved):

PR #3924, opened 2021-07-23
(fix #3913) parquet-avro TestIO includes map function

### LINKED ISSUE #3913: Parquet-Avro could be more unit testing friendly
`JobTest` bypasses any map/flatMaps that immediately succeed `parquetAvroFile`. This can cause confusions while writing unit tests.

Example job:
```
...
sc.parquetAvroFile[Account](args("input"), projection, predicate)
.map(x => AccountOutput(x.getId, x.getName.toString))
.saveAsTextFile(args("output"))
```
The JobTest using `ParquetAvroIO` will completely ignore the line `.map(x => AccountOutput(x.getId, x.getName.toString))`

Solutions:
1. Either pass in the final mapped form as an input in the unit tests. (Not elegant)
2. Make sure the map/flatMaps succeeding `parquetAvroFile` are reapplied to the test IO result. (preferred)

What do you think?

clairemcginty:
the way `ParquetAvroIO` is implemented incorporates the subsequently supplied `map` function into the ScioIO itself ([as the projection function](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/syntax/ScioContextSyntax.scala#L70)), so it gets ignored in `JobTest`, which can be confusing/counterintuitive for users. this PR overrides `readTest` to apply the mapping function directly to the test input.

### DIFF
--- a/scio-examples/src/test/scala/com/spotify/scio/examples/extra/ParquetExampleTest.scala
+++ b/scio-examples/src/test/scala/com/spotify/scio/examples/extra/ParquetExampleTest.scala
@@ -42,7 +42,7 @@ class ParquetExampleTest extends PipelineSpec {
       Seq(new Account(1, "checking", "Alice", 1000.0), new Account(2, "checking", "Bob", 1500.0))
 
     val expected = input
-      // .map(x => AccountOutput(x.getId(), x.getName.toString))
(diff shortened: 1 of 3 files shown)


live-wire (approved):
Looks great to me :)

regadas (approved):

PR #3941, opened 2021-08-13
Add outputParallelization for JdbcReadOptions

LeReunionais:
### What
Be able to set `outputParallelization` to false
 https://github.com/apache/beam/pull/7956

### Why
In some cases, reshuffling step after a JDBC read is not necessary: very small dataset or we know that a shuffle will occur right after the JDBC read.


nevillelyh:
Thanks!

PR #3946, opened 2021-08-16
update scalafix to 0.9.30

nevillelyh:


regadas:
don't fully understand why this kills GC lol

regadas:
oh this actually fails a test

```
[info] - should statically check interpolated queries *** FAILED *** (2 seconds, 132 milliseconds)
1891
[info]   Expected no compiler error, but got the following type error: "not found: value sc", for code: 
1892
[info]       def coll: SCollection[Int] = sc.parallelize((1 to 10).toList)
1893
[info]       val r: SCollection[Int] = tsql"SELECT * FROM $coll".as[Int] (TypedBeamSQLTest.scala:118)
1894
[info] - should support inline scollection definition *** FAILED *** (5 milliseconds)
1895
[info]   Expected no compiler error, but got the following type error: "not found: value sc", for code: 
1896
[info]       val a = sc.parallelize(users)
1897
[info]       val r: SCollection[String] =
1898
[info]         tsql"SELECT A._1 FROM ${a.map(u => (u.username, u.age))} A".as[String] (TypedBeamSQLTest.scala:152)
1899
[info] - should not require type ascription *** FAILED *** (5 milliseconds)
1900
[info]   Expected no compiler error, but got the following type error: "not found: value sc", for code: 
1901
[info]       val a = sc.parallelize(users)
1902
[info]       val r = tsql"SELECT A._1 FROM ${a.map(u => (u.username, u.age))} A".as[String] (TypedBeamSQLTest.scala:170)
```

scio-sql/src/test/scala/com/spotify/scio/sql/TypedBeamSQLTest.scala
```diff
@@ -111,7 +111,7 @@ class TypedBeamSQLTest extends PipelineSpec {
     """To.safe[FooWithEnum, FooWithInt]""" shouldNot compile
   }
 
-  "String interpolation" should "statically check interpolated queries" in runWithContext { sc =>
+  "String interpolation" should "statically check interpolated queries" in runWithContext { _ =>
```
regadas, line 114: maybe adding 
```suggestion
  "String interpolation" should "statically check interpolated queries" in runWithContext { _ => // scalafix:ok
```

PR #3949, opened 2021-08-18
make SMB rename more fault-tolerant, fix #3766

### LINKED ISSUE #3766: FileSystems.rename time out in SMB
This potentially affects vanilla IOs too. We can either tweak the time out or enable a way to use these sinks without a temp->rename since we already hash output directory to workaround GCS eventual consistency.

https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java#L655

Beam's `FileBasedSink` seems to handle it in a more "retry tolerant" way, need to look into how these options handle different failure scenarios.
https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/io/FileBasedSink.java#L773

nevillelyh:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -651,15 +647,36 @@ static void moveFiles(
...
-        cleanupTempFiles(e, filesToCleanUp);
-        throw new RuntimeException("Failed to rename temporary files", e);
-      }
-
+      // During a failure case, files may have been deleted in an earlier step. Thus
+      // we ignore missing files here.
+      FileSystems.rename(
+          srcFiles,
+          dstFiles,
+          StandardMoveOptions.IGNORE_MISSING_FILES,
```
clairemcginty, line 655: so would an SMB write ever get into a state where some bucket filse are missing in the final output directory?

nevillelyh: Yeah it's possible If the job is killed after 4 failures of this DoFn. But in that case the job will fail and output shouldn't be pushed to our data catalog.

Also this is assuming GCS rename of single files is atomic, which I believe is the case. So we won't end up with truncated files due to failure.

clairemcginty (approved):

PR #3954, opened 2021-08-24
Defer ResourceId materialization in SMB to get around Flink FileSystems init issue

stormy-ua:
SMB `sortMergeJoin` is failing on FlinkRunner with the following exception:

```
Caused by: java.lang.IllegalArgumentException: No filesystem found for scheme gs
	at org.apache.beam.sdk.io.FileSystems.getFileSystemInternal(FileSystems.java:497)
	at org.apache.beam.sdk.io.FileSystems.matchNewResource(FileSystems.java:567)
	at org.apache.beam.sdk.io.fs.ResourceIdCoder.decode(ResourceIdCoder.java:49)
	at org.apache.beam.sdk.io.fs.ResourceIdCoder.decode(ResourceIdCoder.java:30)
	at org.apache.beam.sdk.coders.IterableLikeCoder.decode(IterableLikeCoder.java:140)
	at org.apache.beam.sdk.extensions.smb.SortedBucketSource$BucketedInput.readObject(SortedBucketSource.java:743)
```

The root cause is that `FileSystems` aren't initialized properly in the FlinkRunner before the pipeline gets translated which triggers serde. On deserialization of `BucketedInput` it fails because BucketedInput deserialization triggers `ResourceId` [serialization](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/io/fs/ResourceIdCoder.java#L49), but the latter requires FileSystems to be initialized. 

The best approach would be to fix this once and for all in the Flink runner itself by adding `FileSystems. setDefaultPipelineOptions` to this method before a pipeline [gets translated](https://github.com/apache/beam/blob/511fb05e5925fe48947efdd872433d621c85e52d/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkRunner.java#L98), but at the same time it looks like other Beam IO transforms don't have this problem because they translate paths from a string to a `ResourceId` during the actual read operations, so there is no problem with the serde of resource ids.

Let me know what you think. The PR is just a draft and require additional work before merging. Creating it just for the initial review. Tested it on a Flink cluster on my local and this seems to resolve the issue.


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadataUtil.java
```diff
@@ -84,14 +84,17 @@ public static PartitionMetadata create(
...
     final int total = directories.size();
     final Map<ResourceId, PartitionMetadata> partitionMetadata = new HashMap<>();
     BucketMetadata<K, V> canonicalMetadata = null;
     ResourceId canonicalMetadataDir = null;
     int start = 0;
     while (start < total) {
-      final List<ResourceId> input = directories.subList(start, Math.min(total, start + batchSize));
+      final List<ResourceId> input = directories
+          .subList(start, Math.min(total, start + batchSize))
+          .stream().map(dir -> FileSystems.matchNewResource(dir, true))
```
regadas, line 96: 🎨 you can move this to the stream bellow and avoid allocation here.

regadas (commented):
This seems reasonable to me!

clairemcginty (approved):
lgtm! and it's nice that this isn't even a breaking change since the user-facing APIs use `String`s already :)

PR #3956, opened 2021-08-25
(fix #3944) ToTable implements selectedFields

### LINKED ISSUE #3944: `@ToTable` support for `selectedFields` for use in BQ Storage API
this was a request from @sisidra to be able to use `@ToTable` with BQ Storage API -- since `@FromStorage` macro requires an authenticated API lookup to generate a companion object, it's more lightweight to be able to just use `@ToTable` when possible. Right now it does _technically work_ with storage API (in that it compiles and runs), but since it doesn't implement [selectedFields](https://github.com/spotify/scio/blob/9375cc0a5416b26c582046ce92330d5de2cff137/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/BigQueryType.scala#L110), no projection is being applied and all columns are fully read.

clairemcginty:
spotify/scio#3944

offers limited BQ Storage support for `BigQueryType.ToTable` for users who don't want to use the credential-required `BigQueryType.FromStorage`. it's limited in that it simply limits the data returned by the API call to the fields specified in the case class; doesn't support row restriction; since as-is, users can already use `typedBigQueryStorage` with a `BigQueryType.ToTable`-annotated class, I think this is reasonable to add as it just makes an already-existing functionality more efficient.

verified with a unit test & by running scio-examples jobs on real BQ data, to verify that only the selected fields are getting read.

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
@@ -195,6 +195,14 @@ private[types] object TypeProvider {
           desc.headOption.map(d => q"override def tableDescription: _root_.java.lang.String = $d")
         val defToPrettyString =
           q"override def toPrettyString(indent: Int = 0): String = ${p(c, s"$SBQ.types.SchemaUtil")}.toPrettyString(this.schema, ${cName.toString}, indent)"
+
+        val fieldValNames = fields.flatMap {
+          case ValDef(_, name, _, _) => Some(name.toString)
+          case _                     => None
+        }
+        val defSelectedFields =
+          q"def selectedFields: _root_.scala.List[_root_.java.lang.String] = $fieldValNames"
+
         val fnTrait =
           tq"${TypeName(s"Function${fields.size}")}[..${fields.map(_.children.head)}, $cName]"
         val traits = (if (fields.size <= 22) Seq(fnTrait) else Seq()) ++ defTblDesc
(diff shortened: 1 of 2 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -195,6 +195,12 @@ private[types] object TypeProvider {
           desc.headOption.map(d => q"override def tableDescription: _root_.java.lang.String = $d")
         val defToPrettyString =
           q"override def toPrettyString(indent: Int = 0): String = ${p(c, s"$SBQ.types.SchemaUtil")}.toPrettyString(this.schema, ${cName.toString}, indent)"
+
+        val defSelectedFields =
+          q"def selectedFields: _root_.scala.List[_root_.java.lang.String] = ${fields.map {
+            case ValDef(_, fname, _, _) => fname.toString
```
nevillelyh, line 201: Haven't looked at the macro code in forever 😂 How does this handle nested fields?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -195,6 +195,12 @@ private[types] object TypeProvider {
           desc.headOption.map(d => q"override def tableDescription: _root_.java.lang.String = $d")
         val defToPrettyString =
           q"override def toPrettyString(indent: Int = 0): String = ${p(c, s"$SBQ.types.SchemaUtil")}.toPrettyString(this.schema, ${cName.toString}, indent)"
+
+        val defSelectedFields =
+          q"def selectedFields: _root_.scala.List[_root_.java.lang.String] = ${fields.map {
```
stormy-ua, line 200: Is `${fields.map(...)}` substituted with `Seq(field1, field2)`? Not sure we get a List on the right side of the expression.

Issue #3772, opened 2021-04-20
Deprecate case-app


jto:
Probably won't work anyway with scala3 and we should just use Beam options.

stormy-ua:
Closed by https://github.com/spotify/scio/pull/3970

PR #3970, opened 2021-08-30
Remove support for the case-app typed arguments parsing #3772

stormy-ua:


nevillelyh (approved):
:tada:

regadas (approved):
Alright, let's do this.

redbaron:
Without case-app, do you have an example how to parse into Option[T] and List[T] ?

Issue #3494, opened 2020-11-20
Deprecate pacakage `com.spotify.scio.extra.bigquery`


regadas:
With the addition of `avro` `read`/`write` support in `Bigquery` there's no need to have these extra conversions and extension methods laying around as they might add confusion to the user.

patil2097:
I would love help to help on closing this. @regadas  Can I work on this? I have to remove all the files for this package ?

PR #3974, opened 2021-08-31
Deprecate pacakage com.spotify.scio.extra.bigquery #3494

stormy-ua:


regadas (commented):
I think we can deprecate as well the converters within the same package.

scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/AvroConverters.scala
```diff
@@ -28,6 +28,10 @@ import scala.jdk.CollectionConverters._
 object AvroConverters extends ToTableRow with ToTableSchema {
 
   @experimental
+  @deprecated(
+    "use saveAsBigQueryTable from the com.spotify.scio.bigquery package instead",
```
clairemcginty, line 32: should we add that it will be removed in 0.12.0? (i think that's what we agreed on...?)

scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -50,6 +50,11 @@ final class AvroToBigQuerySCollectionOps[T <: IndexedRecord: ClassTag](
    * [[com.spotify.scio.bigquery.TableRow TableRow]].
    */
   @experimental
+  @deprecated(
+    "com.spotify.scio.extra.bigquery package will be removed in 0.12.0," +
```
regadas, line 54:
```suggestion
    "com.spotify.scio.extra.bigquery package will be removed in 0.12.0, " +
```

scio-extra/src/main/scala/com/spotify/scio/extra/bigquery/AvroConverters.scala
```diff
@@ -49,6 +54,11 @@ object AvroConverters extends ToTableRow with ToTableSchema {
    *   the equivalent BigQuery schema
    */
   @experimental
+  @deprecated(
+    "com.spotify.scio.extra.bigquery package will be removed in 0.12.0," +
```
regadas, line 58:
```suggestion
    "com.spotify.scio.extra.bigquery package will be removed in 0.12.0, " +
```

PR #3992, opened 2021-09-08
Fix contains in any order matcher

jwiklund:
Matchers.describedAs converts %(index) into the argument at position index ( http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/core/DescribedAs.html#describedAs(java.lang.String,%20org.hamcrest.Matcher,%20java.lang.Object...) ), this will crash if the message contains % followed by a number somewhere (eg as a result of url encoding) because of an index out of bounds exception.


regadas (commented):
Hi @jwiklund Thank you for this!

Do you think you can provide a test for this? 🙏

jwiklund:
@regadas Added.

Issue #3994, opened 2021-09-10
BQ Storage IO do not validate selectedFields before job execution


sisidra:
When invalid field is specified as selectedFields value, job will be sent for execution to runner (Dataflow) and fail at IO execution time.

Preferred behaviour: fail at job validation time, as job graph is constructed, or any other time before job is actually executed.

```scala
import com.spotify.scio.bigquery._
import com.spotify.scio._

val sc = ScioContext()

sc.bigQueryStorage(
    Table.Spec("bigquery-public-data:samples.shakespeare"),
    selectedFields = List("word", "invalid_col")
)

sc.run().waitUntilDone() // fails here on executor
```

Issue #3993, opened 2021-09-09
@ToTable support for nested selectedFields for use in BQ Storage API 


sisidra:
Cont for https://github.com/spotify/scio/issues/3944 ;)

Current implementation supplies top level fields to the BQ Storage API even if selected field is a record with only small subset of nested fields (quite often the case for different export tables).

Expected behaviour is to enumerate only nested fields in `selectedFields`.

sisidra:
Here is test to demonstrate: https://github.com/sisidra/scio/commit/df9e9ef5d701eb7ab1f72f6fe96100532cf9be74#diff-c3f7d7df2dcfd1a7c4a6b1e9084044eee6f2f1d4022646cb6bddc2073bebae3b

clairemcginty:
hi! Unfortunately the Scio offering reflects the limitations of the current [BQ Storage API](https://github.com/googleapis/java-bigquerystorage/blob/d910a8903d0eed6e8a8a4b7183dc7a0f401f66da/proto-google-cloud-bigquerystorage-v1/src/main/proto/google/cloud/bigquery/storage/v1/stream.proto#L59-L64) (what Beam's BigQueryIO uses for Storage reads):

```proto
  message TableReadOptions {
    // Names of the fields in the table that should be read. If empty, all
    // fields will be read. If the specified field is a nested field, all
    // the sub-fields in the field will be selected. The output field order is
    // unrelated to the order of fields in selected_fields.
    repeated string selected_fields = 1;
```

I'm not sure if nested field support is on the roadmap or not

sisidra:
The documentation actually is poor in this regard, I noticed that. ;)

It states that selecting root field will include all child fields, it does not say that child fields can not be specified. 🤷

My tests show that storage API will accept nested fields with notation `"rootF.childF"` and performance will be better than just by specifying root field only.

sisidra:
Here is example with raw BQ Storage API.
There is a caveat tho - currently selecting nested field on repeated record does not seem to work.


```scala
import com.google.cloud.bigquery.storage.v1._
import com.google.cloud.bigquery.storage.v1.ReadSession._
import org.apache.avro._
import org.apache.avro.generic._
import org.apache.avro.io.DecoderFactory;

val client = BigQueryReadClient.create()
val srcTable = "projects/bigquery-public-data/datasets/samples/tables/github_nested"
val options = TableReadOptions.newBuilder()
  .addSelectedFields("payload.name").addSelectedFields("payload.url")
[...]

PR #4002, opened 2021-09-15
BQ Storage IO do not validate selectedFields before job execution #3994

stormy-ua:


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -86,6 +86,20 @@ private object Reads {
     selectedFields: List[String] = BigQueryStorage.ReadParam.DefaultSelectFields,
     rowRestriction: Option[String] = BigQueryStorage.ReadParam.DefaultRowRestriction
   ): SCollection[T] = {
+    val bqClient = client(sc)
```
clairemcginty, line 89:
```suggestion

```

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -86,6 +86,20 @@ private object Reads {
...
+
+    // check if selected fields are defined in the schema of the table
+    if (selectedFields.nonEmpty) {
+      val schemaFields = bqClient.tables.schema(table.ref).getFields.asScala.map(_.getName).toSet
```
clairemcginty, line 93:
```suggestion
      val bqClient = client(sc)
      val schemaFields = bqClient.tables.schema(table.ref).getFields.asScala.map(_.getName).toSet
```

clairemcginty (commented):
I think it would be better to put this in the macro compilation code--it's in the same credentials scope and will catch validation errors before the user tries to execute the pipeline. wdyt?

stormy-ua:
@clairemcginty I was thinking about adding the logic to macro, but the same `bqReadStorage` function is used by the other IO [reading](https://github.com/spotify/scio/blob/73d731d99bef38d1eefcfd5c89e3a44648550ec4/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L397) generic `TableRow` from a table. So we either need to duplicate the code or call the same validation function in two different places. I decided deferring the validation to the   graph creation stage would be ok since it will eventually catch the error with a wrong field. wdyt?

clairemcginty:
@stormy-ua yeah I think this gets complicated because there are so many ways to read from BQ :sob: let me write out my mental picture/expectations of each BigQueryType so we can align on what each one should do

- `@BigQueryType.toTable` - doesn't do any validation/authentication at macro compile time; generates a bigquery ETL job at runtime before launching scio job
- `@BigQueryType.fromTable`/ `@BigQueryType.fromQuery` - requires authentication/validation at macro compile time; generates a bigquery ETL job at runtime
- Generic `TableRow` - no compile-time auth/validation; generates a bigquery ETL job at runtime
- `@BigQueryType.fromStorage` - requires authentication at macro compile time, doesn't need to do a pre-job bq load 

So I feel like it makes sense for only `@BigQueryType.fromStorage` to do this validation when the macro is being compiled (in the `TypeProvider#storageImpl` method); it's consistent with the behavior of the other types IMO and wouldn't change the scope of what user would expect from using `toTable`/`TableRow`. what do you think?

stormy-ua:
@clairemcginty thanks for the overview of the BQ zoo! I think I agree, my only concern now is that in [the related issue](https://github.com/spotify/scio/issues/3994) an example with tableRow was used.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryUtil.scala
```diff
@@ -36,4 +36,22 @@ object BigQueryUtil {
...
+
+  /** Validates if all selectedFields are in the set of schemaFieldNames */
+  def validateSelectedFields(
+    tableSpec: String,
+    schemaFieldNames: Iterable[String],
+    selectedFields: List[String]
+  ): Unit = {
+    // check if selected fields are defined in the schema of the table
+    if (selectedFields.nonEmpty) {
+      val notFoundFields = selectedFields.toSet -- schemaFieldNames.toSet
```
clairemcginty, line 48: looks good! 

I remembered @sisidra left a comment on #3993 that even though the javadoc doesn't state that nested fields are supported in BQ Storage API, it seems that unofficially, it Just Works ™️    using dot-separated nesting syntax. (https://github.com/spotify/scio/issues/3993#issuecomment-916699743)

I feel a bit hesitant because this doesn't seem to be an official/documented feature but I guess this change would then break Kalvans's code... perhaps it's worth either fully supporting nesting in this `validateSelectedFields` function just in case, or at least validating top-level fields? I'm not sure how nested fields are represented in `bigquery.tables.schema(tableSpec).getFields` though, maybe this is a non issue.

stormy-ua: ah, right, we need to check how this would work with nested fields. going to run some experiments.

stormy-ua:
I checked if nested fields are supported and it appears they are indeed supported even though it [isn't documented in the API](https://github.com/googleapis/java-bigquerystorage/blob/4ab9ec6a32995aa6661241eaf24dbd841538e6d2/proto-google-cloud-bigquerystorage-v1beta1/src/main/proto/google/cloud/bigquery/storage/v1beta1/read_options.proto#L24-L27). It looks like there are users who rely on this feature (not sure how many though) and if we merge the PR w/o a logic for handling nested fields it would break their pipelines. Added support for handling nested fields to the validation logic. @clairemcginty @regadas wdyt?

sisidra:
> it doesn't look like a good idea to do the validation on the worker submitting job

This is unfortunate statement and conclusion. It is exactly what I would expect the pipeline to do. Otherwise other inputs (potentially with TBs of data) can be executed before even validation issue is discovered (and I actually do not remember if the error message made sense at runtime).
When it comes to permission issues - I think it can be circumvented with dedicated PipelineOptions parameter to disable this check.

PR #4030, opened 2021-10-05
Add support for multi-files read IOs

stormy-ua:
The PR adds support for multi-files reads for file based IOs.

### Context/Problem

It is possible to read from files in several ways:

* Read from a single file path or file path pattern matching multiple files.
* Read from multiple disjoint directories and union individual read results. Disjoint means there is no way to cover all directories with a single file path pattern.

Today all file based IOs like Avro, Text, etc. don't support reading from multiple file paths/file patterns out-of-the-box. To read from multiple paths scio users usually do:

```scala
val paths: List[String] = //list of paths
sc.unionAll(paths.map(sc.avroFile[T]))
```

Which is ok for the cases when when an individual file path/pattern doesn't match lots of files. According to [Beam javadocs](https://github.com/apache/beam/blob/1d4a9ccd11c14ac6f0a2de1cc438a881244ede0a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java#L187): 

> If it is known that the filepattern will match a very large number of files (e.g. tens of thousands or more), use {@linkRead#withHintMatchesManyFiles} for better performance and scalability. Note that it may decrease performance if the filepattern matches only a small number of files.

Setting `withHintMatchesManyFiles` to true for Beam's AvroIO would effectively replace a single IO read from a filepattern matching lots of files with a sequence of transform which should scale better: `Paths -> FileIO.matchAll()/FileIO.readMatches() -> ReadableFiles -> IO Read`.

### Proposed API
The idea is to add complementary functions for reading multiple files for all single file read IO functions:

```scala
avroFile(path: String) // exists
avroFiles(paths: List[String], withHintMatchesManyFiles: Boolean = false) //new
```
If `withHintMatchesManyFiles` is false then it just unions individual single file IO results otherwise it applies `Paths -> FileIO.matchAll()/FileIO.readMatches() -> ReadableFiles -> IO Read` transform. All file-based Beam IO support `ReadableFiles -> IO Read` reads, so it integrates smoothly with the Beam API. 

### Design Considerations

Due to the current design of `ScioIO` it doesn't make sense to replace single path parameters with a list of paths because all file-based scio IOs combine read and write functionality in a single class and if a single path replaced with the list of paths then it doesn't work for the write part of the IO because write is always directed to a single file prefix/location. At the same time new multi file based read IOs must integrate with scio job test framework. The most reasonable approach (but not the single one) is to create RO IOs for multi-file reads and apply them if an API user sets `withHintMatchesManyFiles` to true otherwise the existing single file IO is used in conjunction with `unionAll`. The other way of implementing would be to avoid new RO multi-file IOs creation and just add this logic directly to the new API functions. The last approach is applied in this [initial commit](https://github.com/spotify/scio/pull/4030/commits/e800ec14f6795fbde0cc689d90a74b5cced451c7), but was abandoned in favor of the former since it is idiomatically more inline with scio IO design and automatically eliminates any potential issues with the job test integration. 

### Performance Testing/Observations

I executed tests for two major use cases:

* Wide, but shallow input file paths. Wide means there are lots of input file patterns, but each file pattern matches less than 10K.
* Narrow, but deep input file paths. Just a few input file patterns, but at least one matches tens of thousands files.

For the first one I didn't observe any performance differences between using `unionAll` of individual IOs and `readFiles` approach (when `withHintMatchesManyFiles` is true).

For the second one a single AvroIO read wasn't able to parse that many file patterns (several 10Ks) and just failed with OOM on a worker. Whereas the `readFiles` (when `withHintMatchesManyFiles` is true) succeeded. 

#### Concerns

Ideally there a should be a way to apply `readFiles`/`withHintMatchesManyFiles` even to existing single file pattern read IOs, because in the case when even a single file pattern matches tens of 10Ks of files it would fail.

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
@@ -85,14 +85,9 @@ final case class ObjectFileIO[T: Coder](path: String) extends ScioIO[T] {
 final case class ObjectFileMultiFileReadIO[T: Coder](paths: List[String]) extends ScioIO[T] {
   override type ReadP = Unit
   override type WriteP = Nothing
-  override val tapT: TapT.Aux[T, T] = TapOf[T]
 
-  /**
-   * Get an SCollection for an object file using default serialization.
-   *
-   * Serialized objects are stored in Avro files to leverage Avro's block file format. Note that
-   * serialization is not guaranteed to be compatible across Scio releases.
-   */
+  override val tapT: TapT.Aux[T, Nothing] = EmptyTapOf[T]
+
   override protected def read(sc: ScioContext, params: ReadP): SCollection[T] = {
     val coder = CoderMaterializer.beamWithDefault(Coder[T])
     sc.parallelize(paths)
@@ -109,10 +104,10 @@ final case class ObjectFileMultiFileReadIO[T: Coder](paths: List[String]) extend
       )
   }
 
-  override def tap(read: ReadP): Tap[T] =
-    ObjectFileMultiFileTap[T](paths.map(path => ScioUtil.addPartSuffix(path)))
+  override protected def write(data: SCollection[T], params: WriteP): Tap[Nothing] =
+    throw new UnsupportedOperationException("ObjectFileMultiFileReadIO is read-only")
 
-  override protected def write(data: SCollection[T], params: Nothing): Tap[T] = ???
+  override def tap(read: ReadP): Tap[tapT.T] = EmptyTap
 }
 
 object ObjectFileIO {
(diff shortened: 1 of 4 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
```diff
@@ -43,9 +49,49 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
...
 
+  /**
+   * Same as [[avroFile]] but reads from a multi-file input. The way how files are read depends on
+   * the hintMatchesManyFiles parameter.
+   *
+   * If the number of files is large (e.g. tens of thousands or more), set hintMatchesManyFiles
+   * to true for better performance and scalability. Note that it may decrease performance if the
+   * number of files is small.
+   */
+//  def avroFiles(paths: List[String],
```
stormy-ua, line 87: had to comment this out for now bc it conflicts the other `avroFiles` overload. They both have default args. I guess the only way to get around this would be to give them different names?

PR #4052, opened 2021-10-19
add unit test for Avro Source+PTransform serialization

clairemcginty:
Tests fixes from BEAM-12628.

The reason I didn't put it in the `CoderTest` suite is bc we encountered this bug not just in the `Coder#encode`/`Coder#decode` that happens between `PTransform`s, but also when reading from `AvroSource` as it [manually sets an AvroCoder](https://github.com/apache/beam/blob/243128a8fc52798e1b58b0cf1a271d95ee7aa241/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroSource.java#L195) (rather than use one from the coder registry). So even after adding the shaded `AvroCoder` fix for Beam < 2.33.0, the bug was still occurring when reading from Avro source.

Verified that this test does fail for Beam < 2.33.0.

I'm not sure about the test naming/package placement; wdyt?

### DIFF
--- a/scio-schemas/src/main/avro/schema.avsc
+++ b/scio-schemas/src/main/avro/schema.avsc
@@ -58,4 +58,19 @@
         {"name": "string_field", "type": ["null", "string"], "default": null},
         {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": null}
     ]
+},
+{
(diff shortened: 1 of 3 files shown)


regadas (approved):
Looks good!

PR #4090, opened 2021-11-12
Whitelist ~/.m2 ~/.ivy2 and ~/.cache/coursier, fixes #4016

### LINKED ISSUE #4016: Scio pipelines should be executable with a classpath referencing a local repository
My team has upgraded from scio 9.3 to scio 10.4 and now we are unable to execute pipelines straight from intellij because our entire `~/.m2` cache is being filtered out in local execution. In the intellij runtime, most classpath jars are in `user.home`. When we try to run pipelines straight from IJ, we get `NoClassDefFoundError` because the entire classpath is filtered out of `PipelineOptions.filesToStage`.

Example exception in a dataflow job:
```
Caused by: java.lang.NoClassDefFoundError: org/apache/beam/sdk/options/PipelineOptions
```

Relevant logs in the pipeline launch process:
```
{"timestamp":"2021-09-21T20:25:51.510Z","level":"INFO","thread":"main","logger":"org.apache.beam.runners.dataflow.util.PackageUtil","message":"Uploading 4 files from PipelineOptions.filesToStage to staging location to prepare for execution.","context":"default"}
```

^ As you can see, only 4 files are uploaded (our classpath should contain ~200 dependencies)

We've tracked this down to the difference in behavior between 9.3 and 10.4. This is the functionality in 9.3
https://github.com/spotify/scio/blob/v0.9.3/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L141

```
CoreSysProps.UserDir.value
> /Users/zacp/path/to/repo/root
```

And this is in 10.4
https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L150

```
sys.props("user.home")
> /Users/zacp
```

So, anything in my `~` directory is not being staged. This includes the entirety of `~/.m2` and `~/.ivy`. We're manually setting our `filesToStage` for now, but it seems like this is possibly undesired default behavior.

kellen:
Fix #4016

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -142,12 +142,16 @@ private object RunnerContext {
     classLoader: ClassLoader
   ): Iterable[String] = {
     val sanitize: String => String = _.replace("\\", "/")
-    val matchesEnvDir: String => Boolean =
-      _.matches(s"${sanitize(sys.props("user.home"))}/\\..+/.+")
-    val classPathJars = PipelineResources
+    val sanitizedHome = sanitize(sys.props("user.home"))
+    val matchesWhitelistDir: String => Boolean = s => s.matches(s"${sanitizedHome}/\\.(ivy2|m2)/.+") || s.matches(s"${sanitizedHome}/\\.cache/coursier/.+")
+    val matchesEnvDir: String => Boolean = _.matches(s"${sanitizedHome}/\\..+/.+")
+    val classPathJars = org.apache.beam.runners.core.construction.resources.PipelineResources
       .detectClassPathResourcesToStage(classLoader, pipelineOptions)
       .asScala
-      .filterNot(sanitize.andThen(matchesEnvDir))
+      .filterNot { s =>
+        val x = sanitize(s)
+        matchesEnvDir(x) && !matchesWhitelistDir(x)
+      }
 
     logger.debug(s"Classpath jars: ${classPathJars.mkString(":")}")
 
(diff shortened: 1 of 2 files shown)


jto (approved):

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -136,19 +136,24 @@ private object RunnerContext {
     finalLocalArtifacts
   }
 
+  private val sanitizePath: String => String = _.replace("\\", "/")
+
+  private[scio] def isNonRepositoryEnvDir(s: String): Boolean = {
+    val sanitizedUserHome: String = sanitizePath(sys.props("user.home"))
+    s.matches(s"${sanitizedUserHome}/\\..+/.+") && !s.matches(
+      s"${sanitizedUserHome}/\\.(ivy2|m2|cache/coursier)/.+"
```
regadas, line 144: We need to include `$HOME/Library/Caches/Coursier` as coursier as different directory on MacOs

regadas (commented):
I can see this being an issue and I was trying to remember why I removed them but ... nothing; I guess it was smth related to `bloop` executions ...

Anyway I guess we can ahead after we address coursier cache path on macos

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
kellen: That won't be caught by the regex though since it's under `Library` and not a dot-directory, right?

regadas: Sorry, you are right I misread this! all good!

regadas (approved):

danielnorberg:
Should we also whitelist `.sbt` ?

On my classpath and returned by `PipelineResources.detectClassPathResourcesToStage`: 

```
336 = "/Users/dano/.sbt/boot/scala-2.12.15/lib/scala-compiler.jar"
337 = "/Users/dano/.sbt/boot/scala-2.12.15/lib/scala-reflect.jar"
338 = "/Users/dano/.sbt/boot/scala-2.12.15/lib/scala-library.jar"
```

But they are not getting staged and execution fails on dataflow. This is when running a job using intellij. Not sure if `sbt runMain` has the same problem.

PR #4091, opened 2021-11-15
Add methods for sortMergeTransform to read additional sources

bkalra23:
- This PR extends sortMergeTransform to read up to 22 sources.
- Modified the smbMultijoin script to add sortMergeTransform.
- Similar past [PR](https://github.com/spotify/scio/pull/3904) for SMBCogroup


scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
```diff
@@ -3347,6 +3350,2622 @@ final class SMBMultiJoin(private val self: ScioContext) {
...
     )
 
+  def sortMergeTransform[KEY, A, B, C, D](
+    keyClass: Class[KEY],
+    a: SortedBucketIO.Read[A],
+    b: SortedBucketIO.Read[B],
+    c: SortedBucketIO.Read[C],
+    d: SortedBucketIO.Read[D],
+    targetParallelism: TargetParallelism
+  ): sortedBucketScioContext.SortMergeTransformReadBuilder[
```
regadas, line 3360: return type should be `SortedBucketScioContext#SortMergeTransformReadBuilder`

scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
```diff
@@ -23,11 +23,14 @@ import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.values._
 import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, TargetParallelism}
+import com.spotify.scio.smb.syntax.SortedBucketScioContext
 
 import scala.jdk.CollectionConverters._
 
 final class SMBMultiJoin(private val self: ScioContext) {
 
+  val sortedBucketScioContext = new SortedBucketScioContext(self)
```
regadas, line 32: this should be `private`. Though, I think `SMBMultiJoin` should have dependency only on `SortedBucketScioContext` avoiding this. It's possible to change it and still keep backwards compat If I'm looking at this correctly.

regadas (changes requested):
Thanks for this @bkalra23 ! Sorry for the late review to it.

scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
bkalra23: Thanks for the feedback @regadas!
Since the SortedBucketScioContext class is final, I couldn't extend it so had to initialize it here. Am I missing something here? Can you please explain what changes can I make here?

clairemcginty: I don't think removing the `final` keyword from `SortedBucketScioContext` would affect binary compatibility (at least in Java...@regadas maybe you know differently for Scala).

regadas: Sorry for the late reply on this! yeah removing `final` wouldn't affect it; here we also don't want to extend the `SortedBucketScioContext`.

After a second look, let's keep like this but making it `private`. Since it's my fault that this is delayed I'll follow up with the changes.

PR #4093, opened 2021-11-16
(fix #4082) numBuckets should be a required SMB param

### LINKED ISSUE #4082: remove default numBuckets for SMB writes?
if `numBuckets` is left unset it defaults to 128:

https://github.com/spotify/scio/search?q=DEFAULT_NUM_BUCKETS

IMO this is confusing and we should just make it a required field. 128 is pretty randomly chosen and for a large dataset is probably too small; and in unit/integration tests it's going to be needlessly big.

thoughts?

clairemcginty:
I don't find this to be a particularly elegant solution (making `getNumBuckets` a boxed `@Nullable` to comply with AutoValue limitations, as we do [a lot of intermediate building](https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroSortedBucketIO.java#L75) before the final `expand` call) but I think it's workable enough and has little user impact. although it's technically a breaking change, I don't expect there are many cases where this is not set!

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketIO.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketIO.java
@@ -58,7 +58,6 @@ public class TensorFlowBucketIO {
    */
   public static <K> Write<K> write(Class<K> keyClass, String keyField) {
     return new AutoValue_TensorFlowBucketIO_Write.Builder<K>()
-        .setNumBuckets(SortedBucketIO.DEFAULT_NUM_BUCKETS)
         .setNumShards(SortedBucketIO.DEFAULT_NUM_SHARDS)
         .setHashType(SortedBucketIO.DEFAULT_HASH_TYPE)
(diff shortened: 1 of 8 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/TensorFlowBucketIO.java
```diff
@@ -58,7 +58,6 @@ public static Read read(TupleTag<Example> tupleTag) {
    */
   public static <K> Write<K> write(Class<K> keyClass, String keyField) {
```
stormy-ua, line 59: I guess another option would be to require `numBuckets` in the Write constructor, so it isn't possible to create a Write w/o numBuckets:

```
public static <K> Write<K> write(Class<K> keyClass, String keyField, int numBuckets)
```

wdyt?

clairemcginty: I thought about that! but I didn't want to break the API too much, and also keep it consistent with other params (such as the `outputDirectory` param which is implemented the same way).

PR #4099, opened 2021-11-19
update GcsConnectorUtil to remove deprecated auth & search for ADC

clairemcginty:
the PR updates a couple things in `setCredentials` (this is called for reads on `DirectRunner` only; it doesn't affect Dataflow).

- If `GOOGLE_APPLICATION_CREDENTIALS` is not set, instead of giving up, using google's oAuth library to find if `~/.config/gcloud/application_default_credentials.json` exists and whether it contains service account credentials.
- If a service account ADC cannot be found use `fs.gs.auth.access.token.provider.impl`, as described in GCS connector's [CONFIGURATION.md](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/CONFIGURATION.md#authentication). I feel iffy about this because I had to define my own `AccessTokenProvider` class using Google credentials; but I verified it works. 
- the "fallback" option of using the default CLOUDSDK_PROJECT client id/secret no longer works and the configuration is now invalid; support was removed in https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/511 .

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
@@ -17,20 +17,39 @@
 
 package com.spotify.scio.parquet
 
+import com.google.auth.oauth2.{GoogleCredentials, ServiceAccountCredentials}
 import com.spotify.scio.ScioContext
 import com.spotify.scio.util.ScioUtil
 import org.apache.hadoop.mapreduce.Job
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat
+import org.slf4j.LoggerFactory
+
+import java.io.File
+import java.util.Locale
+import scala.util.Try
 
 private[parquet] object GcsConnectorUtil {
-  def setCredentials(job: Job): Unit =
+  private lazy val log = LoggerFactory.getLogger(getClass)
+
+  def setCredentials(job: Job): Unit = {
     // These are needed since `FileInputFormat.setInputPaths` validates paths locally and
     // requires the user's GCP credentials.
-    sys.env.get("GOOGLE_APPLICATION_CREDENTIALS") match {
-      case Some(json) =>
-        job.getConfiguration
-          .set("fs.gs.auth.service.account.json.keyfile", json)
+    val keyFile = sys.env
+      .get("GOOGLE_APPLICATION_CREDENTIALS")
+      .filter(_.nonEmpty)
+      .orElse {
+        Try(GoogleCredentials.getApplicationDefault()).toOption match {
+          case Some(_: ServiceAccountCredentials) => getWellKnownCredentialFile.map(_.toString)
+          case _ => None // ADC was not set, or was a personal user credential
+        }
+      }
+
+    keyFile match {
+      case Some(path) => job.getConfiguration.set("fs.gs.auth.service.account.json.keyfile", path)
       case None =>
+        log.warn(
+          "Application default credentials could not be resolved, using default Cloud SDK credentials"
+        )
         // Client id/secret of Google-managed project associated with the Cloud SDK
         job.getConfiguration
           .setBoolean("fs.gs.auth.service.account.enable", false)
@@ -38,11 +57,13 @@ private[parquet] object GcsConnectorUtil {
         job.getConfiguration
           .set("fs.gs.auth.client.secret", "ZmssLNjJy2998hD4CTg2ejr2")
     }
+  }
 
   def unsetCredentials(job: Job): Unit = {
     job.getConfiguration.unset("fs.gs.auth.service.account.json.keyfile")
     job.getConfiguration.unset("fs.gs.auth.service.account.enable")
-    job.getConfiguration.unset("fs.gs.auth.client.id")
+    job.getConfiguration.unset("fs.gs.auth.service.account.enable")
+    job.getConfiguration.unset("fs.gs.auth.null.enable")
     job.getConfiguration.unset("fs.gs.auth.client.secret")
   }
 
@@ -57,5 +78,19 @@ private[parquet] object GcsConnectorUtil {
     if (!ScioUtil.isLocalRunner(sc.options.getRunner)) {
       GcsConnectorUtil.unsetCredentials(job)
     }
+  }
+
+  // Adapted from com.google.auth.oauth2.DefaultCredentialsProvider
+  private def getWellKnownCredentialFile: Option[File] = {
+    val os = sys.props.getOrElse("os.name", "").toLowerCase(Locale.US)
+    val cloudRootPath = if (os.contains("windows")) {
+      new File(sys.env("APPDATA"))
+    } else {
+      new File(sys.props.getOrElse("user.home", ""), ".config")
+    }
+    val credentialFilePath =
+      new File(cloudRootPath, "gcloud/application_default_credentials.json")
+
+    if (credentialFilePath.exists()) Some(credentialFilePath) else None
   }
 }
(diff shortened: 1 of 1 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
```diff
@@ -17,33 +17,52 @@
...
+          "fs.gs.auth.access.token.provider.impl",
+          "com.spotify.scio.parquet.ApplicationDefaultTokenProvider"
+        )
+      case _ =>
         job.getConfiguration
           .setBoolean("fs.gs.auth.service.account.enable", false)
-        job.getConfiguration.set("fs.gs.auth.client.id", "32555940559.apps.googleusercontent.com")
-        job.getConfiguration
-          .set("fs.gs.auth.client.secret", "ZmssLNjJy2998hD4CTg2ejr2")
+        job.getConfiguration.unset("fs.gs.auth.null.enable")
```
clairemcginty, line 57: this option is required to be set to produce a valid config in this case: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/28c40997412991a16bff0e512c6dc6ed9c606874/util/src/main/java/com/google/cloud/hadoop/util/CredentialOptions.java#L111

it's only meant to be used for unit testing, as there are no credentials. (it will log a warning about this on job launch as well)

stormy-ua (approved):
looks good to me!

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
```diff
@@ -17,33 +17,52 @@
...
+  def setCredentials(job: Job): Unit = {
     // These are needed since `FileInputFormat.setInputPaths` validates paths locally and
     // requires the user's GCP credentials.
-    sys.env.get("GOOGLE_APPLICATION_CREDENTIALS") match {
-      case Some(json) =>
-        job.getConfiguration
-          .set("fs.gs.auth.service.account.json.keyfile", json)
-      case None =>
-        // Client id/secret of Google-managed project associated with the Cloud SDK
+    lazy val applicationDefault = Try(GoogleCredentials.getApplicationDefault())
```
regadas, line 36: `Try` is by nature `lazy`; we can remove `lazy` here.

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
```diff
@@ -17,33 +17,52 @@
...
+      .filter(_.nonEmpty)
+      .orElse {
+        applicationDefault.toOption match {
+          case Some(_: ServiceAccountCredentials) => getWellKnownCredentialFile.map(_.toString)
+          case _                                  => None
+        }
+      } match {
+      case Some(serviceAccountKeyFile) =>
+        job.getConfiguration.set("fs.gs.auth.service.account.json.keyfile", serviceAccountKeyFile)
+      case _ if applicationDefault.isSuccess => // ADC exists but isn't a service account
```
regadas, line 49: I would prefer this to be more explicit / easier to read with smth like:
```scala
      case None =>
        Try(GoogleCredentials.getApplicationDefault()) match {
          case Success(_: ServiceAccountCredentials) =>
            job.getConfiguration.set(
              "fs.gs.auth.access.token.provider.impl",
              "com.spotify.scio.parquet.ApplicationDefaultTokenProvider"
            )
          case _ =>
            job.getConfiguration
              .setBoolean("fs.gs.auth.service.account.enable", false)
            job.getConfiguration.unset("fs.gs.auth.null.enable")
        }
```
this avoids the `.orElse` above

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
```diff
@@ -17,33 +17,52 @@
...
+        job.getConfiguration.set(
+          "fs.gs.auth.access.token.provider.impl",
+          "com.spotify.scio.parquet.ApplicationDefaultTokenProvider"
+        )
+      case _ =>
         job.getConfiguration
           .setBoolean("fs.gs.auth.service.account.enable", false)
-        job.getConfiguration.set("fs.gs.auth.client.id", "32555940559.apps.googleusercontent.com")
-        job.getConfiguration
-          .set("fs.gs.auth.client.secret", "ZmssLNjJy2998hD4CTg2ejr2")
```
regadas, line 39: I guess by using ADC now we don't really need this? 😄

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
clairemcginty: that is cleaner to read! It changes the logic a little bit though... in my version if either `GOOGLE_APPLICATION_CREDENTIALS` or `~/.config/gcloud/application_default_credentials.json` exists (_and_ it's a service account credential), then we set `fs.gs.auth.service.account.json.keyfile`. Else, if `~/.config/gcloud/application_default_credentials.json` exists (and it's not a service account credential) then we set `fs.gs.auth.access.token.provider.impl`. So, the reason I had factored out `Try(GoogleCredentials.getApplicationDefault())` into a local variable was to avoid evaluating it twice.

regadas: Ahah I see yes I did miss that! 😄 my comment edit skills are not yet there 😄

regadas: So keeping the prev line of thought we just need to add a `map` there. So smth like this (didn't test).

```
        Try(GoogleCredentials.getApplicationDefault()).map {
          case _: ServiceAccountCredentials => getWellKnownCredentialFile.map(_.toString)
          case _                            => None
        } match {
          case Success(Some(sa)) =>
            job.getConfiguration.set("fs.gs.auth.service.account.json.keyfile", sa) // it's ok to be duped because here it exposes the intent.
          case Success(None) =>
            job.getConfiguration.set(
              "fs.gs.auth.access.token.provider.impl",
              "com.spotify.scio.parquet.ApplicationDefaultTokenProvider"
            )
          case _ =>
            job.getConfiguration.setBoolean("fs.gs.auth.service.account.enable", false)
            job.getConfiguration.unset("fs.gs.auth.null.enable")
        }
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
clairemcginty: yep, plus they [removed support for those configuration properties](https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/511) so it wasn't working anyway 😓

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
clairemcginty: cool, I like that! it does remove the `sys.env.get("GOOGLE_APPLICATION_CREDENTIALS")` call on our end but, that's fine because that's the first thing `GoogleCredentials.getApplicationDefault()` does anyway. However in that case I should also update the `getWellKnownCredentialFile` method to check for a file existing at `$GOOGLE_APPLICATION_CREDENTIALS` location first 👍

scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
clairemcginty: ah this should be `.setBoolean` to true, somehow got messed up refactoring 😅

regadas (approved):

PR #4100, opened 2021-11-19
Deduplicate async lookup requests

### LINKED ISSUE #3711: Avoid duplicated work in `BaseAsyncLookupDoFn`
## The problem 

The current implementation of `processElement`

1. Check if cache existence for `input`
1. If the cache returns value return it
1. otherwise, aquire a semaphore
1. call `asyncLookup` and pass it `input`. It return a future: `F`
1. Maintain a list of the currently executing futures
1. release a semaphore at the end each future execution
1. make sure all the future are completely executed when "finishing" the bundle. 

However, because Beam processes elements per bundle, the same lookup may be executed multiple time before a value is actually cached.

Consider the following example: We lookup some value per `userId`.

We may have to process a bundle a values looking like this:

```
uid0
uid0
uid0
uid1
uid0
uid0
```

Assuming `maxPendingRequests > 6`, all those elements will be processed simultaneously. So the following futures will be created:

- A future doing a lookup with key `uid0`
- A future doing a lookup with key `uid0`
- A future doing a lookup with key `uid0`
- A future doing a lookup with key `uid1`
- A future doing a lookup with key `uid0`
- A future doing a lookup with key `uid0`

We can see the same work is done 5 times. This issue is likely to happen if 
- The dataset being processed is heavily skewed
- lookup execution takes a while

## Proposed solution

Maintain a `Map` of all the currently executing future indexed by `A` and just return the same future all five time, avoiding unnecessary work.
 
## Caveat

Probably need a good retry strategy ?

kellen:
Fixes #3711

This is a minimal attempt at deduplicating async lookup requests, which results in improved but not optimal behavior.

Specifically, there is a race condition which can result in multiple lookups for the same item being in-flight simultaneously, but adding an additional lock for this case seems like unnecessary complexity.


scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -131,30 +133,63 @@ public void processElement(ProcessContext c, BoundedWindow window) {
...
-            },
-            throwable -> {
-              semaphore.release();
-              results.add(new Result(input, failure(throwable), key, c.timestamp(), window));
-              return null;
-            }
-          )
+          key -> {
+            final F future = asyncLookup((C) client.get(instanceId), input);
+            final boolean shouldRemove = future.equals(inFlightRequests.computeIfAbsent(input, k -> future));
```
stormy-ua, line 170: info: why do we need this check here? it seems like we need to always remove a future from inflight map

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -61,6 +61,7 @@
   // Data structures for handling async requests
   private final Semaphore semaphore;
   private final ConcurrentMap<UUID, F> futures = new ConcurrentHashMap<>();
+  private final ConcurrentMap<A, F> inFlightRequests = new ConcurrentHashMap<>();
```
stormy-ua, line 64: if I remember correctly a runner might actually create multiple cloned copies of a DoFn and in this case inFlight requests dedup would be scoped only to a single instance of a DoFn bc these DoFns wouldn't share their in flight requests. I wonder if this is somehow should be changed to static like `cache` defined above

scio-test/src/test/scala/com/spotify/scio/transforms/AsyncLookupDoFnTest.scala
```diff
@@ -69,6 +77,22 @@ class AsyncLookupDoFnTest extends PipelineSpec {
...
+  "BaseAsyncDoFn" should "deduplicate simultaneous lookups on the same item" in {
+    val n = 100
+    val sc = ScioContext(PipelineOptionsFactory.fromArgs("--targetParallelism=1").create())
+    val f = sc
+      .parallelize(List.fill(n)(10))
+      .parDo(new CountingGuavaLookupDoFn)
+      .materialize
+    val result: ScioResult = sc.run().waitUntilFinish()
+    val maxOutput = result
+      .tap(f)
```
stormy-ua, line 89: that's why we need taps!

scio-test/src/test/scala/com/spotify/scio/transforms/AsyncLookupDoFnTest.scala
```diff
@@ -69,6 +77,22 @@ class AsyncLookupDoFnTest extends PipelineSpec {
     ()
   }
 
+  "BaseAsyncDoFn" should "deduplicate simultaneous lookups on the same item" in {
+    val n = 100
+    val sc = ScioContext(PipelineOptionsFactory.fromArgs("--targetParallelism=1").create())
```
stormy-ua, line 82: is there a specific reason the parallelism is being limited to 1?

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
kellen: Because we only insert if there isn't already a future present for `input`; this avoids a specific race case in which a future might be removed 'early' by some other future for the same `input`

scio-test/src/test/scala/com/spotify/scio/transforms/AsyncLookupDoFnTest.scala
kellen: Just to increase contention in the test.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
kellen: That'd work, but in the face of a failure we'd be propagating it to all threads and potentially failing multiple bundles.

stormy-ua: ah, that's a good point. The choice might depend on a specific use case. Would be nice to have this as an option (e.g. `dedupPerWorkerThread`) exposed to users. Not saying it should be added in this PR, just thinking about possible solutions.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
kellen: Yeah, that's the idea, but to avoid too much locking & keeping the implementation simpler, I'm allowing a race where multiple threads can issue requests simultaneously. 

Basically this check keeps the Future around until completion instead of having another earlier-completing request removing it "early" at the cost of a single comparison.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
clairemcginty: This is a good question. In fact, looking at the cache, it's seems to be duplicated per-instance... I wonder if we would get comparable performance benefits from just making the _cache_ [per-class](https://github.com/spotify/scio/blob/d54ffd473501bcb610502636784676a9ecf4dd28/scio-core/src/main/java/com/spotify/scio/transforms/DoFnWithResource.java#L73) rather than per-instance (or in line with @stormy-ua's suggestion, making it configurable by the user)

kellen: I don't think per-class caching works since there can be various subclasses and static fields in abstract classes are shared across all subclasses.

Now thinking that we may want to make this opt-in since users may expect each processed item to side-effect in the external system.

clairemcginty: yeah I was thinking about that; we could possibly work around it. for example, instead of using  `com.spotify.scio.transforms.BaseAsyncLookupDoFn` as the per-class cache key, using something like [CallSites#getCurrentName](https://github.com/spotify/scio/blob/be7166c58450bd59ae5d0048d39ef3ea0d5ed107/scio-core/src/main/scala/com/spotify/scio/util/CallSites.scala#L76) to use file+line number as a key. that way the cache would be unique across different async lookup transforms, but still shared across all instances+clones of that specific transform.

that's a good point about making this opt-in 👍

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
clairemcginty: makes sense... an alternative to the `inFlightRequests` being a `ConcurrentMap<A, F>` would be as a `ConcurrentMap<UUID, Long>` which just maps the Future `UUID` to the # of times you should call `context.output` on in the `flush` method 🤔 . then we could simply remove keys from `inFlightRequests` at the same time you remove them from `futures` and there shouldn't be a race condition (I think, lol). wdyt?

there could be an issue if the same input has a different `Window` in different requests though.

kellen: Eh, I'm disinclined to add any additional complexity. This change uses the rest of the code unchanged which is a benefit IMO.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
kellen: Does that provide any different functionality than what we're offering now? I feel like I'm probably missing something...

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -131,30 +147,65 @@ public void processElement(ProcessContext c, BoundedWindow window) {
     }
 
     final UUID uuid = UUID.randomUUID();
```
jto, line 149: Just a thought: If there was a way to deterministically derive a `UUID` from `input`, `inFlightRequests` would not be required as `futures` would be enough to de-duplicate.

jto: There's probably no need for an `UUID` so a simple hash function would do the trick.

kellen: You mean like uuid3/uuid5, right? Won't we still need to close over the timestamp & window for this particular input?

We'd still need to emit the output multiple times... I guess in that case we could still do the logic in lines 150-178 but with a new random uuid for the deduplicated request, accepting the very low chance of collision between our random uuid and the deterministic ones

stormy-ua:
Merging to include this into the `0.11.2 RC1`.

Issue #1949, opened 2019-06-04
Add avro 1.9 support


regadas:
Initial work has been done here #1946.

Blocked by `Apache Beam` https://issues.apache.org/jira/browse/BEAM-8388.

jto:
I think we need to use the same version of Avro as Beam since `1.8.x` and `1.9.x` are not compatible. See https://issues.apache.org/jira/browse/BEAM-7328

jto:
@regadas seems like this will never happen. Close the issue ?

regadas:
The above jira ticket was superseded by https://issues.apache.org/jira/browse/BEAM-8388.

i can close or we can just add a blocked label here to track things.

PR #4101, opened 2021-11-23
Avro logical types support

RustedBones:
Scio does run with avro model defining logical types as defined in the [avro specification](https://avro.apache.org/docs/1.9.2/spec.html#Logical+Types).

Avro 1.8 generates broken java code which does not allow full support of the feature but avro 1.9.2 works fine.

Since Beam can work with both avro 1.8 and avro 1.9, this is an attempt to run Scio with avro 1.9 and check logical type support


scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -158,8 +169,17 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
...
-    val t = beam.AvroIO.read(cls).from(path)
-    sc.applyTransform(t)
+
+    // Adapt Beam to accept datum writer + coders
+    val source: AvroSource[T] = AvroSource
+      .from(StaticValueProvider.of(path))
+      .withEmptyMatchTreatment(EmptyMatchTreatment.DISALLOW)
+      .withDatumReaderFactory(new SpecificDatumReaderFactory[T]())
+      .withSchema(cls)
+      .withParseFn(_.asInstanceOf[T], AvroCoder.of(cls, false))
```
RustedBones, line 179: Ideally, beam [AvroIO.Read](https://github.com/apache/beam/blob/b78de493a4a5f56cf8339ab2f174cc25cc57beec/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java#L591) should be updated so it can accept a `DatumReaderFactory` and a `Coder`

scio-macros/src/main/scala/com/spotify/scio/coders/AvroCoderMacros.scala
```diff
@@ -34,7 +34,7 @@ private[coders] object AvroCoderMacros {
       _root_.org.apache.beam.sdk.coders.AvroCoder.of[$companioned](
         classOf[$companioned],
         new $companioned().getSchema,
-        true
+        false
```
RustedBones, line 37: This should probably stay `true` so current code doesn't change behaviour if relying on internal coders as explained by @clairemcginty in https://github.com/apache/beam/pull/15616 but the reflect API breaks logical type support.

RustedBones:
See https://github.com/apache/beam/pull/16271

RustedBones:
Would unblock #1949 for the avro 1.9 possible support

RustedBones:
Closing. waiting for upstream fix

PR #4104, opened 2021-11-26
Upgrade Beam to 2.34

jto:
Beam 2.33 depends on a older version of the BQ client which does not RST_STREAM errors correctly when using the BQ storage API.  Depending on the runner, this could either cause *a lot* of bundles to be retried (Dataflow) and generate a ton of error logs or it can completely prevent the job to execute properly (Flink).

Beam 2.34 ships with an updated version of BQ which should fix the issue.

### DIFF
--- a/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/ChannelPoolCreator.java
+++ b/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/ChannelPoolCreator.java
@@ -51,6 +51,6 @@ public class ChannelPoolCreator {
     final ClientInterceptor[] interceptors = getClientInterceptors(options);
 
     return new ChannelPool(
-        () -> BigtableSession.createNettyChannel(options.getAdminHost(), options, interceptors), 1);
+        () -> BigtableSession.createNettyChannel(options.getAdminHost(), options, false, interceptors), 1);
   }
 }
--- a/build.sbt
+++ b/build.sbt
@@ -33,7 +33,7 @@ val autoServiceVersion = "1.0"
 val autoValueVersion = "1.8.2"
 val avroVersion = "1.8.2"
 val beamVendorVersion = "0.1"
-val beamVersion = "2.33.0"
+val beamVersion = "2.34.0"
 val bigdataossVersion = "2.2.2"
 val bigQueryStorageVersion = "1.21.1"
 val bigtableClientVersion = "1.19.1"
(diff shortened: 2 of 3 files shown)


scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/ChannelPoolCreator.java
```diff
@@ -51,6 +51,6 @@ public static ChannelPool createPool(final BigtableOptions options) throws IOExc
     final ClientInterceptor[] interceptors = getClientInterceptors(options);
 
     return new ChannelPool(
-        () -> BigtableSession.createNettyChannel(options.getAdminHost(), options, interceptors), 1);
+        () -> BigtableSession.createNettyChannel(options.getAdminHost(), options, false, interceptors), 1);
```
jto, line 54: I set the `DirectPath` option to `false` bc. it was the default value before the upgrade. Note that I don't actually understand what this option actually does.  `BigtableSession.createNettyChannel` is an internal method and we should not be using it in the first place. On top of that `BigtableSession` has a method to directly create a `ChannelPool` so I don't understand why this class exists at all. @regadas would you knowwhy we have this ?

regadas: Right, things are a bit fuzzy but I believe this was mainly created to support BT admin tasks. Haven't look at BigtableSession in awhile but it's possible that that method was not present when this was created. If it does exist and allows us to set the interceptors as well than we should replace this.

regadas (approved):
LGTM! @jto so just to confirm this new release doesn't have a lot of version bumps right?

clairemcginty (commented):
I think there are some version upgrades to include, just looking at the changes to `BeamModulePlugin.groovy` [here](https://github.com/apache/beam/compare/v2.33.0...v2.34.0): 

- grpcVersion -> 1.40.1
- guavaVersion -> 30.1.1-jre
- httpCoreVersion -> 4.4.14
- jacksonVersion -> 2.12.4
- protobufVersion -> 3.17.3
- googleClientsVersion -> 1.32.1
- commons-codec -> 1.15

and some of the `googleApiServices*Version`s too

jto:
Yeah you're right. I'll should have been more thorough. Setting the PR to draft and I'll come back to it soon.

jto:
Some of the dependencies were already upgraded by Scala Steward to versions newer than what Beam uses. I upgraded the few deps that were not upgraded yet.

PR #4108, opened 2021-11-29
Set value coder for HadoopIO in ParquetTypeIO

lofifnc:
This fixes the issue that HadoopIO as it is utilised by ParquetTypeIO almost always uses `SerializableCoder` for the output. This has performance and compatibility implications. The only way to configure the coder currently is by manually registering one at the beam coder registry.

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
@@ -19,22 +19,17 @@ package com.spotify.scio.parquet.types
 
 import java.lang.{Boolean => JBoolean}
 import com.spotify.scio.ScioContext
-import com.spotify.scio.coders.Coder
+import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.spotify.scio.io.{ScioIO, Tap, TapOf, TapT}
 import com.spotify.scio.parquet.{BeamInputFile, GcsConnectorUtil}
 import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.SCollection
 import magnolify.parquet.ParquetType
-import org.apache.beam.sdk.io.{
-  DefaultFilenamePolicy,
-  DynamicFileDestinations,
-  FileBasedSink,
-  FileSystems,
-  WriteFiles
-}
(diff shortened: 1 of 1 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
```diff
@@ -69,6 +65,11 @@ final case class ParquetTypeIO[T: ClassTag: Coder: ParquetType](path: String) ex
       .withKeyTranslation(new SimpleFunction[Void, JBoolean]() {
         override def apply(input: Void): JBoolean = true
       })
+      .withValueTranslation(new SimpleFunction[T, T]() {
+        override def apply(input: T): T = input
+
+        override def getInputTypeDescriptor: TypeDescriptor[T] = TypeDescriptor.of(cls)
+      }, CoderMaterializer.beam(sc, coder))
```
clairemcginty, line 72: you should be able to just write `CoderMaterializer.beam(sc, Coder[T])` without the implicit summons on line 49, since ParquetTypeIO has context bound `T: Coder`

otherwise, this PR makes sense, thanks for surfacing this! I just want to look into HadoopFormatIO + do some manual verification of this before merging.

stormy-ua: +1, the same approach is actually used in the [ParquetAvroIO](https://github.com/spotify/scio/blob/73d731d99bef38d1eefcfd5c89e3a44648550ec4/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala#L226) so it makes a lot of sense to do the same. Would be interesting to see some read benchmarks w/ and w/o the fix!

stormy-ua: Executed a sample job w/ and w/o coder override and observed difference in the total vCPU hrs:

* 8.764 vCPU hr - w/o coder override
* 3.747 vCPU hr - w/ coder fix

which is quite significant - about 2.3x times better with the fix

lofifnc:
I also would like to take a look at enabling `.withSkipValueClone`. The cloning is a safeguard against emitting mutated objects from `HadoopIO`, but we should be working with immutable data structures most of the time. But I will open another PR if there is any performance to gain. The plan would be to make it configurable for ParquetTypeIO, with a sane default.

stormy-ua (approved):

PR #4120, opened 2021-12-07
first draft of a Dataflow pre-release check

clairemcginty:
the idea is to add a step that we can trigger before releasing that runs some sample jobs in Dataflow, so we can have increased confidence in the release - similar to what Beam does. The jobs use small, fake datasets and only take about 5 min....but it is still a blocking operation for the GHA task, unfortunately. For now I just picked a job from scio-examples to run; eventually we might want to write custom jobs to test (i.e for popular/frequently-evolving IOs like SMB or Parquet). I validated the yaml for this with [act](https://github.com/nektos/act).

### DIFF
--- a/.github/workflows/pre-release-check.yml
+++ b/.github/workflows/pre-release-check.yml
@@ -1,0 +1,40 @@
+name: pre-release-check
+
+env:
+  ENCRYPTION_KEY: ${{ secrets.ENCRYPTION_KEY }}
+  GOOGLE_APPLICATION_CREDENTIALS: scripts/data-integration-test-key.json
+  GOOGLE_PROJECT_ID: data-integration-test
+  SBT_OPTS: -DbeamRunners=DataflowRunner
+  DEFAULT_DATAFLOW_ARGS: --runner=DataflowRunner --project=data-integration-test --region=europe-west1
+
+on:
+  workflow_dispatch # Manually triggered
+jobs:
+  test-dataflow:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.4.0
+      - name: cache SBT
+        uses: coursier/cache-action@v6
+      - name: Java setup
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: ${{matrix.java}}
+      - name: Setup project
+        run: scripts/gha_setup.sh
+      - name: (Dataflow) IO Writes
+        run: sbt ${{ env.SBT_OPTS }} "++${{ matrix.scala }} scio-examples/runMain ${{ matrix.io_write }} ${{ env.DEFAULT_DATAFLOW_ARGS }} "
+      - name: (Dataflow) IO Reads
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-release-check.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: sbt ${{ env.SBT_OPTS }} "++${{ matrix.scala }} scio-examples/runMain ${{ matrix.io_write }} ${{ env.DEFAULT_DATAFLOW_ARGS }} "
+      - name: (Dataflow) IO Reads
+        run: sbt ${{ env.SBT_OPTS }} "++${{ matrix.scala }} scio-examples/runMain ${{ matrix.io_read }} ${{ env.DEFAULT_DATAFLOW_ARGS }}"
+    strategy:
+      matrix:
+        java:
+          - 8
+          - 11
+        scala:
+          - 2.13.6
```
stormy-ua, line 36: do we want to execute it for scala 2.12 also or it will take even more time?

.github/workflows/pre-release-check.yml
```diff
@@ -0,0 +1,40 @@
...
+        io_write:
+          - com.spotify.scio.examples.extra.AvroExample --method=specificOut --output=gs://data-integration-test-eu/AvroExample/Write/${{ env.GITHUB_RUN_ID }}
```
stormy-ua, line 38: would the output require a separate cleanup job?

.github/workflows/pre-release-check.yml
regadas: Yeah I don't think it's worth it tbh. `2.13.6` only is fine and i would test this with 11 only.

.github/workflows/pre-release-check.yml
regadas: I believe we have a retention policy set; we need to double check.

.github/workflows/pre-release-check.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: cache SBT
+        uses: coursier/cache-action@v6
+      - name: Java setup
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: ${{matrix.java}}
+      - name: Setup project
+        run: scripts/gha_setup.sh
+      - name: (Dataflow) IO Writes
```
regadas, line 26: wdyt of keep this simple by not making a distinction? Job name (`--jobName`) could be sufficient here. My main thought is jobs that might do writing & reading

regadas (commented):
I like the idea. We need to make sure though that the entire process will no trip GHA timeouts.

.github/workflows/pre-release-check.yml
clairemcginty: that's a good point! I did it this way because, in this specific AvroExample,  the `--specificOut` configuration needs to run first, since its output will be fed into the `--specificIn` job's input. So there's some notion of dependencies between jobs 🤔 

Maybe we could write specific jobs for these prelease tests that are designed to have self-contained dependencies, like:

```scala
object DataflowPreReleaseAvroTest {
  def main(args: Array[String])
     val (sc, _) = ...
     // do an Avro write
     sc.run.waitUntilDone
 
     val (sc2, _) = ...
     // do an Avro read using the data that was written
     sc2.run.waitUntilDone 
}

PR #4127, opened 2021-12-09
Fix SMBTransform callsite name for Dataflow graph

clairemcginty:
Fixes the callsite name of the top-level SMB Transform from "via" to "sortMergeTransform" so it's less confusing in the Dataflow job graph.

before:
Screen Shot 2021-12-09 at 11 30 04 AM

after:
Screen Shot 2021-12-09 at 11 30 12 AM


clairemcginty:
discovered a bug - will close for the time being.

stormy-ua (approved):

PR #4128, opened 2021-12-09
Clarify ParquetExample inputs and outputs

clairemcginty:
when I was running the `ParquetExample`s in Dataflow I noticed that the `--typedOut` job doesn't create a valid input for `--typedIn`. I cleaned up the input/output types and added some clarifying comments so that the `--{avroSpecific|avroGeneric|typed}In` jobs have valid dependencies on the `--{avro|typed}Out ` jobs.

### DIFF
--- a/scio-examples/src/test/scala/com/spotify/scio/examples/extra/ParquetExampleTest.scala
+++ b/scio-examples/src/test/scala/com/spotify/scio/examples/extra/ParquetExampleTest.scala
@@ -18,74 +18,57 @@ package com.spotify.scio.examples.extra
  */
 
 import com.spotify.scio.avro.Account
-import com.spotify.scio.examples.extra.ParquetExample.{AccountInput, AccountOutput}
+import com.spotify.scio.examples.extra.ParquetExample.{AccountFull, AccountProjection}
 import com.spotify.scio.testing.PipelineSpec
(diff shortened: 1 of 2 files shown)


scio-examples/src/test/scala/com/spotify/scio/examples/extra/ParquetExampleTest.scala
```diff
@@ -18,74 +18,57 @@ package com.spotify.scio.examples.extra
...
-      .input(ParquetTypeIO[AccountInput]("in.parquet"), input)
+      .input(ParquetTypeIO[AccountProjection]("in.parquet"), input)
       .output(TextIO("out.txt"))(coll => coll should containInAnyOrder(expected))
       .run()
   }
 
-  "ParquetExample" should "work for specific output" in {
-    val expected = dummyData
-
+  it should "work for specific output" in {
```
stormy-ua, line 54: maybe reword it to "work for Avro specific output"?

stormy-ua (approved):

scio-examples/src/test/scala/com/spotify/scio/examples/extra/ParquetExampleTest.scala
clairemcginty: good idea! I will 👍

PR #4156, opened 2021-12-21
Update socco-ng to v0.1.6

regadas:


build.sbt
```diff
@@ -1157,6 +1157,7 @@ lazy val siteSettings = Def.settings(
     "org.apache.beam" % "beam-runners-google-cloud-dataflow-java" % beamVersion,
     "com.nrinaudo" %% "kantan.csv" % kantanCsvVersion
   ),
+  dependencyOverrides += "com.lihaoyi" %% "pprint" % "0.6.6",
```
regadas, line 1160: required since `mdoc` still relies on `v0.6.x`

regadas: `mdoc` should be updated soon; then we can remove this.

PR #4157, opened 2021-12-21
Add helper methods for org.scalacheck.Gen to PipelineSpec 

kellen:
Provides a standard way to log the seed on test failure and to handle generation failure.

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
@@ -220,10 +220,10 @@ trait PipelineTestUtils {
         )
         None
       case Some(a) =>
-        val r = Try(fn(a))
+        val r = Try(fn(a)).map(Some(_))
         if (r.isFailure)
           logger.error(s"Failure at ${name.value}:${line.value}. Seed: ${seed.toBase64}")
-        r.map(Option(_)).get
+        r.get
     }
   }
 }
--- a/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
@@ -21,9 +21,15 @@ import org.apache.beam.sdk.options._
 import com.spotify.scio._
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.values.SCollection
+import org.scalacheck.Gen
+import org.scalacheck.rng.Seed
+import org.slf4j.LoggerFactory
+
+import scala.util.Try
 
 /** Trait with utility methods for unit testing pipelines. */
 trait PipelineTestUtils {
+  private[this] val logger = LoggerFactory.getLogger(this.getClass)
 
   /**
    * Test pipeline components with a [[ScioContext]].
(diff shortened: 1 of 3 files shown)


scio-test/src/test/scala/com/spotify/scio/testing/PipelineTestUtilsTest.scala
```diff
@@ -0,0 +1,39 @@
+package com.spotify.scio.testing
+
+import org.scalacheck.Gen
+import org.scalacheck.rng.Seed
+
+class PipelineTestUtilsTest extends PipelineSpec {
+  "PipelineTestUtils" should "generate a test case and have an optional return value" in {
+    val xxx = withGen(Gen.const(1))(i => i)
```
stormy-ua, line 8: maybe just `x` and `y`?

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -173,4 +179,51 @@ trait PipelineTestUtils {
...
+          s"Failed to generate a valid value at ${name.value}:${line.value}. " +
+            s"Consider rewriting Gen instances to be less failure-prone. " +
+            s"Seed: ${seed.toBase64}"
+        )
+        None
+      case Some(a) =>
+        val r = Try(fn(a))
+        if (r.isFailure)
+          logger.error(s"Failure at ${name.value}:${line.value}. Seed: ${seed.toBase64}")
+        r.map(Option(_)).get
```
stormy-ua, line 226: I think `Some(r.get)` would be easier to read

stormy-ua (approved):
👍

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -173,4 +179,51 @@ trait PipelineTestUtils {
...
+    genA.apply(Gen.Parameters.default, seed) match {
+      case None =>
+        logger.error(
+          s"Failed to generate a valid value at ${name.value}:${line.value}. " +
+            s"Consider rewriting Gen instances to be less failure-prone. " +
+            s"Seed: ${seed.toBase64}"
+        )
+        None
+      case Some(a) =>
+        val r = Try(fn(a)).map(Some(_))
```
clairemcginty, line 223: should this be `.map(Option(_))` ? It's unlikely but a successful `fn` call could return null.

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -173,4 +179,51 @@ trait PipelineTestUtils {
...
+   * @return
+   *   The result of `fn` applied to the generated `A`, or `None` on generation failure.
+   */
+  def withGen[A, T](genA: Gen[A], seed: Seed)(
+    fn: A => T
+  )(implicit line: sourcecode.Line, name: sourcecode.Name): Option[T] = {
+    genA.apply(Gen.Parameters.default, seed) match {
+      case None =>
+        logger.error(
+          s"Failed to generate a valid value at ${name.value}:${line.value}. " +
```
clairemcginty, line 217:
```suggestion
          s"$genA failed to generate a valid value at ${name.value}:${line.value}. " +
```

clairemcginty (commented):
this is a cool idea, thanks for adding it!

were you able to validate the [sourcecode](https://github.com/com-lihaoyi/sourcecode) implicits on your end? I tried running the tests in the SBT shell and see `[pool-1-thread-1-ScalaTest-running-PipelineTestUtilsTest] ERROR com.spotify.scio.testing.PipelineTestUtilsTest - Failed to generate a valid value at x:13` 🤔

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
kellen: In that case ... I think `null` would be the expected/correct return value?

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
kellen: This gives non-actionable output e.g.
```
org.scalacheck.Gen$$anon$1@7c10c3de failed to generate a valid value at PipelineTestUtilsTest.scala:14. Consider rewriting Gen instances to be less failure-prone. Seed: 7nn4lp9qpZM6Q51a0l9PzXFV2TEf_VfplLbN9km20ZF=
```

kellen:
Thanks @clairemcginty, I meant to use `FileName`

PR #4162, opened 2021-12-27
Move from log4j to log4j-core (log4shell)

galuszkak:


galuszkak:
Given existing security vulnerability in log4j 1.2.17 ( https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-4104 ) aka log4shell, my patch propose move into log4j-core that is already patched.

regadas (changes requested):
Thanks for this but this is not really adding the referred dependencies only overriding. This is only worth adding if there are transitive deps bringing them in ... I need to double check this. The rm log4j might still be relevant as well

jto:
@regadas I agree with your comment and I think we can just close this PR.

PR #4176, opened 2022-01-06
use google-github-actions/auth in pre-release check

clairemcginty:
Switches to https://github.com/google-github-actions/auth with a GHA-managed secret rather than manually decrypting a service account key. We can switch the other steps to use this too.

proof it works!: https://github.com/spotify/scio/runs/4732201000


.github/workflows/pre-release-check.yml
```diff
@@ -1,28 +1,30 @@
...
+  DEFAULT_DATAFLOW_ARGS: --runner=DataflowRunner --project=data-integration-test --region=europe-west1 --tempLocation=gs://dataflow-tmp-europe-west1/gha
 
 on:
   workflow_dispatch # Manually triggered
 jobs:
   test-dataflow:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
```
regadas, line 14: it seems that if we use SA key setting permissions is not required?

regadas (commented):
I like this! and yeah we should then update the other workflows

.github/workflows/pre-release-check.yml
clairemcginty: you're right! thanks 👍

clairemcginty:
@regadas thanks, updated the other workflows!

PR #4183, opened 2022-01-10
bump gcs-connector to 2.2.4

clairemcginty:
addresses ParquetIO connection hanging issue. cc @stormy-ua


stormy-ua (approved):
🙏 

I hope the new version of bigdata oss won't break something else 😄

PR #4184, opened 2022-01-10
Rework GHA pre-release job setup

clairemcginty:
this re-organizes the pre-release GHA setup to optimize the amount of concurrency and minimize dead time on the GHA worker. GHA doesn't have great built-in support for concurrent execution of tasks, so I created a Scala runnable class that uses vanilla Scala Futures to express this task graph (i.e. Avro read example depends on Avro write example). Also, I added Parquet jobs.

a downside of it is that you can't run individual jobs/parts of the workflow...you have to run the whole thing every time

Tested this out on GHA to verify it works as expected! lmk what you think.

### DIFF
--- a/scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
+++ b/scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
@@ -34,7 +34,7 @@ object RunPreReleaseIT {
           parquet(runId) :+ avro(runId)
         )
         .map(_ => log.info("All Dataflow jobs ran successfully."))
-        .recover(e => log.error("At least one Dataflow job failed", e)),
+        .recover(e => throw new RuntimeException("At least one Dataflow job failed", e)),
       Duration(1, TimeUnit.HOURS)
     )
   }
--- a/scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
+++ b/scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2022 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
```diff
@@ -0,0 +1,96 @@
...
+        .sequence(
+          parquet(runId) :+ avro(runId)
+        )
+        .map(_ => log.info("All Dataflow jobs ran successfully."))
+        .recover(e => throw new RuntimeException("At least one Dataflow job failed", e)),
+      Duration(1, TimeUnit.HOURS)
+    )
+  }
+
+  private def avro(runId: String): Future[Unit] = {
```
stormy-ua, line 42: Minor thing, more like a nitpick. Maybe it's better to return `List` even now when `avro` has only one job? It will create an implicit job execution interface `String => List[Future[Unit]]` so all primitive job executions could be uniformly composed into a single list of futures?

scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
```diff
@@ -0,0 +1,96 @@
...
+    val out1 = gcsPath[AvroExample.type]("specificOut", runId)
+    val out2 = gcsPath[AvroExample.type]("specificIn", runId)
+
+    Future
```
stormy-ua, line 46: So since two invoke jobs are chained sequentially this is equivalent to:

```scala
      Future {
          log.info("Starting Avro IO tests... ")
          invokeJob[AvroExample.type]("--method=specificOut", s"--output=$out1")
          invokeJob[AvroExample.type]("--method=specificIn", s"--input=$out1/*", s"--output=$out2")
     }
)
```

just an observation, I guess you wanted to express the fact that individual avro example jobs are separate execution units in the graph with a sequential dependency.

stormy-ua (approved):
👍 looks great! I really like that it was able to catch the recent Parquet issue!

scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
clairemcginty: good idea! I'll update that.

kellen (approved):
👍

PR #4185, opened 2022-01-10
Un-deprecate AvroConverters

kellen:


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/avro/ToTableRow.scala
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright 2019 Spotify AB.
```
stormy-ua, line 2: should this be 2022?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/avro/AvroConverters.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.spotify.scio.annotations.experimental
+import com.spotify.scio.bigquery.TableRow
+import org.apache.avro.Schema
+import org.apache.avro.generic.IndexedRecord
+
+import scala.jdk.CollectionConverters._
+
+object AvroConverters extends ToTableRow with ToTableSchema {
+  @experimental
+  def toTableRow[T <: IndexedRecord](record: T): TableRow = {
```
clairemcginty, line 13: can you add a Scaladoc for this method similar to the `toTableSchema` one?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/avro/ToTableRow.scala
clairemcginty: \+ add the copyright header to other new classes?

regadas (changes requested):
hm, what's the use case that makes us keep these around? iirc the reason why this was scheduled for removal was because it was added avro Read/Write support for BigQueryIO.

kellen:
@regadas Several internal users including my former team. Primary issue is that the recommended replacement doesn't appear to be able to create a table without manually specifying BQ schema, which can be burdensome. Perhaps I'm missing something obvious?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/avro/ToTableRow.scala
kellen: I'm copying old code, so I don't think so? I have a vague memory that it's not actually required to have the date in the headers anyway.  Will add to other files.

PR #4192, opened 2022-01-12
Update Beam version matrix + include info on Beam dependencies

clairemcginty:


site/src/main/paradox/Apache-Beam.md
```diff
@@ -30,6 +30,7 @@ Also check out the [SDK Version Support Status](https://cloud.google.com/dataflo
 
 | **Scio Version** | **Beam Version** | **Details** |
 |:----------------:|:----------------:|:------------|
+| 0.11.2           | 2.34.0           | This version will be deprecated on November 11, 2022. |
```
stormy-ua, line 33: good catch, forgot to update it in the previous release

PR #4195, opened 2022-01-13
Fix parquet avro generic records read

stormy-ua:
The following primitive job reading Avro Parquet: 

```scala
val schema: Schema = ...
implicit val genericRecordCoder: Coder[GenericRecord] = Coder.avroGenericRecordCoder(schema)
sc.parquetAvroFile[GenericRecord](args("input"), schema)
      .map(identity)
      .count
```

would fail with the following error:

```
Exception in thread "main" org.apache.avro.AvroTypeException: Not an enum: <***EnumVariant***>
	at org.apache.avro.generic.GenericDatumWriter.writeEnum(GenericDatumWriter.java:177)
	at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:119)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:75)
	at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:166)
	at org.apache.avro.generic.GenericDatumWriter.writeRecord(GenericDatumWriter.java:156)
	at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:118)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:75)
	at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:166)
	at org.apache.avro.generic.GenericDatumWriter.writeRecord(GenericDatumWriter.java:156)
	at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:118)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:75)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:62)
	at org.apache.beam.sdk.coders.AvroCoder.encode(AvroCoder.java:363)
	at org.apache.beam.sdk.coders.Coder.encode(Coder.java:136)
	at org.apache.beam.sdk.util.CoderUtils.encodeToSafeStream(CoderUtils.java:85)
	at org.apache.beam.sdk.util.CoderUtils.encodeToByteArray(CoderUtils.java:69)
	at org.apache.beam.sdk.util.CoderUtils.encodeToByteArray(CoderUtils.java:54)
	at org.apache.beam.sdk.util.CoderUtils.clone(CoderUtils.java:144)
	at org.apache.beam.sdk.io.hadoop.format.HadoopFormatIO$HadoopInputFormatBoundedSource$HadoopInputFormatReader.cloneIfPossiblyMutable(HadoopFormatIO.java:983)
	at org.apache.beam.sdk.io.hadoop.format.HadoopFormatIO$HadoopInputFormatBoundedSource$HadoopInputFormatReader.transformKeyOrValue(HadoopFormatIO.java:972)
	at org.apache.beam.sdk.io.hadoop.format.HadoopFormatIO$HadoopInputFormatBoundedSource$HadoopInputFormatReader.getCurrent(HadoopFormatIO.java:946)
	at org.apache.beam.sdk.io.hadoop.format.HadoopFormatIO$HadoopInputFormatBoundedSource$HadoopInputFormatReader.getCurrent(HadoopFormatIO.java:856)
	at org.apache.beam.runners.direct.BoundedReadEvaluatorFactory$BoundedReadEvaluator.processElement(BoundedReadEvaluatorFactory.java:156)
	at org.apache.beam.runners.direct.DirectTransformExecutor.processElements(DirectTransformExecutor.java:165)
```

Root cause: Beam's AvroCoder expects GenericRecords to be encoded using Avro GenericData model, but `parquet-avro` uses `SpecificData` model unless it is configured to use `GenericData` explicitly.

Fix: infer case when GenericRecord is read and apply Parquet specific configs to switch `parquet-avro` to `GenericData` model so it produces GenericData records compatible with Beam's AvroCoder. To configure Parquet two [read config params](https://github.com/apache/parquet-mr/tree/master/parquet-avro) are used: 
 - `parquet.avro.compatible=false`
 -  `parquet.avro.data.supplier=org.apache.parquet.avro.GenericDataSupplier `. 
 
The former forces Parquet to use [AvroRecordConverter](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroRecordConverter.java) instead of [AvroIndexedRecordConverter](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java) and needed because AvroIndexedRecordConverter doesn't support `GenericData` at all. The latter instructs parquet to use `GenericData` model by forcing it to use `org.apache.parquet.avro.GenericDataSupplier`.

The issue was observed for Avro enums: specific data model would deserialize enums as an instance of Java enum, but AvroCoder expects it to be wrapped into [EnumSymbol](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L511) and fails during runtime.

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -26,6 +26,7 @@ import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.SCollection
 import com.twitter.chill.ClosureCleaner
 import org.apache.avro.Schema
+import org.apache.avro.generic.GenericRecord
 import org.apache.avro.reflect.ReflectData
 import org.apache.avro.specific.SpecificRecordBase
 import org.apache.beam.sdk.io.FileBasedSink.FilenamePolicy
@@ -45,7 +46,6 @@ import org.apache.parquet.hadoop.ParquetInputFormat
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
 
 import java.lang.{Boolean => JBoolean}
-
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
 import scala.util.Either
@@ -200,6 +200,16 @@ object ParquetAvroIO {
       job.setInputFormatClass(classOf[AvroParquetInputFormat[T]])
       job.getConfiguration.setClass("key.class", classOf[Void], classOf[Void])
       job.getConfiguration.setClass("value.class", avroClass, avroClass)
+
+      // Needed to make GenericRecord read by parquet-avro work with Beam's
+      // org.apache.beam.sdk.coders.AvroCoder.
+      if (ScioUtil.classOf[A] == classOf[GenericRecord]) {
+        job.getConfiguration.setBoolean("parquet.avro.compatible", false)
+        job.getConfiguration.set(
+          "parquet.avro.data.supplier",
+          "org.apache.parquet.avro.GenericDataSupplier"
+        )
+      }
 
       AvroParquetInputFormat.setAvroReadSchema(job, readSchema)
       if (projection != null) {
--- a/scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
+++ b/scio-examples/src/it/scala/com/spotify/scio/examples/RunPreReleaseIT.scala
@@ -77,6 +77,7 @@ object RunPreReleaseIT {
     val out1 = gcsPath[ParquetExample.type]("avroOut", runId)
     val out2 = gcsPath[ParquetExample.type]("typedIn", runId)
     val out3 = gcsPath[ParquetExample.type]("avroSpecificIn", runId)
+    val out4 = gcsPath[ParquetExample.type]("avroGenericIn", runId)
 
     val write = Future
       .successful(log.info("Starting Parquet IO tests... "))
@@ -91,6 +92,13 @@ object RunPreReleaseIT {
           "--method=avroSpecificIn",
           s"--input=$out1/*",
           s"--output=$out3"
+        )
+      ),
+      write.flatMap(_ =>
+        invokeJob[ParquetExample.type](
+          "--method=avroGenericIn",
+          s"--input=$out1/*",
+          s"--output=$out4"
         )
       )
     )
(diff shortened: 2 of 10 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -201,6 +201,16 @@ object ParquetAvroIO {
       job.getConfiguration.setClass("key.class", classOf[Void], classOf[Void])
       job.getConfiguration.setClass("value.class", avroClass, avroClass)
 
+      // Needed to make GenericRecord read by parquet-avro work with Beam's
+      // org.apache.beam.sdk.coders.AvroCoder.
+      if (ScioUtil.classOf[A] == classOf[GenericRecord]) {
```
clairemcginty, line 206: Maybe we should use `classOf[GenericRecord].isAssignableFrom(ScioUtil.classOf[A])` ? I've seen some pipelines that are typed `GenericData.Record` -  which is the [default implementation of GenericRecord](https://avro.apache.org/docs/1.8.1/api/java/org/apache/avro/generic/GenericData.Record.html).

clairemcginty: actually, nvm, because that would include `SpecificRecord`s too.

clairemcginty (approved):

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
stormy-ua: yeah, it wouldn't work, but I think you concern is valid

stormy-ua: after thinking about it more, I think it is ok to support only GenericRecord bc `GenericData.Record` is a default implementation and there is no guarantee it is returned by `parquet-avro` anyway + since we are reading the interface is more important because a user don't need to construct an instance of a record.

PR #4208, opened 2022-01-14
remove scalacheck Pipeline utils

clairemcginty:
~scalacheck wasn't being exported as a compile dep of `scio-test` and causing issues with the `PipelineTestUtils` trait. this separates the traits & de-scopes scalacheck dependency.~

Removes Scalacheck utils - they will be put into Ratatool instead.


kellen (approved):
👍

stormy-ua (approved):
👍

regadas (changes requested):
I just realized that this was added to `scio-test` and looking at this I don't think it should be added to this module; Since this is not really enhancing scio testing by itself ,only nice helpers to work with scalacheck, these should be added to `scio-extra`.

Since this is broken in 0.11.3 we should take the opportunity to move them. This also avoids pulling scalacheck in scio-test unnecessarily.

regadas (approved):

PR #4210, opened 2022-01-15
Add support for mocking PTransform in tests

kellen:
Useful for testing e.g. AsyncDoFns

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
@@ -17,11 +17,24 @@
 
 package com.spotify.scio.testing
 
-import com.spotify.scio.coders.Coder
+import com.google.common.util.concurrent.{Futures, ListenableFuture}
+import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.spotify.scio.io.ScioIO
+import com.spotify.scio.transforms.DoFnWithResource.ResourceType
+import com.spotify.scio.transforms.{
+  BaseAsyncLookupDoFn,
+  DoFnWithResource,
+  GuavaAsyncDoFn,
+  GuavaAsyncLookupDoFn
+}
 import com.spotify.scio.values.SCollection
 import com.spotify.scio.{ScioContext, ScioResult}
+import org.apache.beam.sdk.coders
 import org.apache.beam.sdk.testing.TestStream
+import org.apache.beam.sdk.transforms.{PTransform, ParDo}
+import org.apache.beam.sdk.values.{KV, PCollection}
+import org.apache.commons.io.output.NullOutputStream
+
 import scala.collection.concurrent.TrieMap
 import scala.collection.mutable.{Set => MSet}
 import scala.jdk.CollectionConverters._
@@ -119,12 +132,68 @@ private[scio] class TestDistCache(val m: Map[DistCacheIO[_], _]) {
   }
 }
 
+private[scio] class TestTransform(val m: Map[String, Map[_, _]]) {
+  val s: MSet[String] =
+    java.util.concurrent.ConcurrentHashMap.newKeySet[String]().asScala
+
+  def apply[T, U](
+    t: MockTransform[T, U],
+    tCoder: coders.Coder[T],
+    uCoder: coders.Coder[U]
+  ): PTransform[_ >: PCollection[T], PCollection[U]] = {
+    val key = t.name
+    require(
+      m.contains(key),
+      s"Missing mock transform: $key, available: ${m.keys.mkString("[", ", ", "]")}"
+    )
+    require(!s.contains(key), s"Mock transform $key has already been used once.")
+    s.add(key)
+    val xxx = m(key)
+    if (xxx.nonEmpty) {
+      try {
+        tCoder.encode(xxx.keys.head.asInstanceOf[T], NullOutputStream.NULL_OUTPUT_STREAM)
+      } catch {
+        case e: ClassCastException =>
+          throw new IllegalArgumentException(
+            s"Input for mock transform $key does not match pipeline transform. Found: ${xxx.keys.head.getClass}",
+            e
+          )
+      }
+      try {
+        uCoder.encode(xxx.values.head.asInstanceOf[U], NullOutputStream.NULL_OUTPUT_STREAM)
+      } catch {
+        case e: ClassCastException =>
+          throw new IllegalArgumentException(
+            s"Output for mock transform $key does not match pipeline transform. Found: ${xxx.values.head.getClass}",
+            e
+          )
+      }
+    }
+    val values = xxx.asInstanceOf[Map[T, U]]
+    val xform: PTransform[_ >: PCollection[T], PCollection[U]] = ParDo.of(
+      new GuavaAsyncDoFn[T, U, None.type]() {
+        override def processElement(input: T): ListenableFuture[U] =
+          Futures.immediateFuture(values(input))
+        override def getResourceType: ResourceType = ResourceType.PER_CLASS
+        override def createResource(): None.type = None
+      }
+    )
+    xform
+  }
+
+  def validate(): Unit = {
+    val d = m.keySet -- s
+    require(d.isEmpty, "Unmatched mock transform: " + d.mkString(", "))
+  }
+}
+
 private[scio] object TestDataManager {
   private val inputs = TrieMap.empty[String, TestInput]
   private val outputs = TrieMap.empty[String, TestOutput]
   private val distCaches = TrieMap.empty[String, TestDistCache]
   private val closed = TrieMap.empty[String, Boolean]
   private val results = TrieMap.empty[String, ScioResult]
+  private val mockedTransforms = TrieMap.empty[String, TestTransform]
 
   private def getValue[V](key: String, m: TrieMap[String, V], ioMsg: String): V = {
     require(m.contains(key), s"Missing test data. Are you $ioMsg outside of JobTest?")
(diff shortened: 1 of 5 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -193,7 +192,24 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+            t.apply[T, U](
+              MockTransform[T, U](name),
+              internal.getCoder,
+              coder
+            )
+          }
+        optXForm.getOrElse(transform)
+      }
+    }
+    ensureSerializable(coder).fold(throw _, pApply(name, xform).setCoder)
```
jto, line 212: Beam already has an API to override transforms in a Pipeline. see [Pipeline.html#replaceAll](https://beam.apache.org/releases/javadoc/2.0.0/org/apache/beam/sdk/Pipeline.html#replaceAll-java.util.List-)

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -119,12 +127,70 @@ private[scio] class TestDistCache(val m: Map[DistCacheIO[_], _]) {
   }
 }
 
+private[scio] class TestTransform(val m: Map[String, Map[_, _]]) {
```
clairemcginty, line 130: Hm... instead of this `TestTransform` class, I think it would be more idiomatic+flexible to work with actual `PTransform`s as much as possible. What if we had a utility class somewhere in the `testing` package that created mock transforms? Like:

```scala
object TransformMocks {
  def forAsyncLookupDoFn[InputT, OutpuT](m: Map[InputT, OutputT]): PTransform[PCollection[InputT], PCollection[KV[InputT, Try[OutputT]]] = {
     // just use a Beam `MapElements` transform to implement
  }

  // mock for Source transforms, could be useful for, say, SMB
  def forSource[InputT](s: Seq[InputT]): PTransform[PBegin, PCollection[InputT]] = {
     // use Beam `Create` transform
  }

  // etc...
}
```

then the `mockTransform` method you added to JobTest could take in an actual `PTransform` as input and use the `Pipeline#replaceAll` method to drop it in, as Julien suggested. (Would prob be clearer to rename from `mockTransform` to `replaceTransform` in that case). wdyt?

stormy-ua (approved):
👍

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -119,12 +122,17 @@ private[scio] class TestDistCache(val m: Map[DistCacheIO[_], _]) {
...
+  def overrideTransforms(pipeline: Pipeline): Unit = pipeline.replaceAll(overrides.toList.asJava)
+}
+
 private[scio] object TestDataManager {
   private val inputs = TrieMap.empty[String, TestInput]
   private val outputs = TrieMap.empty[String, TestOutput]
   private val distCaches = TrieMap.empty[String, TestDistCache]
   private val closed = TrieMap.empty[String, Boolean]
   private val results = TrieMap.empty[String, ScioResult]
+  private val transformOverrides = TrieMap.empty[String, TestTransform]
```
clairemcginty, line 135:
```suggestion
  private val transformOverrides = TrieMap.empty[String, Set[PTransformOverride]]
```

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -171,6 +182,9 @@ private[scio] object TestDataManager {
     require(closed(testId), "ScioContext was not executed. Did you forget .run()?")
     closed -= testId
   }
+
+  def overrideTransforms(testId: String, pipeline: Pipeline): Unit =
+    transformOverrides.get(testId).foreach(_.overrideTransforms(pipeline))
```
clairemcginty, line 187:
```suggestion
    transformOverrides.get(testId).foreach(overrides => pipeline.replaceAll(overrides.toList.asJava))
```

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -583,6 +583,7 @@ class ScioContext private[scio] (
 
   private[scio] def execute(): ScioExecutionContext = {
     val sc = this
+    testId.foreach(tId => TestDataManager.overrideTransforms(tId, pipeline))
```
clairemcginty, line 586:
```suggestion
    if (isTest) {
      TestDataManager.overrideTransforms(tId, pipeline)
    }
```
I know it's functionally the same but this makes the logic more clear imo.

clairemcginty (approved):

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
kellen: We need to unwrap the actual testId, so it'd need to be a pattern match on the option with an empty `None` case ... which should just be a `foreach`. Or use `testId.get` which 🙅

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
kellen: This is equivalent and I'm all for nuking random classes but it'd also break the pattern used by the other parts of `TestDataManager`.

clairemcginty: eh, I think it's ok to break pattern and get rid of the `TestTransform` class, since it doesn't have much functionality outside of holding the Set of PTransform overrides.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
clairemcginty: heh, you'll hate this but there is a lot of precedent for `testId.get` in Scio 😅  I just think it makes the intention of what's going on clearer.

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
kellen: Done.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
kellen: 😢 Done.

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
```diff
@@ -144,17 +148,20 @@ private[scio] object TestDataManager {
     testId: String,
     ins: Map[String, JobInputSource[_]],
     outs: Map[String, SCollection[_] => Any],
-    dcs: Map[DistCacheIO[_], _]
+    dcs: Map[DistCacheIO[_], _],
+    xformOverrides: Set[PTransformOverride]
   ): Unit = {
     inputs += (testId -> new TestInput(ins))
     outputs += (testId -> new TestOutput(outs))
```
jto, line 155: Theoretically, `ins` and `outs` could also be implemented in terms of transform overrides rather than the ugly ugly ugly implementation we have now. Is that something you've tried @kellen ?

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -841,4 +895,106 @@ class JobTestTest extends PipelineSpec {
     }
     e.getMessage should endWith(" was not greater than or equal to 100")
   }
+
+  "transformOverride" should "pass with a source override" in {
+    JobTest[SourceTransformOverrideJob.type]
+      .args("--input=in.txt", "--output=out.txt")
+      .input(TextIO("in.txt"), Seq("1", "2", "3"))
```
jto, line 902: Unless I'm missing smth there's really not point in setting `input` here ?

site/src/main/paradox/Scio-Unit-Tests.md
```diff
@@ -82,3 +82,43 @@ To run the test, we use the `runWithContext`, this will run calculateTeamScores
...
+Sources can also be overridden with `TransformOverride.ofSource`. For example, this source:
+
+@@snip [JobTestTest.scala](scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala) { #JobTestTest_example_4 }
+
+Can be overridden with static mock data:
+
+@@snip [JobTestTest.scala](scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala) { #JobTestTest_example_5 }
+
+Due to type erasure it is possible to provide the incorrect types for the transform and the error will not be caught until runtime.
+Typically a `ClassCastException` will be thrown whose message contains the correct type:
```
jto, line 118: The error message is rather unhelpful. Would it be possible to catch that specific error and provide a more informative message ?

scio-core/src/main/scala/com/spotify/scio/testing/TestDataManager.scala
kellen: No, this is my first attempt at poking at this code, but that does seem possible to do in follow-up work.

scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
jto: I see. I think we can live with that for now. I expect most people won't override "normal" IOs so they should not run into that. Assuming we replace the current `JobTest` implementation with overrides in the future, that problem will be solved.

jto:
This is pretty great and I ma now wondering why we never thought of using overrides to implement that earlier... 
I'm looking forward to a follow up work where this replaces the `JobTest` implementation completely. Hopefully `JobTest` can just become nothing more than a bunch of helpers implemented on top of this, but without any "special" logic to them.
It should allow us to remove all the `if(sc.isTest)` branches we currently have everywhere in the code.

jto:
@regadas would be nice if you could have a look at this too :)

site/src/main/paradox/Scio-Unit-Tests.md
kellen: Sort of. Basically all we have control over is inside the PTransform, so if we construct it such that it knows what type the user asserts it should have in tests, and we witness an element input during the test that isn't assignable, then we can preemptively throw a nice exception. I've added this.

It however doesn't help for anything with type parameters like `KV` which will still 💥 with `ClassCastException`.

Also, we don't have any information about the true output type, either in the test or in the PTransform, so if we return the wrong output type we'll always get a `ClassCastException` from a coder.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -583,6 +583,7 @@ class ScioContext private[scio] (
 
   private[scio] def execute(): ScioExecutionContext = {
     val sc = this
+    if (isTest) TestDataManager.overrideTransforms(testId.get, pipeline)
```
regadas, line 586: :art: here I would move this to https://github.com/deacondesperado/scio/tree/dd1b576a41014f062546a51af63f01024b0fa4e5/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala

regadas (approved):
@kellen this looks great! well needed. Left just a small nit.

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
kellen: @regadas The overrides need to be applied after the entire pipeline is constructed, so I don't think applying the overrides here would work.

PR #4215, opened 2022-01-18
Add parameter to skip cloning in typedParquetFile

lofifnc:
This adds a parameter to `typedParquetFile` toggling skipValueCoder of the underlying `HadoopIO` and enables skipping by default. Disabling cloning spares us from encoding and decoding every element read with the input. 
> **.withSkipValueClone(true)**:
> Hadoop formats typically work with Writable data structures which are mutable and instances
> are reused by the input format reader. Therefore, to not to have elements which can change value
> after they are emitted from read, this IO will clone each key value read from underlying hadoop
> input format (unless they are in the list of well known immutable types). However, in cases where
> used input format does not reuse instances for key/value or translation functions are used which
> already output immutable types, such clone of values can be needless penalty. In these cases IO
> can be instructed to skip key/value cloning. 

Setting it to `true` by default in contrast to the Apache Beam default should be fine, because we are working with immutable case classes.

### Not skipping clone:
` val parquetInput = sc.typedParquetFile[CaseClass]("gs://bucket/path/*.parquet", skipClone = false)`
Screenshot 2022-01-18 at 21 55 35

### Skipping clone:
`val parquetInput = sc.typedParquetFile[CaseClass]("gs://bucket/path/*.parquet", skipClone = true)`
Screenshot 2022-01-18 at 21 55 51

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
@@ -21,7 +21,6 @@ import java.lang.{Boolean => JBoolean}
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.spotify.scio.io.{ScioIO, Tap, TapOf, TapT}
-import com.spotify.scio.parquet.types.ParquetTypeIO.ReadParam.DefaultSkipClone
 import com.spotify.scio.parquet.{BeamInputFile, GcsConnectorUtil}
 import com.spotify.scio.util.ScioUtil
 import com.spotify.scio.values.SCollection
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
@@ -116,8 +116,8 @@ object ParquetTypeIO {
   }
   final case class ReadParam[T] private (
     predicate: FilterPredicate = null,
-    skipClone: Boolean = ReadParam.DefaultSkipClone,
-    conf: Configuration = ReadParam.DefaultConfiguration
+    conf: Configuration = ReadParam.DefaultConfiguration,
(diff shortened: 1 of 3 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
```diff
@@ -109,9 +107,11 @@ object ParquetTypeIO {
   object ReadParam {
     private[types] val DefaultPredicate = null
     private[types] val DefaultConfiguration = new Configuration()
+    private[types] val DefaultSkipClone = true
   }
   final case class ReadParam[T] private (
     predicate: FilterPredicate = null,
+    skipClone: Boolean = ReadParam.DefaultSkipClone,
```
clairemcginty, line 120: I would put the `skipClone` param at the end of the params list, after `conf`! (same for `def typedParquetFile` below)

clairemcginty (commented):
great timing on this PR; we were just discussing testing out this param!

Were you able to test this in Dataflow for `parquetAvroFile` reads? It seems self-evident for Scala case classes that we don't need to worry about them being shared/mutated, but not sure the same holds true for Avro input format.

lofifnc:
> great timing on this PR; we were just discussing testing out this param!
> 
> Were you able to test this in Dataflow for `parquetAvroFile` reads? It seems self-evident for Scala case classes that we don't need to worry about them being shared/mutated, but not sure the same holds true for Avro input format.

Avro SpecificRecords are mutable by design,  to increase performance. Take the Utf8 class for example: [org.apache.avro.util
Class Utf8](https://avro.apache.org/docs/1.4.1/api/java/org/apache/avro/util/Utf8.html). So it shouldn't work on paper.

scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
lofifnc: Done 👍

clairemcginty (approved):
looks good! I tried this out on Dataflow on a slightly larger dataset and saw some resource usage improvement as well. We'll include this in the upcoming 0.11.5 release!

PR #4229, opened 2022-01-29
Missing inEarlyPane assertion

mkuthan:
Just for better alignment with Beam I added missing inEarlyPane assertion.

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
@@ -307,7 +307,7 @@ trait SCollectionMatchers extends EqInstances {
    * early pane for each key.
    */
   def inEarlyPane[T: ClassTag, B: ClassTag](
-      window: BoundedWindow
+    window: BoundedWindow
   )(matcher: MatcherBuilder[T]): Matcher[T] =
     matcher match {
       case value: SingleMatcher[T, _] =>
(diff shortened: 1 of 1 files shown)


scio-test/src/main/scala/com/spotify/scio/testing/SCollectionMatchers.scala
```diff
@@ -302,6 +302,20 @@ trait SCollectionMatchers extends EqInstances {
         value.matcher(_.inLatePane(window))
     }
 
+  /**
+   * SCollection assertion only applied to the specified window, running the checker only on the
+   * early pane for each key.
+   */
+  def inEarlyPane[T: ClassTag, B: ClassTag](
```
regadas, line 309: I believe we can simplify this with following:
```suggestion
  def inEarlyPane[T](
```

regadas (commented):
Thanks for this @mkuthan

Would it be possible to add a test for it in `SCollectionMatchersTest`?

mkuthan:
> Thanks for this @mkuthan
> 
> Would it be possible to add a test for it in `SCollectionMatchersTest`?

@regadas - Scio just delegates to Beam, so it would be rather Beam test for early panes :)

But let me find some spare time for writing the test (perhaps for other pane related assertions as well). I’m starting a one week winter break on Friday so it could be delayed a bit.

regadas:
@mkuthan yeah the idea is not really to test the logic 😄 it was really to just validate the type inference, the same way we do for the others. Anyway is not really a must-have.

regadas (approved):

mkuthan:
> @mkuthan yeah the idea is not really to test the logic 😄 it was really to just validate the type inference, the same way we do for the others. Anyway is not really a must-have.

Thanks! I totally forgot about this PR :( tough week at work after week on skis

PR #4243, opened 2022-02-08
upgrade to beam 2.36.0

clairemcginty:
References:

- https://github.com/apache/beam/blob/v2.36.0/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
- https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/24.2.0/artifact_details.html

Note that this Beam upgrade contains a breaking to `RedisDoFn` by upgrading from `jedis` 3.x to 4.x in beam-sdks-java-io-redis. I referenced Beam's [upgrade code](https://github.com/apache/beam/pull/16287/files) as well as the [Jedis Release Notes]( https://github.com/redis/jedis/blob/v4.0.0/docs/3to4.md) to upgrade Scio's `RedisDoFn`. that said, I'm not super familiar with the Redis API so PTAL.


clairemcginty:
cc @anish749 can you take a look at the Redis changes?

clairemcginty:
Dataflow IT check ran successfully for this branch: https://github.com/spotify/scio/runs/5113877682?check_suite_focus=true

anish749 (approved):

anish749:
This looks good to me, and I don't think it will introduce any regression.

There is one breaking change for scio, which is not a breaking change in Beam, since Scio exposes the `RedisDoFn` for reading, and it exposes the Jedis Transaction now instead of Pipeline (the [p here](https://github.com/spotify/scio/blob/67a5fe6ba20ec41f65804c5962bc28d41c50a62d/scio-examples/src/main/scala/com/spotify/scio/examples/extra/RedisExamples.scala#L155)), so we might want to document that. I don't think Beam exposes this.
In our team, we are having a different implementation of the RedisDoFn. I'll share a doc with our learnings internally.

kellen (approved):
LGTM

Issue #3458, opened 2020-11-05
Double check / refine post execution hooks


nevillelyh:
Something to execute after `sc.run()` with `ScioContext` & `PipelineResult`, for integration hooks i.e. submitting counters to a dashboard, lineage, or updating Bigtable cluster.

jto:
IIRC we removed those hooks in Scio `0.8.x` since all the `Future` were removed. Now you can just do:

```scala
// ...
sc.waitUntilDone()
executePostProcessing()
```

Are you suggesting we reintroduce hooks ?

amfranklin:
Start a thread around this would work with Flyte/Dataflow.

PR #4246, opened 2022-02-09
Update BigtableExample to waitUntilDone

csteegz:
Per #3458 , `sc.run().waitUntilDone()` is the preferred way to run a task post completion, like updating the size of a Bigtable cluster. 
Update the example to wait for a run to complete before scaling down the Bigtable cluster, which would be necessary for a larger topology.


clairemcginty (approved):

regadas (approved):

regadas:
@csteegz Thanks!

PR #4260, opened 2022-02-18
Parquet read: replace HadoopFormatIO with Splittable DoFn IO to improve read performance

stormy-ua:
### Context

We have been getting complaints about sub-optimal Scio Parquet read performance. Even after some attempts to improve the performance by tweaking [HadoopFormatIO](https://beam.apache.org/documentation/io/built-in/hadoop/) (e.g. this [PR](https://github.com/spotify/scio/pull/4108)) it's still below what we would expect.

### Benchmarking

Parquet read performance benchmarks executed by Spotify confirmed performance issues with using [HadoopFormatIO](https://beam.apache.org/documentation/io/built-in/hadoop/) for reading Parquet. At the same time, built-in Beam's [ParquetIO](https://github.com/apache/beam/blob/master/sdks/java/io/parquet/src/main/java/org/apache/beam/sdk/io/parquet/ParquetIO.java) can't be integrated with scio as-is because of coder issues and lack of support for typed reads we do using `magnolify-parquet`.

image

As it can be seen on the graph, scio 0.11.4 Parquet reads are 4-5x times slower than reads with the tweaked Beam's [ParquetIO](https://github.com/apache/beam/blob/master/sdks/java/io/parquet/src/main/java/org/apache/beam/sdk/io/parquet/ParquetIO.java). Beam's Parquet IO was tweaked to work with scio.

### Solution

The PR introduces a generalized implementation of a brand new scio "Parquet read IO" which supports both Avro and typed reads. The change is backwards compatible and doesn't require any changes in a user code. The new Parquet Read IO is implemented as a splittable Do Fn which can split to read multiple row groups from a Parquet file in parallel.

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -20,6 +20,7 @@ package com.spotify.scio.parquet.avro
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.{Coder, CoderMaterializer}
 import com.spotify.scio.io.{ScioIO, Tap, TapOf, TapT}
+import com.spotify.scio.parquet.read.{ParquetRead, ReadSupportFactory}
 import com.spotify.scio.parquet.{BeamInputFile, GcsConnectorUtil}
 import com.spotify.scio.testing.TestDataManager
 import com.spotify.scio.util.ScioUtil
@@ -33,19 +34,18 @@ import org.apache.beam.sdk.io.FileBasedSink.FilenamePolicy
 import org.apache.beam.sdk.io._
 import org.apache.beam.sdk.io.fs.ResolveOptions.StandardResolveOptions
 import org.apache.beam.sdk.io.fs.ResourceId
-import org.apache.beam.sdk.io.hadoop.format.HadoopFormatIO
+import org.apache.beam.sdk.io.hadoop.SerializableConfiguration
 import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider
-import org.apache.beam.sdk.transforms.SimpleFunction
+import org.apache.beam.sdk.transforms.SerializableFunction
 import org.apache.beam.sdk.transforms.windowing.{BoundedWindow, PaneInfo}
-import org.apache.beam.sdk.values.{TypeDescriptor, WindowingStrategy}
+import org.apache.beam.sdk.values.WindowingStrategy
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.mapreduce.Job
-import org.apache.parquet.avro.{AvroParquetInputFormat, AvroParquetReader}
+import org.apache.parquet.avro.{AvroParquetReader, AvroReadSupport}
 import org.apache.parquet.filter2.predicate.FilterPredicate
 import org.apache.parquet.hadoop.ParquetInputFormat
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
 
-import java.lang.{Boolean => JBoolean}
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
 import scala.util.Either
@@ -60,7 +60,7 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
   override protected def read(sc: ScioContext, params: ReadP): SCollection[T] = {
     val coder = CoderMaterializer.beam(sc, Coder[T])
     sc.pipeline.getCoderRegistry.registerCoderForClass(ScioUtil.classOf[T], coder)
-    sc.applyTransform(params.read(sc, path)).map(_.getValue)
+    params.read(sc, path)
   }
 
   override protected def readTest(sc: ScioContext, params: ReadP): SCollection[T] = {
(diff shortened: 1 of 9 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -186,57 +185,44 @@ object ParquetAvroIO {
...
       // org.apache.beam.sdk.coders.AvroCoder.
-      if (ScioUtil.classOf[A] == classOf[GenericRecord]) {
-        job.getConfiguration.setBoolean("parquet.avro.compatible", false)
-        job.getConfiguration.set(
-          "parquet.avro.data.supplier",
+      if (!isSpecific) {
+        hadoopConf.setBoolean(AvroReadSupport.AVRO_COMPATIBILITY, false)
+        hadoopConf.set(
+          AvroReadSupport.AVRO_DATA_SUPPLIER,
           "org.apache.parquet.avro.GenericDataSupplier"
```
jto, line 200: I would maybe replace this to make it a static dependency checked by the compiler.
```suggestion
          classOf[org.apache.parquet.avro.GenericDataSupplier].getCanonicalName
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -0,0 +1,237 @@
...
+      val parquetFileMetadata = reader.getFooter.getFileMetaData
+      val fileSchema = parquetFileMetadata.getSchema
+      val fileMetadata = parquetFileMetadata.getKeyValueMetaData
+      val readSupport = readSupportFactory.readSupport
+
+      val readContext = readSupport.init(
+        new InitContext(
+          hadoopConf,
+          fileMetadata.asScala.map { case (k, v) =>
+            k -> ImmutableSet.of(v).asInstanceOf[JSet[String]]
```
jto, line 105: Why do you need a cast here ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -0,0 +1,237 @@
...
+      splitRowGroupsWithLimit(
+        restriction.getFrom,
+        restriction.getTo,
+        reader.getRowGroups.asScala.toSeq
+      ).foreach(output.output)
+    }.toEither
+
+    reader.close()
+
+    splitResult.fold(throw _, identity)
```
jto, line 76: If the goal is to guarantee `reader.close()` is always executed, I would probably use `try {} finally {}`

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -0,0 +1,237 @@
...
+      "start {} to {}",
+      tracker.currentRestriction.getFrom,
+      tracker.currentRestriction.getTo
+    )
+    val options: ParquetReadOptions = HadoopReadOptions.builder(conf.get()).build
+
+    val reader = ParquetFileReader.open(BeamInputFile.of(file.openSeekable), options)
+    try {
+      val filter = options.getRecordFilter
+      val hadoopConf = options.asInstanceOf[HadoopReadOptions].getConf
```
jto, line 95: Why not use `as` ?
```suggestion
      val hadoopConf = options.as(classOf[HadoopReadOptions]).getConf
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -0,0 +1,237 @@
...
+      reader.getRowGroups.asScala
+        .slice(restriction.getFrom.toInt, restriction.getTo.toInt)
+        .foldLeft((0.0, 0.0)) { case ((count, size), rowGroup) =>
+          (count + rowGroup.getRowCount, size + rowGroup.getTotalByteSize)
+        }
+    }.toEither
+    reader.close()
+
+    countSize.fold(throw _, identity)
+  }
```
jto, line 214: Same remark here. I would use `try {} finally {}`.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -0,0 +1,237 @@
...
+      )
+      reader.setRequestedSchema(readContext.getRequestedSchema)
+
+      val columnIOFactory = new ColumnIOFactory(parquetFileMetadata.getCreatedBy)
+      val recordConverter =
+        readSupport.prepareForRead(hadoopConf, fileMetadata, fileSchema, readContext)
+      val columnIO = columnIOFactory.getColumnIO(readContext.getRequestedSchema, fileSchema, true)
+
+      var currentRowGroupIndex = tracker.currentRestriction.getFrom
+      (0L until currentRowGroupIndex).foreach(_ => reader.skipNextRowGroup())
```
jto, line 118: Depending on the underlying file storage, seeking can be quite an expensive operation. For example seeking on GCS literally means sending a new HTTP request to the GCS API. Not sure if it's possible but ideally it would be better to directly "jump" to the desired row group.

jto (approved):

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -0,0 +1,237 @@
...
+import org.apache.parquet.hadoop.metadata.BlockMetaData
+import org.apache.parquet.io.{ColumnIOFactory, ParquetDecodingException, RecordReader}
+import org.apache.parquet.{HadoopReadOptions, ParquetReadOptions}
+import org.slf4j.LoggerFactory
+
+import java.util.{Set => JSet}
+import scala.jdk.CollectionConverters._
+import scala.util.Try
+
+class ParquetReadFn[T, R](
```
jto, line 23: Why do we need a new SplittableDoFn implementation ? Beam already have https://beam.apache.org/releases/javadoc/2.36.0/org/apache/beam/sdk/io/parquet/ParquetIO.ReadFiles.html which seems to be fairly close.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
stormy-ua: ah, that's a different type of options - [https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/ParquetReadOptions.java)

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ReadSupportFactory.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.apache.parquet.avro.AvroReadSupport
+import org.apache.parquet.hadoop.api.ReadSupport
+
+sealed trait ReadSupportFactory[T] extends Serializable {
+  def readSupport: ReadSupport[T]
+}
+
+object ReadSupportFactory {
+  def typed[T: ParquetType]: ReadSupportFactory[T] = new ReadSupportFactory[T] {
+    def readSupport: ReadSupport[T] = implicitly[ParquetType[T]].readSupport
```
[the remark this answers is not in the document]
stormy-ua: good point

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
stormy-ua: yeah, that's true, but there is no API for jumping to a row group directly or at least I didn't find it. This shouldn't be a big problem because in Spotify we try to set row group size almost equal to the shard size so the result shards have only one row group per file.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
jto: Would adding a type ascription work ?
```suggestion
            k -> (ImmutableSet.of(v): JSet[String])
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
jto: :+1:

clairemcginty (commented):
sorry for how terribly belated this review is. looks great and I love how streamlined the SplittableDoFn implementation looks in Scala!

I guess I need to update this PR to include the splittable dofn change for the `parquet.tensorflow` package?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -186,57 +185,44 @@ object ParquetAvroIO {
...
-          override def getInputTypeDescriptor = TypeDescriptor.of(aCls)
-          override def getOutputTypeDescriptor = TypeDescriptor.of(oCls)
-        })
-        .withConfiguration(job.getConfiguration)
+      val tCoder = sc.pipeline.getCoderRegistry.getCoder(ScioUtil.classOf[T])
+      val cleanedProjectionFn = ClosureCleaner.clean(projectionFn)
+
+      sc.customInput(
+        "AvroParquetIO",
+        ParquetRead.read[A, T](
```
clairemcginty, line 217:
```suggestion
      sc.applyTransform(
        ParquetRead.read[A, T](
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
```diff
@@ -50,36 +48,24 @@ final case class ParquetTypeIO[T: ClassTag: Coder: ParquetType](path: String) ex
...
-          override def apply(input: T): T = input
-
-          override def getInputTypeDescriptor: TypeDescriptor[T] = TypeDescriptor.of(cls)
-        },
-        CoderMaterializer.beam(sc, Coder[T])
+    val coder = CoderMaterializer.beam(sc, implicitly[Coder[T]])
+
+    sc.customInput(
+      "TypedParquetIO",
+      ParquetRead.read(
```
clairemcginty, line 62:
```suggestion
    sc.applyTransform(
      ParquetRead.read(
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/types/ParquetTypeIO.scala
```diff
@@ -112,12 +96,10 @@ object ParquetTypeIO {
   object ReadParam {
     private[types] val DefaultPredicate = null
     private[types] val DefaultConfiguration = new Configuration()
-    private[types] val DefaultSkipClone = true
   }
   final case class ReadParam[T] private (
     predicate: FilterPredicate = null,
-    conf: Configuration = ReadParam.DefaultConfiguration,
-    skipClone: Boolean = ReadParam.DefaultSkipClone
+    conf: Configuration = ReadParam.DefaultConfiguration
```
clairemcginty, line 102: cc @lofifnc - the `skipClone` param is removed in the new splittable DoFn implementation of Parquet, which doesn't rely on `HadoopFormatIO`.

lofifnc: This is great! Sayonara `HadoopFormatIO`!

stormy-ua:
🎉

clairemcginty:
thanks for all your hard work on this @stormy-ua  :)

PR #4262, opened 2022-02-22
staging: allow .sbt/boot

danielnorberg:
Further fix for https://github.com/spotify/scio/issues/4016

Also see https://github.com/spotify/scio/pull/4090

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -141,7 +141,7 @@ private object RunnerContext {
   private[scio] def isNonRepositoryEnvDir(s: String): Boolean = {
     val sanitizedUserHome: String = sanitizePath(sys.props("user.home"))
     s.matches(s"${sanitizedUserHome}/\\..+/.+") && !s.matches(
-      s"${sanitizedUserHome}/\\.(ivy2|m2|cache/coursier|sbt/boot/.*/lib)/.+"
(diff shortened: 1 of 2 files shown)


RustedBones:
@kellen  This solves the issue when:
- `scalaVersion` is not set -> scala version from sbt is used, staging of the jars may not happen
- job run from `repl` -> here we are forced to run with the scala version for sbt

In the `repl` if you inspect the jars to stage with:
```sh
scio> import scala.jdk.CollectionConverters._
scio> org.apache.beam.runners.core.construction.resources.PipelineResources.detectClassPathResourcesToStage(sc.getClass.getClassLoader, sc.options).asScala.foreach(println)
```

We can see that lots of libs in `.sbt/boot`
However, most of them are relevant for `sbt` only, we should probably focus on the `.sbt/boot/scala-*/lib/` subfloder only

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -141,7 +141,7 @@ private object RunnerContext {
   private[scio] def isNonRepositoryEnvDir(s: String): Boolean = {
     val sanitizedUserHome: String = sanitizePath(sys.props("user.home"))
     s.matches(s"${sanitizedUserHome}/\\..+/.+") && !s.matches(
-      s"${sanitizedUserHome}/\\.(ivy2|m2|cache/coursier)/.+"
+      s"${sanitizedUserHome}/\\.(ivy2|m2|cache/coursier|sbt/boot)/.+"
```
RustedBones, line 144: I think we should limit to the `lib` subfolder

RustedBones (changes requested):

PR #4286, opened 2022-03-12
Add projection for parquet example readTest

martinbomio:
Adding projection for Parquet example read, when running unit tests since now the JobTest and the normal reading have different behaviours.


martinbomio:
Closing this one since it was already fixed in https://github.com/spotify/scio/commit/a51d7a094ac8e4b813f64487f3461c8d8c0c664e

Issue #3617, opened 2021-02-17
Secondary sort key for SMB


nevillelyh:
A `V => K2`, used to tie-break when primary bucket key `K` is the same, so that `Iterable<V>` is already sorted on `K2` for the same `K`. This gives us extra flexibility to join by `(K, K2)` in addition to `K`.

An useful example is `K` = user ID and `K2` = event ID, so that we can join an event stream with user metadata, or 2 event streams.

PR #4288, opened 2022-03-17
SMB secondary sort key

kellen:
Adds optional secondary key sort to SMB saves/reads.

The salient sort occurs in SortedBucketSink, from around line 415. This is an additional sort process. We could potentially rework the sorter to do an in-place sort for both keys simultaneously.

This change requires an additional field in all BucketMetadata subtypes which then needs to propagate type parameters to effectively the entire SMB codebase. The BucketMetadata version is incremented, but is backwards compatible.

Usage:
```scala
    sc.sortMergeJoin(
      classOf[String], classOf[Int],
      AvroSortedBucketIO.read(new TupleTag[A]("a"), aSchema).from(aPath),
      AvroSortedBucketIO.read(new TupleTag[B]("b"), bClass).from(bPath)
    ).map {
      case ((k1, k2), (aData, bData)) =>
        // ...
    }
```

Transform:
```scala
    sc.sortMergeTransform(
      classOf[String], classOf[Int],
      AvroSortedBucketIO.read(new TupleTag[A]("a"), aSchema).from(aPath),
      AvroSortedBucketIO.read(new TupleTag[B]("b"), bClass).from(bPath)
    ).via { case ((k1, k2), (aData, bData), outputCollector) =>
      // ...
    }
```


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -378,18 +366,23 @@ public void populateDisplayData(Builder builder) {
...
-   * and can avoid the additional ser/de operation per element.
-   */
-  private static class SortBytesDoFn<K>
-      extends DoFn<KV<K, Iterable<KV<byte[], byte[]>>>, KV<K, Iterable<KV<byte[], byte[]>>>> {
+  private static class SortShardDoFn<K1, K2, V>
+      extends DoFn<KV<BucketShardId, Iterable<V>>, KV<BucketShardId, Iterable<byte[]>>> {
     private final BufferedExternalSorter.Options sorterOptions;
+    private final BucketMetadata<K1, K2, V> bucketMetadata;
+    final Coder<V> valueCoder;
+    private final Comparator<byte[]> bytesComparator = UnsignedBytes.lexicographicalComparator();
```
clairemcginty, line 374: since we've already copied over the `org.apache.beam.sdk.extensions.sorter` package into Scio... what if we implemented a custom Comparator to do primary and secondary sort in one fell swoop (for the case where the Sink has a secondary key present; otherwise we could use this Comparator as-is)? something like [this](https://stackoverflow.com/a/4805676) but for byte[]?

clairemcginty (commented):
this is my first pass at a review since it's a big PR :) largely re-wording suggestions but a couple performance related questions.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -398,22 +391,81 @@ public void populateDisplayData(Builder builder) {
...
       try {
         bucketsInitiatedSorting.inc();
-
-        for (KV<byte[], byte[]> kv : record.getValue()) {
-          sorter.add(kv);
+        // sort by primary key
+        for (V value : record.getValue()) {
+          // TODO is there some better way to avoid this? it's either shuffle or serde here
+          final byte[] valueBytes = CoderUtils.encodeToByteArray(valueCoder, value);
+          final byte[] keyBytesPrimary = bucketMetadata.getKeyBytesPrimary(value);
```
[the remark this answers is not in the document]
kellen: My concern was that we're shuffling bytes unnecessarily if we extract them early, but you're right that parallelism is more limited here. Shuffling must be more expensive both in $ and time than extracting?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
```diff
@@ -398,22 +391,81 @@ public void populateDisplayData(Builder builder) {
...
         }
-        c.output(KV.of(record.getKey(), sorter.sort()));
+        final Iterable<KV<byte[], byte[]>> primarySorted = primarySorter.sort();
 
+        if (!bucketMetadata.hasSecondaryKey()) {
+          // no secondary sort, so discard key and output
+          c.output(KV.of(bucketShardId, Iterables.transform(primarySorted, kv -> kv.getValue())));
+        } else {
+          // secondary key sort
+          byte[] curKey = null;
```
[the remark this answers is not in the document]
kellen: 🙅

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
kellen: We could definitely do that, was thinking about it as follow up work.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java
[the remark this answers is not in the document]
clairemcginty: tough call--possible to get a benchmark testing this? Also, before merging I think we should run benchmarks for a large single-key Sink and Source, before & after this PR.

clairemcginty:
so I tested this out on a large dataset - about 5TB uncompressed

Results:
(1) SMB sink with a secondary sort key specified
Screen Shot 2022-04-04 at 3 16 22 PM

(2) SMB sink with only a primary sort key specified
Screen Shot 2022-04-04 at 4 56 42 PM

(3) Control (SMB sink using latest Scio release, 0.11.5)
Screen Shot 2022-04-04 at 4 12 09 PM

will follow up with Read/Transform tests tomorrow

scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadataTest.java
```diff
@@ -251,58 +275,89 @@ public void testDisplayData() throws Exception {
...
+
+    Assert.assertTrue(metadata2.isPartitionCompatibleForPrimaryKey(metadata3));
+    Assert.assertTrue(metadata2.isPartitionCompatibleForPrimaryAndSecondaryKey(metadata3));
+
+    Assert.assertFalse(metadata3.isPartitionCompatibleForPrimaryKey(metadata4));
+    Assert.assertFalse(metadata3.isPartitionCompatibleForPrimaryAndSecondaryKey(metadata4));
+
+    Assert.assertTrue(metadata4.isPartitionCompatibleForPrimaryKey(metadata5));
+    Assert.assertFalse(metadata4.isPartitionCompatibleForPrimaryAndSecondaryKey(metadata5));
   }
```
clairemcginty, line 361:
```suggestion
  }
  
  @Test
  public void testCompatiblePreSecondaryKey() throws Exception {
    Assert.assertEquals(null,
        BucketMetadata.from("{\"type\":\"org.apache.beam.sdk.extensions.smb.AvroBucketMetadata\","
            + "\"version\":0,"
            + "\"numBuckets\":8192,"
            + "\"numShards\":1,"
            + "\"keyClass\":\"java.lang.String\","
            + "\"hashType\":\"MURMUR3_32\","
            + "\"keyField\":\"some_key_field\""
            + "}").getKeyClassSecondary());
  }

```

scio-smb/src/test/scala/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadataTest.scala
```diff
@@ -253,14 +255,14 @@ class ParquetBucketMetadataTest extends AnyFlatSpec with Matchers {
...
 
-    metadata1.isPartitionCompatible(metadata2) shouldBe false
-    metadata2.isPartitionCompatible(metadata3) shouldBe true
-    metadata3.isPartitionCompatible(metadata4) shouldBe false
+    metadata1.isPartitionCompatibleForPrimaryKey(metadata2) shouldBe false
+    metadata2.isPartitionCompatibleForPrimaryKey(metadata3) shouldBe true
+    metadata3.isPartitionCompatibleForPrimaryKey(metadata4) shouldBe false
   }
 
   it should "check key type" in {
```
clairemcginty, line 263:
```suggestion
  it should "parse older versions of metadata.json without a secondary key field" in {
    BucketMetadata
      .from("""
      {"type":"org.apache.beam.sdk.extensions.smb.ParquetBucketMetadata",
      "version":0,
      "numBuckets":8192,
      "numShards":1,
      "keyClass":"java.lang.String",
      "hashType":"MURMUR3_32",
      "keyField":"some_key_field"
      }""")
      .getKeyClassSecondary shouldBe null
  }

  it should "check key type" in {
```

kellen:
Rough job cost calculations (ignoring discounts):

SINKS:
(1) SMB sink with a secondary sort key specified: $147.81
(2) SMB sink with only a primary sort key specified: $147.45
(3) Control (SMB sink using latest Scio release, 0.11.5): $151.06

SOURCES:
(1) SMB Source with secondary key: $3.67
(2) SMB Source without secondary key: $3.68
(3) Control (SMB source using latest Scio release, 0.11.5): $3.61

scio-smb/src/test/scala/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadataTest.scala
kellen: There are checks for version compatibility on the base `BucketMetadata` [here](https://github.com/spotify/scio/pull/4288/files#diff-2035aae344dac85e32b09945cccfab365a8ce1f9d39f19a400786a9ac6356632R137)

IIRC `getKeyClassSecondary` has to return `null` since old versions won't have anything set ... and i think being consistent with all other fields being null makes most sense.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -35,50 +36,67 @@
...
 
   public AvroBucketMetadata(
       int numBuckets,
       int numShards,
-      Class<K> keyClass,
+      Class<K1> keyClassPrimary,
+      Class<K2> keyClassSecondary,
       BucketMetadata.HashType hashType,
       String keyField,
+      String keyFieldSecondary,
```
RustedBones, line 54: Wouldn't it make more sense to group the class and the field together?
Field ordering is already not really respected with the base class since `keyField` appear before the `filenamePrefix`.

Also, having primary & secondary as prefix would look more natural on the reading and favorises grouping

```suggestion
      Class<K1> primaryKeyClass,
      String primaryKeyField,
      Class<K2> secondaryKeyClass,
      String secondaryKeyField,
      BucketMetadata.HashType hashType,
```

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroSortedBucketIO.java
```diff
@@ -62,55 +61,116 @@ public static <T extends SpecificRecordBase> Read<T> read(
   }
 
   /** Returns a new {@link Write} for Avro generic records. */
-  public static <K> Write<K, GenericRecord> write(
-      Class<K> keyClass, String keyField, Schema schema) {
-    return AvroSortedBucketIO.newBuilder(keyClass, keyField).setSchema(schema).build();
+  public static <K1> Write<K1, Void, GenericRecord> write(
+      Class<K1> keyClassPrimary, String keyFieldPrimary, Schema schema) {
+    return AvroSortedBucketIO.<K1, Void, GenericRecord>newBuilder(
+            keyClassPrimary, null, keyFieldPrimary, null)
```
RustedBones, line 67: As mentioned above, changing the ordering would look more natural
```suggestion
    return AvroSortedBucketIO.<K1, Void, GenericRecord>newBuilder(
            primaryKeyClass, primaryKeyField, null, null)
```

RustedBones (commented):
I'm seeing a lot of nulls in the java code. How is the policy to work with `java.lang.Optional` ?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
kellen: This is all internal so should basically never be constructed by a user; I've just maintained the existing order.

I evaluated using `primaryKeyClass`/`secondaryKeyClass` but the diff was harder to follow since the primary/secondary info is a secondary concern to whether you're operating on the class vs field etc.

kellen:
Optionals are used mainly as return values in java and are substantially hindered by bad api decisions. Also, in interacting with beam code, the default is to use `null`.

scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeSortedBucketIO.scala
```diff
@@ -35,13 +36,25 @@ object ParquetTypeSortedBucketIO {
...
-  def write[K: ClassTag, T: ClassTag: Coder: ParquetType](keyField: String): Write[K, T] =
-    Write(keyField)
+  def write[K: ClassTag, T: ClassTag: Coder: ParquetType](keyField: String): Write[K, Void, T] =
+    Write(keyField, None)
+
+  def write[K1: ClassTag, K2: ClassTag, T: ClassTag: Coder: ParquetType](
+    keyFieldPrimary: String,
+    keyFieldSecondary: String
+  ): Write[K1, K2, T] =
+    Write(keyFieldPrimary, Option(keyFieldSecondary))
```
RustedBones, line 46: It should be safe to use `Some` here instead

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
RustedBones: > the primary/secondary info is a secondary concern whether you're operating on the class vs field.

IMHO we'll actually operate more often at primary/secondary level rather that key/field

Also in term of validity, in the avro and json wold, the keys and fields are strongly coupled and can't live without the other as seen in the assertion bellow, they both belong to the primary/secondary umbrella (which could actually be a key extractor class). For other storage, we could imagine that other property can come into play.

Agreed for the diff. This is already big enough, we should'nt make it more complex

scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeSortedBucketIO.scala
kellen: Yes, but more idiomatic to use `Option` and it'll save you if `null` is passed for some unknown reason.

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
clairemcginty: yeah this is a tricky one. I understand not wanting to make it more complex, but I think Michel's suggested ordering is more in line with standard Scala/Java APIs - for example   [Map](https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/Map.scala#L289) [constructors](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/util/Map.html#of(K,V,K,V,K,V)) are ordered like `k1, v1, k2, v2, ...` not `k1, k2, v1, v2`.

kellen: Done

RustedBones (approved):

kellen:
Failing only mima checks, will add a note in the compatibility docs for 0.12.0 separately

kellen:
Fixes #3617

PR #4294, opened 2022-03-23
Move PopulateTestData script to GHA

clairemcginty:
This is for the test data that we use in our integration tests, for Avro and BQ. So now, we can repopulate them just by running this new GHA action if/when retention period kicks in for our test datasets, which seems to happen periodically :) I split the script across two modules according to where that test data is used.


kellen (approved):
👍

PR #4311, opened 2022-04-12
Add scio-elasticsearch8 module

### LINKED ISSUE #4253: Add scio-elasticsearch8 module

RustedBones:
Fix #4253
Using the [elasticsearch Java API client](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/8.1/introduction.html)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,7 +27,7 @@ ThisBuild / turbo := true
 
 val beamVendorVersion = "0.1"
 val beamVersion = "2.38.0"
-// scala-steward:off
+
 // check version used by beam
 // https://github.com/apache/beam/blob/master/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
 val animalSnifferAnnotationsVersion = "1.20"
@@ -78,7 +78,6 @@ val googleIAMVersion = "1.2.1"
 val grpcVersion = "1.44.0"
 val jacksonVersion = "2.13.1"
 val protobufVersion = "3.19.3"
-// scala-steward:on
 
 val algebirdVersion = "0.13.9"
 val algebraVersion = "2.7.0"
@@ -86,6 +85,8 @@ val annoy4sVersion = "0.10.0"
 val annoyVersion = "0.2.6"
 val breezeVersion = "2.0"
 val caffeineVersion = "2.9.3"
+val cassandraDriverVersion = "3.11.2"
+val cassandraVersion = "3.11.12"
 val catsVersion = "2.7.0"
 val chillVersion = "0.10.0"
 val circeVersion = "0.14.1"
@@ -112,11 +113,10 @@ val metricsVersion = "3.2.6"
 val opencensusVersion = "0.28.0"
 val parquetExtraVersion = "0.4.3"
 val parquetVersion = "1.12.2"
-val pprintVersion = "0.7.1"
+val pprintVersion = "0.7.3"
 val protobufGenericVersion = "0.2.9"
 val scalacheckVersion = "1.16.0"
 val scalaCollectionCompatVersion = "2.7.0"
-val scalatestplusVersion = "3.1.0.0-RC2"
 val scalacticVersion = "3.2.11"
 val scalaMacrosVersion = "2.1.1"
 val scalatestVersion = "3.2.12"
@@ -126,6 +126,8 @@ val slf4jVersion = "1.7.36"
 val sparkeyVersion = "3.2.2"
 val tensorFlowVersion = "0.4.1"
 val zoltarVersion = "0.6.0"
+// dependent versions
+val scalatestplusVersion = s"$scalatestVersion.0"
 
 ThisBuild / scalafixScalaBinaryVersion := CrossVersion.binaryScalaVersion(scalaVersion.value)
 val excludeLint = SettingKey[Set[Def.KeyedInitialize[_]]]("excludeLintKeys")
@@ -516,7 +518,7 @@ lazy val `scio-test`: Project = project
       "org.apache.beam" % "beam-sdks-java-core" % beamVersion % "test",
       "org.apache.beam" % "beam-sdks-java-core" % beamVersion % "test" classifier "tests",
       "org.scalatest" %% "scalatest" % scalatestVersion,
-      "org.scalatestplus" %% "scalatestplus-scalacheck" % scalatestplusVersion % "test,it",
+      "org.scalatestplus" %% "scalacheck-1-16" % scalatestplusVersion % "test,it",
       "org.scalacheck" %% "scalacheck" % scalacheckVersion % "test,it",
       "com.spotify" %% "magnolify-datastore" % magnolifyVersion % "it",
       "com.spotify" %% "magnolify-guava" % magnolifyVersion,
@@ -590,7 +592,7 @@ lazy val `scio-avro`: Project = project
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-simple" % slf4jVersion % "test,it",
       "org.scalatest" %% "scalatest" % scalatestVersion % "test,it",
-      "org.scalatestplus" %% "scalatestplus-scalacheck" % scalatestplusVersion % "test,it",
+      "org.scalatestplus" %% "scalacheck-1-16" % scalatestplusVersion % "test,it",
       "org.scalacheck" %% "scalacheck" % scalacheckVersion % "test,it",
       "com.spotify" %% "magnolify-cats" % magnolifyVersion % "test",
       "com.spotify" %% "magnolify-scalacheck" % magnolifyVersion % "test"
@@ -652,7 +654,7 @@ lazy val `scio-google-cloud-platform`: Project = project
       "org.hamcrest" % "hamcrest-library" % hamcrestVersion % "test",
       "org.scalacheck" %% "scalacheck" % scalacheckVersion % "test,it",
       "org.scalatest" %% "scalatest" % scalatestVersion % "test,it",
-      "org.scalatestplus" %% "scalatestplus-scalacheck" % scalatestplusVersion % "test,it",
+      "org.scalatestplus" %% "scalacheck-1-16" % scalatestplusVersion % "test,it",
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-simple" % slf4jVersion % "test,it"
     )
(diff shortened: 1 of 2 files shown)


scio-elasticsearch/es8/src/main/scala/com/spotify/scio/elasticsearch/CoderInstances.scala
```diff
@@ -0,0 +1,29 @@
...
+package com.spotify.scio.elasticsearch
+
+import co.elastic.clients.elasticsearch.core.bulk.BulkOperation
+import com.spotify.scio.coders._
+
+trait CoderInstances {
+
+  // BulkOperation only has JsonpSerializable, not @JsonpDeserializable
+  // use kryo instead
+  implicit val bulkOperationCoder: Coder[BulkOperation] = Coder.kryo[BulkOperation]
```
RustedBones, line 27: There is no easy way to deserialize the BulkOperation, fallback with kryo

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,755 @@
...
+
+    /**
+     * An exception that puts information about the failures in the bulk execution.
+     */
+    public static class BulkExecutionException extends IOException {
+
+      private final Iterable<ErrorCause> failures;
+
+      BulkExecutionException(BulkResponse bulkResponse) {
+        // super(bulkResponse.items()); TODO
```
RustedBones, line 718: This does not extend `Throwable` anymore.
Should we at least provide a custom message ?

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,755 @@
...
+      public void finishBundle() throws Exception {
+        flush();
+      }
+
+      @ProcessElement
+      public void processElement(ProcessContext c) throws Exception {
+        final T t = c.element();
+        final Iterable<BulkOperation> operations = toBulkOperations.apply(t);
+        for (BulkOperation operation : operations) {
+          // long bytes = documentSize(operation);
```
RustedBones, line 499: Since the operations are not serialized yet, it is hard to know the size at this point

RustedBones: removing

RustedBones:
Related integration test draft in #4313

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
clairemcginty: hm, it looks like it still has a [buildFailureMessage](https://github.com/elastic/elasticsearch/blob/v8.1.2/server/src/main/java/org/elasticsearch/action/bulk/BulkResponse.java#L98) method we can use?

build.sbt
```diff
@@ -51,6 +51,7 @@ val datastoreV1ProtoClientVersion = "1.6.3"
 // scala-steward:off
 val elasticsearch6Version = "6.8.23"
 val elasticsearch7Version = "7.17.1"
+val elasticsearch8Version = "8.1.2"
```
clairemcginty, line 54: I think we can leave this outside of the `// scala-steward:off` block :) I originally added that because scala-steward kept trying to bump the major versions of ES6 and ES7, but at least until ES9 is released, we can let it update minor versions of ES8.

build.sbt
```diff
@@ -699,7 +701,27 @@ lazy val `scio-elasticsearch7`: Project = project
...
+      "org.elasticsearch" % "elasticsearch" % elasticsearch7Version,
+      "org.scalatest" %% "scalatest" % scalatestVersion % "test"
+    )
+  )
+  .dependsOn(
+    `scio-core`,
+    `scio-test` % "test"
+  )
+
+lazy val `scio-elasticsearch8`: Project = project
```
clairemcginty, line 713: remember to add this to the root [aggregate](https://github.com/spotify/scio/blob/main/build.sbt#L371) so it gets published!

clairemcginty (commented):
this is only the first half of the PR but looks good so far! have a few minor comments. Will continue with the Scala half soon :)

build.sbt
```diff
@@ -675,7 +676,8 @@ lazy val `scio-elasticsearch6`: Project = project
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.elasticsearch" % "elasticsearch" % elasticsearch6Version,
       "org.elasticsearch" % "elasticsearch-x-content" % elasticsearch6Version,
-      "org.elasticsearch.client" % "transport" % elasticsearch6Version
+      "org.elasticsearch.client" % "transport" % elasticsearch6Version,
+      "org.scalatest" %% "scalatest" % scalatestVersion % "test"
```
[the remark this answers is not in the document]
RustedBones: This is not 'needed' per say. This is a direct dependency, do we prefer this to be pulled transitively by `scio-test` ?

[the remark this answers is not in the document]
regadas: if it's a direct dep then it's good to have it declared! 👍

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,711 @@
...
+            FluentBackoff.DEFAULT
+                .withMaxRetries(this.maxRetries)
+                .withInitialBackoff(this.retryPause);
+        this.requestFn = request(clientSupplier, error);
+        this.retryFn = retry(requestFn, backoffConfig);
+      }
+
+      @Teardown
+      public void teardown() throws Exception {
+        this.clientSupplier.close();
```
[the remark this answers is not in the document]
RustedBones: We can't close the client. We need to close the underlying transport. I think we need to change the suplier to return the transport instead. See [here](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/api-conventions.html#_object_life_cycles)

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
RustedBones: This isn't available on the Client Java API: [BulkResponse](https://github.com/elastic/elasticsearch-java/blob/main/java-client/src/main/java/co/elastic/clients/elasticsearch/core/BulkResponse.java).

RustedBones: created our own `buildFailureMessage` on the same model

RustedBones:
Adding back `maxBulkRequestBytes` by checking serialized size of bulk operations. Looks to be required by #1829

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,791 @@
...
+     * Returns a transform for writing to Elasticsearch cluster.
+     *
+     * @param error applies given function if specified in case of Elasticsearch error with bulk
+     *     writes. Default behavior throws IOException.
+     */
+    public static <T> Bound withError(ThrowingConsumer<BulkExecutionException> error) {
+      return new Bound<>().withError(error);
+    }
+
+    public static <T> Bound withMaxBulkRequestSize(int maxBulkRequestSize) {
```
clairemcginty, line 133: add Javadoc for this + `withMaxBulkRequestBytes` and explain how they're used differently?

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
```diff
@@ -0,0 +1,791 @@
...
+      }
+
+      @ProcessElement
+      public void processElement(ProcessContext c) throws Exception {
+        final T t = c.element();
+        final Iterable<BulkOperation> operations = toBulkOperations.apply(t);
+        for (BulkOperation operation : operations) {
+          long bytes = operationSize(clientSupplier, operation);
+          if ((chunk.size() + 1) > maxBulkRequestSize
+              || (chunkBytes + bytes) > maxBulkRequestBytes) {
```
clairemcginty, line 535: I'm wondering if we really need the `maxBulkRequestSize` option -- offering just the `maxBulkRequestBytes` functionality might be good enough for users. wdyt?

scio-elasticsearch/es8/src/main/scala/com/spotify/scio/elasticsearch/CoderInstances.scala
RustedBones: Since the new client API, the model isn't yet in a document form but still a Java object. I think it is not possible get better serialization support, except for the `DeleteOperation` which do not contain document and is annotated with `@JsonpDeserializable`.

The other Operations require to serialize a document, which is done with the internal json mapper from the client or with the provided serializer on the operation (which we should also serialize to reconstruct the operation). I do not know how to do it except by using kryo. Moreover, if a custom serializer is provided, we don't have the associate deserializer to parse the document.

Is it worth to only create a specific case for the `DeleteOperation` ?

scio-elasticsearch/es8/src/main/java/org/apache/beam/sdk/io/elasticsearch/ElasticsearchIO.java
regadas: Well, knowing the docs it's the best scenario but there are use cases where the nature of the payload is somewhat dynamic and vary greatly on size. That's why this combo `maxBulkRequestSize` + `maxBulkRequestBytes` was introduced. That said, not sure about ES8 but previous versions couldn't cope with this and it was really hard to keep/tune the ES clusters to have predictable indexing speeds and be stable.

RustedBones: Will put that back then. Agreed that the doc example mention indexing only.
We can have cases where things are mixed and instead of being aggressive on the `maxBulkRequestSize` to ensure the request never gets too big, we should leave `maxBulkRequestBytes`

We discussed the settings naming with @clairemcginty because we think that `maxBulkRequestSize` vs `maxBulkRequestBytes` is confusing. `size` and `bytes` are often used for the same thing. I would suggest to rename `maxBulkRequestSize` option to `maxBulkRequestOperations` WDYT ?

regadas: no opposition in renaming it; either that or perhaps `maxBulkRequestSizeBytes`

PR #4313, opened 2022-04-14
Add es8 integration test

RustedBones:


regadas (commented):
Looks good! let's see if GHA is ok with it.

PR #4314, opened 2022-04-19
Mock BQ wildcard table

### LINKED ISSUE #3248: Enable partitioned table for mockTable
I am aggregating some events for 7 days from a partitioned table(https://ghe.spotify.net/ads/adserver-data-pipelines/blob/base_ctr/adserver-data-pipelines/src/main/scala/com/spotify/ads/adserver/pipelines/batch/bidadjustment/CreativeBaseCTRJob.scala), and want to use mockTable(https://ghe.spotify.net/ads/adserver-data-pipelines/blob/644014c0f3f06c6ebf37c2f7544c037c5c54fdc3/adserver-data-pipelines/src/test/scala/com/spotify/ads/adserver/pipelines/batch/bidadjustment/CreativeBaseCTRJobTest.scala#L133) for integration tests of my SQL.

I got this error for table not found. I guess this is because my original table is partitioned. 
`
com.google.api.client.googleapis.json.GoogleJsonResponseException: 404 Not Found
{
  "code" : 404,
  "errors" : [ {
    "domain" : "global",
    "message" : "Not found: Table ad-selfserve:ad_platform_services.clean_flights",
    "reason" : "notFound"
  } ],
  "message" : "Not found: Table ad-selfserve:ad_platform_services.clean_flights",
  "status" : "NOT_FOUND"
}
`

RustedBones:
Fix #3248 

Adding `mockWildcardTable` API on `MockBigQuery` 

Eg: To get the highest recorded temperature from the 2020s decade 

```sql
SELECT
  max,
  year,
  mo,
  da,
FROM
  `bigquery-public-data.noaa_gsod.gsod202*`
WHERE
  max != 9999.9 # code for missing data
  AND _TABLE_SUFFIX BETWEEN '0' AND '9'
ORDER BY
  max DESC
LIMIT 1
```

We can mock `bigquery-public-data.noaa_gsod.gsod2022` and use it as only matched table by the query with 

```
mockWildcardTable("bigquery-public-data.noaa_gsod.gsod202", "2").withData(...)
```

### DIFF
--- a/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
+++ b/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
@@ -65,6 +65,46 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
     mbq.queryResult(sqlQuery) should contain theSameElementsAs expected
   }
 
+  // see https://cloud.google.com/bigquery/docs/querying-wildcard-tables
+  it should "support wildcard tables" in {
+    val prefix = "bigquery-public-data.noaa_gsod.gsod202"
+    val suffix = "2"
+    val wildcardSqlQuery =
+      """SELECT
+        |  max,
+        |  year,
+        |  mo,
+        |  da,
+        |FROM
+        |  `bigquery-public-data.noaa_gsod.gsod202*`
+        |WHERE
+        |  max != 9999.9 # code for missing data
+        |  AND _TABLE_SUFFIX BETWEEN '0' AND '9'
+        |ORDER BY
+        |  max DESC
+        |LIMIT 1
+        |""".stripMargin
+
+    def gsod(max: Double, year: Int, mo: Int, da: Int): TableRow =
+      TableRow(
+        "max" -> max,
+        "year" -> year.toString,
+        "mo" -> mo.toString,
+        "da" -> da.toString
+      )
+
+    val inData = Seq(
+      gsod(55.0, 2022, 2, 11),
+      gsod(50.7, 2022, 1, 13),
+      gsod(50.5, 2022, 1, 13),
+      gsod(9999.9, 2022, 1, 1)
+    )
+
+    val mbq = MockBigQuery()
+    mbq.mockWildcardTable(prefix, suffix).withData(inData)
+    mbq.queryResult(wildcardSqlQuery) should contain only inData.head
+  }
+
   // =======================================================================
   // Integration test with type-safe mock data
   // =======================================================================
--- a/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
+++ b/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
@@ -81,7 +81,6 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
         |  AND _TABLE_SUFFIX BETWEEN '20' AND '29'
         |ORDER BY
         |  max DESC
-        |LIMIT 1
         |""".stripMargin
 
     def gsod(max: Double, year: Int, mo: Int, da: Int): TableRow =
@@ -118,7 +117,14 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
     suffixData.foreach { case (suffix, inData) =>
       mbq.mockWildcardTable(prefix, suffix).withData(inData)
     }
-    mbq.queryResult(wildcardSqlQuery) should contain only suffixData("21").head
+    mbq.queryResult(wildcardSqlQuery) should contain theSameElementsInOrderAs Seq(
+      gsod(54.2, 2021, 6, 30),
+      gsod(54.0, 2020, 6, 8),
+      gsod(53.2, 2021, 6, 22),
+      gsod(53.1, 2021, 7, 1),
+      gsod(52.2, 2020, 7, 31),
+      gsod(52.1, 2020, 7, 30)
+    )
   }
 
   // =======================================================================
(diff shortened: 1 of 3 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -64,6 +64,32 @@ class MockBigQuery private (private val bq: BigQuery) {
     new MockTable(bq, t.getSchema, original, temp)
   }
 
+  /**
+   * Mock a BigQuery wildcard table using the table prefix and the suffix matched by the wildcard.
+   * Each table can be mocked only once in a test class.
```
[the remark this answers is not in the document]
RustedBones: It is hard to separate the prefix from the suffix in the API since we need to check if the original table exists.
I kept the same API with `def mockWildcardTable(prefix: String, suffix: String)` but it is now possible to call it several times.

Moreover, having `withData(suffix1, data1)` will force to create a table on the fly, and doesn't allign well with the original API. We should have had smth like `mockTableData(original, data)` directly on `MockBigQuery` to be aligned.

We could imagine smth like `mockWildcard(prefix).mockTable(suffix1).withData(data1)` but I feel this indirection isn't helping

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -70,23 +70,29 @@ class MockBigQuery private (private val bq: BigQuery) {
...
     )
 
-    val t = bq.tables.table(original)
-    val temp = bq.tables.createTemporary(t.getLocation, Some(suffix)).getTableReference
-    val tmpWildcard = new TableReference()
-      .setProjectId(temp.getProjectId)
-      .setDatasetId(temp.getDatasetId)
-      .setTableId(temp.getTableId.dropRight(suffix.length) + "*")
-    mapping += (wildcard -> tmpWildcard)
+    val ref = tempWildcard.clone().setTableId(tempWildcard.getTableId.dropRight(1) + suffix)
```
clairemcginty, line 93: maybe I'm missing something, how is `ref` different from `original`?

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -93,16 +92,33 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
...
+      )
     )
 
     val mbq = MockBigQuery()
-    mbq.mockWildcardTable(prefix, suffix).withData(inData)
-    mbq.queryResult(wildcardSqlQuery) should contain only inData.head
+    suffixData.foreach { case (suffix, inData) =>
+      mbq.mockWildcardTable(prefix, suffix).withData(inData)
+    }
+    mbq.queryResult(wildcardSqlQuery) should contain only suffixData("21").head
```
clairemcginty, line 121: I thought BigQuery `BETWEEN` was inclusive, shouldn't the query also contain `suffixData("20")`?

clairemcginty (commented):
this API looks good! just have a few questions

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
RustedBones: the query retrieves the max temperature of the decade (`LIMIT 1`)
The highest one in the test data is in 2021, 1st element

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
RustedBones: ref is actually pointing to the test table in the staging dataset:
- prefix: `bigquery-public-data.noaa_gsod.gsod20`
- suffix: `20`
- original: `bigquery-public-data.noaa_gsod.gsod2020`
- wildcard: `bigquery-public-data.noaa_gsod.gsod20*`
- tempWildcard: `my-test-project.scio_bigquery_staging_us_east_1.scio_query_20210422164412_*`
- ref:  `my-test-project.scio_bigquery_staging_us_east_1.scio_query_20210422164412_20`

RustedBones: could also be named `temp` since we do not have to get the reference after creation

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
kellen: Seems like removing `LIMIT 1` would provide a better test?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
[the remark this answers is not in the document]
kellen: IMO simplicity is a benefit here, so keeping a single `mockWildcardTable` call, analogous to the existing `mockTable`, is better than adding complexity. Most users will likely have a map of suffix -> data anyway, so they'll just be `foreach`ing over it.

kellen (approved):

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
clairemcginty: ah gotcha! Maybe rename to `tempRef`? 🎨  Otherwise looks good!

clairemcginty (approved):

PR #4315, opened 2022-04-20
Add Avro primitive type bytes to tests

marconilanna:
`bytes` is one of Avro's primitive types (https://avro.apache.org/docs/1.8.2/spec.html#schema_primitive), however it is currently not covered by tests

### DIFF
--- a/scio-schemas/src/main/avro/schema.avsc
+++ b/scio-schemas/src/main/avro/schema.avsc
@@ -8,6 +8,7 @@
         {"name": "type", "type": "string"},
         {"name": "name", "type": ["null", "string"]},
         {"name": "amount", "type": "double"},
+        {"name": "bytes", "type": "bytes"},
         {
          "name": "account_status",
          "type": [
@@ -63,6 +64,7 @@
         {"name": "float_field", "type": ["null", "float"], "default": null},
         {"name": "double_field", "type": ["null", "double"], "default": null},
         {"name": "boolean_field", "type": ["null", "boolean"], "default": null},
+        {"name": "bytes_field", "type": ["null", "bytes"], "default": null},
         {"name": "string_field", "type": ["null", "string"], "default": null},
         {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": null}
     ]
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/AvroExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/AvroExample.scala
@@ -32,6 +32,7 @@ import com.spotify.scio.io.ClosedTap
 import org.apache.avro.Schema
 import org.apache.avro.generic.{GenericData, GenericRecord}
 
+import java.nio.ByteBuffer
 import scala.jdk.CollectionConverters._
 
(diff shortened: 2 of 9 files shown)


scio-schemas/src/main/avro/schema.avsc
```diff
@@ -8,6 +8,7 @@
         {"name": "type", "type": "string"},
         {"name": "name", "type": ["null", "string"]},
         {"name": "amount", "type": "double"},
+        {"name": "bytes", "type": "bytes"},
```
RustedBones, line 11: I don't think this is relevant to add the `bytes` field in this record

scio-schemas/src/main/avro/schema.avsc
```diff
@@ -63,6 +64,7 @@
         {"name": "float_field", "type": ["null", "float"], "default": null},
         {"name": "double_field", "type": ["null", "double"], "default": null},
         {"name": "boolean_field", "type": ["null", "boolean"], "default": null},
+        {"name": "bytes_field", "type": ["null", "bytes"], "default": null},
         {"name": "string_field", "type": ["null", "string"], "default": null},
         {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": null}
```
RustedBones, line 69: We are also missing a `map_field` and a `fixed_field`. Would you also like to add those ?
I'm seeing for array that the default value is not correct. this should be set to `[]`. IIRC this won't be accepted anymore when updating avro version

scio-schemas/src/main/avro/schema.avsc
clairemcginty: agreed, this schema is used heavily in tests + scio-examples so let's not add this new field. IMO adding the bytes field to `TestRecord` should be a sufficient test case.

marconilanna: @RustedBones @clairemcginty I agree with you both that the `bytes` field feels out of place in `Account`.

However, this PR was motivated by an issue we had where we could verify that a field of type `bytes` was set properly just before being serialized by `JobTest`, but nonetheless was empty after being read (turned out the issue was not related to Scio and we already fixed it).

I think there is value in a full round-trip test: verifying that values of all types can be serialized and read back again. If we remove the `bytes` field from `Account`, what do you suggest? Maybe we could modify `AvroInOutTest` to output `TestRecord` (or a variation thereof)

RustedBones: `AvroInOutTest` is more a test example that a real test. This also does not check serialization, since it is using a test skink.
Since our AvroIO is mostly a wrapper around the beam `AvroIO` which is tested with this [schema](https://github.com/apache/beam/blob/243128a8fc52798e1b58b0cf1a271d95ee7aa241/sdks/java/core/src/test/avro/org/apache/beam/sdk/schemas/test.avsc) I think we can rely on their unit testing (and the avro lib one) ensuring that all types are supported.

However I'm not sure if `scio-schemas` is relevant anymore. Since`sbt-avro` 2.0.0, it is possible to define schemas in the module's test config. Projects could be more independent and have their own schemas matching their use case.

PR #4340, opened 2022-04-28
Move ES7 and ES8 to scala-steward:off block

clairemcginty:


RustedBones:
Ok, I see those [rules](https://github.com/spotify/scio/blob/main/.scala-steward.conf) are [not working](https://github.com/spotify/scio/pull/4339/files) because `scala-steward` only checks the new versions in the rules, and finds a candidate.

We should remove those pin configs since they conflict.

We should however not move them into the beam group since those are not part of beam's dependency.
We can simply add the `// scala-steward:off` marker on them.

I opened https://github.com/scala-steward-org/scala-steward/issues/2593 to see if there is a way to let `scala-steward` update several versions of the same artifact

clairemcginty:
> We should however not move them into the beam group since those are not part of beam's dependency.

ah, it wasn't clear to me that that block was intended for beam-specific dependencies. IMO it's starting to feel a little over-complicated to have multiple sections of scala-steward ignore rules in the build :/

RustedBones:
Yes, @regadas raised the problem [here](https://github.com/spotify/scio/pull/4318#discussion_r861713650).
I've already opened #4345 , we can simply add elasticsearch there

clairemcginty:
sounds good then! do you want to add it or shall I?

PR #4368, opened 2022-04-29
Fix java waitForFutures impl

kellen:
`waitForFutures` shouldn't fail because one of the underlying futures failed. [See also](https://github.com/spotify/scio/blob/main/scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java#L48)


kellen:
Wrong, closing.

PR #4370, opened 2022-05-02
Enable MiMa for scio-parquet and scio-elasticsearch8

RustedBones:


clairemcginty (approved):
looks good. if we re-merge the breaking Parquet change for Scio 0.12.0 we'll have to add this back, I guess, but that can be added to the PR itself.

RustedBones:
Since we are going for a MINOR update with 0 prefix, `0.12.0` we could disable MiMa globally

PR #4372, opened 2022-05-03
Fix BQ table pagination in tableReferences

RustedBones:
If BQ table has more that 50 partition, we go into an infinite loop

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala
@@ -168,19 +168,20 @@ final private[client] class TableOps(client: Client) {
 
   /** Get list of tables in a dataset. */
   def tableReferences(projectId: Option[String], datasetId: String): Seq[TableReference] = {
-    var nextPageToken: String = null
-    def getNextPage: TableList = {
+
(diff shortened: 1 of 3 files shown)


kellen (approved):

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala
```diff
@@ -168,19 +168,20 @@ final private[client] class TableOps(client: Client) {
...
       }
     }
-    var rep = getNextPage
+
+    var rep = getNextPage(None)
     val b = Seq.newBuilder[TableReference]
     Option(rep.getTables).foreach(_.asScala.foreach(b += _.getTableReference))
     while (rep.getNextPageToken != null) {
-      rep = getNextPage
+      rep = getNextPage(Some(rep.getNextPageToken))
```
clairemcginty, line 184: do we know this will never be null?

kellen: Checked 2 lines up :)

clairemcginty: nvm

clairemcginty (approved):

PR #4373, opened 2022-05-03
Fix bullet points

wscheep:
(@ wouts)
Broken on the [scio docs site](https://spotify.github.io/scio/Scio-Unit-Tests.html), even though it seems to render properly on github.


RustedBones:
Thank you !

Issue #4392, opened 2022-05-19
Reduce coder memory footprint


RustedBones:
Coders are responsible for a non negligible memory consumption at runtime.

When analysing heap-dump on a dataflow worker, we can see that coders are duplicated in every TaskExecutors.

Most of the memory allocated from the coders is for their [materializationStackTrace](https://github.com/spotify/scio/blob/3d149ea5076dc35b963d557457bf07360631e8cb/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala#L252).

Since this one is shared by coder type, we should find a way to reduce the memory allocation by using an intern string representation for instance, instead of duplication the stack trace for every instance

kellen:
Closed in #4397

PR #4397, opened 2022-05-25
Reduce coder memory footprint

RustedBones:
See #4392 

The goal of the PR is to create only one `materializationStackTrace` on the top-most `WrappedBCoder` created during coder materialization.

Before the change, in the new implementation of `Coders should have a useful stacktrace when a Coder throws`, 8 materialisation stacks were appended to the exception. It is reduced to 1 with the change.

Dumping the memory on a worker and comparing with previous one: The size taken by `java.lang.StackTraceElement` is reduced from 4Mb down to 300kb. Coder serialization/deserialization should also be much after now.

Since `LazyCoder` do not contain the `materializationStackTrace`, it is also now wrapped by a `WrappedBCoder`. This can increase the graph size.


scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -319,9 +336,7 @@ final class CoderTest extends AnyFlatSpec with Matchers {
     jInstant.now coderShould notFallback()
   }
 
-  // Broken because of a bug in Beam
-  // See: https://issues.apache.org/jira/browse/BEAM-5645
-  ignore should "Serialize Row (see: BEAM-5645)" in {
```
RustedBones, line 324: `Row` now have proper equality. The JIRA is still open, maybe worth closing it

RustedBones:
The graph size wasn't changing much after the change.
I'm however trying to re-purpose the `WrappedBCoder` to only contain the materialized stacktrace. With that, the graph size would even be a bit smaller than before

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -39,62 +41,65 @@ object CoderMaterializer {
...
+    case _: KVCoder[_, _]   => false // KV cannot be made nullable
+    case _: Transform[_, _] => false // nullability should be deferred to transformed coders
+    case _: Ref[_]          => false // nullability should be deferred to underlying coder
+    case _                  => o.nullableCoders
+  }
+
+  private def isWrappableCoder(topLevel: Boolean, c: Coder[_]): Boolean = c match {
+    case _: RawBeam[_]    => false // raw should not be wrapped
+    case _: KVCoder[_, _] => false // KV should not be wrapped, but independent k,v can
+    case _                => topLevel
```
jto, line 60: Same here. Please avoid refection match.

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -39,62 +41,65 @@ object CoderMaterializer {
...
+    case _: RawBeam[_]      => false // raw cannot be made nullable
+    case _: KVCoder[_, _]   => false // KV cannot be made nullable
+    case _: Transform[_, _] => false // nullability should be deferred to transformed coders
+    case _: Ref[_]          => false // nullability should be deferred to underlying coder
+    case _                  => o.nullableCoders
+  }
+
+  private def isWrappableCoder(topLevel: Boolean, c: Coder[_]): Boolean = c match {
+    case _: RawBeam[_]    => false // raw should not be wrapped
+    case _: KVCoder[_, _] => false // KV should not be wrapped, but independent k,v can
```
jto, line 59: Why the change ? `KVCoder` was wrapped before.

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
RustedBones: You can do this on sealed trait without issues. No need to use `unapply`. Eg. if I remove the default case, the compiler tells
```sh
 match may not be exhaustive.
[warn] It would fail on the following inputs: Beam(_), Disjunction(_, _, _, _), Fallback(_), Record(_, _, _, _), Ref(), Transform(_, _)
```

RustedBones:
`BackcompatibilityTest` are failing. Will look into that

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
jto: It'll still compile to runtime reflection which I'd try to avoid.

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -262,26 +262,27 @@ private[scio] case class WrappedBCoder[T](u: BCoder[T]) extends BCoder[T] {
...
 
+  override def encode(value: T, os: OutputStream, context: BCoder.Context): Unit =
+    catching(u.encode(value, os, context))
+
   override def decode(is: InputStream): T =
     catching(u.decode(is))
 
+  override def decode(is: InputStream, context: BCoder.Context): T =
+    catching(u.decode(is, context))
+
```
RustedBones, line 279: `WrappedBCoder` was not delegating the deprecated `encode` and `decode` methods to the underlying coder.
This can cause issue, eg with the `ProtoCoder`: context is important for the coder, and if not provided, `Context.NESTED` is used. However, when parsing a top level element in a DoFn, `Context.OUTER` must be used instead.

Since `ProtoCoder` is a `org.apache.beam` coder, it wasn't wrapped before. So the fix to reduce size graph has a side effect to hid the `WrappedBCoder` bug

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -458,10 +472,8 @@ final class CoderTest extends AnyFlatSpec with Matchers {
       Short => jShort
     }
 
-    def opts: PipelineOptions =
-      PipelineOptionsFactory
-        .fromArgs("--nullableCoders=true")
-        .create()
```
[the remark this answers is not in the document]
jto: `ScioOptions` is a subclass of `PipelineOptions` so this _should_ work.

jto:
> @f-loris concerning the option to disable the stack trace, we had a discussion with @jto about it. We'd like to try fixing the coder issue without introducing a new option, as this increases config complexity and will probably only be used by few power users.
On top of that there's probably something else going on if you graph size it too big. We have 1000's of Scio workflows here at Spotify and this issue I've only seen when people were abusing `union`.

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
RustedBones: I'm not sure I follow. The scala compiler will translate those in both cases with an `isInstanceOf` check regardless if pattern match is used with the type or the `unapply`.

f-loris:
> > @f-loris concerning the option to disable the stack trace, we had a discussion with @jto about it. We'd like to try fixing the coder issue without introducing a new option, as this increases config complexity and will probably only be used by few power users.
> 
> On top of that there's probably something else going on if you graph size it too big. We have 1000's of Scio workflows here at Spotify and this issue I've only seen when people were abusing `union`.

Indeed, the pipeline which is making troubles uses `union` in 2 places. So, are unions in general problematic or is there something to pay attention to when using them? 

In my case, replacing some of the coders with kryo, reduced the graph size from 16MB to 9 MB, so 7 MB reduction. The affected transforms contained some groupBy and a leftOuterJoin.

jto:
@f-loris 
> Indeed, the pipeline which is making troubles uses union in 2 places. So, are unions in general problematic or is there something to pay attention to when using them?

There are OK but you don't want to have "big" unions. The typical issue is that people would union 30+ IO reads which generates a giant graph. In that case you should use SplittableDofn instead. That's typically seen in batch pipelines though so  for streaming, maybe something else is the cause.

> In my case, replacing some of the coders with kryo, reduced the graph size from 16MB to 9 MB, so 7 MB reduction. The affected transforms contained some groupBy and a leftOuterJoin.

This is likely to have negative consequence on the performances of your job.

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
jto: Looks like you're right :) It does not matter then.

kellen (approved):

PR #4401, opened 2022-06-07
Fix run pre-release gha by stating jobs synchonously

RustedBones:
GitHub pre-release action is [failing](https://github.com/spotify/scio/actions/runs/2453588109) because too many jar staging operations are run in parallel.

Change test parallelism so jobs are started sequentially, and only execution wait is done in parallel


RustedBones:
I had to delete the staging folder. there was probably a corrupted jar due to a parallel upload.
New execution [here](https://github.com/spotify/scio/runs/6774394966?check_suite_focus=true)

clairemcginty (approved):

PR #4415, opened 2022-06-09
Replace deprecated upload-release-asset by action-gh-release

RustedBones:


.github/workflows/deploy.yml
```diff
@@ -20,24 +20,11 @@ jobs:
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ github.ref }}
-          release_name: Release ${{ github.ref }}
-          draft: true
       - name: Upload Repl
-        uses: actions/upload-release-asset@v1.0.2
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: softprops/action-gh-release@v1
```
clairemcginty, line 24: so the token isn't needed anymore?

RustedBones: passing the toke as env variable is now deprecated. The new action has a default param set to the token:
> token: Secret GitHub Personal Access Token. Defaults to ${{ github.token }}

clairemcginty (approved):

Issue #4341, opened 2022-04-28
Deprecation Timeline.  0.11.5 > 0.11.0 ... BeamSQL - remove or update deprecation timeline.  


brucearctor:
https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala#L147

https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala#L143

Currently Scio is on version 0.11.5

Messaging for BeamSQL to be deprecated in 0.11.0 ...

SO, either update Deprecation plan/wording ( what future version), or actually remove support?

brucearctor:
See:  https://github.com/spotify/scio/pull/4421

PR #4421, opened 2022-06-14
deprecate at 0.12.0

brucearctor:


brucearctor:
Addressing: https://github.com/spotify/scio/issues/4341

brucearctor:
@RustedBones -- updating timeline/messaging in code, in case 0.12.0 is not the next release.

RustedBones:
Thanks for the PR @brucearctor.
We're already preparing the 0.12.0 release. Except if we find some bugs, we are not going to have another 0.11.x release.
I think you can already cleanup the deprecated code, without changing the message

brucearctor:
I jumped on something else -- not likely to relook at this for probably at least a week.

PR #4422, opened 2022-06-14
Migrate to magnolia1

### LINKED ISSUE #4420: Update magnolia artifact
Scio is using

```
"com.softwaremill.magnolia" %% "magnolia-core" % "1.0.0-M4"
```

but magnolia v1 for scala 2 has been released under another artifact:

```
libraryDependencies += "com.softwaremill.magnolia1_2" %% "magnolia" % "1.x.x"
```

Scio (and magnolify) should be updated to this newer artifact

RustedBones:
Fix #4420

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -109,7 +109,7 @@ val kantanCodecsVersion = "0.5.1"
 val kantanCsvVersion = "0.6.2"
 val kryoVersion = "4.0.2"
 val magnoliaVersion = "1.1.2"
-val magnolifyVersion = "0.4.8+10-0c6f9747-SNAPSHOT"
+val magnolifyVersion = "0.5.0"
 val metricsVersion = "3.2.6"
 val parquetExtraVersion = "0.4.3"
 val parquetVersion = "1.12.3"
@@ -219,7 +219,7 @@ val commonSettings = Def
     excludeDependencies ++= Seq(
       "org.apache.beam" % "beam-sdks-java-io-kafka"
     ),
-    resolvers ++= Seq(Resolver.sonatypeRepo("public"), Resolver.sonatypeRepo("snapshots")),
+    resolvers += Resolver.sonatypeRepo("public"),
     Test / javaOptions += "-Dscio.ignoreVersionWarning=true",
     Test / testOptions += Tests.Argument("-oD"),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v", "-a"),
--- a/build.sbt
+++ b/build.sbt
@@ -108,8 +108,8 @@ val junitVersion = "4.13.2"
 val kantanCodecsVersion = "0.5.1"
 val kantanCsvVersion = "0.6.2"
 val kryoVersion = "4.0.2"
-val magnoliaVersion = "1.0.0-M4"
-val magnolifyVersion = "0.4.8"
+val magnoliaVersion = "1.1.2"
+val magnolifyVersion = "0.4.8+10-0c6f9747-SNAPSHOT"
 val metricsVersion = "3.2.6"
 val parquetExtraVersion = "0.4.3"
 val parquetVersion = "1.12.3"
@@ -219,7 +219,7 @@ val commonSettings = Def
     excludeDependencies ++= Seq(
       "org.apache.beam" % "beam-sdks-java-io-kafka"
     ),
-    resolvers += Resolver.sonatypeRepo("public"),
+    resolvers ++= Seq(Resolver.sonatypeRepo("public"), Resolver.sonatypeRepo("snapshots")),
     Test / javaOptions += "-Dscio.ignoreVersionWarning=true",
     Test / testOptions += Tests.Argument("-oD"),
     testOptions += Tests.Argument(TestFrameworks.JUnit, "-q", "-v", "-a"),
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -107,8 +107,8 @@ val junitVersion = "4.13.2"
 val kantanCodecsVersion = "0.5.1"
 val kantanCsvVersion = "0.6.2"
 val kryoVersion = "4.0.2"
-val magnoliaVersion = "1.0.0-M4"
-val magnolifyVersion = "0.4.8"
+val magnoliaVersion = "1.1.2"
+val magnolifyVersion = "0.4.8+10-0c6f9747-SNAPSHOT"
```
kellen, line 111: Assuming you'll update this before merge

build.sbt
```diff
@@ -190,7 +190,7 @@ val commonSettings = Def
       "com.google.protobuf" % "protobuf-lite",
       "org.apache.beam" % "beam-sdks-java-io-kafka"
     ),
-    resolvers += Resolver.sonatypeRepo("public"),
+    resolvers ++= Seq(Resolver.sonatypeRepo("public"), Resolver.sonatypeRepo("snapshots")),
```
kellen, line 193: Assuming you'll remove this before merge

scio-macros/src/main/scala/com/spotify/scio/MagnoliaMacros.scala
```diff
@@ -33,35 +35,36 @@ private[scio] object MagnoliaMacros {
...
+              q"_root_.magnolia1.Param[$tc_, $t_, $p_]($name, $typeNameParam, $idx, $isRepeated, $typeclassParam, $defaultVal, Array.empty[Any], Array.empty[Any], Array.empty[Any])"
             )
-          case q"new SealedTrait($typeName, $subtypes, $_, $_)" =>
+          case q"new SealedTrait($typeName, $subtypesArray, $_, $_, $_)" =>
             super.transform(
-              q"new _root_.magnolia.SealedTrait($typeName, $subtypes, Array(), Array())"
+              q"new _root_.magnolia1.SealedTrait($typeName, $subtypesArray, Array.empty[Any], Array.empty[Any], Array.empty[Any])"
             )
-          case q"Subtype[$tc, $t, $p]($typeName, $id, $_, $_, $coder, $cast0, $cast1)" =>
+          case q"Subtype[$tc_, $t_, $s_]($name, $idx, $_, $_, $_, $tc, $isType, $asType)" =>
```
kellen, line 58: Both `$tc_` and `$tc` ? I know this is existing code but maybe more verbosity would be clearer

scio-macros/src/main/scala/com/spotify/scio/MagnoliaMacros.scala
```diff
@@ -33,35 +35,36 @@ private[scio] object MagnoliaMacros {
...
             super.transform(
-              q"_root_.magnolia.Subtype[$tc, $t, $p]($typeName, $id, Array(), Array(), $coder, $cast0, $cast1)"
+              q"_root_.magnolia1.Subtype[$tc_, $t_,$s_]($name, $idx, Array.empty[Any], Array.empty[Any], Array.empty[Any], $tc, $isType, $asType)"
```
kellen, line 60:
```suggestion
              q"_root_.magnolia1.Subtype[$tc_, $t_, $s_]($name, $idx, Array.empty[Any], Array.empty[Any], Array.empty[Any], $tc, $isType, $asType)"
```

scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
```diff
@@ -77,23 +77,23 @@ private object Derived extends Serializable {
 }
 
 trait LowPriorityCoderDerivation {
-  import magnolia._
+  import magnolia1._
 
   type Typeclass[T] = Coder[T]
 
-  def combine[T](ctx: CaseClass[Coder, T]): Coder[T] =
```
kellen, line 84: What's the logic on this rename?

kellen (approved):
Approving to unblock a merge, but see comments

scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
RustedBones: `magnolia1` changed its API:
- `combine` -> `join`
- `dispatch` -> `split`

The magnolia macro expects those functions to be defined. This code shouldn't be used directly by users anyway

scio-macros/src/main/scala/com/spotify/scio/MagnoliaMacros.scala
RustedBones: I tried to stick with the `magnolify` naming. See the `apply` signature of `Subtype`:

```scala
  def apply[Tc[_], T, S <: T](
      name: TypeName,
      idx: Int,
      anns: Array[Any],
      inheritedAnns: Array[Any],
      tpeAnns: Array[Any],
      tc: CallByNeed[Tc[S]],
      isType: T => Boolean,
      asType: T => S
  ): Subtype[Tc, T] =
 ```
 I maybe should try to use some capital names for types, even if it is not usually used in matching. Since we are in the macro/quasiquote scope, it should be OK

PR #4425, opened 2022-06-18
added example that shows how to select top N records from a BigQuery table

siddjain:


kellen:
Thanks for the contribution. This replicates things demonstrated in other examples, so closing.

PR #4433, opened 2022-06-28
Update beam to 2.40

RustedBones:


build.sbt
```diff
@@ -348,14 +345,8 @@ lazy val sparkRunnerDependencies = Seq(
...
 
-// only available for scala 2.11 & 2.12
-// https://issues.apache.org/jira/browse/FLINK-13414
-// beam-runners-flink-1.14 by default pulls 2.12 scala libs,
-// exclude those libs and add them manually with proper scala version
 lazy val flinkRunnerDependencies = Seq(
-  "org.apache.beam" % "beam-runners-flink-1.14" % beamVersion excludeAll (
-    ExclusionRule("com.twitter", "chill_2.12")
-  ),
+  "org.apache.beam" % "beam-runners-flink-1.15" % beamVersion,
```
RustedBones, line 349: @jto heads up for flink update in Scio `0.12.0`. Is that fine ?

build.sbt
```diff
@@ -611,12 +600,8 @@ lazy val `scio-google-cloud-platform`: Project = project
   .settings(
     description := "Scio add-on for Google Cloud Platform",
     libraryDependencies ++= Seq(
-      "com.google.cloud" % "google-cloud-spanner" % googleCloudSpannerVersion excludeAll (
-        ExclusionRule(organization = "io.grpc")
-      ),
-      "com.google.cloud.bigtable" % "bigtable-client-core" % bigtableClientVersion excludeAll (
-        ExclusionRule(organization = "io.grpc")
-      ),
```
RustedBones, line 619: Conflicts are handled by global `dependencyOverrides`. It would be preferable to see conflict in evicted rather that dropping a required transitive grpc dep

build.sbt
```diff
@@ -57,29 +57,28 @@ val googleApiServicesStorageVersion = s"v1-rev20211201-$googleClientsVersion"
...
-val googleCloudDatastore = "0.93.4"
-val googleCloudMonitoringVersion = "3.2.4"
-val googleCloudSpannerVersion = "6.20.0"
-val googleCloudVersion = "2.4.0"
-val googleCommonsProtoVersion = "2.7.2"
-val googleHttpClientsVersion = "1.41.2"
-val googleIAMVersion = "1.2.1"
-val grpcVersion = "1.44.0"
-val jacksonVersion = "2.13.1"
-val opencensusContribGrpcMetricsVersion = "0.12.3"
```
RustedBones, line 80: This one is wrong in the library-bom and cause conflict. default `opencensusVersion` must be use too

build.sbt
```diff
@@ -185,9 +184,7 @@ val commonSettings = Def
     Compile / doc / scalacOptions := Scalac.docOptions.value,
     javacOptions ++= Seq("-source", "1.8", "-target", "1.8", "-Xlint:unchecked"),
     Compile / doc / javacOptions := Seq("-source", "1.8"),
-    // protobuf-lite is an older subset of protobuf-java and causes issues
     excludeDependencies ++= Seq(
-      "com.google.protobuf" % "protobuf-lite",
```
RustedBones, line 190: Dep is not pulled anymore. Just cleaning

build.sbt
```diff
@@ -348,17 +345,10 @@ lazy val sparkRunnerDependencies = Seq(
   "org.apache.spark" %% "spark-streaming" % sparkVersion
 )
 
-// only available for scala 2.11 & 2.12
-// https://issues.apache.org/jira/browse/FLINK-13414
-// beam-runners-flink-1.14 by default pulls 2.12 scala libs,
-// exclude those libs and add them manually with proper scala version
```
RustedBones, line 354: `beam-runners-flink-1.15` only depends on java lib now.
`flink-clients` and `flink-streaming-java` are now also java libs

build.sbt
kellen: Do you mean the jackson version? It's `2.13.0` in the [beam build](https://github.com/apache/beam/blob/release-2.40.0/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L480)

`opencensusContribGrpcMetricsVersion` seems to have been unused

build.sbt
```diff
@@ -57,29 +57,28 @@ val googleApiServicesStorageVersion = s"v1-rev20211201-$googleClientsVersion"
...
-val googleIAMVersion = "1.2.1"
-val grpcVersion = "1.44.0"
-val jacksonVersion = "2.13.1"
-val opencensusContribGrpcMetricsVersion = "0.12.3"
+val gaxHttpJsonVersion = "0.101.0"
+val gaxVersion = "2.16.0"
+val googleApiCommonVersion = "2.1.5"
+val googleAuthVersion = "1.6.0"
+val googleCloudBigTableVersion = "2.6.2"
+val googleCloudDatastore = "0.95.0"
```
kellen, line 71: Looks like beam overrides this to be `0.93.10`:
https://github.com/apache/beam/blob/release-2.40.0/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L683

Seems to be due to some explicit request: https://github.com/apache/beam/pull/17497#pullrequestreview-959120468

kellen (changes requested):

build.sbt
RustedBones: google library-bom declares
```
io.opencensus:opencensus-contrib-grpc-metrics:0.12.3
```
However, if used with this version it doesn't work. `opencensus-contrib-grpc-metrics` should share the `opencensusVersion`

kellen (approved):

PR #4438, opened 2022-06-30
Add granularity option for Parquet-SDF reads

clairemcginty:
1. The default Parquet block size for Scio writes is [1GB](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/resources/core-site.xml). Thus, the vast majority of Parquet writes end up with 1 row-group per file.
2. We observed a long startup lag for [Parquet-SplittableDoFn](https://github.com/spotify/scio/pull/4436) of many files (in the thousands). Logging showed this time was being taken up opening and closing thousands of [ParquetFileReader](https://github.com/spotify/scio/blob/ebf7b15b09b687d74202d31f168bef6154781e04/scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala#L40)s at job startup, in order to fetch metadata required to compute "splits" of row-group ranges. Since this happens before the source is split and thus before auto-scaling can kick in, it ends up being thousands of almost sequential operations.
3. Since most Parquet files are only 1 row group, this metadata computation isn't tangibly helpful at all because it'll always return the range [0,1) anyway.
4. Thus, this PR adds a `Configuration` option `scio.parquet.read.splitgranularity` that allows you to choose if you want file or row-group splittability, with a default of `file`. In this case, the SDF will always return the bound of an entire file rather than attempting metadata lookups.
5. In practice, for a dataset of ~10k files this reduced startup lag from 10 min to ~3min.

I don't love the name of the Configuration key/value that I added, suggestions welcome there :)


clairemcginty:
this builds off of @stormy-ua's work on #4260 :)

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -112,40 +130,51 @@ class ParquetReadFn[T, R](
...
+                0,
+                pages.getRowCount,
+                file,
+                recordReader,
+                outputReceiver,
+                projectionFn
+              )
+            } finally {
+              pages = reader.readNextRowGroup()
+            }
```
RustedBones, line 154: Since the implementation is not really symmetric, we can probably go further and not check `tryClaim` every cycle.
What is the reason we wrap into a try in the `File` path compared to the `RowGroup` ?

clairemcginty: I added the `tryClaim` check because I noticed that without it, Dataflow would redundantly process files because multiple workers would be assigned the same piece of work... it actually duplicate-read like 10mil records without 😬 

the logic I really wanted was something Java-style like 

```java
if (!tracker.tryClaim(tracker.currentRestriction().getFrom)) {
  return;
}
while (...) {
  ..
}
```

but...scala 😅 

you're right the try/finally isn't necessary, I'll remove!

kellen (approved):

kellen:
Merging for 0.12.0-RC1

PR #4458, opened 2022-07-14
Improve KV batch API

RustedBones:
I realized that sio is only having an API for beam `GroupIntoBatches` with `ofSize`

1. this can be problematic in streaming pipelines when keys have few elements and batch takes time to fill. Adding the `maxBufferingDuration` to the API
2. `GroupIntoBatches` also allows to batch by byte size, using by default the coder size
3. 'Abusing' byteSize with custom weight, so users can batch on arbitrary element metric

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -803,7 +803,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
   /**
    * Batches inputs to a desired weight. Batches will contain only elements of a single key.
    *
-   * The weight of each element is computer from the provided function.
+   * The weight of each element is computer from the provided cost function.
    *
    * Elements are buffered until the weight is reached, at which point they are outputed to the
    * output [[SCollection]].
@@ -815,7 +815,8 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
    * buffered can be set. Once a batch is flushed to output, the timer is reset. The provided limit
    * must be a positive duration or zero; a zero buffering duration effectively means no limit.
    *
-   * @param batchByteSize
+   * @param weight
+   * @param cost
    * @param maxBufferingDuration
    *
    * @group per_key
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -826,7 +826,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
     cost: V => Long,
     maxBufferingDuration: Duration = Duration.ZERO
   ): SCollection[(K, Iterable[V])] = {
-    val weigher: SerializableFunction[V, java.lang.Long] = cost(_).asInstanceOf[java.lang.Long]
+    val weigher = Functions.serializableFn(cost.andThen(_.asInstanceOf[java.lang.Long]))
     val groupIntoBatches = GroupIntoBatches
       .ofByteSize[K, V](weight, weigher)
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -749,12 +750,87 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+   * output [[SCollection]].
+   *
+   * Windows are preserved (batches contain elements from the same window). Batches may contain
+   * elements from more than one bundle.
+   *
+   * A time limit (in processing time) on how long an incomplete batch of elements is allowed to be
+   * buffered can be set. Once a batch is flushed to output, the timer is reset. The provided limit
+   * must be a positive duration or zero; a zero buffering duration effectively means no limit.
+   *
+   * @param batchByteSize
```
farzad-sedghi, line 818: update params

RustedBones: Yes, Thanks!

scio-test/src/test/scala/com/spotify/scio/values/PairSCollectionFunctionsTest.scala
```diff
@@ -407,6 +407,26 @@ class PairSCollectionFunctionsTest extends PipelineSpec {
     }
   }
 
+  it should "support batchByteSizedByKey()" in {
+    runWithContext { sc =>
+      val p = sc
```
clairemcginty, line 412: 🎨 remove the local variable `p` ? (same with below test)

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -749,12 +750,88 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+   * @param maxBufferingDuration
+   *
+   * @group per_key
+   */
+  def batchWeightedByKey(
+    weight: Long,
+    cost: V => Long,
+    maxBufferingDuration: Duration = Duration.ZERO
+  ): SCollection[(K, Iterable[V])] = {
+    val weigher: SerializableFunction[V, java.lang.Long] = cost(_).asInstanceOf[java.lang.Long]
```
clairemcginty, line 829: Scio has a [Functions#serializableFn](https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala#L248) that can be used for this

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -749,12 +750,88 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
+   * buffered can be set. Once a batch is flushed to output, the timer is reset. The provided limit
+   * must be a positive duration or zero; a zero buffering duration effectively means no limit.
+   *
+   * @param weight
+   * @param cost
+   * @param maxBufferingDuration
+   *
+   * @group per_key
+   */
+  def batchWeightedByKey(
```
clairemcginty, line 824: just semantically, this function name kind of seems like "per key" refers to the weighting, not the batching ... wdyt about something like `weightedBatchByKey`/`byteSizeBatchByKey`? I guess the latter sounds a bit clunky.

clairemcginty (commented):
looks good, I'm a fan of these improvements!

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
farzad-sedghi: +1 to keeping the prefix. Another alternative is keeping all `batchByKey` as the prefix so we'll get `batchByKeyWeighted` and `batchByKeyByteSized` to avoid the confusion Claire mentioned.

PR #4461, opened 2022-07-14
IO Windowing

kellen:
Adds windowing filename support to file-based IOs. Additionally, makes file-based IOs more consistent in their `saveAs*` APIs.

Adds to most file-based IOs:
*  `shardNameTemplate: String`, pattern-based filename format used by `DefaultFilenamePolicy.fromStandardParameters`
* `filenamePolicyCreator: FilenamePolicyCreator`, a Single-Abstract-Method-able class taking a `path` and a `suffix` (required by most IO implementations) and returning a `FilenamePolicy`.
* `tempDirectory: String`, an optional temp directory location. Was previously supported by some IOs, but not all.

Supporting a more consistent interface required moving some sinks (e.g. `BytesSink`) from `FileIO.Sink` to `FileBasedSink` to be used via `WriteFiles`.


scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -177,7 +198,10 @@ final case class SpecificRecordIO[T <: SpecificRecord: ClassTag: Coder](path: St
         params.suffix,
         params.codec,
         params.metadata,
-        params.tempDirectory
+        params.shardNameTemplate,
+        ScioUtil.tempDirOrDefault(params.tempDirectory, data.context),
```
kellen, line 202: Here is a choice I am not 100% on; do we want to use the pipeline default temp dir for the write when none is provided (as we do in SMB) or is there a better choice?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/syntax/SCollectionSyntax.scala
```diff
@@ -73,29 +48,27 @@ class SCollectionOps[T](private val self: SCollection[T]) extends AnyVal {
    * @param compression
    *   defaults to snappy
    */
-  def saveAsDynamicParquetAvroFile(
+  def saveAsParquetAvroFile(
```
kellen, line 51: This is a breaking change, the old `saveAsDynamicParquetAvroFile` was actually _not_ a 'dynamic' per the usage of that term in the rest of the API. Also the API was a little _too_ scala-y; requiring an `Either`.

scio-tensorflow/src/main/java/org/apache/beam/sdk/io/TFRecordFileBasedSink.java
```diff
@@ -0,0 +1,51 @@
...
+          private WritableByteChannel outChannel;
+          /* package private :( */
+          private /* TFRecordIO.TFRecordCodec */ Object codec;
+          private Method writeMethod;
+
+          @Override
+          protected void prepareWrite(final WritableByteChannel channel) throws Exception {
+            this.outChannel = channel;
+            // FIXME why! make! things! private?
+            Class<?> inner = Class.forName("org.apache.beam.sdk.io.TFRecordIO$TFRecordCodec");
```
kellen, line 34: I really really don't want to copy-paste this codec into scio. Also I believe access modifiers to be bad, except as suggestions.

scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
```diff
@@ -88,27 +89,47 @@ import org.apache.beam.sdk.values.PCollection
  */
 object CsvIO {
 
-  final val DefaultCsvConfig: CsvConfiguration = CsvConfiguration(
+  final private val _DefaultCsvConfig: CsvConfiguration = CsvConfiguration(
```
RustedBones, line 92: I think the `_` prefix convention is for python.
I there are no name conflict it could have stayed without

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -86,8 +91,12 @@ private[scio] object ScioUtil {
     }
   }
 
-  def pathWithShards(path: String): String =
-    path.replaceAll("\\/+$", "") + "/part"
+  def pathWithPrefix(path: String, prefix: String = null): String = {
+    val replacement = s"/${Option(prefix).getOrElse("part")}"
+    // replaceAll surprisingly does the wrong thing with regex /*$
+    val stripped = StringUtils.stripEnd(path, "/")
+    s"${stripped}${replacement}"
```
RustedBones, line 98: This part misses some cleaning
```suggestion
    val filePrefix = Option(prefix).getOrElse("part")
    val stripped = StringUtils.stripEnd(path, "/")
    s"${stripped}/${filePrefix}"
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -0,0 +1,49 @@
...
+    numShards: Int = ParquetAvroIO.WriteParam.DefaultNumShards,
+    suffix: String = ParquetAvroIO.WriteParam.DefaultSuffix,
+    compression: CompressionCodecName = ParquetAvroIO.WriteParam.DefaultCompression,
+    conf: Configuration = ParquetAvroIO.WriteParam.DefaultConfiguration,
+    tempDirectory: String = ParquetAvroIO.WriteParam.DefaultTempDirectory
+  )(
+    destinationFn: T => String
+  )(implicit ct: ClassTag[T], coder: Coder[T]): ClosedTap[Nothing] = {
+    if (self.context.isTest) {
+      throw new NotImplementedError(
```
kellen, line 32: We should consider how we can support dynamic destinations in test.

scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroFileBasedSink.java
```diff
@@ -32,13 +32,13 @@
 
 import java.nio.channels.WritableByteChannel;
 
-public class ParquetAvroSink<T> extends FileBasedSink<T, Void, T> {
+public class ParquetAvroFileBasedSink<T> extends FileBasedSink<T, Void, T> {
```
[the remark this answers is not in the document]
kellen: We have naming collisions between `FileIO.Sink`s and `FileBasedSink`s, so yeah making the convention consistent for these two types.

scio-extra/src/test/scala/com/spotify/scio/extra/csv/CsvIOTest.scala
```diff
@@ -199,22 +204,18 @@ class CsvIOTest extends ScioIOSpec with TapSpec with BeforeAndAfterEach {
...
-    SerializableUtils.serializeToByteArray(CsvIO.ReadDoFn[TestTuple](CsvIO.DefaultCsvConfig))
+    SerializableUtils.serializeToByteArray(
+      CsvIO.ReadDoFn[TestTuple](CsvIO.WriteParam.DefaultCsvConfig)
+    )
   }
 
   private def writeAsCsvAndReadLines[T: HeaderEncoder: Coder](
-    items: Seq[T],
-    params: CsvIO.WriteParam = CsvIO.DefaultWriteParams
+    xform: (ScioContext, String) => ClosedTap[Nothing]
```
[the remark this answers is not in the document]
kellen: short for `transform` but I'll just rename it

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
kellen: Don't these do the same thing?

scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
kellen: Was colliding with the WriteParam versions of `DefaultCsvConfig`, but using the full `CsvIO.DefaultCsvConfig` is better 👍

scio-core/src/main/scala/com/spotify/scio/util/FilenamePolicyCreator.scala
```diff
@@ -0,0 +1,50 @@
+package com.spotify.scio.util
+
+import com.twitter.chill.ClosureCleaner
+import org.apache.beam.sdk.io.FileBasedSink
+import org.apache.beam.sdk.io.FileBasedSink.FilenamePolicy
+import org.apache.beam.sdk.io.fs.ResolveOptions.StandardResolveOptions
+import org.apache.beam.sdk.io.fs.ResourceId
+import org.apache.beam.sdk.transforms.windowing.{BoundedWindow, PaneInfo}
+
+trait FilenamePolicyCreator {
```
clairemcginty, line 10: I would call this `FilenamePolicySupplier`, more in line with Beam naming

scio-core/src/main/scala/com/spotify/scio/util/FilenamePolicyCreator.scala
```diff
@@ -0,0 +1,50 @@
...
+  def apply(path: String, suffix: String): FilenamePolicy
+}
+
+object FilenamePolicyCreator {
+  def filenamePolicyCreatorOf(
```
clairemcginty, line 15: I feel like the `Int, Int`s in the windowed and unwindowed signatures could be confusing for users because it's not immediately clear what they represent 🤔  I get not wanting to add more classes but even some syntactic sugar like this could help?

```scala
type ShardNum = Int
type NumShards = Int

def filenamePolicyCreatorOf(
    windowed: (ShardNum, NumShards, BoundedWindow, PaneInfo) => String = null,
    unwindowed: (ShardNum, NumShards) => String = null
  ): FilenamePolicyCreator = ...
```

scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
```diff
@@ -45,32 +47,68 @@ final case class TextIO(path: String) extends ScioIO[String] {
...
+    if (tempDirectory == null) throw new IllegalArgumentException("tempDirectory must not be null")
+    if (shardNameTemplate != null && filenamePolicyCreator != null)
+      throw new IllegalArgumentException(
+        "shardNameTemplate and filenamePolicyCreator may not be used together"
+      )
+
+    val fp = Option(filenamePolicyCreator)
+      .map(c => c.apply(ScioUtil.pathWithPrefix(path, ""), suffix))
+      .getOrElse(
+        ScioUtil.defaultFilenamePolicy(
```
clairemcginty, line 72: Since this logic block seems to get repeated for each IO, what if we added more of it to the `ScioUtil.defaultFilenamePolicy` helper? Something like 

```scala
object ScioUtil {
  def resolveFilenamePolicy(
    policy: Option[FilenamePolicyCreator],
    path: String,
    suffix: String,
    shardNameTemplate: String,
    isWindowed: Boolean
  ): FilenamePolicyCreator =
    Option(filenamePolicyCreator)
      .map(c => c.apply(pathWithPrefix(path, ""), suffix))
      .getOrElse(
        defaultFilenamePolicy(pathWithPrefix(path),
          shardNameTemplate,
          suffix,
          isWindowed
        )
      )
```

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -267,20 +293,25 @@ final case class GenericRecordParseIO[T](path: String, parseFn: GenericRecord =>
...
-    private[avro] val DefaultNumShards = 0
-    private[avro] val DefaultSuffix = ""
-    private[avro] val DefaultCodec: CodecFactory = CodecFactory.deflateCodec(6)
-    private[avro] val DefaultMetadata: Map[String, AnyRef] = Map.empty
-    private[avro] val DefaultTempDirectory = null
+    private[scio] val DefaultNumShards = 0
+    private[scio] val DefaultSuffix = ""
+    private[scio] val DefaultCodec: CodecFactory = CodecFactory.deflateCodec(6)
+    private[scio] val DefaultMetadata: Map[String, AnyRef] = Map.empty
+    private[scio] val DefaultShardNameTemplate: String = null
```
clairemcginty, line 300: Another thought about consolidating some pipeline logic: what if we defined a single case class in scio-core, `WindowParams(shardNameTemplate: String, tempDirectory: String, filenamePolicyCreator: FilenamePolicyCreator)?`

Then either all of the `WriteParam` implementations could just include the single `WindowParams` instance; or in `ScioIO`, alongside `def write`, we could have an overloaded `def write(data: SCollection[T], params: WriteP, windowParams: WindowParams)` ... wdyt?

scio-core/src/main/scala/com/spotify/scio/util/FilenamePolicyCreator.scala
kellen: This is really only for use in the case of tests where brevity is prized. Otherwise users should just do `new FilenamePolicy` and make an impl, which is equivalent to what you're asking for, I think

scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
kellen: Sadly this isn't exactly the same in all cases to maintain some semblance of backwards compatibility

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
kellen: None of these are strictly related to windowing; you can use them all in an unwindowed context. Maybe `FileWriteParams`? I guess `suffix` would also end up in this class. 

It'd help somewhat for consistency across IOs but we'd still (at least per our current `saveAs*` design) need to replicate the parameters to all the save methods. It seems like only the very earliest `saveAs` methods took the entire `WriteParam` class as an argument, and IMO those should be migrated to the current status quo if there's any left.

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
RustedBones: I'm more focusing on the naming: 
- `replacement` -> there isn't any replacement
- `// replaceAll surprisingly does the wrong thing with regex /*$` we do not call `replaceAll`

kellen:
Ready for re-review and/or further comments @clairemcginty @RustedBones

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -86,8 +91,11 @@ private[scio] object ScioUtil {
     }
   }
 
-  def pathWithShards(path: String): String =
-    path.replaceAll("\\/+$", "") + "/part"
+  def pathWithPrefix(path: String, prefix: String = null): String = {
+    val filePrefix = Option(prefix).getOrElse("part")
+    val stripped = StringUtils.stripEnd(path, "/")
+    s"${stripped}/${filePrefix}"
+  }
```
RustedBones, line 98: I'm a bit confused by this signature. not giving a `prefix` will end up building with `part`.

I think this could be more logical:
```suggestion
  def pathFolder(path: String): String = pathWithPrefix(path, "")
  def pathWithPrefix(path: String, prefix: String = "part"): String = {
    val stripped = StringUtils.stripEnd(path, "/")
    s"${stripped}/${prefix}"
  }
```
We'll however have to change all occurrences of
```diff
-pathWithPrefix(path, "")
+pathFolder(path)
```
And where `null` is unchecked (I think only one place)
```diff
-pathWithPrefix(path, prefix)
+Option(prefix) match {
+  case Some(p) => pathWithPrefix(path, p)
+  case None => pathWithPrefix(path)
+}
```
To ensure that even with null, a prefix is added.
We could also work with options, with default to `Some("part")`:
```diff
-pathWithPrefix(path, "")
+pathWithPrefix(path, None)
```
and
```diff
-pathWithPrefix(path, prefix)
+Option(prefix) match {
+  case p: Some => pathWithPrefix(path, p)
+  case None => pathWithPrefix(path) // use default prefix
+}
```

farzad-sedghi (commented):
This is an impressive refactoring and unifying of he API, well done! a little bit hard to review in all the details (for me at least) as it is a huge PR. Maybe one way to break it into smaller PRs was to have one PR only for restructurings the packages and project, which is easy to review and merge, and then the PR with the main changes could be a bit less complicated.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -123,25 +125,42 @@ object ProtobufIO {
...
     codec: CodecFactory,
     metadata: Map[String, AnyRef],
-    tempDirectory: String
-  ) = {
+    shardNameTemplate: String,
+    tempDirectory: ResourceId,
+    filenamePolicySupplier: FilenamePolicySupplier,
+    isWindowed: Boolean
+  ): beam.AvroIO.Write[U] = {
+    if (tempDirectory == null) throw new IllegalArgumentException("tempDirectory must not be null")
```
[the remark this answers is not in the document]
kellen: `avroOut` is called from multiple locations, so this is effectively just ensuring that they are using this (internal) method correctly

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
kellen: The original variant always appended `part` so this was some attempt at maintaining backwards compatibility, but you're right that it'd be better to have a clearer api here, will address.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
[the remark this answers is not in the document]
farzad-sedghi: can't `isWindowed` be null?

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -123,25 +125,42 @@ object ProtobufIO {
...
+    if (shardNameTemplate != null && filenamePolicySupplier != null)
+      throw new IllegalArgumentException(
+        "shardNameTemplate and filenamePolicySupplier may not be used together"
+      )
```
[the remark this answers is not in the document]
kellen: Done.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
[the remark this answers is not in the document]
kellen: ah, I see what you mean. `isWindowed` is purely constructed by scio, while the tempdir is partially constructed by user code with null defaults, so the risk is greater.

scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
kellen: Looking at this again, was able to simplify

scio-core/src/main/scala/com/spotify/scio/util/FilenamePolicySupplier.scala
```diff
@@ -12,6 +12,37 @@ trait FilenamePolicySupplier {
 }
 
 object FilenamePolicySupplier {
+  def resolve(
```
RustedBones, line 15: neat!

RustedBones (approved):

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/syntax/SCollectionSyntax.scala
clairemcginty: I think it's ok for this to be a breaking change; it doesn't seem to have much, if any, adoption in its current state :)

scio-core/src/main/scala/com/spotify/scio/util/FilenamePolicySupplier.scala
```diff
@@ -0,0 +1,81 @@
...
+    path: String,
+    suffix: String,
+    shardNameTemplate: String,
+    tempDirectory: ResourceId,
+    filenamePolicySupplier: FilenamePolicySupplier,
+    isWindowed: Boolean,
+    defaultPrefix: String = null
+  ): FilenamePolicy = {
+    require(tempDirectory != null, "tempDirectory must not be null")
```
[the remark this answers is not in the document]
clairemcginty: On second thought, does `resolve` use `tempDirectory` at all? could it just be removed from method signature?

PR #4468, opened 2022-07-21
update run instructions for examples

farzad-sedghi:
in the instructions to run example codes zone/region are set by `--zone` argument, while it is expected to set `--region` from [this list](https://cloud.google.com/dataflow/docs/resources/locations). `Getting started` page is updated accordingly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 # sbt
 dist/*
 target/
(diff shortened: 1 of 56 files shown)


clairemcginty (approved):
nice! thanks for updating these 🎨

Issue #4385, opened 2022-05-12
Make the temporary table prefix and target dataset configurable in the MockBigQuery class


azaroui:
Actually temporary table prefix and staging dataset are defined in TableOps. This prevents us from specifying them when calling the mockTable method in the MockBigQuery.
https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala#L48
We need this because we don't necessarily have the authorizations to create datasets on the fly.

farzad-sedghi:
https://github.com/spotify/scio/pull/4474

clairemcginty:
@azaroui can you expand a little more on why you need temp table prefix to be configurable?

azaroui:
@clairemcginty In our case, the devs do not have the authorization to create tables on the source dataset and we have created a dedicated dataset for tests purpose to which all the devs have write authorization. So if I can't pass the dataset as a parameter the test will fail with unauthorized error.

PR #4474, opened 2022-07-26
optional project and dataset for mock bq

farzad-sedghi:
fixing https://github.com/spotify/scio/issues/4385

### DIFF
--- a/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
+++ b/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
@@ -60,14 +60,14 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
     val expected = Seq(wordCount("i", 10), wordCount("thou", 20), wordCount("thy", 30))
   }
 
-  "MockBigQuery" should "support mock data" in {
+  it should "support mock data" in {
     val mbq = MockBigQuery()
     mbq.mockTable(tableRef).withData(MockBQData.inData)
     mbq.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
     mbq.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
   }
 
-  "MockBigQuery" should "support mock data with custom staging dataset" in {
+  it should "support mock data with custom staging dataset" in {
     val mbq = MockBigQuery(stagingDataset = Some("scio_bigquery_staging_custom"))
     mbq.mockTable(tableRef).withData(MockBQData.inData)
     mbq.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
--- a/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
+++ b/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
@@ -18,8 +18,9 @@
 package com.spotify.scio.bigquery
 
 import org.scalatest.Inspectors.forAll
-import org.scalatest.matchers.should.Matchers
 import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach}
 
 object BigQueryIT {
   val tableRef = "bigquery-public-data:samples.shakespeare"
@@ -37,12 +38,22 @@ object BigQueryIT {
   class WordCount
 }
 
-class BigQueryIT extends AnyFlatSpec with Matchers {
+class BigQueryIT extends AnyFlatSpec with Matchers with BeforeAndAfterEach with BeforeAndAfterAll {
   import BigQueryIT._
 
   // =======================================================================
   // Integration test with mock data
   // =======================================================================
+
+  val StagingDatasetPrefix = "scio_bigquery_staging_custom"
+
+  override def beforeEach() {
+    sys.props.subtractOne(BigQuerySysProps.StagingDatasetPrefix.flag)
+  }
+
+  override def afterAll() {
+    sys.props.subtractOne(BigQuerySysProps.StagingDatasetPrefix.flag)
+  }
 
   object MockBQData {
     def shakespeare(w: String, wc: Long, c: String, cd: Long): TableRow =
@@ -68,10 +79,18 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
   }
 
   it should "support mock data with custom staging dataset" in {
-    val mbq = MockBigQuery(stagingDataset = Some("scio_bigquery_staging_custom"))
-    mbq.mockTable(tableRef).withData(MockBQData.inData)
-    mbq.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
-    mbq.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
+    val defaultMockedBQ = MockBigQuery()
+    defaultMockedBQ.mockTable(tableRef).withData(Seq())
+
+    sys.props.addOne(BigQuerySysProps.StagingDatasetPrefix.flag -> StagingDatasetPrefix)
+    val mockedBQ = MockBigQuery()
+    mockedBQ.mockTable(tableRef).withData(MockBQData.inData)
+    mockedBQ.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
+    mockedBQ.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
+
+    // making sure nothing is written using default staging tables
+    defaultMockedBQ.queryResult(legacyQuery) should contain theSameElementsAs Seq()
+    defaultMockedBQ.queryResult(sqlQuery) should contain theSameElementsAs Seq()
   }
 
   object MockBQWildcardData {
@@ -143,18 +162,26 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
   }
 
   it should "support wildcard tables with custom staging dataset" in {
-    val mbq = MockBigQuery(stagingDataset = Some("scio_bigquery_staging_custom"))
+    val defaultMockedBQ = MockBigQuery()
+    defaultMockedBQ.mockTable(tableRef).withData(Seq())
+
+    sys.props.addOne(BigQuerySysProps.StagingDatasetPrefix.flag -> StagingDatasetPrefix)
+    val mockedBQ = MockBigQuery()
     MockBQWildcardData.suffixData.foreach { case (suffix, inData) =>
-      mbq.mockWildcardTable(MockBQWildcardData.prefix, suffix).withData(inData)
+      mockedBQ.mockWildcardTable(MockBQWildcardData.prefix, suffix).withData(inData)
     }
-    mbq.queryResult(
+    mockedBQ.queryResult(
       MockBQWildcardData.wildcardSqlQuery
     ) should contain theSameElementsInOrderAs MockBQWildcardData.expected
+
+    // making sure nothing is written using default staging tables
+    defaultMockedBQ.queryResult(legacyQuery) should contain theSameElementsAs Seq()
+    defaultMockedBQ.queryResult(sqlQuery) should contain theSameElementsAs Seq()
   }
 
-  // =======================================================================
-  // Integration test with type-safe mock data
-  // =======================================================================
+//  // =======================================================================
+//  // Integration test with type-safe mock data
+//  // =======================================================================
 
   it should "support typed BigQuery" in {
     val inData = Seq(
--- a/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
+++ b/scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
@@ -44,7 +44,7 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
   // Integration test with mock data
   // =======================================================================
 
-  "MockBigQuery" should "support mock data" in {
+  object MockBQData {
     def shakespeare(w: String, wc: Long, c: String, cd: Long): TableRow =
       TableRow("word" -> w, "word_count" -> wc, "corpus" -> c, "corpus_date" -> cd)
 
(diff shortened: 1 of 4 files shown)


scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -44,7 +44,7 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
   // Integration test with mock data
   // =======================================================================
 
-  "MockBigQuery" should "support mock data" in {
+  object MockBQData {
```
RustedBones, line 47: As this is static, it can go in the `BigQueryIT` object

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -58,15 +58,23 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
...
-    mbq.queryResult(legacyQuery) should contain theSameElementsAs expected
-    mbq.queryResult(sqlQuery) should contain theSameElementsAs expected
+    mbq.mockTable(tableRef).withData(MockBQData.inData)
+    mbq.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
+    mbq.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
   }
 
-  // see https://cloud.google.com/bigquery/docs/querying-wildcard-tables
-  it should "support wildcard tables" in {
+  "MockBigQuery" should "support mock data with custom staging dataset" in {
```
RustedBones, line 70: use `it` if the main test group does not change

```suggestion
  it should "support mock data with custom staging dataset" in {
```

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -58,15 +58,23 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
...
+    val mbq = MockBigQuery(stagingDataset = Some("scio_bigquery_staging_custom"))
+    mbq.mockTable(tableRef).withData(MockBQData.inData)
+    mbq.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
+    mbq.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
+  }
+
+  object MockBQWildcardData {
```
RustedBones, line 77: same as other test data

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -58,15 +58,23 @@ class BigQueryIT extends AnyFlatSpec with Matchers {
...
+    mbq.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
   }
 
-  // see https://cloud.google.com/bigquery/docs/querying-wildcard-tables
-  it should "support wildcard tables" in {
+  "MockBigQuery" should "support mock data with custom staging dataset" in {
+    val mbq = MockBigQuery(stagingDataset = Some("scio_bigquery_staging_custom"))
+    mbq.mockTable(tableRef).withData(MockBQData.inData)
+    mbq.queryResult(legacyQuery) should contain theSameElementsAs MockBQData.expected
+    mbq.queryResult(sqlQuery) should contain theSameElementsAs MockBQData.expected
```
RustedBones, line 74: Isn't this querying the data from the previous test ?
We should have custom queries with the overridden dataset here

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -60,6 +66,11 @@ class MockBigQuery private (private val bq: BigQuery) {
 
     val t = bq.tables.table(original)
     val temp = bq.tables.createTemporary(t.getLocation).getTableReference
+
+    // override project and dataset in the mutable temp object
+    stagingProject.foreach(temp.setProjectId(_))
+    stagingDataset.foreach(temp.setDatasetId(_))
```
RustedBones, line 72: Wasn't this working ?
```suggestion
    stagingProject.foreach(temp.setProjectId)
    stagingDataset.foreach(temp.setDatasetId)
```

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -85,6 +96,11 @@ class MockBigQuery private (private val bq: BigQuery) {
       wildcard, {
         // fake table reference, only used in the mapping for query replacement
         val w = bq.tables.temporaryTableReference(t.getLocation)
+
+        // override project and dataset in the mutable temp object
+        stagingProject.foreach(w.setProjectId(_))
+        stagingDataset.foreach(w.setDatasetId(_))
```
RustedBones, line 102:
```suggestion
        stagingProject.foreach(w.setProjectId)
        stagingDataset.foreach(w.setDatasetId)
```

RustedBones (changes requested):

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
farzad-sedghi: do you mean instead of wrapping it in `MockBQData`, put it directly under `BigQueryIT`?

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
farzad-sedghi: can you help me understand how it is different from before? We used to have two tests (mock and wildcard mock) both hitting the same staging dataset `scio_bigquery_staging_us`, and now both of them are duplicated to run the same tests on the new parallel dataset `scio_bigquery_staging_custom`

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -44,7 +46,11 @@ object MockBigQuery {
  * Use [[mockTable(original:String)* mockTable]] to feed data into live BigQuery service and
  * [[queryResult]] to query them.
  */
-class MockBigQuery private (private val bq: BigQuery) {
+class MockBigQuery private (
+  private val bq: BigQuery,
+  stagingProject: Option[String] = None,
```
clairemcginty, line 51: I don't think we need these `None` defaults because they're already covered by the constructor you added in the `MockBigQuery` companion object.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -31,11 +31,13 @@ import scala.reflect.runtime.universe._
...
-  def apply(): MockBigQuery = new MockBigQuery(BigQuery.defaultInstance())
-
   /** Create a new MockBigQuery instance with the given BigQueryClient. */
-  def apply(bq: BigQuery): MockBigQuery = new MockBigQuery(bq)
+  def apply(
+    bq: BigQuery = BigQuery.defaultInstance(),
+    stagingProject: Option[String] = None,
+    stagingDataset: Option[String] = None
+  ): MockBigQuery =
+    new MockBigQuery(bq = bq, stagingProject = stagingProject, stagingDataset = stagingDataset)
```
clairemcginty, line 40:
```suggestion
    new MockBigQuery(bq, stagingProject, stagingDataset)
```

just a suggestion to simplify bc i don't think the shadowed field names are needed for clarity.

clairemcginty (commented):
what's the use case for making `stagingProject` configurable as well as `stagingDataset`? In #4385 the issue is that dataset/temp table prefix aren't configurable; I don't think we've gotten a complaint about project before.

clairemcginty:
> > what's the use case for making `stagingProject` configurable as well as `stagingDataset`? In #4385 the issue is that dataset/temp table prefix aren't configurable; I don't think we've gotten a complaint about project before.
> 
> it's a good point you are right. I'm just thinking now that we are going to give control to provide their dataset, why not giving full control? kind of prefer it over giving half way control. do you see a risk with that?

In general, I think we want to be pretty opinionated on what we expose in our user-facing APIs, and if there's not a strong use case for it, we shouldn't add it. I can't think of a good reason staging dataset would be in another project - it's not a precedent in any other IO type really either :)

thinking more about the implementation of this, for just dataset naming, instead of overriding the option in `MockBigQuery` class, it might be cleaner to add `bigquery.staging_dataset_prefix` as a [BigQuerySysProp](https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQuerySysProps.scala). That way we can just directly reference it in [TableOps](https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala#L50) so the override logic isn't decoupled from the temp table logic. What do you think? cc @RustedBones

farzad-sedghi:
> 

I see sure I can remove project override. regarding sysprop, I initially thought about that but then after our chats I found out this is only to support mocking. then I thought it is ok and maybe better to do it directly in the mock code

farzad-sedghi:
> I see sure I can remove project override. regarding sysprop, I initially thought about that but then after our chats I found out this is only to support mocking. then I thought it is ok and maybe better to do it directly in the mock code e.g. easier to write/read tests

I see sure I can remove project override. regarding sysprop, I initially thought about that but then after our chats I found out this is only to support mocking. then I thought it is ok and maybe better to do it directly in the mock code e.g. easier to write/read tests

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
RustedBones: you can keep the object grouping nested in  `BigQueryIT`. I think this is convenient.

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
RustedBones: Ok, my bad.
I was confused that the query contains a dataset which is not the same as the one given to the `mbq` but this is actually what we expect -> the tableref is swapped with one in the `stagingDataset`

However, how do you actually make sure in this test that the provided `stagingDataset` is used instead of the default one?

RustedBones:
I think the sysprop makes sense so the setting is set once and used for all `MockBigQuery` instantiation.
I feel that users do not want to repeat the setting on every integration test.

As for the project, I don't really know. Some users may want to separate test costs from production one. Hence using another project to run test load jobs and queries can make sense.

clairemcginty:
> I think the sysprop makes sense so the setting is set once and used for all `MockBigQuery` instantiation. I feel that users do not want to repeat the setting on every integration test.
> 
> As for the project, I don't really know. Some users may want to separate test costs from production one. Hence using another project to run test load jobs and queries can make sense.

yeah... in that case though they could just have separate containers running the jobs with different GCP project configurations, I guess. Maybe for now let's just do dataset -- we can always add support for project later if it's asked for.

farzad-sedghi:
Ok then just to make sure we are all on the same page, we will allow the users to use sysprops for setting staging dataset only, instead of doing it programmatically per `MockBigQuery` object.

Then the question is will this property be only picked up by MockBigQuery for overriding, or do we want to allow users to choose their staging dataset in prod too? I don't see why not. wdyt? @clairemcginty @RustedBones

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
RustedBones: If it is not too hard, we can try to issue an extra BQ describe command on the dataset/table to be sure it is there.
Otherwise I trust the good implementation :)

RustedBones:
we can scope the setting in a test namespace like `bigquery.test.staging_dataset_prefix` to hint this isn't related to main codebase

farzad-sedghi:
> we can scope the setting in a test namespace like `bigquery.test.staging_dataset_prefix` to hint this isn't related to main codebase

thanks @RustedBones yes in case we decide to only support test, that's the best way to do it. I guess the main question is if we want to support this in prod or not. I will go ahead and implement this only for test only ...

farzad-sedghi:
> > > what's the use case for making `stagingProject` configurable as well as `stagingDataset`? In #4385 the issue is that dataset/temp table prefix aren't configurable; I don't think we've gotten a complaint about project before.
> > 
> > 
> > it's a good point you are right. I'm just thinking now that we are going to give control to provide their dataset, why not giving full control? kind of prefer it over giving half way control. do you see a risk with that?
> 
> In general, I think we want to be pretty opinionated on what we expose in our user-facing APIs, and if there's not a strong use case for it, we shouldn't add it. I can't think of a good reason staging dataset would be in another project - it's not a precedent in any other IO type really either :)
> 
> thinking more about the implementation of this, for just dataset naming, instead of overriding the option in `MockBigQuery` class, it might be cleaner to add `bigquery.staging_dataset_prefix` as a [BigQuerySysProp](https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQuerySysProps.scala). That way we can just directly reference it in [TableOps](https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/TableOps.scala#L50) so the override logic isn't decoupled from the temp table logic. What do you think? cc @RustedBones

on a second thought, there seems to be a problem/conflict between:
- setting a separate SysProp for test staging dataset e.g. `bigquery.test.staging_dataset_prefix`
- directly referencing the config in TableOps, so we don't have separate logic for handling this in mock and prod

the reason is since TableOps utils class is not aware of the context, the caller either needs to read the relevant SysProp and pass to table ops or needs to pass a parameters like `isTest` to table ops, so it knows what to read. Both of these require more than one layer of API change and the logic for test and prod is different.

clairemcginty:
I was thinking the sys prop simply be for `bigquery.staging_dataset_prefix` and be referenced in `TableOps` -- no need to differentiate anything for `MockBigQuery`. that seems simpler and would still solve #4385 if I'm understanding it correctly?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQuerySysProps.scala
```diff
@@ -38,6 +38,11 @@ object BigQuerySysProps {
 
   val Project: SysProp = SysProp("bigquery.project", "System property key for billing project.")
 
+  val StagingDatasetPrefix: SysProp = SysProp(
+    "bigquery.staging_dataset_prefix",
+    "System property key for setting the prefix for staging table in test"
```
clairemcginty, line 43:
```suggestion
    "System property key for setting the prefix for staging table"
```

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -37,14 +38,24 @@ object BigQueryIT {
...
   // =======================================================================
 
-  "MockBigQuery" should "support mock data" in {
+  val StagingDatasetPrefix = "scio_bigquery_staging_custom"
+
+  override def beforeEach() {
+    sys.props.subtractOne(BigQuerySysProps.StagingDatasetPrefix.flag)
+  }
+
+  override def afterAll() {
```
clairemcginty, line 54: Instead of overriding `beforeEach` and `afterAll` to set and unset the property, I think it would be simpler to just add this setting/unsetting logic to the beginning/end of the `it should "support mock data with custom staging dataset"` unit test 👍 If we were testing this property in more than one unit test,`beforeEach` would make sense, but I vote for keeping it simple for just the single unit test :)

otherwise, this looks good!

clairemcginty (approved):

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
farzad-sedghi: Two tests use it :) I don't think beginning and after each test is a good idea because in case an exception happens during the test, it won't clean after itself. This is more future proof, as it resets the state before each test, regardless of whether they care about the property or not. and one clean up after the entire suite.

RustedBones (approved):

clairemcginty (approved):

scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/BigQueryIT.scala
```diff
@@ -37,14 +38,24 @@ object BigQueryIT {
...
 
   // =======================================================================
   // Integration test with mock data
   // =======================================================================
 
-  "MockBigQuery" should "support mock data" in {
+  val StagingDatasetPrefix = "scio_bigquery_staging_custom"
+
+  override def beforeEach() {
+    sys.props.subtractOne(BigQuerySysProps.StagingDatasetPrefix.flag)
```
RustedBones, line 51: `SystemProperties` APIs `addOne` and `subtractOne` are only available for scala 2.13
Scala 2.12 only supports `+=` and `-=` operators

Issue #4400, opened 2022-06-01
transformOverride should support 1-to-many


kellen:
Currently transformOverride supports 1-to-1 mapping of input to output but should be flatMappable.

brucearctor:
@farzad-sedghi -- should we close this issue?  Looks to be addressed?

PR #4484, opened 2022-08-02
support 1-to-n mapping function for transformOverride

farzad-sedghi:
fixing https://github.com/spotify/scio/issues/4400

### DIFF
--- a/scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
@@ -241,8 +241,10 @@ object TransformOverride {
    *   with a transform flat-mapping keys of `mapping` to corresponding values in `mapping`, and
    *   wrapping the result in a [[BaseAsyncLookupDoFn.Try]] in a [[KV]].
    */
-  def ofIterAsyncLookup[T: ClassTag, U](name: String, mapping: Map[T, Iterable[U]])
-  : PTransformOverride =
+  def ofIterAsyncLookup[T: ClassTag, U](
+    name: String,
+    mapping: Map[T, Iterable[U]]
+  ): PTransformOverride =
     ofIterKV[T, BaseAsyncLookupDoFn.Try[U]](
       name,
       mapping.map { case (k, v) => k -> v.map(i => new BaseAsyncLookupDoFn.Try(i)) }
--- a/scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
+++ b/scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
@@ -1,6 +1,7 @@
 package com.spotify.scio.testing
 
-import java.{lang, util}
+import java.lang.{Iterable => JIterable}
+import java.util
 
 import com.spotify.scio.transforms.BaseAsyncLookupDoFn
 import com.spotify.scio.util.Functions
@@ -97,7 +98,7 @@ object TransformOverride {
 
     val overrideFactory =
       factory[PCollection[T], PCollection[U], PTransform[PCollection[T], PCollection[U]]](
-        t => PTransformReplacements.getSingletonMainInput(t),
+        PTransformReplacements.getSingletonMainInput,
         new PTransform[PCollection[T], PCollection[U]]() {
           override def expand(input: PCollection[T]): PCollection[U] =
             input.apply(MapElements.via(Functions.simpleFn(wrappedFn)))
@@ -112,7 +113,7 @@ object TransformOverride {
    *   with a transform flat-mapping elements via `fn`.
    */
   def ofIter[T: ClassTag, U](name: String, fn: T => Iterable[U]): PTransformOverride = {
-    val wrappedFn: T => lang.Iterable[U] = fn
+    val wrappedFn: T => JIterable[U] = fn
       .compose { t: T =>
         typeValidation(
           s"Input for override transform $name does not match pipeline transform.",
@@ -128,8 +129,8 @@ object TransformOverride {
         t => PTransformReplacements.getSingletonMainInput(t),
         new PTransform[PCollection[T], PCollection[U]]() {
           override def expand(input: PCollection[T]): PCollection[U] = {
-            val inferableFn = new InferableFunction[T, lang.Iterable[U]] {
-              override def apply(input: T): lang.Iterable[U] = wrappedFn.apply(input)
+            val inferableFn = new InferableFunction[T, JIterable[U]] {
+              override def apply(input: T): JIterable[U] = wrappedFn.apply(input)
             }
             input.apply(FlatMapElements.via(inferableFn))
           }
@@ -187,7 +188,7 @@ object TransformOverride {
    *   wrapping the result in a [[KV]]
    */
   def ofIterKV[T: ClassTag, U](name: String, mapping: Map[T, Iterable[U]]): PTransformOverride =
-    ofIter[T, KV[T, U]](name, mapping.map { case (k, v) => k -> v.map(KV.of(k, _)) })
+    ofIter[T, KV[T, U]](name, mapping.map { case (k, iterV) => k -> iterV.map(KV.of(k, _)) })
 
   /**
    * @return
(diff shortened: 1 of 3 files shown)


scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
```diff
@@ -105,6 +98,42 @@ object TransformOverride {
     PTransformOverride.of(new EqualNamePTransformMatcher(name), overrideFactory)
   }
 
+  /**
+   * @return
+   *   A [[PTransformOverride]] which when applied will override a [[PTransform]] with name `name`
+   *   with a transform flat-mapping elements via `fn`.
+   */
+  def off[T: ClassTag, U](name: String, fn: T => Iterable[U]): PTransformOverride = {
```
farzad-sedghi, line 106: To be fixed: when I use the same function name `of`, I get `double definition` because it will have the same signature with the previous function after erasure:
```
have same type after erasure: (name: String, fn: Function1, evidence$1: scala.reflect.ClassTag): org.apache.beam.sdk.runners.PTransformOverride
```

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
```diff
@@ -105,6 +98,43 @@ object TransformOverride {
...
+      t
+    }.andThen(_.asJava)
+
+    val overrideFactory =
+      factory[PCollection[T], PCollection[U], PTransform[PCollection[T], PCollection[U]]](
+        t => PTransformReplacements.getSingletonMainInput(t),
+        new PTransform[PCollection[T], PCollection[U]]() {
+          override def expand(input: PCollection[T]): PCollection[U] = {
+            val sf: SimpleFunction[T, lang.Iterable[U]] = Functions.simpleFn(wrappedFn)
+            input.apply(FlatMapElements.via(sf))
```
farzad-sedghi, line 122: this line causes `interface java.lang.Iterable is not a super class of U`. IIUC  the correct type is passed:

`  public static <InputT, OutputT> FlatMapElements<InputT, OutputT> via(`
 `SimpleFunction<? super InputT, ? extends Iterable<OutputT>> fn) {` 

And I make sure I pass a java iterable. Any ideas?

farzad-sedghi: Update: this one is fixed by creating an InferableFunction directly, instead of using `Functions.simpleFn`

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
farzad-sedghi: Solved: chose new name `ofIter`

kellen: Using a separate name is a good choice, `ofN` or `ofSeq` might be other options but `ofIter` is reasonable.

kellen (commented):
Looks good! Can you add an example usage to the docs in `Scio-Unit-Tests`?

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
```diff
@@ -105,6 +106,38 @@ object TransformOverride {
...
+        t
+      }
+      .andThen(_.asJava)
+
+    val overrideFactory =
+      factory[PCollection[T], PCollection[U], PTransform[PCollection[T], PCollection[U]]](
+        t => PTransformReplacements.getSingletonMainInput(t),
+        new PTransform[PCollection[T], PCollection[U]]() {
+          override def expand(input: PCollection[T]): PCollection[U] = {
+            val inferableFn = new InferableFunction[T, lang.Iterable[U]] {
```
clairemcginty, line 131: I think `FlatMapElements#via` also accepts a `SimpleFunction` - you could leverage our utility class [Functions#simpleFn](https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/util/Functions.scala#L254) to wrap `wrappedFn` into the `SimpleFunction` Java interface and pass that to `via`.

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
```diff
@@ -105,6 +106,38 @@ object TransformOverride {
...
+          t.getClass,
+          implicitly[ClassTag[T]].runtimeClass
+        )
+        t
+      }
+      .andThen(_.asJava)
+
+    val overrideFactory =
+      factory[PCollection[T], PCollection[U], PTransform[PCollection[T], PCollection[U]]](
+        t => PTransformReplacements.getSingletonMainInput(t),
```
clairemcginty, line 128: 🎨 

```suggestion
        PTransformReplacements.getSingletonMainInput,
```

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
farzad-sedghi: Didn't work for some reason, that's what I tried first. Please look at this: https://github.com/spotify/scio/pull/4484#discussion_r935819671

site/src/main/paradox/Scio-Unit-Tests.md
```diff
@@ -98,11 +98,27 @@ Scio provides convenience methods for constructing `PTransformOverride`s in the
 Continuing the example above, `TransformOverride.ofAsyncLookup` can be used to map static mock data into the expected output format
 for the transform, here `KV[Int, BaseAsyncLookupDoFn.Try[String]]`.
 
-@@snip [JobTestTest.scala](/scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala) { #JobTestTest_example_2 }
+@@snip [JobTestTest.scala](/scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala) { #JobTestTest_example_2_1 }
+
+or use a T to Iterable[U] mapping instead, if it is more convenient:
```
kellen, line 103: Per our discussion on Friday this isn't correct; it's not a matter of convenience it's the nature of the transform that dictates which you use.

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
clairemcginty: ah, i see! Maybe worth adding `InferableFunction` support to our `Functions` class :)

site/src/main/paradox/Scio-Unit-Tests.md
farzad-sedghi: In this case it is convenience, because the original transform being overridden is 1-to-1. The example at line 115 is for the the case you have to use this override. Wdyt?

scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala
```diff
@@ -105,6 +106,38 @@ object TransformOverride {
...
+   *   A [[PTransformOverride]] which when applied will override a [[PTransform]] with name `name`
+   *   with a transform flat-mapping elements via `fn`.
+   */
+  def ofIter[T: ClassTag, U](name: String, fn: T => Iterable[U]): PTransformOverride = {
+    val wrappedFn: T => lang.Iterable[U] = fn
+      .compose { t: T =>
+        typeValidation(
+          s"Input for override transform $name does not match pipeline transform.",
+          t.getClass,
+          implicitly[ClassTag[T]].runtimeClass
```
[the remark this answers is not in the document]
farzad-sedghi: How important is it? It doesn't play well with generics need to spend more time on it if important:
``` 
17: class type required but T found
           classOf[T].runtimeClass
                   ^
```

site/src/main/paradox/Scio-Unit-Tests.md
farzad-sedghi: found the actual misassumption in my impl. working to fix it. thanks for precise review.

kellen (approved):
Nice work!

clairemcginty (approved):
looks great! 🎉

PR #4488, opened 2022-08-05
Support Neo4jIO

turb:
Support the experimental `org.apache.beam.sdk.io.neo4j.Neo4jIO` introduced in Beam 2.38. This is inspired by `JdbcIO`.

Request for comment: Beam only provides `readAll`, hence the usage of `sc.parallelize(Seq(())).applyTransform` to simulate a `read`.

1. I don't know if it is the proper way to do this.
2. Maybe @mattcasters knows why a simple `read` is not provided in Beam.


mattcasters:
> Maybe @mattcasters knows why a simple read is not provided in Beam.

Frankly I couldn't come up with a good way to parallelise a Cypher query in a generic way. If you would read just nodes or relationships you could split up the natural internal ID address space. For a generic cypher query it's not that trivial.
The "solution" then is to provide sets of parameters to the query which is then executed multiple times in parallel. With a bit of luck you can get to the same embarrassingly parallel solution.

turb:
> Frankly I couldn't come up with a good way to parallelise a Cypher query in a generic way.

Thanks for your answer!

I may be mistaken, but that indeed prevents using Splitable DoFn to get parallel IO, however it is still possible to have a naïve `read` that simply runs on a single node. I think that's what `JdbcIO` does — it only includes an `outputParallelization` parameter to shuffle the output.

With `readAll`, if we have only one cypher query to execute, we need to provide a one-discarded-element `PCollection`, which is kind of ugly.

To say things differently, this is not about the parallelisation feature, but the availability of `read` in the API, for convenience.

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIoOptions.scala
```diff
@@ -0,0 +1,90 @@
...
+ * @param url
+ *   connection url, e.g. "neo4j://neo4j.com:7687"
+ * @param username
+ *   login username
+ * @param password
+ *   login password
+ */
+final case class Neo4jConnectionOptions(url: String, username: String, password: String)
+
+sealed trait Neo4jIoOptions
```
RustedBones, line 34: Shouldn't we define common options here ?

```suggestion
sealed trait Neo4jIoOptions {
  def connectionOptions: Neo4jConnectionOptions
  def cypher: String
}  
```
It will remove the need to pattern match to create the id in `neo4jIoId`

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIoOptions.scala
```diff
@@ -0,0 +1,90 @@
...
+ * @param sessionConfig
+ *   [[org.neo4j.driver.SessionConfig]] for specifying Neo4J driver session config
+ * @param transactionConfig
+ *   [[org.neo4j.driver.TransactionConfig]] for specifying Neo4J driver transaction config
+ */
+final case class Neo4jWriteOptions[T](
+  connectionOptions: Neo4jConnectionOptions,
+  cypher: String,
+  unwindMapName: String,
+  parametersFunction: T => java.util.Map[String, AnyRef],
```
RustedBones, line 86: I would prefer to have it as scala `Map` and convert to a java later on in scio

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/syntax/AllSyntax.scala
```diff
@@ -0,0 +1,20 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.neo4j.syntax
+
+trait AllSyntax extends ScioContextSyntax with SCollectionSyntax
```
RustedBones, line 20: Is this required ? or can we directly make the `neo4j` package extend `ScioContextSyntax` and `SCollectionSyntax`

RustedBones (commented):
Thanks a lot for the contribution!
This looks quite good.

Maybe we can also create a very basic use-case in `scio-example` if you know one

RustedBones:
@turb for parallel reading in jdbc, we have the `jdbcShardedSelect`:
It takes a `shardColumn` and a `numShards` with
- `shardColumn` being uniform and linear, where we can know the min and max values
- `numShards` to split the `shardColum` space above into x buckets

we can potentially do smth like this
```scala
sc.paralellize(Range(min, shard1), Range(shard1, shard2), Range(shard2, max)) // to be computed ?
   .applyTransform(
      beam.Neo4jIO
        .readAll[Unit, T]()
        ...
        .withCypher(shardedCypher) // ensures we have $min $max filter
        .withParametersFunction((r: Range) => Map("min" -> r.min, "max" -> r.max)) 
```

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIoOptions.scala
turb: Modified in fixup d4cfa0a18abd3c767a741f7a7367f1baaf3ecc48

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIoOptions.scala
turb: Modified in fixup d23faf501efaf57461f7f84e94963fb2df093905

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/syntax/AllSyntax.scala
turb: To be honest I took it directly from `JdbcIo`, so I don't know what is the proper pattern in Scio.

RustedBones: Ok! I'm not too familiar either with what is the best archetype for the modules :) We can leave it like this

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIoOptions.scala
RustedBones: Since this is a sealed trait it does not matter, but in traits, it is preferable to use the least powerful abstraction.
Here `def` instead of `val` so impl can have more freedom when overriding. I think it is also better for binary compatibility too

turb:
> Maybe we can also create a very basic use-case in `scio-example` if you know one

Added one in fixup 7669f0abc01562933efd32a9108935667b74b6f4

turb:
> for parallel reading in jdbc, we have the `jdbcShardedSelect`

I suppose @mattcasters has on opinion it.

However, since the Beam impl is still experimental, is it a good idea to build this kind of pattern on top of it yet?

RustedBones:
IMHO it's more for further direction.
I'd prefer incremental development, where we can start with the propose `read` and `write` and continue discussion in other issue/PR

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIoOptions.scala
turb: Good catch => e65ed92df37139fbab06f6401ebddc2d7d7732fa

mattcasters:
I'd be happy to consider any and all changes to Neo4jIO. It was really hard to get it across the finish line because of all the new rules and regulations that are in effect in the Beam project which do not apply to existing code like JdbcIO. Because of this, both IOs are not the same. Neo4jIO was a bit of a challenge since there was no other clean code for an IO available. Months was spent on it.
However, now that the code exists we can expand functionality. A Neo4j shard selector would indeed be something cool to consider.  Feel free to create an issue in the beam project and I'd be happy to investigate.

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
```diff
@@ -0,0 +1,117 @@
...
+    val neo4jOptions = Neo4jWriteOptions[(String, String)](
+      Neo4jConnectionOptions(container.boltUrl, container.username, container.password),
+      """UNWIND $origin AS origin
+        |MATCH
+        |  (m:Movie {name: origin.movie}),
+        |  (c:Country {name: origin.country})
+        |CREATE (m)-[:ORIGIN]->(c)
+        |""".stripMargin,
+      "origin",
+      { case (movie, country) => Map("movie" -> movie, "country" -> country) }
```
RustedBones, line 101: Same thing here
```scala
.saveAsNeo4j(neo4jOptions, cypher)(x => Map(...))
```
As a neo4j newbie, I'm also confused by the requirement of the `UNWIND` and the obligation to declare the `unwindMapName`
As I understand, this is to enable batching.

Can this be hidden to the user ?
```
MATCH
  (m:Movie {name: $movie}),
  (c:Country {name: $country})
CREATE (m)-[:ORIGIN]->(c)
```
could be automatically transformed with a carefully picked unwinded name, replacing the `$`

RustedBones (commented):
I took some liberty and pushed on your branch some integration tests for neo4j.
Playing with the API raised few questions.

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
turb: > As a neo4j newbie, I'm also confused by the requirement of the UNWIND and the obligation to declare the unwindMapName
> As I understand, this is to enable batching.
> 
> Can this be hidden to the user ?

Maybe so. however, I hereby notify — bother — @mattcasters again, since he is the best expert we can find here, I believe!

turb:
> I took some liberty and pushed on your branch some integration tests for neo4j.

That's great! Many thanks.

I have locally updated the Neo4j docker image pulled by testContainers to 4.4.10 from 3.5.0, as the later has no Apple Silicon version, hence runs forever and timeouts on my laptop. Should I share this change?

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
mattcasters: Just to clarify the UNWIND situation: I should mention that without UNWIND you really can't get decent performance with Neo4j.  It is also painful to note that the dialect for UNWIND is a bit different from regular Cypher statements. However, I'm told this is something my colleagues in engineering are working on for the upcoming version 5 of Neo4j.  I propose to wait until that situation becomes a bit clearer, before making additional changes. Fingers crossed.

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
```diff
@@ -0,0 +1,117 @@
...
+
+    val neo4jOptions = Neo4jReadOptions(
+      Neo4jConnectionOptions(container.boltUrl, container.username, container.password),
+      s"MATCH (p)-[r: ACTED_IN]->(m) WHERE p.name='${martin.name}' RETURN p, r, m",
+      row => {
+        val p = Person(row.get("p").get("name").asString())
+        val m =
+          Movie(row.get("m").get("name").asString(), Year.of(row.get("m").get("year").asInt()))
+        Role(p, m, row.get("r").get("role").asString())
+      }
```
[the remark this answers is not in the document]
clairemcginty: agreed, I think @RustedBones's proposal is more in line with overall Scio API design 👍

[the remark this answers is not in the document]
turb: @clairemcginty thanks! in this design, should the parameters be passed using:
1. `override type ReadP = Neo4jReadParam` and in `read` parameter, `Neo4jReadParam` being a case class containing the cypher query and the row mapper.
2. or separate properties of `Neo4jCypher` case class?

(same for write)

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
```diff
@@ -6,22 +6,28 @@ import org.apache.beam.runners.direct.DirectRunner
...
-import java.time.Year
+import org.testcontainers.utility.DockerImageName
 
 import scala.jdk.CollectionConverters._
 
 object Neo4jIOIT {
+
+  val ImageName: DockerImageName = {
+    val tag = classOf[Driver].getPackage.getImplementationVersion.take("M.m.p".length)
+    DockerImageName.parse("neo4j").withTag(tag)
```
RustedBones, line 17: version is now taken from the drive. @turb can your try if this works on M1 ?

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIO.scala
```diff
@@ -22,22 +22,30 @@ import com.spotify.scio.coders.Coder
...
 
 sealed trait Neo4jIO[T] extends ScioIO[T]
 
 object Neo4jIO {
-  final def apply[T](opts: Neo4jIoOptions): Neo4jIO[T] =
+
+  object WriteParam {
+    private[neo4j] val BeamDefaultBatchSize = 5000L
+  }
+  final case class WriteParam(batchSize: Long = WriteParam.BeamDefaultBatchSize)
```
RustedBones, line 36: If we use `batchSize` as a write param, we can have common `Neo4jOptions`

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIO.scala
```diff
@@ -81,12 +87,24 @@ case class Neo4jCypher[T: Coder](readOptions: Neo4jReadOptions[T]) extends Neo4j
...
+  private[neo4j] val unwindMapName: String = cypher match {
+    case Neo4jIO.UnwindParameterRegex(name) => name
+    case _ =>
+      throw new IllegalArgumentException(
+        s"""Expected unwind cypher with parameter but got:
+         |$cypher
+         |See: https://neo4j.com/docs/cypher-manual/current/clauses/unwind/#unwind-creating-nodes-from-a-list-parameter
+         |""".stripMargin
+      )
+  }
```
RustedBones, line 105: Trying to be smart here: we can skip passing `unwindMapName` manually and extract if from the query.
Since we only support `writeUnwind` I think this is safe

RustedBones (commented):
As many of our API actually uses typeclasses, I think we can leverage this in also for the `neo4jIO`

- `RowMapper` would be the read typeclass to transform `Record` to `T`
- `ParametersBuilder` would extract the `Map[String, AnyRef]` parameter map from `T`

I'm less sure of the 2nd, if we should actually leave if as a function instead of a typeclass

@clairemcginty @farzad-sedghi

scio-examples/src/main/scala/com/spotify/scio/examples/extra/Neo4JExample.scala
```diff
@@ -23,46 +23,43 @@
...
     val (sc, _) = ContextAndArgs(cmdlineArgs)
 
-    val entities =
-      sc.read(Neo4jCypher(Neo4jReadOptions(neo4jConf, readCypher, neo4JRecordToEntity)))
+    val entities = sc.neo4jCypher(neo4jConf, readCypher)
 
     val modifiedEntities = entities.map(e => e.copy(property = e.property.map(_ + " modified")))
 
-    modifiedEntities.write(Neo4jWrite(Neo4jWriteOptions(neo4jConf, writeCypher, "rows", setter)))
+    modifiedEntities.saveAsNeo4j(neo4jConf, writeCypher)
```
turb, line 63: Adapted `Neo4JExample` with the new interface.

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
turb: It seems to work on the Docker side, however I get:

```
[info] - should read cypher query from the graph database *** FAILED ***
[info]   java.lang.IllegalArgumentException: unable to serialize DoFnWithExecutionInformation{doFn=org.apache.beam.sdk.io.neo4j.Neo4jIO$ReadFn@2b3d73e2, mainOutputTag=Tag<output>, sideInputMapping={}, schemaInformation=DoFnSchemaInformation{elementConverters=[], fieldAccessDescriptor=*}}
[info]   at org.apache.beam.sdk.util.SerializableUtils.serializeToByteArray(SerializableUtils.java:59)
[info]   at org.apache.beam.repackaged.direct_java.runners.core.construction.ParDoTranslation.translateDoFn(ParDoTranslation.java:736)
<snip>
[info]   ...
[info]   Cause: java.io.NotSerializableException: com.spotify.scio.neo4j.Neo4jIOIT$$anonfun$1$$anonfun$2
[info]   at java.base/java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1175)
<snip, same for write>
```

I tried force `rowMapper` and `paramsBuilder` to be static by putting it in the object, but that did not work.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/Neo4JExample.scala
RustedBones: Thanks I missed that part. Also noticed that there already a exists a typeclass out there as you mentioned in the comment.

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
turb: Fixed with 49362e2c5a48cbdeabb9a7a231276c9948cc55b3

RustedBones: I pushed my refactoring too quickly without testing. We need to make the `RowMapper` and `ParametersBuilder` serializable. sorry about that

RustedBones:
Thanks a lot for fixing the issues @turb 
I think the PR is good to go now. I'll open an issue in [magnolify](https://github.com/spotify/magnolify) to support the `RowMapper` automatic derivation as `neotype` core package comes with a lot of extras.

RustedBones (approved):

turb:
> Thanks a lot for fixing the issues @turb I think the PR is good to go now. I'll open an issue in [magnolify](https://github.com/spotify/magnolify) to support the `RowMapper` automatic derivation as `neotype` core package comes with a lot of extras.

Great! Should we squash the fixups, or will this be managed by the merger?

RustedBones:
we'll do a squash merge for the whole PR

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/syntax/ScioContextSyntax.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Get an SCollection for a Neo4J cypher query
+   *
+   * @param neo4jOptions
+   *   options for configuring a Neo4J driver
+   * @param cypher
+   *   Neo4J cypher query
+   */
+  def neo4jCypher[T: RowMapper: Coder](neo4jOptions: Neo4jOptions, cypher: String): SCollection[T] =
```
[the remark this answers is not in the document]
RustedBones: See https://github.com/spotify/magnolify/pull/588

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIO.scala
```diff
@@ -0,0 +1,129 @@
...
+        .withSessionConfig(neo4jOptions.sessionConfig)
+        .withTransactionConfig(neo4jOptions.transactionConfig)
+        .withBatchSize(params.batchSize)
+        .withCypher(cypher)
+        .withParametersFunction(ParametersBuilder[T].apply(_).asJava)
+    )
+    EmptyTap
+  }
+
+  override def tap(params: ReadP): Tap[Nothing] =
```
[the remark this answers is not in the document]
RustedBones: It is too difficult to get the query that represent the written data I think

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIO.scala
```diff
@@ -0,0 +1,111 @@
...
+
+object Neo4jIO {
+
+  object WriteParam {
+    private[neo4j] val BeamDefaultBatchSize = 5000L
+  }
+  final case class WriteParam(batchSize: Long = WriteParam.BeamDefaultBatchSize)
+
+  implicit private def recordConverter(record: Record): Value =
+    Values.value(record.asMap(identity))
```
RustedBones, line 38: By doing this, we loose indexed based access capability

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIO.scala
```diff
@@ -0,0 +1,111 @@
...
+    sc.parallelize(Seq(()))
+      .applyTransform(
+        beam.Neo4jIO
+          .readAll[Unit, T]()
+          .withDriverConfiguration(dataSourceConfiguration(neo4jOptions.connectionOptions))
+          .withSessionConfig(neo4jOptions.sessionConfig)
+          .withTransactionConfig(neo4jOptions.transactionConfig)
+          .withCypher(cypher)
+          .withRowMapper(neo4jType.from(_))
+          .withCoder(CoderMaterializer.beam(sc, coder))
```
RustedBones, line 81: coder was missing

RustedBones (commented):
We now have a single `Neo4jIO` supporting both read and write.
Conversion is handled by `magnolify` typeclass for `Value`.

`Record` result is a 'neighbor' of `Value`, and only allows indexed based access.
Here we loose this feature. Single value or tuples aren't supported on the read side

PR #4489, opened 2022-08-08
BatchDoFn and sio batch API on SCollection

RustedBones:
Amortize processing cost by local batching of elements
Batching respects windowing

This aims to give symetric API with the KV batching in #4458 
As the batch is emitted on `finishBundle`, no `maxBufferingDuration` is required

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
@@ -1,0 +1,120 @@
+package com.spotify.scio.transforms;
+
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * Batches input into a desired batch size.
+ *
+ * <p>Elements are buffered until there are enough elements for a batch,
+ * at which point they are emitted to the output PCollection
+ *
+ * <p>Windows are preserved (batches contain elements from the same window).
+ * Batches are not spanning over bundles. Once a bundle is finished, the batch is emitted even if not full.
+ * This function can only batch 10 parallel windows. If new element comes from an 11th window,
+ * the bigger batch will be emitted to give room for this new element.
+ *
+ * @param <InputT>
+ */
+public class BatchDoFn<InputT> extends DoFn<InputT, Iterable<InputT>> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BatchDoFn.class);
+
+    private static final int MAX_LIVE_WINDOWS = 10;
+
+    private static class Buffer<InputT> {
+
+        private final List<InputT> elements;
+
+        private long weight;
+
+        Buffer() {
+            this.elements = new ArrayList<>();
+            this.weight = 0;
+        }
+
+        public List<InputT> getElements() {
+            return elements;
+        }
+
+        public long getWeight() {
+            return weight;
+        }
+
+        public void add(InputT element, long weight) {
+            elements.add(element);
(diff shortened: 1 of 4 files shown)


scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
```diff
@@ -0,0 +1,120 @@
...
+ * This function can only batch 10 parallel windows. If new element comes from an 11th window,
+ * the bigger batch will be emitted to give room for this new element.
+ *
+ * @param <InputT>
+ */
+public class BatchDoFn<InputT> extends DoFn<InputT, Iterable<InputT>> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BatchDoFn.class);
+
+    private static final int MAX_LIVE_WINDOWS = 10;
```
farzad-sedghi, line 28: someone would ask this anyway :) why 10 and why not making it the default value instead of constant?

scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
```diff
@@ -0,0 +1,120 @@
...
+            BoundedWindow window = wb.getKey();
+            Buffer<InputT> buffer = wb.getValue();
+            context.output(buffer.getElements(), window.maxTimestamp(), window);
+        }
+        buffers.clear();
+    }
+
+    private void flushBatch(
+            BoundedWindow window,
+            Buffer<InputT> buffer,
```
farzad-sedghi, line 109: isn't it less error prone to only pass the key, and do the look up inside flush method?

farzad-sedghi: is it to avoid one extra look up?

scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
RustedBones: Shamelessly stole beam's python implementation. the constant is defined [here](https://github.com/apache/beam/blob/00bbc72e005d741f632bb171b36d470f6544be55/sdks/python/apache_beam/transforms/util.py#L566)

scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
RustedBones: yes, `remove` returns the value, we can do with the key only

scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
farzad-sedghi: hehe so maybe we need to do a better job :))

RustedBones: Do you think we should also expose this in the scala API, or we only give it as convenience if a user has a problem with the default, he could use the `BatchDoFn` with a custom param ?

RustedBones:
I ran some test on dataflow with `gs://apache-beam-samples/shakespeare/kinglear.txt` input:
- fixed size batch are respected
- weighted batches are respected

clairemcginty (commented):
the new API lgtm! I guess all that's left is each SCollection method needs a scaladoc explaining the parameters (and particularly the semantic that the batches are guaranteed to be *at most* size `batchSize`, not always *exactly* size `batchSize`. Can the `SCollectionTest` tests be uncommented?

scio-core/src/main/java/com/spotify/scio/transforms/BatchDoFn.java
farzad-sedghi: I don't see why not exposing it? with a default value so they don't have to worry about it if they don't want?

RustedBones:
@clairemcginty all comments should be addressed.
I managed to trick the tests to get a single bundle. This ensures batching is working within the bundle

PR #4490, opened 2022-08-09
fix BigQueryType.fromQuery query string formatting examples

luster:
Since `Row.query` is a `string`, you have to call `Row.query.format(args(0))`, etc.


luster:
opened a new PR #4503 ! thanks @clairemcginty

PR #4483, opened 2022-08-01
Migrate from sbt-protobuf to sbt-protoc

RustedBones:
This enables `grpc` java codegen support,

Clean useless `scio-schemas`. Move test schemas to project's Test configuration.

PR #4492, opened 2022-08-10
Implement gRPC lookup API

RustedBones:
Set base to `sbt-protoc` #4483 

Create an API to facilitate lookup over GRPC on `SCollection` elements with idiomatic scala


scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcDoFn.java
```diff
@@ -0,0 +1,121 @@
...
+    private int maxPendingRequests = DEFAULT_MAX_PENDING_REQUESTS;
+    private ChannelSupplier channelSupplier;
+    private SerializableFunction<Channel, ClientT> newClientFn;
+    private SerializableBiFunction<ClientT, RequestT, ListenableFuture<ResponseT>> lookupFn;
+
+    protected Builder() {}
+
+    public Builder<RequestT, ResponseT, ClientT> withChannelSupplier(
+        ChannelSupplier channelSupplier) {
+      this.channelSupplier = channelSupplier;
```
clairemcginty, line 92: I think it would be nice to provide some kind of helper functions for constructing a channel supplier -- i.e. allowing users to supply a NameResolverFactory, keepAliveTimeout, LB policy, etc. wdyt?

clairemcginty (commented):
looks good, glad this is going into Scio!

scio-grpc/src/test/scala/com/spotify/scio/grpc/GrpcDoFnTest.scala
```diff
@@ -0,0 +1,179 @@
...
+  override def beforeAll(): Unit = server.start()
+
+  override def afterAll(): Unit = server.shutdown()
+
+  "GrpcDoFn" should "issue request and propagate response" in {
+    runWithContext { sc =>
+      val result = sc
+        .parallelize(0 to 3)
+        .grpcLookup[ConcatResponse, ConcatServiceFutureStub](
+          () => NettyChannelBuilder.forTarget(ServiceUri).usePlaintext().build(),
```
farzad-sedghi, line 107: do you think we need to make convenience methods to make such lines simpler?

scio-grpc/src/test/scala/com/spotify/scio/grpc/GrpcDoFnTest.scala
```diff
@@ -0,0 +1,179 @@
...
+          ConcatServiceGrpc.newStub,
+          2
+        ) { case (client, observer, i) =>
+          val request = ConcatRequest
+            .newBuilder()
+            .setStringOne(i.toString)
+            .setStringTwo((Char.char2int('a') + i).toChar.toString)
+            .build()
+          client.concatServerStreaming(request, observer)
```
farzad-sedghi, line 144: this is a neat way to map the collection with lookup. still user has to do this last line manually. what about instead providing two functions for building the request and passing the client method? e.g.

```
.grpcLookupStream[ConcatResponse, ConcatServiceStub](
          () => NettyChannelBuilder.forTarget(ServiceUri).usePlaintext().build(),
          ConcatServiceGrpc.newStub,
          2
        )(
           i => ConcatRequest
               .newBuilder()
               .setStringOne(i.toString)
               .setStringTwo((Char.char2int('a') + i).toChar.toString)
               .build(),
           ConcatServiceStub.concatServerStreaming
       )
```

RustedBones: In that case, for symetry I would also change the `grpcLookup` with smth like this:
```scala
  def grpcLookup[GrpcRequest <: Message, GrpcResponse <: Message: Coder, Client <: AbstractFutureStub[Client]](
    channelSupplier: () => Channel,
    clientFactory: Channel => Client,
    maxPendingRequests: Int
  )(
  r: Request => GrpcRequest,
  f: Client => GrpcRequest => ListenableFuture[GrpcResponse]
  )
  
 def grpcLookupStream[GrpcRequest <: Message, GrpcResponse <: Message: Coder, Client <: AbstractStub[Client]](
    channelSupplier: () => Channel,
    clientFactory: Channel => Client,
    maxPendingRequests: Int
  )(
  r: Request => GrpcRequest,
  f: Client => (GrpcRequest, StreamObserver[GrpcResponse]) => Unit
  ) 
```
I have the feeling that this signature will be counter intuitive for some users. @clairemcginty @kellen WDYT ?

build.sbt
```diff
@@ -207,7 +208,7 @@ lazy val keepExistingHeader =
 val commonSettings = Def
   .settings(
     organization := "com.spotify",
-    headerLicense := Some(HeaderLicense.ALv2("2020", "Spotify AB")),
+    headerLicense := Some(HeaderLicense.ALv2(currentYear.toString, "Spotify AB")),
```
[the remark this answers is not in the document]
RustedBones: yes, got disappointed when generating headers and getting 2020 :upside_down_face:

scio-grpc/src/test/scala/com/spotify/scio/grpc/GrpcDoFnTest.scala
farzad-sedghi: oh I see. is the problem that calling the client method (`concatServerStreaming` in this case) is not trivial to be done by Scio, because it has more than one parameter?

farzad-sedghi: what I was thinking with this is that users can keep a list of such pairs, even in a separate file if they want which can help with readability (they can name them, and easy to know where the mappings are) and reusability. but I understand it can be a bit confusing. Maybe we can provide two flavours of the methods

RustedBones: Not really, I think the signature `A => B => C`, meaning this expects a function that returns a function may not be well understood by users with are not familiar with FP

farzad-sedghi: oh ok, I thought we think FP first in scio

farzad-sedghi: neat!

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -46,18 +50,12 @@
...
+    extends DoFnWithResource<A, KV<A, T>, Pair<C, Cache<A, B>>>
     implements FutureHandlers.Base<F, B> {
   private static final Logger LOG = LoggerFactory.getLogger(BaseAsyncLookupDoFn.class);
 
-  // DoFn is deserialized once per CPU core. We assign a unique UUID to each DoFn instance upon
-  // creation, so that all cloned instances share the same ID. This ensures all cores share the
-  // same Client and Cache.
-  private static final ConcurrentMap<UUID, Object> client = new ConcurrentHashMap<>();
-  private static final ConcurrentMap<UUID, Cache> cache = new ConcurrentHashMap<>();
-  private final UUID instanceId;
```
RustedBones, line 58: client and cache are basically resources. Delegate logic to `DoFnWithResource` instead

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -345,44 +357,26 @@ public boolean equals(Object obj) {
...
    */
-  public abstract static class CacheSupplier<A, B, K> implements Serializable {
-    /**
-     * Create a new {@link Cache} instance. This is called once per {@link BaseAsyncLookupDoFn}
-     * instance.
-     */
-    public abstract Cache<K, B> createCache();
-
-    /** Get cache key for the input element. */
-    public abstract K getKey(A input);
```
RustedBones, line 360: :warning: This is a breaking change :warning: 
I feel this API was very complex. To support this, users would have to provide their own cache implementation that would delegate to an underlying cache.

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcDoFn.java
```diff
@@ -0,0 +1,127 @@
...
+    this.channelSupplier = channelSupplier;
+    this.newClientFn = newClientFn;
+    this.lookupFn = lookupFn;
+  }
+
+  @Override
+  public ResourceType getResourceType() {
+    // gRPC stubs are thread safe, we can share the client per instance
+    return ResourceType.PER_INSTANCE;
+  }
```
RustedBones, line 61: This should be a nice optimization. I feel we can create the interceptors directly in the `newClientFn` if need be since the client will be shared anyway

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcDoFn.java
RustedBones: I feel this should be done in the scala layer, not here

scio-test/src/test/scala/com/spotify/scio/transforms/AsyncLookupDoFnTest.scala
```diff
@@ -277,8 +268,11 @@ class CallbackFailingScalaLookupDoFn extends ScalaAsyncLookupDoFn[Int, String, A
...
-  override def createCache(): Cache[java.lang.Long, String] =
-    CacheBuilder.newBuilder().build[java.lang.Long, String]()
-  override def getKey(input: Int): java.lang.Long = input.toLong
+// Here we need a custom supplier because guava cache only supports object
+// We can overcome by using a TransformingCache and boxing
+class TestCacheSupplier extends CacheSupplier[Int, String] {
+  override def get(): Cache[Int, String] =
+    new SimpleTransformingCache[Int, java.lang.Integer, String](CacheBuilder.newBuilder().build()) {
+      override protected def transformKey(key: Int): java.lang.Integer = Int.box(key)
+    }
```
RustedBones, line 277: Here is an example of what can be done if the input type of the `AsyncLookupDoFn` can't be supported by the guava cache.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -114,20 +114,29 @@ public <K> BaseAsyncLookupDoFn(int maxPendingRequests, CacheSupplier<A, B, K> ca
...
     this.cacheSupplier = cacheSupplier;
     this.deduplicate = deduplicate;
     this.semaphore = new Semaphore(maxPendingRequests);
   }
 
-  protected abstract C newClient();
+  @Override
+  public ResourceType getResourceType() {
+    return ResourceType.PER_CLONE;
+  }
```
RustedBones, line 127: Maybe we should leave this abstract.
I was conservative here. Previous code looked to be `PER_INSTANCE` but this won't work if client is not thread safe

build.sbt
```diff
@@ -850,6 +851,25 @@ lazy val `scio-extra`: Project = project
...
 
+lazy val `scio-grpc`: Project = project
+  .in(file("scio-grpc"))
+  .settings(commonSettings)
+  .settings(publishSettings)
+  .settings(protobufSettings)
+  .settings(
+    description := "Scio add-on for gRPC",
+    libraryDependencies ++= Seq(
+      "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
```
clairemcginty, line 862: Do we need this dep? It should already be in `scio-core`

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
clairemcginty: Hmm.. maybe we should not override `getResourceType` here and let implementations be explicit about their level of resource-sharing dependent on client type, wdyt? I'm thinking about common use cases like adding rate-limiting, where going from per-instance-->per-clone could be dangerous to do opaquely to the user.

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
clairemcginty: sounds good, since `BaseAsyncLookupDoFn` now extends `DoFnWithResource` and users don't have to worry about the `UUID` key, I think it makes sense to get rid of this complex logic!

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcDoFn.java
```diff
@@ -0,0 +1,127 @@
...
+      return this;
+    }
+
+    public Builder<RequestT, ResponseT, ClientT> withLookupFn(
+        SerializableBiFunction<ClientT, RequestT, ListenableFuture<ResponseT>> lookupFn) {
+      this.lookupFn = lookupFn;
+      return this;
+    }
+
+    public Builder<RequestT, ResponseT, ClientT> withMaxPendingRequests(int maxPendingRequests) {
```
clairemcginty, line 114: Maybe add a short doc about what this means and that it's per-instance, not global?

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcDoFn.java
```diff
@@ -0,0 +1,127 @@
...
+      implements Serializable {
+
+    private int maxPendingRequests = DEFAULT_MAX_PENDING_REQUESTS;
+    private ChannelSupplier channelSupplier;
+    private SerializableFunction<Channel, ClientT> newClientFn;
+    private SerializableBiFunction<ClientT, RequestT, ListenableFuture<ResponseT>> lookupFn;
+
+    protected Builder() {}
+
+    public Builder<RequestT, ResponseT, ClientT> withChannelSupplier(
```
clairemcginty, line 96: what do you think about adding a method to the Builder that allows users to pass in `Interceptor`s (or serializable Interceptor suppliers...)?

RustedBones: I'm not really in favor of this. `Interceptors` are set on the client, it we expose them here, we should probably also the same for every call options. I'd prefer letting this up to the user to configure the client in the `newClientFn`

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcDoFn.java
RustedBones: this is actually per clone since handled by the semaphore in the `BaseAsyncLookupDoFn` so it is per clone. Will update the doc everywhere.

build.sbt
RustedBones: Almost all modules declare it.
As it is also a direct dependency, I think it is better to keep the declaration explicit

PR #4495, opened 2022-08-12
Fix: Capturing ProcessContext in AsyncLookupDoFn leads to NPE in Runner V2

### LINKED ISSUE #4496: Capturing ProcessContext in AsyncLookupDoFn leads to NPE in Runner V2
Beam version: 2.39.0
Scio version: 0.11.9

---

Capturing the `ProcessContext` via `c.timestamp()` in the callbacks within `BaseAsyncLookupDoFn` results in an error when using the Runner V2:

```
    at org.apache.beam.fn.harness.FnApiDoFnRunner$ProcessBundleContextBase.timestamp(FnApiDoFnRunner.java:2496)
    at com.spotify.scio.transforms.BaseAsyncLookupDoFn.lambda$processElement$7(BaseAsyncLookupDoFn.java:203)
    at com.spotify.scio.transforms.FutureHandlers$Guava$1.onSuccess(FutureHandlers.java:66)
    at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1139)
    at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
    at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1277)
    at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:1038)
    at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:783)
    at com.google.common.util.concurrent.SettableFuture.set(SettableFuture.java:49)
    ...
```
which points to [1]
```java
  @Override
  public Instant timestamp() {
    return currentElement.getTimestamp();
  }
```
and `currentElement` appears to be `null` as mentioned in the comment in [2],
```java
  // The member variables below are only valid for the lifetime of certain methods.
  /** Only valid during {@code processElement...} methods, null otherwise. */
  private WindowedValue<InputT> currentElement;
```
as the callback can be invoked asynchronously outside of the lifetime of the `processElement` call.

[1] https://github.com/apache/beam/blob/v2.39.0/sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java#L2494-L2497
[2] https://github.com/apache/beam/blob/v2.39.0/sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java#L258-L260

Illedran:
Closes #4496.

---

I've chosen to refactor the arguments and inject exactly what's needed instead of the whole `ProcessContext`, which seems to no longer even be mentioned as a valid argument in the public docs anymore: https://beam.apache.org/releases/javadoc/2.39.0/org/apache/beam/sdk/transforms/DoFn.ProcessElement.html

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
@@ -140,14 +140,16 @@ public abstract class BaseAsyncLookupDoFn<A, B, C, F, T> extends DoFn<A, KV<A, T
 
   @SuppressWarnings("unchecked")
   @ProcessElement
-  public void processElement(ProcessContext c, BoundedWindow window) {
-    flush(r -> c.output(KV.of(r.input, r.output)));
+  public void processElement(@Element A input,
+                             @Timestamp Instant timestamp,
+                             OutputReceiver<KV<A, T>> outputReceiver,
+                             BoundedWindow window) {
+    flush(r -> outputReceiver.output(KV.of(r.input, r.output)));
 
     // found in cache
-    final A input = c.element();
     B cached = cacheSupplier.get(instanceId, input);
     if (cached != null) {
-      c.output(KV.of(input, success(cached)));
+      outputReceiver.output(KV.of(input, success(cached)));
       return;
     }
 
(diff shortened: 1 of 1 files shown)


RustedBones (approved):
Thanks for the fix.
I was skimming through this file recently and also thought this looking like 'old school' beam.

@clairemcginty should we update the codebase to newer beam style where posible ?
@farzad-sedghi also pointed out that some performance optimization may not happen in the future if we keep accessing the context

clairemcginty (approved):
looks good! @RustedBones , agreed, we should create a ticket to update any outstanding `ProcessContext` usage.

PR #4503, opened 2022-08-18
fix BigQueryType.fromQuery query string formatting examples

luster:
Since `Row.query` is a string, you have to call `Row.query.format(args(0))`, etc.


luster:
ping @clairemcginty

kellen:
This should not be necessary, see: https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala#L142

Issue #4502, opened 2022-08-17
add v0.12.0 scalafix rules


clairemcginty:
add scalafix rules for breaking changes in 0.12.0: https://spotify.github.io/scio/migrations/v0.12.0-Migration-Guide.html

PR #4504, opened 2022-08-23
setup scalafix

farzad-sedghi:
fix https://github.com/spotify/scio/issues/4502


scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -0,0 +1,54 @@
...
+                }
+                val allArgs = (headParam :: tail).mkString(", ")
+                // TODO(farzad): remove ""
+                Patch.replaceTree(a, s"$qual.saveAsBigQueryTable($allArgs)")
+              }
+            case _ =>
+              Patch.empty
+          }
+      }
+      case _ @ Importer(q"com.spotify.scio.extra.bigquery", imps) =>
```
RustedBones, line 37: don't use alias if discarded.
Won't the following quasi-quote work ?
```scala
case q"import com.spotify.scio.extra.bigquery.{..$imps}" =>
```

RustedBones (commented):
For readability I think that would be way clearer to always pattern match using quasiquotes instead of the using the ADT tree.

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -0,0 +1,54 @@
...
+
+class FixBqSaveAsTable extends SemanticRule("FixBqSaveAsTable") {
+  private val scoll = "com/spotify/scio/values/SCollection#"
+
+  override def fix(implicit doc: SemanticDocument): Patch = {
+    doc.tree.collect { case a @ Term.Apply(fun, head :: tail) =>
+      fun match {
+        case Term.Select(qual, name) =>
+          name match {
+            case _ @Term.Name("saveAvroAsBigQuery") if expectedType(qual, scoll)=>
```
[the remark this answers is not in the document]
farzad-sedghi: makes capturing all the vars in all the cases a bit too complicated, as some match with Name and some with Type... might be a better way to do it but I don't think it's worth spending more time :)

scalafix/input-0_12/src/main/scala/fix/LintBqSaveAsTable.scala
```diff
@@ -0,0 +1,30 @@
+/*
+rule = LintBqSaveAsTable
+*/
+package fix
+package v0_12_0
+
+import com.google.api.services.bigquery.model.TableReference
+import com.spotify.scio.bigquery._
+import com.spotify.scio.extra.bigquery._ //TODO(farzad)
```
RustedBones, line 9: Still a TODO ?

RustedBones (approved):

PR #4509, opened 2022-08-29
Do not reference scio coder after materialization

RustedBones:
When materializing coders, we tried to cut ties with scio coders and macro generated magnolia typeclasses.
Unfortunately, the materialized beam coders contain lambda function with outer references to the scio coders.

At serialization time, everything gets written into the graph. we then have:
- large graph size (some users have already reached the dataflow graph size limit)
- streaming pipeline incompatibility. As serialized graph can change from builds (macro code identifiers), updating a streaming pipeline may fail with `The Coder or type for step ... has changed.` even if there wasn't any code change
- slow pipeline start: All the extra info has to be deseriallized by every worker on startup

Previous `xmap` coder transformation could also be problematic: 2 different transforms with the same type signature could be considered as equal. In dataflow v2, we saw that the runner can swap coders if they are equal

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -93,7 +93,7 @@ final case class Record[T] private (
   typeName: String,
   cs: Array[(String, Coder[Any])],
   construct: Seq[Any] => T,
-  destruct: T => Seq[Any]
+  destruct: T => Iterator[Any]
 ) extends Coder[T] {
   override def toString: String = {
     val str = cs.map { case (k, v) => s"($k, $v)" }.mkString(", ")
@@ -282,7 +282,7 @@ final private[scio] case class RecordCoder[T](
   typeName: String,
   cs: Array[(String, BCoder[Any])],
   construct: Seq[Any] => T,
-  destruct: T => Seq[Any]
+  destruct: T => Iterator[Any]
 ) extends StructuredCoder[T] {
 
   @inline def onErrorMsg[A](msg: => String)(f: => A): A =
@@ -293,55 +293,50 @@ final private[scio] case class RecordCoder[T](
     }
 
   override def encode(value: T, os: OutputStream): Unit = {
-    destruct(value)
-      .zip(cs)
-      .foreach { case (v, (l, c)) =>
-        onErrorMsg(
-          s"Exception while trying to `encode` an instance of $typeName: Can't encode field $l value $v"
-        ) {
-          c.encode(v, os)
-        }
+    destruct(value).zip(cs).foreach { case (v, (l, c)) =>
+      onErrorMsg(
+        s"Exception while trying to `encode` an instance of $typeName: Can't encode field $l value $v"
+      ) {
+        c.encode(v, os)
       }
-
+    }
   }
 
   override def decode(is: InputStream): T = {
-    val vs = cs
-      .foldLeft(Seq.newBuilder[Any]) { case (b, (l, c)) =>
-        onErrorMsg(
-          s"Exception while trying to `decode` an instance of $typeName: Can't decode field $l"
-        ) {
-          b += c.decode(is)
-        }
+    val b = Seq.newBuilder[Any]
+    cs.foreach { case (l, c) =>
+      val v = onErrorMsg(
+        s"Exception while trying to `decode` an instance of $typeName: Can't decode field $l"
+      ) {
+        c.decode(is)
       }
-      .result()
-    construct(vs)
+      b += v
+    }
+    construct(b.result())
   }
 
   // delegate methods for determinism and equality checks
 
   override def verifyDeterministic(): Unit = {
-    val problems = cs.toList.flatMap { case (label, c) =>
+    var cause = Option.empty[NonDeterministicException]
+    val reasons = List.newBuilder[String]
+    cs.foreach { case (l, c) =>
       try {
         c.verifyDeterministic()
-        Nil
       } catch {
         case e: NonDeterministicException =>
-          val reason = s"field $label is using non-deterministic $c"
-          List(reason -> e)
+          cause = Some(e)
+          reasons += s"field $l is using non-deterministic $c"
       }
     }
 
-    problems match {
-      case (_, e) :: _ =>
-        val reasons = problems.map { case (reason, _) => reason }
-        throw new NonDeterministicException(this, reasons.asJava, e)
-      case Nil =>
+    cause.foreach { e =>
+      throw new NonDeterministicException(this, reasons.result().asJava, e)
     }
   }
 
   override def toString: String = {
-    val body = cs.map { case (label, c) => s"$label -> $c" }.mkString(", ")
+    val body = cs.map { case (l, c) => s"$l -> $c" }.mkString(", ")
     s"RecordCoder[$typeName]($body)"
   }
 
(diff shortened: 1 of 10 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -173,88 +165,86 @@ final private case class DisjunctionCoder[T, Id](
...
+private[scio] case class RefCoder[T](var bcoder: BCoder[T]) extends DelegatingCoder[T]
+
+private[scio] object RefCoder {
+  def apply[T](): RefCoder[T] = new RefCoder[T](null)
+}
+
+final private[scio] class LazyCoder[T](
   typeName: String,
-  o: CoderMaterializer.CoderOptions
-)(coder: Coder[T])
```
RustedBones, line 179: serializing this may not be deterministic as `coder` could have been derived by a macro, depending on the compiler order, some `anonfun$x` could change.
This also increases serialized graph size.

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -73,29 +75,37 @@ object CoderMaterializer {
...
-        LazyCoder[T](t, o)(c)
+      case r @ Ref(t, c) =>
+        refs.get(r) match {
+          case Some(rc) =>
+            LazyCoder(t, rc.bcoder.asInstanceOf[BCoder[T]])
+          case None =>
+            val rc = RefCoder[T]()
+            refs += r -> rc
+            rc.bcoder = beamImpl(o, c, refs)
+            rc
```
RustedBones, line 107: inspired from file history. See #2605 
@jto was there an issue with this ?

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -173,88 +165,86 @@ final private case class DisjunctionCoder[T, Id](
   override def getCoderArguments: JList[_ <: BCoder[_]] = coders.values.toList.asJava
 }
 
-final private[scio] case class LazyCoder[T](
+sealed private[scio] trait DelegatingCoder[T] extends BCoder[T] {
```
RustedBones, line 168: Makes sure we delegate all public calls to the `bcoder`

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -173,88 +165,86 @@ final private case class DisjunctionCoder[T, Id](
...
+    bcoder.getEncodedTypeDescriptor
+}
+
+private[scio] case class RefCoder[T](var bcoder: BCoder[T]) extends DelegatingCoder[T]
+
+private[scio] object RefCoder {
+  def apply[T](): RefCoder[T] = new RefCoder[T](null)
+}
+
+final private[scio] class LazyCoder[T](
```
RustedBones, line 202: This plays the role of the 'nothing' coder in the previous `uncycled` to avoid `StackOverflow` in case of recursive type.

scio-macros/src/main/scala/com/spotify/scio/MagnoliaMacros.scala
```diff
@@ -40,32 +40,33 @@ private[scio] object MagnoliaMacros {
...
+        q"new _root_.magnolia1.SealedTrait($typeName, $subtypesArray, Array.empty[Any], Array.empty[Any], Array.empty[Any])"
+      case q"Subtype[$tpTC, $tpT, $tpS]($name, $idx, $_, $_, $_, $tc, $isType, $asType)" =>
+        q"_root_.magnolia1.Subtype[$tpTC, $tpT, $tpS]($name, $idx, Array.empty[Any], Array.empty[Any], Array.empty[Any], $tc, $isType, $asType)"
+    }
+
+    // remove all outer references used in rawConstruct
+    // so method serialization do not take any closure
+    val inlineRawConstruct: PartialFunction[Tree, Tree] = {
+      case q"def rawConstruct(..$params): $tpT = {..$statements}" =>
+        val closureCleansed = if (statements.size > 1) statements.tail else statements
```
RustedBones, line 58: This eliminated extra call of 
```scala
MagnoliaUtil.checkParamLengths(fieldValues, params.length, typeName.full)
```
before object creation. The function does not access any of the class members

This should also come with a performance boost

scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
```diff
@@ -17,74 +17,38 @@
...
-              rawConstruct(ps)
-            )
+      // calling patched rawConstruct on empty object should work
+      val emptyCtx = ClosureCleaner.instantiateClass(ctx.getClass).asInstanceOf[CaseClass[Coder, T]]
+      val construct = emptyCtx.rawConstruct _
 
-        Coder.record[T](typeName.full, cs, constructor, destruct)
+      val destruct = if (ctx.isValueClass) {
+        // TODO do not access ctx
+        v: T => Seq(ctx.parameters.head.dereference(v))
```
RustedBones, line 42: Here it gets extremely hard to get access to the wrapped value without using macros again. Since this is a ValueClass, the `ctx` should be relatively small and ok to serialize

scio-core/src/main/scala/com/spotify/scio/coders/DerivedCoder.scala
```diff
@@ -17,74 +17,38 @@
...
+        .result()
 
-        val constructor: Seq[Any] => T =
-          ps =>
-            catching(s"Error while constructing object from parameters $ps")(
-              rawConstruct(ps)
-            )
+      // calling patched rawConstruct on empty object should work
+      val emptyCtx = ClosureCleaner.instantiateClass(ctx.getClass).asInstanceOf[CaseClass[Coder, T]]
+      val construct = emptyCtx.rawConstruct _
```
RustedBones, line 38: As `rawConstruct` is patched, we can call it on a empty object of the same class.

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -318,33 +293,29 @@ final private[scio] case class RecordCoder[T](
...
-    while (i < array.length) {
-      val (label, c) = cs(i)
-      val v = array(i)
-      onErrorMsg(
-        s"Exception while trying to `encode` an instance of $typeName:  Can't encode field $label value $v"
-      ) {
-        c.encode(v, os)
+    destruct(value)
+      .zip(cs)
+      .foreach { case (v, (l, c)) =>
```
RustedBones, line 298: Not sure if working with arrays was done fore performance reasons

RustedBones (commented):
This should probably get bench-marked.

On the graph serialization side, I ran the changes on a relatively complex pipeline. The size decreased from `4.4M` to `3.9M`

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
jto: The PR adds a couple of new tests so I suppose that was the issue :)

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
jto: Yep

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -318,105 +248,249 @@ final private[scio] case class RecordCoder[T](
...
+  // stop call stack and only compare on typeName
+  override def equals(obj: Any): Boolean = obj match {
+    case that: LazyCoder[_] => typeName == that.typeName
+    case _                  => false
+  }
+
+  override def hashCode(): Int = typeName.hashCode
+
+  // stop call stack and not interfere with other result
+  override def verifyDeterministic(): Unit = {}
```
clairemcginty, line 453: I'm trying to think about possible ramifications for not making the `verifyDeterministic` call and whether that could cause issues.  I wonder if we could invoke it when `bCoder` is initialized, like:

```scala
override lazy val bcoder: BCoder[T] = {
  verifyDeterministic()
  bc
}
```

wdyt?

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -65,61 +65,117 @@ Cannot find an implicit Coder instance for type:
...
-    s"Record($typeName, $str)"
+    val body = cs.map { case (k, v) => s"($k, $v)" }.mkString(", ")
+    s"Record[$typeName]($body)"
   }
 }
 
 // KV are special in beam and need to be serialized using an instance of KvCoder.
-final case class KVCoder[K, V] private (koder: Coder[K], voder: Coder[V]) extends Coder[KV[K, V]] {
-  override def toString: String = s"KVCoder($koder, $voder)"
+final case class KVCoder[K, V] private (koder: Coder[K], voder: Coder[V]) extends Coder[KV[K, V]]
```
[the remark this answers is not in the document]
RustedBones: it was equivalent to the default one provided by `case class`

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
RustedBones: This is not an initialization problem. `verifyDeterministic` on a structured coder will call recursively on every field, if we reach a `LazyCoder`, we do not want to call again on the base struct. we just stop here. It is fine to not throw. determisistic decision is left to the other fields

clairemcginty: ah right, of course 👍

clairemcginty (approved):

Issue #4497, opened 2022-08-12
Check and update usage of Beam ProcessContext in ProcessElement functions


RustedBones:
Beam's new way of writing custom do function is to use annotations in the `@ProcessElement` function.

Check that `ProcessContext` is only used where relevant
Update to newer beam style when possible or indicate in code why we still require to access the ProcessContext

PR #4511, opened 2022-08-30
Changed processElement to newer annotated style in scio-avro, scio-core

shnapz:
Original request:
> Beam's new way of writing custom do function is to use annotations in the @ProcessElement function.
>
> Check that ProcessContext is only used where relevant
> Update to newer beam style when possible or indicate in code why we still require to access the ProcessContext

I have committed only one change to discuss and elaborate request details:
SortedBucketTransform.processElement signature is changed from ProcessContext argument to "element" and "outputReceiver"

[Original ticket](https://github.com/spotify/scio/issues/4497.)

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
@@ -22,6 +22,7 @@ import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.beam.sdk.transforms.display.DisplayData;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
+import org.apache.beam.sdk.values.KV;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -80,10 +81,15 @@ public class FileDownloadDoFn<OutputT> extends DoFn<URI, OutputT> {
   }
 
   @ProcessElement
-  public void processElement(ProcessContext c, BoundedWindow window) {
-    batch.add(new Element(c.element(), c.timestamp(), window));
+  public void processElement(
+      @DoFn.Element URI element,
+      @Timestamp Instant timestamp,
+      OutputReceiver<OutputT> outputReceiver,
+      BoundedWindow window
+  ) {
+    batch.add(new Element(element, timestamp, window));
     if (batch.size() >= batchSize) {
-      processBatch(c);
+      processBatch(outputReceiver);
     }
   }
 
@@ -100,13 +106,13 @@ public class FileDownloadDoFn<OutputT> extends DoFn<URI, OutputT> {
         .add(DisplayData.item("Keep Downloaded Files", keep));
   }
 
-  private void processBatch(ProcessContext c) {
+  private void processBatch(OutputReceiver<OutputT> outputReceiver) {
     if (batch.isEmpty()) {
       return;
     }
     LOG.info("Processing batch of {}", batch.size());
     List<URI> uris = batch.stream().map(e -> e.uri).collect(Collectors.toList());
-    remoteFileUtil.download(uris).stream().map(fn::apply).forEach(c::output);
+    remoteFileUtil.download(uris).stream().map(fn::apply).forEach(outputReceiver::output);
     if (!keep) {
       LOG.info("Deleting batch of {}", batch.size());
       remoteFileUtil.delete(uris);
--- a/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
@@ -22,7 +22,6 @@ import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.beam.sdk.transforms.display.DisplayData;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
-import org.apache.beam.sdk.values.KV;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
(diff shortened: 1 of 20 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -216,8 +216,9 @@ public FinalizeTransformedBuckets(
     }
 
     @ProcessElement
-    public void processElement(ProcessContext c) throws IOException {
-      final Iterator<MergedBucket> mergedBuckets = c.element().iterator();
+    public void processElement(@DoFn.Element Iterable<MergedBucket> element,
```
clairemcginty, line 219: 🎨 Import `org.apache.beam.sdk.transforms.DoFn.Element` directly and just reference it as `@Element`?

clairemcginty (approved):
looks great! :tada:

shnapz:
The scope of https://github.com/spotify/scio/issues/4497 is pretty wide and it will require change of 10+ occurrences of `@ProcessElement`. I am concerned that mostly it is not covered with Unit tests. I tried to write unit test and failed it for:
scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
Is it even possible to mock something, and write unit test on AvroIO?
cc: @clairemcginty @RustedBones @kellen

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
```diff
@@ -470,7 +475,7 @@ public void processElement(
               false,
               bucketId,
               effectiveParallelism,
-              context.getPipelineOptions());
+              pipelineOptions);
```
RustedBones, line 478: any reason to change this if context is still required ?

RustedBones (approved):
Very nice cleanup !

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketTransform.java
shnapz: I will revert this

shnapz:
Covered scio-avro and scio-core

scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.beam.sdk.transforms.display.DisplayData;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
+import org.apache.beam.sdk.values.KV;
```
RustedBones, line 25: unused import ?

scio-core/src/main/scala/com/spotify/scio/util/FunctionsWithSideInput.scala
```diff
@@ -35,6 +35,10 @@ private[scio] object FunctionsWithSideInput {
   def filterFn[T](f: (T, SideInputContext[T]) => Boolean): DoFn[T, T] =
     new SideInputDoFn[T, T] {
       val g = ClosureCleaner.clean(f) // defeat closure
+
+      /*
+       * ProcessContext is required as an argument because it is passed to SideInputContext
+       * */
```
RustedBones, line 41: Thanks for adding those on every processElement where new syntax can't be applied :heart:

scio-test/src/test/scala/com/spotify/scio/transforms/AsyncDoFnSpec.scala
```diff
@@ -215,7 +199,14 @@ abstract class AsyncDoFnTester[P[_], F[_]] extends BaseDoFnTester {
...
   override def request(): Unit = {
-    fn.processElement(processContext, null)
+    val input: Int = nextElement
+    val timestamp = DateTime.parse("2022-08-31").toInstant
+    val outputReceiver = new OutputReceiver[String] {
+      override def output(output: String): Unit = outputBuffer.append(output)
+      override def outputWithTimestamp(output: String, timestamp: Instant): Unit = ???
+    }
+    val window: BoundedWindow = null
+    fn.processElement(input, timestamp, outputReceiver, window)
```
RustedBones, line 209: Test simplification as side effect is cool

RustedBones (approved):

scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
shnapz: Fixed!

clairemcginty (approved):
Looks great and thorough! 🎉

shnapz:
Code coverage has been run in this PR. Assuming that all changes are covered by existing tests, because no new code blocks were written

scio-core/src/main/scala/com/spotify/scio/util/FunctionsWithSideInput.scala
clairemcginty: Just checking, these outstanding usages shouldn't be impacted by #4496 because we're not calling `c.timestamp()`, right? :)

shnapz: Not sure. In this particular case ProcessContext is exposed with implementations of `trait SideInput[T]`:
https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala#L31

clairemcginty: I did a quick test with [SideInOutExample](https://github.com/spotify/scio/blob/main/scio-examples/src/main/scala/com/spotify/scio/examples/extra/SideInOutExample.scala) on Dataflow with RunnerV2 and it worked fine, so we should be good! I'll do a quick check on other `c.timestamp()` usages

PR #4533, opened 2022-09-22
Fix example page using showdown

### LINKED ISSUE #4501: Examples page won't load
The scio examples page no longer loads, at least not in Chrome or Safari: https://spotify.github.io/scio/examples/index.html

Looking at the page source, the content seems to be there, but never rendered. Chrome gives me the following error:

`index.html:79 Cross-Origin Read Blocking (CORB) blocked cross-origin response https://strapdownjs.com/v/0.2/strapdown.js with MIME type text/html. See https://www.chromestatus.com/feature/5629709824032768 for more details.
`

RustedBones:
Fix #4501 

Preview:
image


saucoide:
@RustedBones This doesn't seem to be deployed?  
Didn't want to open a new issue, but the site is still serving the strapdownjs version

RustedBones:
Hi @saucoide. Thanks for the heads-up. It was deployed, until we released a patch for scio 0.11. The branch has been started before the fix and we forgot to cherry-pick it. Will redeploy the doc
CC @clairemcginty

PR #4536, opened 2022-09-27
Detect and retrieve GCP temp location (fixes #4535)

### LINKED ISSUE #4535: NPE on 0.12.0 SNAPSHOT when setting gcpTempLocation
We use gcpTempLocation to locate the temp directory in the same region than runners. The parameter is set like this: `--gcpTempLocation=gs://our_bucket_in_the_right_region/temp`.

When trying 0.12.0 (last one being 0.12.0-RC2+1-e8854f73-SNAPSHOT), I get an NPE:

```
Exception in thread "main" java.lang.NullPointerException
	at java.base/java.util.regex.Matcher.getTextLength(Matcher.java:1770)
	at java.base/java.util.regex.Matcher.reset(Matcher.java:416)
	at java.base/java.util.regex.Matcher.<init>(Matcher.java:253)
	at java.base/java.util.regex.Pattern.matcher(Pattern.java:1133)
	at org.apache.beam.sdk.io.FileSystems.parseScheme(FileSystems.java:501)
	at org.apache.beam.sdk.io.FileSystems.matchNewResource(FileSystems.java:587)
	at com.spotify.scio.util.ScioUtil$$anonfun$tempDirOrDefault$2.apply(ScioUtil.scala:121)
	at com.spotify.scio.util.ScioUtil$$anonfun$tempDirOrDefault$2.apply(ScioUtil.scala:119)
	at scala.Option.getOrElse(Option.scala:201)
	at com.spotify.scio.util.ScioUtil$.tempDirOrDefault(ScioUtil.scala:119)
	at com.spotify.scio.io.TextIO.write(TextIO.scala:93)
	<snip>
```

Looks like `ScioUtil` has been changed in @kellen's PR #4461 a few days ago but I don't see how the `tempDirOrDefault` function has been affected.

turb:
Detects if pipeline options is a `GcpOptions` and retrieve `gcpTempLocation` instead of `tempLocation` if it is the case.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -115,12 +115,16 @@ private[scio] object ScioUtil {
     DefaultFilenamePolicy.fromStandardParameters(prefix, shardTemplate, suffix, isWindowed)
   }
 
-  def tempDirOrDefault(tempDirectory: String, sc: ScioContext): ResourceId = {
-    Option(tempDirectory).map(toResourceId).getOrElse {
-      val tempLocationOpt: String = sc.options.getTempLocation
-      FileSystems.matchNewResource(tempLocationOpt, true)
-    }
-  }
+  def tempDirOrDefault(tempDirectory: String, sc: ScioContext): ResourceId =
+    Option(tempDirectory)
+      .map(toResourceId)
+      .orElse(
+        Try(sc.optionsAs[GcpOptions]).toOption
+          .flatMap(gcpOptions => Option(gcpOptions.getGcpTempLocation))
+          .orElse(Option(sc.options.getTempLocation))
+          .map(FileSystems.matchNewResource(_, true))
+      )
+      .getOrElse(throw new IllegalStateException("Unable to find any temp location"))
 
   def isWindowed(coll: SCollection[_]): Boolean =
     coll.internal.getWindowingStrategy != WindowingStrategy.globalDefault()
--- a/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
@@ -115,16 +115,13 @@ private[scio] object ScioUtil {
     DefaultFilenamePolicy.fromStandardParameters(prefix, shardTemplate, suffix, isWindowed)
   }
 
-  def tempDirOrDefault(tempDirectory: String, sc: ScioContext): ResourceId =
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -115,12 +115,16 @@ private[scio] object ScioUtil {
...
+  def tempDirOrDefault(tempDirectory: String, sc: ScioContext): ResourceId =
+    Option(tempDirectory)
+      .map(toResourceId)
+      .orElse(
+        Try(sc.optionsAs[GcpOptions]).toOption
+          .flatMap(gcpOptions => Option(gcpOptions.getGcpTempLocation))
+          .orElse(Option(sc.options.getTempLocation))
+          .map(FileSystems.matchNewResource(_, true))
+      )
+      .getOrElse(throw new IllegalStateException("Unable to find any temp location"))
```
kellen, line 127: We can re-use `toResourceId`. I think we'd also prefer to use `tempLocation` over `gcpTempLocation`?

```suggestion
  def tempDirOrDefault(tempDirectory: String, sc: ScioContext): ResourceId = {
    Option(tempDirectory)
      .orElse(Option(sc.options.getTempLocation))
      .orElse(Try(sc.optionsAs[GcpOptions]).toOption.flatMap(x => Option(x.getGcpTempLocation)))
      .map(toResourceId)
      .getOrElse(throw new IllegalStateException("Unable to find any temp location"))
  }
```

scio-core/src/main/scala/com/spotify/scio/util/ScioUtil.scala
```diff
@@ -116,10 +116,11 @@ private[scio] object ScioUtil {
...
   def tempDirOrDefault(tempDirectory: String, sc: ScioContext): ResourceId = {
-    Option(tempDirectory).map(toResourceId).getOrElse {
-      val tempLocationOpt: String = sc.options.getTempLocation
-      FileSystems.matchNewResource(tempLocationOpt, true)
-    }
+    Option(tempDirectory)
+      .orElse(Option(sc.options.getTempLocation))
+      .orElse(Try(sc.optionsAs[GcpOptions]).toOption.flatMap(x => Option(x.getGcpTempLocation)))
+      .map(toResourceId)
+      .getOrElse(throw new IllegalStateException("Unable to find any temp location"))
```
kellen, line 123: Apologies missed this on the first go. `IllegalArgumentException` is what we use elsewhere, also updated the message to reflect what gets thrown in `getTempFile`:
```suggestion
      .getOrElse(throw new IllegalArgumentException("No temporary location was specified. Specify a temporary location via --tempLocation or PipelineOptions.setTempLocation."))
```

PR #4544, opened 2022-10-17
First ensures table exists

michaelyangbling:


michaelyangbling:
looks like gcp cannot creates instance if ensureTables happens before, need to create instance manually,,,,

PR #4547, opened 2022-10-19
Fix record coder structural value equality

### LINKED ISSUE #4541: Coder: Mutation detector warning about structural values
I'm currently testing with 0.12-RC2 and in my tests I see a lot of messages from `org.apache.beam.sdk.util.MutationDetectors` are appearing. I could trace it down due to the usage of floating point number in my case classes.

Please see the following reproducer:

```
import com.spotify.scio.coders.{Coder, CoderMaterializer}
import org.apache.beam.sdk.util.MutationDetectors

case class Test(value: Double)

object CoderTest {

  implicit val testCoder: Coder[Test] = Coder.gen[Test]

  def main(args: Array[String]): Unit = {
    val value = Test(value = 3.9)
    val detector = MutationDetectors.forValueWithCoder(value, CoderMaterializer.beamWithDefault(testCoder))
    detector.verifyUnmodified()
  }

}
```

This produces: 
> WARN org.apache.beam.sdk.util.MutationDetectors - Coder of type class com.spotify.scio.coders.MaterializedCoder has a #structuralValue method which does not return true when the encoding of the elements is equal. Element Test(3.9)

I don't know if the root cause issue originates from scio or Apache Beam. 

Would be great if you can take a look at that before releasing 0.12. Thanks a lot!

RustedBones:
Fix #4541

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -26,7 +26,7 @@ import org.apache.beam.sdk.coders.{Coder => BCoder, CustomCoder, StructuredCoder
 import org.apache.beam.sdk.util.common.ElementByteSizeObserver
 import org.apache.beam.sdk.values.KV
 
-import java.util.{Arrays => JArrays, Collections, List => JList, Objects}
+import java.util.{Collections, List => JList, Objects}
 import scala.annotation.implicitNotFound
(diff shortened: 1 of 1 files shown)


clairemcginty (approved):

RustedBones:
Confirmed in the [logs](https://github.com/spotify/scio/actions/runs/3282874450/jobs/5406862718) that the message disappeared

PR #4549, opened 2022-10-20
Update to beam 2.42

RustedBones:


build.sbt
```diff
@@ -40,46 +40,46 @@ val commonsCodecVersion = "1.15"
...
-val googleApiServicesDataflowVersion = s"v1b3-rev20210818-$googleClientsVersion"
-val googleApiServicesPubsubVersion = s"v1-rev20211130-$googleClientsVersion"
-val googleApiServicesStorageVersion = s"v1-rev20211201-$googleClientsVersion"
+val googleApiServicesBigQueryVersion = s"v2-rev20220827-$googleClientsVersion"
+val googleApiServicesDataflowVersion = s"v1b3-rev20220812-$googleClientsVersion"
+val googleApiServicesPubsubVersion = s"v1-rev20220829-$googleClientsVersion"
+val googleApiServicesStorageVersion = s"v1-rev20220705-$googleClientsVersion"
 
 // check versions from libraries-bom
 // https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/index.html
```
shnapz, line 60: Do you think it is worth to specify in comments that "com.google.cloud:libraries-bom:26.1.1" is used? Otherwise it is not directly clear

shnapz (commented):
I see that tests are failing. Have couple of questions

build.sbt
```diff
@@ -1438,6 +1438,5 @@ ThisBuild / dependencyOverrides ++= Seq(
   "io.opencensus" % "opencensus-contrib-grpc-metrics" % opencensusVersion,
   "org.checkerframework" % "checker-qual" % checkerFrameworkVersion,
   "org.codehaus.mojo" % "animal-sniffer-annotations" % animalSnifferAnnotationsVersion,
-  "org.scala-lang.modules" %% "scala-xml" % scalaXmlVersion,
-  "org.threeten" % "threetenbp" % threetenbpVersion
```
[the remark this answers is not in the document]
RustedBones: We're indeed not supporting java 6/7 since long time. This dep was added because it showed as conflict in the `evicted` in previous beam version, so we favored the one shipped with beam.
Since 2.41, this dep is not used anymore by beam, and there is no conflicts. Hence removing the definition here

build.sbt
RustedBones: I know. I wanted to avoid doing like in beam build file, where they also have a link which is very often not updated and points to the wrong version.

RustedBones:
Tests are failing due to a new extra check on output coders, expecting strict `KV`

RustedBones:
The PR to incriminate: https://github.com/apache/beam/commit/ec44ac9a9e305a813be54f81bef8eac36df02d0c
It adds a check that `GroupByKey` output value coder is a strict iterable wrapping wrapping:

input: `KV(MaterializedCoder(StringUtf8Coder), MaterializedCoder(VarIntCoder))`
expected output `KV(MaterializedCoder(StringUtf8Coder), Iterable(MaterializedCoder(VarIntCoder))`
actual output `KV(MaterializedCoder(StringUtf8Coder), MaterializedCoder(Iterable(VarIntCoder))`

lukecwik:
> The PR to incriminate: [apache/beam#22702](https://github.com/apache/beam/pull/22702) It adds a check that `GroupByKey` output value coder is a strict iterable wrapping:
> 
> input: `KV(MaterializedCoder(StringUtf8Coder), MaterializedCoder(VarIntCoder))` expected output `KV(MaterializedCoder(StringUtf8Coder), Iterable(MaterializedCoder(VarIntCoder))` actual output `KV(MaterializedCoder(StringUtf8Coder), MaterializedCoder(Iterable(VarIntCoder))`

It seems as though there is logic already in SCIO to handle the case where the input/output is a KvCoder and is not wrapped with a MaterializedCoder. Can the logic be extended for the value coder on the output PCollection to wrap with Iterable first?

shnapz:
`should work with a windowed collection` fails on Java 11, 2.12.17, but not in java 8, 2.13.8.
I wonder why we don't test on:
```
        exclude:
          - scala: 2.12.17
            java: 8
```

RustedBones:
Tests are failing due to concurrency issue from beam [here](https://github.com/apache/beam/pull/23811)
we only cross build for latest supported JVM and normal test on previous JVM (2.13 is the default prject's scala version)

lukecwik:
> Tests are failing due to concurrency issue from beam [here](https://github.com/apache/beam/pull/23811) we only cross build for latest supported JVM and normal test on previous JVM (2.13 is the default prject's scala version)

Did you mean to say that https://github.com/apache/beam/pull/23811 resolved the issue or created a new issue?

RustedBones:
Moving straight to beam 2.43 with #4607

PR #4550, opened 2022-10-20
Fix NullPointerException with extendedDebugInfo=true

RustedBones:


RustedBones:
Updated desc

RustedBones:
After change, jhm bench are still on par with 0.11 baseline
Generated graph size decreased:
- 0.11: 4.4M
- 0.12.0-RC1: 2.9M
- PR: 2.1M

clairemcginty (approved):

PR #4555, opened 2022-10-25
Concurrently support legacy Parquet reads for 0.12.x

clairemcginty:
To make the 0.12.x migration easier, we're letting users opt into the new SDF-based Parquet implementation, while keeping the legacy implementation as the default throughout 0.12.x. Hopefully we can fully deprecate+remove legacy code by 0.13.x.

the `ParquetConfiguration` convenience method is a little out of scope of this PR, I can move it into a new PR if that's easier - I was just getting annoyed at constructing `Configuration` objects in Scala....

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -171,15 +171,15 @@ object ParquetAvroIO {
 
     def read(sc: ScioContext, path: String)(implicit coder: Coder[T]): SCollection[T] = {
       val jobConf = Option(conf).getOrElse(new Configuration())
-      if (
-        jobConf.getBoolean(
-          ParquetReadConfiguration.UseSplittableDoFn,
-          ParquetReadConfiguration.UseSplittableDoFnDefault
-        )
-      ) {
-        readSplittableDoFn(sc, jobConf, path)(coder)
+      val useSplittableDoFn = jobConf.getBoolean(
+        ParquetReadConfiguration.UseSplittableDoFn,
+        ParquetReadConfiguration.UseSplittableDoFnDefault
+      )
+
+      if (useSplittableDoFn) {
+        readSplittableDoFn(sc, jobConf, path)
       } else {
-        readLegacy(sc, jobConf, path)(coder)
+        readLegacy(sc, jobConf, path)
       }
     }
 
@@ -204,7 +204,7 @@ object ParquetAvroIO {
         ParquetInputFormat.setFilterPredicate(conf, predicate)
       }
 
-      val tCoder = CoderMaterializer.beam(sc, coder)
+      val bCoder = CoderMaterializer.beam(sc, coder)
       val cleanedProjectionFn = ClosureCleaner.clean(projectionFn)
 
       sc.applyTransform(
@@ -214,7 +214,7 @@ object ParquetAvroIO {
           path,
           Functions.serializableFn(cleanedProjectionFn)
         )
-      ).setCoder(tCoder)
+      ).setCoder(bCoder)
     }
 
     private def readLegacy(sc: ScioContext, conf: Configuration, path: String)(implicit
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -250,9 +250,7 @@ object ParquetAvroIO {
       val transform = HadoopFormatIO
         .read[JBoolean, T]()
         // Hadoop input always emit key-value, and `Void` causes NPE in Beam coder
-        .withKeyTranslation(new SimpleFunction[Void, JBoolean]() {
-          override def apply(input: Void): JBoolean = true
-        })
+        .withKeyTranslation(Functions.simpleFn[Void, JBoolean](_ => true))
         .withValueTranslation(new SimpleFunction[A, T]() {
           // Workaround for incomplete Avro objects
           // `SCollection#map` might throw NPE on incomplete Avro objects when the runner tries
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -172,7 +172,12 @@ object ParquetAvroIO {
 
     def read(sc: ScioContext, path: String)(implicit coder: Coder[T]): SCollection[T] = {
       val jobConf = Option(conf).getOrElse(new Configuration())
-      if (jobConf.getBoolean(ParquetReadConfiguration.UseSplittableDoFn, false)) {
+      if (
+        jobConf.getBoolean(
+          ParquetReadConfiguration.UseSplittableDoFn,
+          ParquetReadConfiguration.UseSplittableDoFnDefault
+        )
+      ) {
         readSplittableDoFn(sc, jobConf, path)(coder)
       } else {
         readLegacy(sc, jobConf, path)(coder)
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -31,7 +31,6 @@ import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.reflect.ReflectData
 import org.apache.avro.specific.SpecificRecordBase
-import org.apache.beam.sdk.coders
 import org.apache.beam.sdk.io._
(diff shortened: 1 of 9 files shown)


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,38 +169,102 @@ object ParquetAvroIO {
     val readSchema: Schema =
       if (isSpecific) ReflectData.get().getSchema(avroClass) else projection
 
-    def read(sc: ScioContext, path: String): SCollection[T] = {
-      val hadoopConf = Option(conf).getOrElse(new Configuration())
+    def read(sc: ScioContext, path: String)(implicit coder: Coder[T]): SCollection[T] = {
```
RustedBones, line 172: Can't we set the coder as implicit at the class level, where the type parameter is ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,38 +169,102 @@ object ParquetAvroIO {
...
+      val jobConf = Option(conf).getOrElse(new Configuration())
+      if (
+        jobConf.getBoolean(
+          ParquetReadConfiguration.UseSplittableDoFn,
+          ParquetReadConfiguration.UseSplittableDoFnDefault
+        )
```
RustedBones, line 178: nit, the condition is a bit long. save it in a val for better readability

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,38 +169,102 @@ object ParquetAvroIO {
...
+      ) {
+        readSplittableDoFn(sc, jobConf, path)(coder)
+      } else {
+        readLegacy(sc, jobConf, path)(coder)
```
RustedBones, line 182: why giving the coder explicitly ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,38 +169,102 @@ object ParquetAvroIO {
...
-        AvroReadSupport.setRequestedProjection(hadoopConf, projection)
+        AvroReadSupport.setRequestedProjection(conf, projection)
       }
       if (predicate != null) {
-        ParquetInputFormat.setFilterPredicate(hadoopConf, predicate)
+        ParquetInputFormat.setFilterPredicate(conf, predicate)
       }
 
-      val tCoder = sc.pipeline.getCoderRegistry.getCoder(ScioUtil.classOf[T])
+      val tCoder = CoderMaterializer.beam(sc, coder)
```
RustedBones, line 207:
```suggestion
      val bCoder = CoderMaterializer.beam(sc, coder)
```
For consisteny

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,38 +169,102 @@ object ParquetAvroIO {
...
+
+      val g = ClosureCleaner.clean(projectionFn) // defeat closure
+      val aCls = avroClass
+      val oCls = ScioUtil.classOf[T]
+      val transform = HadoopFormatIO
+        .read[JBoolean, T]()
+        // Hadoop input always emit key-value, and `Void` causes NPE in Beam coder
+        .withKeyTranslation(new SimpleFunction[Void, JBoolean]() {
+          override def apply(input: Void): JBoolean = true
+        })
```
RustedBones, line 255: Can't we write ?
```suggestion
        .withKeyTranslation((input: Void) => true: JBoolean)
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -48,6 +57,23 @@ final case class ParquetExampleIO(path: String) extends ScioIO[Example] {
 
   override protected def read(sc: ScioContext, params: ReadP): SCollection[Example] = {
     val conf = Option(params.conf).getOrElse(new Configuration())
+    if (
+      conf.getBoolean(
+        ParquetReadConfiguration.UseSplittableDoFn,
+        ParquetReadConfiguration.UseSplittableDoFnDefault
+      )
```
RustedBones, line 64: same here

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -71,6 +97,35 @@ final case class ParquetExampleIO(path: String) extends ScioIO[Example] {
...
+    if (params.predicate != null) {
+      ParquetInputFormat.setFilterPredicate(job.getConfiguration, params.predicate)
+    }
+
+    val source = HadoopFormatIO
+      .read[JBoolean, Example]()
+      // Hadoop input always emit key-value, and `Void` causes NPE in Beam coder
+      .withKeyTranslation(new SimpleFunction[Void, JBoolean]() {
+        override def apply(input: Void): JBoolean = true
+      })
```
RustedBones, line 124: same here

site/src/main/paradox/migrations/v0.12.0-Migration-Guide.md
```diff
@@ -79,3 +79,35 @@ case class MyClass(s: String, i: Int)
 val scoll: SCollection[MyClass] = ???
 scoll.saveAsDynamicParquetAvroFile("gs://output/") { m => s"/${m.s}/${m.i}"}
 ```
+
+## Parquet Reads
+
+Alongside the existing Parquet read implementation ("legacy Parquet"), we're concurrently offering a new Parquet read implementation that uses Beam's new [SplittableDoFn](https://beam.apache.org/blog/splittable-do-fn/)
+API. Legacy Parquet is still the default read format, but can enable the new implementation in your `Configuration`:
```
RustedBones, line 86: strange to state we default to legacy. would be preferable that we offer new implementation instead :)

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,38 +169,102 @@ object ParquetAvroIO {
...
+      }
+      if (predicate != null) {
+        ParquetInputFormat.setFilterPredicate(job.getConfiguration, predicate)
+      }
+
+      val g = ClosureCleaner.clean(projectionFn) // defeat closure
+      val aCls = avroClass
+      val oCls = ScioUtil.classOf[T]
+      val transform = HadoopFormatIO
+        .read[JBoolean, T]()
```
shnapz, line 251: Why do we call it "legacy" and "default" if it is a newly written code?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
clairemcginty: sadly no, we need to fully write it out for Scala 2.12 interop 😭

site/src/main/paradox/migrations/v0.12.0-Migration-Guide.md
clairemcginty: I know 😢 I'm just concerned about performance degradation if users can't immediately update to RunnerV2. Let's aim to make it the default by 0.13.x?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
clairemcginty: it's actually copy-pasted from the 0.11.12 branch :) https://github.com/spotify/scio/tree/v0.11.12/scio-parquet/src/main/scala/com/spotify/scio/parquet this is the code that the SplittabledoFn Parquet implementation replaced.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
clairemcginty: I thought about that.... This code was set up really oddly -- the read logic is implemented within the `ReadParam` case class -- and I thought from a design perspective, it's extra strange to make a ReadParam[T] require an implicit T: Coder (plus, could be a breaking change from the user perspective if they're constructing a ReadParam outside the Scio job class?). So, this felt like the least weird way to implement this, lol.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
RustedBones: Yes, `SimpleFunction` is not an interface but we have an helper for that
```suggestion
        .withKeyTranslation(Functions.simpleFn[Void, JBoolean](_ => true))
```

PR #4576, opened 2022-10-26
Increase apache.avro version to aling with Magnolify

shnapz:
Magnolify depends on the newer version of org.apache.avro. Updating the current one to the latest.

Assuming that it is fully backward compatible with "1.8.2". After implementing TypedParquet annotations Magnolify code depends on some new features of org.apache.avro and fails with MethodNotFound in the current Scio. This can be quickly fixed by explicit dependency in users' pipeline


shnapz:
Looks like we can't upgrade that version. Will try to decrease version in magnolify

clairemcginty:
Magnolify should be fine too since avro version is Provided :) https://github.com/spotify/magnolify/blob/6b536a9465cf8afcfbdabc08b7f8fc918228c23a/build.sbt#L336

PR #4590, opened 2022-11-07
Remove unnecessary AvroCoderMacros

RustedBones:
specific record class schema can be loaded using the `SpecificData.getSchema`. No need of a macro


scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
```diff
@@ -125,8 +126,12 @@ trait AvroCoders {
     Coder.beam(new SlowGenericRecordCoder)
 
   import org.apache.avro.specific.SpecificRecordBase
-  implicit def genAvro[T <: SpecificRecordBase]: Coder[T] =
-    macro AvroCoderMacros.staticInvokeCoder[T]
+  implicit def genAvro[T <: SpecificRecordBase: ClassTag]: Coder[T] = {
```
shnapz, line 129: Is `SpecificRecord` type applicable here?

RustedBones: Do you mean instead of `SpecificRecordBase` ?
It should work. We can probably relax our type bound everywhere

shnapz: Yes, and it remains backward compatible

RustedBones:
Mima error
```
[error] scio-macros: Failed binary compatibility check against com.spotify:scio-macros_2.13:0.12.0! Found 2 potential problems
[error]  * class com.spotify.scio.coders.AvroCoderMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("com.spotify.scio.coders.AvroCoderMacros")
[error]  * object com.spotify.scio.coders.AvroCoderMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("com.spotify.scio.coders.AvroCoderMacros$")
```

shnapz:
So `Mima error` is fine as long as it's 0.12?

RustedBones:
Done in #4713

PR #4603, opened 2022-11-16
Do not wrap already nullable coders

RustedBones:
When `nullableCoders` is enabled, we can end-up with multiple wrapping of the coder and failure to work on pair collections for instance:
```log
Failed to extract key-value coders from Coder[(K, V)]: NullableCoder(NullableCoder(...
```

This is mainly due to coder extraction on pair collections, where the unwrapped materialized code is reused. In this case, we should not re-wrap the coder if it is already nullable


kellen (approved):
👍

PR #4607, opened 2022-11-18
Update beam to v2.43.0

RustedBones:


scio-core/src/main/scala/com/spotify/scio/schemas/instances/JavaInstances.scala
```diff
@@ -19,12 +19,10 @@ package com.spotify.scio.schemas.instances
 
 import java.util
 import java.time.LocalDate
-
 import com.spotify.scio.IsJavaBean
 import com.spotify.scio.schemas.{ArrayType, MapType, RawRecord, Schema, Type}
 import org.apache.beam.sdk.schemas.JavaBeanSchema
 import org.apache.beam.sdk.schemas.Schema.{FieldType, LogicalType}
-import org.apache.beam.sdk.extensions.sql.impl.utils.CalciteUtils
```
RustedBones, line 27: accessing beam sql extension in core pulls in the google-cloud sdk

build.sbt
```diff
@@ -337,22 +337,26 @@ lazy val assemblySettings = Seq(
   assembly / test := {},
   assembly / assemblyMergeStrategy ~= { old =>
     {
-      case s if s.endsWith(".properties")            => MergeStrategy.filterDistinctLines
-      case s if s.endsWith("public-suffix-list.txt") => MergeStrategy.filterDistinctLines
-      case s if s.endsWith("pom.xml")                => MergeStrategy.last
-      case s if s.endsWith(".class")                 => MergeStrategy.last
```
RustedBones, line 343: this is very risky. we want to fail when any new class conflict arises

build.sbt
```diff
@@ -604,48 +609,58 @@ lazy val `scio-macros`: Project = project
   .settings(
     description := "Scio macros",
     libraryDependencies ++= Seq(
+      // compile
       "com.chuusai" %% "shapeless" % shapelessVersion,
       "com.esotericsoftware" % "kryo-shaded" % kryoVersion,
-      "org.apache.beam" % "beam-sdks-java-extensions-sql" % beamVersion,
```
RustedBones, line 609: This is not used in `macros` and pull too many things

build.sbt
```diff
@@ -654,108 +669,111 @@ lazy val `scio-google-cloud-platform`: Project = project
...
       "joda-time" % "joda-time" % jodaTimeVersion,
-      "junit" % "junit" % junitVersion % "test",
       "org.apache.avro" % "avro" % avroVersion,
       "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
       "org.apache.beam" % "beam-sdks-java-extensions-google-cloud-platform-core" % beamVersion,
-      "org.apache.beam" % "beam-sdks-java-io-google-cloud-platform" % beamVersion,
+      "org.apache.beam" % "beam-sdks-java-io-google-cloud-platform" % beamVersion excludeAll (
+        // remove transitive test dependencies at compile time
+        "junit" % "junit",
+        "org.hamcrest" % "hamcrest-core"
```
RustedBones, line 703: `beam-sdks-java-io-google-cloud-platform` packages some junit test suite in the main artifact.
Just exclude `junit` and `hamcrest-core` (and also, we should prefer `hamcrest` instead of sub modules artifacts)

colmsnowplow:
👋 this PR looks to solve some issues we have with dependencies. 

Any idea how close/far it is to making it into a release? Just trying to decide if a temporary fix on our side is worth the effort. :)

RustedBones:
Hi @colmsnowplow. We just had some bug fixes merged. This will be the next step for `0.12.1` and probably `0.11.14`. We expect to get that out next week

clairemcginty (approved):

PR #4615, opened 2022-11-29
PubsubIO: expose more options & correctly pass write params

Illedran:
I'm interested in exposing the [PubsubGrpcClient](https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/pubsub/PubsubGrpcClient.html) which is likely to be a more optimized transport than JSON (which is the [Beam default](https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.Read.html#withClientFactory-org.apache.beam.sdk.io.gcp.pubsub.PubsubClient.PubsubClientFactory-)).

While I was at it I also added params for other options that have been added over time and fixed the `PubsubIO.WriteParam`, which were not actually passed to Beam as the `with*` methods create a new object instead of mutating it.


scio-google-cloud-platform/src/main/scala/com/spotify/scio/pubsub/PubsubIO.scala
```diff
@@ -108,28 +111,40 @@ object PubsubIO {
...
-      .getOrElse(r0)
+  private[pubsub] def configureWrite[T](
+    write: beam.PubsubIO.Write[T]
+  )(
+    name: String,
+    params: WriteParam,
+    idAttribute: String,
+    timestampAttribute: String
+  ): beam.PubsubIO.Write[T] = {
+    val w0 = write.to(name)
```
clairemcginty, line 142: thanks for catching the bug with `.with`! Not sure how that one snuck in there.

Usually in our IOs we use a pattern of `var` + `fold` to apply successive `.withX...` statements, like this: https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L84-L89

clairemcginty (commented):
looks good! I have a minor comment about how the .with statements are applied.

On another thought, it looks like PubsubClientFactory is [instantiated](https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/pubsub/PubsubClient.PubsubClientFactory.html) with idAttribute, timestampAttribute, and PubsubOptions: the first two are available in `PubsubIO` and the third in `ScioContext`. Maybe we could support some kind of `default` factory that constructs `PubsubClientFactory` from those three available parmeters?

Illedran:
For the factories, Beam provides `FACTORY` objects as public singletons for each client implementation that can be passed here e.g.:
* Json factory: https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/pubsub/PubsubJsonClient.html#FACTORY
* Grpc factory: https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/gcp/pubsub/PubsubGrpcClient.html#FACTORY

I think it makes sense not to provide a default factory from Scio side so we can reuse [Beam's default factory](https://github.com/apache/beam/blob/0dd083f81e4c01c0b374ff52f967de26db8a188b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java#L114-L115) and stay in sync with that, but it's your call! :smiley:

scio-google-cloud-platform/src/main/scala/com/spotify/scio/pubsub/PubsubIO.scala
Illedran: Fixed!

clairemcginty (approved):
makes sense - thanks for your contribution :)

Issue #4591, opened 2022-11-08
Support service account impersonation in BigQuery client


clairemcginty:
Currently we support authentication to the BQ client (which is used for performing the BigQuery -> GCS export tasks that support non-storage BQ reads) using Application Default Credentials, as well as by a supplied secret file:

https://github.com/spotify/scio/blob/main/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/BigQuery.scala#L196-L213

Ideally we could also support [account impersonation](https://cloud.google.com/iam/docs/impersonating-service-accounts) via a new SBT option, something like `bigquery.actAs` that allows you to specify the name of a service account to impersonate?

Note: I haven't done much research on this, if it's possible with the existing options we should document it on the site.

shnapz:
Apparently it requires a code change:
```
ImpersonatedCredentials.create(sourceCredentials, "target_SA_principal")
```
and a new `apply` overload there. Will dig into details more

PR #4616, opened 2022-11-29
Added SA impersonation for BigQuery client

shnapz:
Implementing [#4591](https://github.com/spotify/scio/issues/4591)
In order to configure impersonation two new SBT options were added:

- bigquery.act_as (target SA principal)
- bigquery.act_as_lifetime (a duration in seconds of a target SA temporary credentials lifetime)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
```diff
@@ -70,6 +72,14 @@ object BigQueryConfig {
   def readTimeoutMs: Option[Int] =
     BigQuerySysProps.ReadTimeoutMs.valueOption.map(_.toInt)
 
+  def impersonationInfo: Option[ImpersonationInfo] = {
+    BigQuerySysProps.ActAs.valueOption.map { actAs =>
+      val lifetime =
+        BigQuerySysProps.ImpersonationLifetimeSec.valueOption.flatMap(x => Try(x.toInt).toOption)
+      ImpersonationInfo(actAs, lifetime.getOrElse(3600))
```
clairemcginty, line 79:
```suggestion
      ImpersonationInfo(actAs, lifetime.getOrElse(0))
```

I would set the default here to 0, that seems to be the Google library's hint to use its default value: https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L679

clairemcginty (approved):
one small suggestion but otherwise, looks great! 🎉

clairemcginty (approved):

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
```diff
@@ -70,6 +72,14 @@ object BigQueryConfig {
   def readTimeoutMs: Option[Int] =
     BigQuerySysProps.ReadTimeoutMs.valueOption.map(_.toInt)
 
+  def impersonationInfo: Option[ImpersonationInfo] = {
+    BigQuerySysProps.ActAs.valueOption.map { actAs =>
+      val lifetime =
+        BigQuerySysProps.ImpersonationLifetimeSec.valueOption.flatMap(x => Try(x.toInt).toOption)
```
RustedBones, line 78: do we want to silently default the `lifetime` if invalid ? can't we let it crash ?

RustedBones (commented):
Nice work. And as always good effort on the doc :+1:

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/BigQueryConfig.scala
shnapz: I think you are right

shnapz:
@clairemcginty @RustedBones made all config reads parsing fail fast

PR #4617, opened 2022-11-30
Fix SMB metadata compatibility issues

clairemcginty:
SMB BucketMetadata decoding was failing on an evolved metadata schema because we hadn't set `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` to false :( 

This is a tricky fix to deploy because basically, if a writer wants to add extra properties to their metadata.json file (i.e., if they want to use SMB secondary sort), all consumers will first have to upgrade to a version of Scio that includes this fix so that they won't break at read time. To ameliorate this somewhat, I set the BucketMetadata impls to _not_ write out `"keyClassSecondary":"null"` in the metadata.json file if the user hasn't set it. That way at least the default pathway won't be broken for consumers.


kellen (approved):
🙌

Issue #4598, opened 2022-11-14
Allow SMB ExternalSorter temp location to be customized with SBT opt


clairemcginty:
currently it uses a static location, `/tmp`: https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/sorter/ExternalSorter.java#L31

If users are running the job using a custom container setup, they might want to mount a specific volume w/ specific storage capacity to be used for SMB sort temp location. We should add a custom SBT opt for this (`smb.sorter.tmpdir`), and leave `/tmp` as the default location if it's left unspecified.

PR #4618, opened 2022-12-01
Exposed ExternalSorter tempLocation as System Property

shnapz:
Implementing [issue 4598](https://github.com/spotify/scio/issues/4598)

Question: do we also want to parametrize **sorter memory** option? It is more tricky:
- it is sourced in different places in the code
- there are different designations of memory option: for external sorter and in-memory sorted it means different things

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/sorter/ExternalSorter.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/sorter/ExternalSorter.java
@@ -93,7 +93,7 @@ public abstract class ExternalSorter implements Sorter {
   /** Returns a {@link Sorter} configured with the given {@link Options}. */
   public static ExternalSorter create(Options options) {
     checkArgument(
-        options.getSorterType() == Options.SorterType.NATIVE, "Only Native sorter is supported");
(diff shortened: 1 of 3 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/sorter/ExternalSorter.java
```diff
@@ -92,7 +92,8 @@ private void checkMemoryMB() {
 
   /** Returns a {@link Sorter} configured with the given {@link Options}. */
   public static ExternalSorter create(Options options) {
-    checkArgument(options.getSorterType() == Options.SorterType.NATIVE);
+    checkArgument(
+        options.getSorterType() == Options.SorterType.NATIVE, "Only Native sorter is supported");
```
clairemcginty, line 96:
```suggestion
        options.getSorterType() == Options.SorterType.NATIVE, "NativeExternalSorter is the only supported external sorter");
```

clairemcginty (approved):
looks good to me! I think we can hold off sorter memory for now -- we do have the existing sortMemoryMb [option](https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroSortedBucketIO.java#L304).

gliter (approved):

PR #4625, opened 2022-12-07
scalafix for the rest of the BQ API changes

farzad-sedghi:
In the [previous attempt](https://github.com/spotify/scio/pull/4504) the cases where schema was provided were ignored in fixing and was just marked by lint. It turns out those cases are pretty easy to fix too.


scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -11,28 +11,43 @@ class FixBqSaveAsTable extends SemanticRule("FixBqSaveAsTable") {
 
   override def fix(implicit doc: SemanticDocument): Patch = {
     doc.tree.collect {
-      case a @ Term.Apply(fun, head :: tail) =>
+      case a @ Term.Apply(fun, params) =>
         if (fun.symbol.normalized.toString.contains(methodName)) {
```
RustedBones, line 15: `methodName` and `scoll` should be defined as constants in companion object

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -11,28 +11,43 @@ class FixBqSaveAsTable extends SemanticRule("FixBqSaveAsTable") {
...
-                      s"table = Table.Ref(${head.toString.split("=").last.trim})"
-                    } else {
-                      s"Table.Ref($head)"
-                    }
-                    val allArgs = (headParam :: tail).mkString(", ")
-                    Patch.replaceTree(a, s"$qual.saveAsBigQueryTable($allArgs)")
-                  }
+                  val paramsUpdated =
+                    params
+                      .map(_.toString)
```
RustedBones, line 22: the goal of meta programming is to avoid working with strings.

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -11,28 +11,43 @@ class FixBqSaveAsTable extends SemanticRule("FixBqSaveAsTable") {
...
+                      .zipWithIndex
+                      .foldLeft(List[String]()) { case (accumulator, (param, index)) =>
+                        accumulator :+ (
+                          index match {
+                            // table is always the first param and without default value
+                            case 0 =>
+                              if (param.contains("=")) {
```
RustedBones, line 29: `Term.Asign` should be used instead of  looking for `=`

RustedBones (changes requested):

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
farzad-sedghi: not sure how important it is in this case, these files are standalone scripts

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -11,28 +11,50 @@ class FixBqSaveAsTable extends SemanticRule("FixBqSaveAsTable") {
...
-                    } else {
-                      s"Table.Ref($head)"
-                    }
-                    val allArgs = (headParam :: tail).mkString(", ")
-                    Patch.replaceTree(a, s"$qual.saveAsBigQueryTable($allArgs)")
-                  }
+                  val paramsUpdated =
+                    params
+                      .zipWithIndex
+                      .foldLeft(List[Object]()) { case (accumulator, (param, index)) =>
```
RustedBones, line 23: Why folding into a `List`. Can't we map ?

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
```diff
@@ -11,28 +11,50 @@ class FixBqSaveAsTable extends SemanticRule("FixBqSaveAsTable") {
...
+                                  } else {
+                                    param
+                                  }
+                                case _ =>
+                                  // if not a named param, `avroSchema` param should come second
+                                  if (index == 1) {
+                                    s"toTableSchema($param)"
+                                  } else {
+                                    param
+                                  }
```
RustedBones, line 49: Here you are mixing Tree and String
We should work with quasiquotes to return trees only.

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
farzad-sedghi: What is the problem of folding in a list?

scalafix/rules/src/main/scala/fix/FixBqSaveAsTable.scala
farzad-sedghi: can you explain what can go wrong doing so?

shnapz (approved):
Not sure about all details, but change in FixBqSaveAsTable.scala looks consistent

scalafix/build.sbt
```diff
@@ -84,7 +92,7 @@ lazy val `input-0_12` = project
 
 lazy val `output-0_12` = project
   .settings(
-    libraryDependencies ++= scio10(Scio.`0.12`)
+    libraryDependencies ++= scio12(Scio.`0.12`)
```
clairemcginty, line 95:
```suggestion
    libraryDependencies ++= scio12
```

scalafix/build.sbt
```diff
@@ -47,6 +47,14 @@ def scio11(version: String) =
     "com.spotify" %% "scio-avro"
   ).map(_ % version)
 
+def scio12(version: String) =
+  List(
+    "com.spotify" %% "scio-core",
+    "com.spotify" %% "scio-avro",
+    "com.spotify" %% "scio-google-cloud-platform",
+    "com.spotify" %% "scio-extra"
+  ).map(_ % version)
```
clairemcginty, line 56: 🎨 imo we shouldn't parameterize version when the method is already called `scio12`:

```suggestion
lazy val scio12 =
  List(
    "com.spotify" %% "scio-core",
    "com.spotify" %% "scio-avro",
    "com.spotify" %% "scio-google-cloud-platform",
    "com.spotify" %% "scio-extra"
  ).map(_ % Scio.`0.12`)
```

clairemcginty (approved):

PR #4644, opened 2023-01-03
Support EmptyMatchTreatment on TextIO

turb:
This supports Beam `emptyMatchTreatment` param in `TextIO` end `sc.textFile`.


turb:
I see the change affect binary compatibility, however I am unsure what is the right way to fix it. I thought that the use of default parameters implies several function overloads in the generated bytecode.

RustedBones:
This is expected:
Modifying a method with an additional parameter with default value is source compatible but not binary compatible.
The bytecode will not contain the method signature without the parameter. This will fail if called from a lib compiled with previous scio version.
As scio is still on a v0, we can be flexible with [version policy](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#early-semver-and-sbt-version-policy) but I'd rather plan that for 0.13.

@spotify/flatmap WDYT ?

turb:
> @spotify/flatmap WDYT ?

Not sure this notification works @RustedBones

RustedBones:
@clairemcginty reduced breaking changes on the class. i ignored the following:
- breaking API due to new default param -> we've allowed that before
- `copy` signature change but never used
- `MissingTypesProblem` because it is a 1 param case class anymore, losing unused `AbstractFunction1` extention

shnapz:
Need to target this to v0.13.x branch

RustedBones:
Hi @turb. Beam 2.47 has just been released. We'll prepare the 0.13 to be out ASAP now!

PR #4659, opened 2023-01-19
Add Java 17 to GHA setup

### LINKED ISSUE #4597: Java 17 support
Java 17 has been supported in Dataflow as of Beam [2.37.0](https://beam.apache.org/blog/beam-2.37.0/). Currently several of Scio's unit tests fail when run with java 17, mostly Kryo related.

clairemcginty:
(fix #4597) testing GHA setup w/ Java 17.

See linked ticket for more info on why the JVM opts were added to both .jvmopts and to build.sbt: basically, the build.sbt options are ignored for tests where `fork := false`:

```sh
[warn] javaOptions will be ignored, fork is set to false
```

Instead, opts are picked up from .jvmopts. Conversely, for forked tests, the JVM options don't seem to get passed from .jvmopts to the forked process 🤷‍♀️ so it seems like we need both to support both forked and non-forked tests.

Another option is just setting fork := true for all tests


scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -211,10 +211,9 @@ final class CoderTest extends AnyFlatSpec with Matchers {
       cw.runWithImplicit
       throw new Throwable("Is expected to throw when passing implicit from outer class")
     } catch {
-      case e: NullPointerException =>
-        // In this case outer field is called "$cw" and it is hard to wrap it with proper exception
-        // so we allow it to fail with NullPointerException
-        e.getMessage should be(null)
+      case _: NullPointerException =>
```
clairemcginty, line 214: Interestingly in Java 17, this now throws an NPE with the message `Cannot throw exception because "null" is null`. cc @shnapz @RustedBones

shnapz: @clairemcginty will take a look

clairemcginty: sounds good!

.github/workflows/release.yml
```diff
@@ -12,11 +12,11 @@ jobs:
           fetch-depth: 0
       - name: cache SBT
         uses: coursier/cache-action@v6
-      - name: Java 11 setup
+      - name: Java 17 setup
```
clairemcginty, line 15: We can also save this for 0.13 milestone

regadas (approved):
LGTM @clairemcginty

PR #4664, opened 2023-01-20
Improve coder tests

shnapz:


scio-core/src/main/scala/com/spotify/scio/coders/LowPriorityCoderDerivation.scala
```diff
@@ -93,9 +93,9 @@ object LowPriorityCoderDerivation {
     def id(v: T): Int = subTypes.collectFirst { case (isType, index) if isType(v) => index }.get
   }
 
-  def colsureFunction[E, D, R](enclosed: E)(gen: E => D => R): D => R = gen(enclosed)
+  def closureFunction[E, D, R](enclosed: E)(gen: E => D => R): D => R = gen(enclosed)
```
shnapz, line 96: Fixing a typo

scio-core/src/main/scala/com/spotify/scio/coders/CustomCoder.scala
```diff
@@ -0,0 +1,285 @@
+package com.spotify.scio.coders
```
shnapz, line 1: Just extracted this from `Coder.scala`

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -683,12 +315,13 @@ private[coders] object CoderStackTrace {
...
   )
 
-  def prepare: Array[StackTraceElement] =
+  def prepare: Array[StackTraceElement] = {
     CoderStackElemMarker +: Thread
       .currentThread()
       .getStackTrace
       .dropWhile(!_.getClassName.contains(CoderMaterializer.getClass.getName))
-      .take(10)
+      .take(15)
```
shnapz, line 323: Otherwise unit test fails, because now it has larger callstack of test code

scio-core/src/main/scala/com/spotify/scio/coders/WrappedCoder.scala
```diff
@@ -0,0 +1,99 @@
+package com.spotify.scio.coders
```
shnapz, line 1: Extracted from `Coder.scala`

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +31,145 @@ import scala.reflect.ClassTag
...
+  private def checkRoundtripWithCoder[T](
+    beamCoder: BCoder[T],
+    actualValue: T,
+    expectedBytes: Array[Byte] = null
+  )(implicit
     eq: Equality[T]
   ): Assertion = {
-    val bytes = CoderUtils.encodeToByteArray(beamCoder, value)
+    val bytes = CoderUtils.encodeToByteArray(beamCoder, actualValue)
+    if (expectedBytes != null) {
```
shnapz, line 169: added an ability to assert exact bytes serialized

scio-test/src/test/scala/com/spotify/scio/coders/AvroCoderPipelineTest.scala
```diff
@@ -0,0 +1,81 @@
...
+import com.spotify.scio.avro._
+import com.spotify.scio.avro.StringFieldTest
+import com.spotify.scio.testing.PipelineSpec
+import com.spotify.scio.values.SCollection
+import org.scalatest.BeforeAndAfterAll
+
+import java.nio.file.Files
+import scala.jdk.CollectionConverters._
+
+class AvroCoderPipelineTest extends PipelineSpec with BeforeAndAfterAll {
```
shnapz, line 31: All test cases with avro extracted to the separate spec

scio-test/src/test/scala/com/spotify/scio/coders/AvroCoderTest.scala
```diff
@@ -1,81 +1,47 @@
...
-import java.nio.file.Files
+import org.scalatest.flatspec.AnyFlatSpec
+import com.spotify.scio.testing.CoderAssertions._
+import org.apache.avro.generic.GenericRecord
+import org.scalactic.Equality
+import org.scalatest.matchers.should.Matchers
 import scala.jdk.CollectionConverters._
 
-class AvroCoderTest extends PipelineSpec with BeforeAndAfterAll {
+final class AvroCoderTest extends AnyFlatSpec with Matchers {
```
shnapz, line 10: All test cases with avro extracted from CodersTest to this spec

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -314,7 +200,11 @@ final class CoderTest extends AnyFlatSpec with Matchers {
     t22.take(22) coderShould roundtrip()
   }
 
-  it should "have a Coder for Nothing" in {
+  it should "have a Coder for Nothing, Unit, Void" in {
+    "Coder[Void]" should compile
```
shnapz, line 204: moved them here from another test case

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -398,22 +262,16 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
-  // FIXME: implement the missing coders
+  // FIXME: TableRowJsonCoder cannot be tested in scio-test because of circular dependency on scio-google-cloud-platform
   it should "support all the already supported types" in {
     import java.math.{BigInteger, BigDecimal => jBigDecimal}
     import java.nio.file.FileSystems
 
     import org.apache.beam.sdk.transforms.windowing.IntervalWindow
 
-    // TableRowJsonCoder
-    // SpecificRecordBase
```
shnapz, line 409: SpecificRecordBase and Message are verified in other test cases

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -466,29 +324,22 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
-  it should "provide a fallback if no safe coder is available" in {
-    val record: GenericRecord = Avro.user
-    record coderShould fallback()
+    CaseClassWithExplicitCoder(1, "hello") coderShould notFallback() and haveCoderInstance(
+      CaseClassWithExplicitCoder.caseClassWithExplicitCoderCoder
+    )
   }
 
   it should "support classes with private constructors" in {
-    Coder.gen[PrivateClass]
```
shnapz, line 486: has 0 effect on the next statement

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -585,28 +430,28 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
-    null.asInstanceOf[jDouble] coderShould roundtrip(opts)
-    null.asInstanceOf[jLong] coderShould roundtrip(opts)
-    null.asInstanceOf[jShort] coderShould roundtrip(opts)
-    null.asInstanceOf[(String, Top)] coderShould roundtrip(opts)
-    (null, null).asInstanceOf[(String, Top)] coderShould roundtrip(opts)
-    null.asInstanceOf[DummyCC] coderShould roundtrip(opts)
-    DummyCC(null) coderShould roundtrip(opts)
-    null.asInstanceOf[Top] coderShould roundtrip(opts)
-    null.asInstanceOf[Either[String, Int]] coderShould roundtrip(opts)
+    null.asInstanceOf[String] coderShouldWithOpts opts and roundtrip()
```
shnapz, line 433: probably not the best syntax :(

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -43,128 +41,22 @@ import java.time.{Instant, LocalDate}
 import java.util.UUID
 import java.time.format.DateTimeFormatter
 
-// record
-final case class UserId(bytes: Seq[Byte])
```
shnapz, line 47: moved those classes after the test spec, to easier navigate to each test case

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -690,8 +534,7 @@ final class CoderTest extends AnyFlatSpec with Matchers {
 
   it should "#2595: work with parameterized types" in {
     case class Example(stringT: Either[Array[Byte], String], longT: Either[Array[Byte], Long])
-    val sc = com.spotify.scio.ScioContext.forTest()
```
shnapz, line 693: this `.forTest()` has no effect on coder

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +31,128 @@ import scala.reflect.ClassTag
...
 
-  def notFallback[T: ClassTag](opts: PipelineOptions = DefaultPipelineOptions): CoderAssertion[T] =
-    new CoderAssertion[T] {
-      override def assert(value: T)(implicit c: Coder[T], eq: Equality[T]): Assertion = {
-        c should !==(Coder.kryo[T])
-        val beamCoder = CoderMaterializer.beamWithDefault(c, o = opts)
-        checkRoundtripWithCoder[T](beamCoder, value)
-      }
-    }
+  trait AssertionContextBase {
```
shnapz, line 78: need this trait without type parameter to support case like:
```
def beOfType[ExpectedCoder: ClassTag](): CoderAssertion
```
to avoid passing two type parameters

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -43,161 +50,63 @@ import java.time.{Instant, LocalDate}
...
     CoderMaterializer.beam(PipelineOptionsFactory.create(), coder)
 
-  "Coders" should "support primitives" in {
-    1 coderShould roundtrip()
-    'a' coderShould roundtrip()
-    "yolo" coderShould roundtrip()
-    4.5 coderShould roundtrip()
+  it should "support primitives" in {
+    false coderShould roundtripToBytes(Array(0)) and beOfType[Beam[_]] and
+      materializeTo[BooleanCoder]
```
shnapz, line 63: I think it makes sense to assert types of underlying coders to make sure implicits + pattern matching in `materialize` worked as expected

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +31,148 @@ import scala.reflect.ClassTag
...
+  type CoderAssertionT[T] = AssertionContext[T] => Assertion
+  type CoderAssertion = AssertionContextBase => Assertion
+
+  implicit class ValueShouldSyntax[T](value: T) {
     def coderShould(
-      coderAssertion: CoderAssertion[T]
-    )(implicit c: Coder[T], eq: Equality[T]): Assertion =
-      coderAssertion.assert(value)
-  }
+      coderAssertion: CoderAssertionT[T]
```
shnapz, line 39: we can accept both `CoderAssertion` and `CoderAssertionT[T]` due to contra-variance

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -43,161 +58,100 @@ import java.time.{Instant, LocalDate}
...
+      beOfType[CoderTransform[_, _]] and
+      materializeTo("SeqCoder")
 
-  def run(): Unit =
-    InnerCaseClass("51") coderShould roundtrip()
-}
+    s coderShould roundtrip() and
+      beOfType[CoderTransform[_, _]] and
+      materializeTo("SeqCoder") and
+      beFullyCompliant()
```
shnapz, line 119: an easy way to run coder against all checks

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
RustedBones: That's damn good syntax!

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -43,161 +58,100 @@ import java.time.{Instant, LocalDate}
...
+      beOfType[Transform[_, _]] and
+      materializeToTransformOf[ByteCoder]
 
-// private
-class PrivateClass private (val value: Long) extends AnyVal
-object PrivateClass {
-  def apply(l: Long): PrivateClass = new PrivateClass(l)
-}
-case class UsesPrivateClass(privateClass: PrivateClass)
+    "yolo" coderShould roundtripToBytes(Array(121, 111, 108, 111)) and
```
RustedBones, line 86: :laughing:

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -43,161 +58,100 @@ import java.time.{Instant, LocalDate}
...
-@SerialVersionUID(1)
-sealed trait TraitWithAnnotation
-@SerialVersionUID(2)
-final case class FirstImplementationWithAnnotation(s: String) extends TraitWithAnnotation
-@SerialVersionUID(3)
-final case class SecondImplementationWithAnnotation(i: Int) extends TraitWithAnnotation
-
-// AnyVal
-final case class AnyValExample(value: String) extends AnyVal
+    4.5.asInstanceOf[Float] coderShould roundtripToBytes(Array(64, -112, 0, 0)) and
```
RustedBones, line 97:
```suggestion
    4.5f coderShould roundtripToBytes(Array(64, -112, 0, 0)) and
```

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +32,185 @@ import scala.reflect.ClassTag
...
+      ctx.beamCoder.asInstanceOf[MaterializedCoder[_]].bcoder shouldBe a[ExpectedBeamCoder]
     }
 
-  def fallback[T: ClassTag](opts: PipelineOptions = DefaultPipelineOptions): CoderAssertion[T] =
-    new CoderAssertion[T] {
-      override def assert(value: T)(implicit c: Coder[T], eq: Equality[T]): Assertion = {
-        c should ===(Coder.kryo[T])
-        roundtripKryo(opts).assert(value)
-      }
+  def materializeTo(className: String): CoderAssertion =
```
RustedBones, line 133: Why is this needed ? is there some shortcoming with using the type parameter  `def materializeTo[ExpectedBeamCoder: ClassTag]` ?

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -43,161 +58,100 @@ import java.time.{Instant, LocalDate}
...
-case class UsesPrivateClass(privateClass: PrivateClass)
+    "yolo" coderShould roundtripToBytes(Array(121, 111, 108, 111)) and
+      beOfType[Beam[_]] and
+      materializeTo[StringUtf8Coder]
 
-// proto
-case class ClassWithProtoEnum(s: String, `enum`: OuterClassForProto.EnumExample)
+    4.5 coderShould roundtripToBytes(Array(64, 18, 0, 0, 0, 0, 0, 0)) and
+      beOfType[Beam[_]] and
+      materializeTo("SDoubleCoder$") and
```
RustedBones, line 92: not a fan of the string param here. We can probably work out smth

RustedBones (approved):
Great refactoring.
Nit, find a way to avoid using `materializeTo`  with string class name param

shnapz:
> Great refactoring. Nit, find a way to avoid using `materializeTo` with string class name param

@RustedBones 
Yeah :( I hesitated to open those private classes/objects. But I can lift their visibility to a package level - so tests can access them

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -19,22 +19,37 @@ package com.spotify.scio.coders
...
+  ByteCoder,
+  Coder => BCoder,
+  CoderException,
+  InstantCoder,
+  ListCoder,
+  NullableCoder,
+  RowCoder,
+  StringUtf8Coder,
+  VarIntCoder
+}
```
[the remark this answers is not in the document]
shnapz: Indeed :)

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
shnapz: @RustedBones so now I have changed `private` to `private[coders]` for multiple classes to avoid this

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +32,185 @@ import scala.reflect.ClassTag
...
+  }
+
+  def roundtrip[T: Equality](): CoderAssertionT[T] = ctx =>
+    checkRoundtripWithCoder[T](ctx.beamCoder, ctx.actualValue.get)
+
+  def roundtripToBytes[T: Equality](expectedBytes: Array[Byte]): CoderAssertionT[T] = ctx =>
+    checkRoundtripWithCoder[T](ctx.beamCoder, ctx.actualValue.get, expectedBytes)
+
+  def haveCoderInstance(expectedCoder: Coder[_]): CoderAssertion = ctx =>
+    ctx.coder should ===(expectedCoder)
```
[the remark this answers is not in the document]
shnapz: `===` also allows customizing equality and is the most strict:
```
result should equal (3) // can customize equality
result should === (3)   // can customize equality and enforce type constraints
result should be (3)    // cannot customize equality, so fastest to compile
result shouldEqual 3    // can customize equality, no parentheses required
result shouldBe 3       // cannot customize equality, so fastest to compile, no parentheses required
```
probably each case should be applied according to a use case. So `===` is good for a compile-time check, I would aim to use it everywhere but it slows compilation (not significantly?)

scio-core/src/main/scala/com/spotify/scio/coders/CustomCoder.scala
```diff
@@ -0,0 +1,301 @@
+/*
+ * Copyright 2019 Spotify AB.
```
clairemcginty, line 2: update date on license header :)

scio-core/src/main/scala/com/spotify/scio/coders/WrappedCoder.scala
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright 2019 Spotify AB.
```
clairemcginty, line 2: same

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -420,6 +425,7 @@ private object SDoubleCoder extends BCoder[Double] {
     } else {
       java.lang.Double.valueOf(value)
     }
+  override def consistentWithEquals() = bc.consistentWithEquals()
```
clairemcginty, line 428: nice catch with this and `FloatCoder`!

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +32,175 @@ import scala.reflect.ClassTag
 object CoderAssertions {
   private lazy val DefaultPipelineOptions = PipelineOptionsFactory.create()
 
-  implicit class CoderAssertionsImplicits[T](private val value: T) extends AnyVal {
+  type CoderAssertionT[T] = AssertionContext[T] => Assertion
+  type CoderAssertion = AssertionContextBase => Assertion
```
clairemcginty, line 36: I'm wondering if this might be overloading the name `CoderAssertion` too much... maybe we could rename this something like `GenericCoderAssertion`? Although that might be confusing too if people hear "Generic" and think we're talking about Avro 😓 wdyt?

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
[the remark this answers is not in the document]
clairemcginty: that's a good point about slowing down compilation. Maybe we shouldn't adopt it globally but should audit the rest of our unit tests and substitute in `===` where appropriate?

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
```diff
@@ -30,102 +32,175 @@ import scala.reflect.ClassTag
...
+      structuralValueConsistentWithEquals()(ctx)
+      beSerializable()(ctx)
+      beConsistentWithEquals()(ctx)
+      bytesCountTested[T]().apply(ctx)
+      beNonDeterministic()(ctx)
+    }
 
-    result should ===(value)
-  }
+  def beFullyCompliantNotConsistentWithEquals[T <: Object: ClassTag](): CoderAssertionT[T] =
```
clairemcginty, line 185: do we need `T <: Object` ?

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -585,28 +588,28 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
-    null.asInstanceOf[jDouble] coderShould roundtrip(opts)
-    null.asInstanceOf[jLong] coderShould roundtrip(opts)
-    null.asInstanceOf[jShort] coderShould roundtrip(opts)
-    null.asInstanceOf[(String, Top)] coderShould roundtrip(opts)
-    (null, null).asInstanceOf[(String, Top)] coderShould roundtrip(opts)
-    null.asInstanceOf[DummyCC] coderShould roundtrip(opts)
-    DummyCC(null) coderShould roundtrip(opts)
-    null.asInstanceOf[Top] coderShould roundtrip(opts)
-    null.asInstanceOf[Either[String, Int]] coderShould roundtrip(opts)
+    null.asInstanceOf[String] coderShouldWithOpts opts and notFallback()
```
clairemcginty, line 591: just from a semantics perspective, I feel like having to include the `and` makes things a bit confusing--because `coderShouldWithOpts opts` on its own is not making any assertions. Possible to rework to enable syntax like `coderShouldWithOpts(opts) notFallback()`?

clairemcginty (commented):
this is amazing and so thorough 🎉 just a few minor comments.

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
RustedBones: Yes, package visibility for coders sounds appropriate

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
shnapz: Hmm.. What about `CoderAssertionBase`, like I did with `AssertionContextBase`?

scio-test/src/main/scala/com/spotify/scio/testing/CoderAssertions.scala
shnapz: Unfortunately yes, it is needed by a nested check `def bytesCountTested[T <: Object: ClassTag]()` that requires it for:
```
val arr = Array(ctx.actualValue.get)
```
where `.get` returns `T`

scio-core/src/main/scala/com/spotify/scio/coders/CustomCoder.scala
shnapz: fixed

scio-core/src/main/scala/com/spotify/scio/coders/WrappedCoder.scala
shnapz: fixed

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
shnapz: changed to:
```
null.asInstanceOf[String] coder WithOptions(opts) should notFallback()
```

scio-test/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -585,28 +588,29 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
-    null.asInstanceOf[jDouble] coderShould roundtrip(opts)
-    null.asInstanceOf[jLong] coderShould roundtrip(opts)
-    null.asInstanceOf[jShort] coderShould roundtrip(opts)
-    null.asInstanceOf[(String, Top)] coderShould roundtrip(opts)
-    (null, null).asInstanceOf[(String, Top)] coderShould roundtrip(opts)
-    null.asInstanceOf[DummyCC] coderShould roundtrip(opts)
-    DummyCC(null) coderShould roundtrip(opts)
-    null.asInstanceOf[Top] coderShould roundtrip(opts)
-    null.asInstanceOf[Either[String, Int]] coderShould roundtrip(opts)
+    null.asInstanceOf[String] coder WithOptions(opts) should notFallback()
```
shnapz, line 591: @clairemcginty does it look better?

clairemcginty: yes, looks better I think!

PR #4667, opened 2023-01-24
Fix SMB producer compatibility issues

clairemcginty:
Unfortunately the concept of "version" in scio-smb was not well designed to start with so we find ourselves running into issues where consumers on older versions will break on producer updates. This is not an ideal solution but we can go forward with the mandate that all changes to the SMB metadata format *must* be backwards compatible... at least pending a thorough reworking of versioning semantics.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
@@ -75,7 +75,12 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.hash.Hashing;
 @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "type")
 public abstract class BucketMetadata<K1, K2, V> implements Serializable, HasDisplayData {
 
-  @JsonIgnore public static final int CURRENT_VERSION = 1;
+  /**
+   * The VERSION parameter is included in the metadata.json file for every SMB. Take extreme care in
+   * bumping version: Scio versions prior to 0.12.1 perform a version compatibility check on SMB
+   * partitions which may fail if the SMB producer bumps to an incompatible version.
+   */
+  @JsonIgnore public static final int CURRENT_VERSION = 0;
 
   // Represents the current major version of the Beam SMB module. Storage format may differ
   // across versions and require internal code branching to ensure backwards compatibility.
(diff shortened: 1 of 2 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -206,8 +211,6 @@ public HashFunction create() {
 
   boolean isCompatibleWith(BucketMetadata other) {
     return other != null
-        // version 1 is backwards compatible with version 0
-        && (this.version <= 1 && other.version <= 1)
```
RustedBones, line 210: Shouldn't this be rolled back to original 0.11 code instead of being deleted ?
```scala
&& this.version == other.version
```

clairemcginty: My thought process was that if a producer has already migrated to 0.12.1 and produced a few days of data, then upgrades to 0.12.2, they’ll have a mix of version: 1 and version: 0 in their metadata. If we make the version check strict again the consumers will upgrade to 0.12.2 and break on version: 1 data :(

maybe i can keep the check the way it was (0 and 1 are compatible) and just make a note that the next version bump should be to 2.

RustedBones: yes we should re-introduce some checks otherwise the release will accept any version

clairemcginty: Sounds good - updated

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -75,7 +75,14 @@
 @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "type")
 public abstract class BucketMetadata<K1, K2, V> implements Serializable, HasDisplayData {
 
-  @JsonIgnore public static final int CURRENT_VERSION = 1;
+  /**
+   * The VERSION parameter is included in the metadata.json file for every SMB. Take extreme care in
+   * bumping version: Scio versions prior to 0.12.1 perform a version compatibility check on SMB
+   * partitions which may fail if the SMB producer bumps to an incompatible version.
+   *
+   * <p>The next version bump should be to: 2
```
clairemcginty, line 83: the `<p>` tag was added by javafmt 🤷‍♀️

RustedBones (approved):

PR #4690, opened 2023-02-02
Unwrap internal SCollection coders to be reused in transforms

RustedBones:
When `NullableCoders` is tuned on, users could get the following error
```log
java.lang.IllegalArgumentException: Failed to extract key-value coders from Coder[(K, V)]: 
  NullableCoder(NullableCode(...
```

This happens when `keyBy` is used. It only requires the key coder as implicit. The value coder is propagated for the materialized coder set on the internal collection.
When using pre-materialized coders, We should make sure those are completely unwrapped.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
@@ -20,6 +20,7 @@ package com.spotify.scio.coders
 import com.spotify.scio.coders.{instances => scio}
 import com.spotify.scio.values.{SCollection, SideInput}
 import org.apache.beam.sdk.{coders => beam}
+import org.apache.beam.sdk.values.PCollection
 
 import scala.annotation.tailrec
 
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PCollectionWrapper.scala
```diff
@@ -27,7 +27,7 @@ private[values] trait PCollectionWrapper[T] extends TransformNameable {
   /** The [[org.apache.beam.sdk.values.PCollection PCollection]] being wrapped internally. */
   val internal: PCollection[T]
 
-  implicit def coder: Coder[T] = Coder.beam(internal.getCoder)
```
RustedBones, line 30: Here we were propagating wrapped coders

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
```diff
@@ -33,38 +34,49 @@ private[scio] object BeamCoders {
...
   /** Get key-value coders from an `SCollection[(K, V)]`. */
   def getTupleCoders[K, V](coll: SCollection[(K, V)]): (Coder[K], Coder[V]) = {
     val coder = coll.internal.getCoder
     val (k, v) = unwrap(coder) match {
-      case c: scio.Tuple2Coder[K, V] => (c.ac, c.bc)
+      case c: scio.Tuple2Coder[K, V] =>
+        (c.ac, c.bc)
       case c: RecordCoder[(K, V)] =>
-        (
-          c.cs.find(_._1 == "_1").get._2.asInstanceOf[beam.Coder[K]],
```
RustedBones, line 43: Prefer to rely on type position that on the field name

clairemcginty (approved):

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
```diff
@@ -33,38 +34,49 @@ private[scio] object BeamCoders {
...
+        val ac = coderElement[K](c)(0)
+        val bc = coderElement[V](c)(1)
+        (ac, bc)
       case _ =>
         throw new IllegalArgumentException(
           s"Failed to extract key-value coders from Coder[(K, V)]: $coder"
         )
     }
-    (Coder.beam(k), Coder.beam(v))
+    (Coder.beam(unwrap(k)), Coder.beam(unwrap(v)))
```
kellen, line 62: Is calling `unwrap` required everywhere `Coder.beam` is called?

kellen (approved):

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
RustedBones: No. Most of the usage of `Coder.beam` are not with unwrap like [here](https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala).
We have to unwrap when moving back from materilalized coder back to scio coders. (reuse of internal coders in partial transformation with `keyBy`, `mapValues`, ...)

PR #4691, opened 2023-02-02
Fix SingletonCoder serialization

RustedBones:
If `singleton` in `SingletonCoder` was initialized, the coder could not be deserialized.

```log
java.lang.IllegalArgumentException: unable to deserialize Custom DoFn With Execution Info
...
Cause: java.io.InvalidClassException: com.example.MyObject$; no valid constructor
```

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -140,7 +140,7 @@ final private class SingletonCoder[T](
   val typeName: String,
   supply: () => T
 ) extends CustomCoder[T] {
-  private lazy val singleton = supply()
+  @transient private lazy val singleton = supply()
 
   override def toString: String = s"SingletonCoder[$typeName]"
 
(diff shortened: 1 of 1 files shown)


shnapz:
Interesting, good catch on LazyCoder! 
But what is the use case? I thought Singleton is used only for scala Object

clairemcginty (approved):

RustedBones:
@shnapz The problem comes when users have object in their model.
This can happen with ADTs, eg:
```scala
sealed trait Color
case object Red extends Color
case object Green extends Color
case object Blue extends Color
```
The disjunction coder for `Color` will delegate to singleton coders.
The previous code was working in most cases as case objects are serializable. However, it does not when:
- the object is not a `case object`
- the case object extends a class with parameter as explained [here](https://github.com/lloydmeta/enumeratum/issues/140)

The trick here is to only serialize the supplier in the coder.

In practice, `Coder` are serialized before being used. So the lazy val won't be initialized, (serialized as null).
But in some test cases, beam serializes the coder after usage, hence the observed failure.

PR #4701, opened 2023-02-09
Avoid using outputfield cache

shnapz:
Removed call to ClosureCleaner that potentially can cause race condition


clairemcginty (approved):

RustedBones:
Nice. Glad we can keep the error message

Issue #4692, opened 2023-02-02
TupleTag collides in sortMergeTransform of multiple sorted bucket IO


shnapz:
1. Improve documentation to explain that TupleTag should have unique name in the context of SMB transform: https://spotify.github.io/scio/extras/Sort-Merge-Bucket.html
2. On Scala API layer validate that TupleTags are unique, like on `sc.sortMergeTransform` etc.

clairemcginty:
fixed in #4711

PR #4709, opened 2023-02-13
Tuple tag validation

farzad-sedghi:
https://github.com/spotify/scio/issues/4692

### DIFF
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
@@ -21,6 +21,7 @@ package com.spotify.scio.smb.util
 
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.Coder
+import com.spotify.scio.smb.validateTupleTags
 import com.spotify.scio.values._
 import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, TargetParallelism}
 import com.spotify.scio.smb.syntax.SortedBucketScioContext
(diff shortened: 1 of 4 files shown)


clairemcginty (commented):
I think it would be simpler to do this in the SortedBucketIO.java class (both in the [`CoGbk` constructor + the `and` method](https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java#L138-L160). That way the changes are constrained to a single file, work for both SortedBucketSource + SortedBucketTransform, and will still be applied even if the user chooses to forego using our Scala SMB APIs. wdyt?

farzad-sedghi:
> I think it would be simpler to do this in the SortedBucketIO.java class (both in the [`CoGbk` constructor + the `and` method](https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java#L138-L160). That way the changes are constrained to a single file, work for both SortedBucketSource + SortedBucketTransform, and will still be applied even if the user chooses to forego using our Scala SMB APIs. wdyt?

totally forgot this is our own IO layer :)

scio-smb/src/main/scala/com/spotify/scio/smb/util/SMBMultiJoin.scala
```diff
@@ -32,20 +33,21 @@ final class SMBMultiJoin(private val self: ScioContext) {
...
   def sortMergeCoGroup[KEY: Coder, A: Coder, B: Coder, C: Coder, D: Coder, E: Coder](
-    keyClass: Class[KEY],
-    a: SortedBucketIO.Read[A],
-    b: SortedBucketIO.Read[B],
-    c: SortedBucketIO.Read[C],
-    d: SortedBucketIO.Read[D],
-    e: SortedBucketIO.Read[E],
-    targetParallelism: TargetParallelism
-  ): SCollection[(KEY, (Iterable[A], Iterable[B], Iterable[C], Iterable[D], Iterable[E]))] = {
+                                                                                      keyClass: Class[KEY],
```
shnapz, line 36: hmmmm.... why is there so much whitespace? is everything ok with styling?

farzad-sedghi:
Closed because none of the changes in this PR were needed

PR #4710, opened 2023-02-13
Set default JdkAddOpenModules for Java 17

clairemcginty:
Required for Kryo coders to function.

If users sets the `JdkAddOpenModules` option in their job specifically, it will completely override these default options from Scio.

### DIFF
--- a/.github/workflows/pre-release-check.yml
+++ b/.github/workflows/pre-release-check.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 11
+          java-version: 17
       - name: set JVM opts
         run: scripts/gha_setup.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-release-check.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 11
+          java-version: 17
```
shnapz, line 23: don't we want to run it in both 11 and 17?

clairemcginty: Ideally yes, but not sure it's worth the extra Dataflow cost... imo since our release config still targets Java 8 I think we're pretty well-covered as far as writing compatible code. wdyt?

regadas (approved):

PR #4711, opened 2023-02-14
Unique tuple tags

farzad-sedghi:
https://github.com/spotify/scio/issues/4692


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
```diff
@@ -76,6 +77,23 @@ public static <K1, K2> CoGbkWithSecondaryBuilder<K1, K2> read(
     return new CoGbkWithSecondaryBuilder<>(primaryKeyClass, secondaryKeyClass);
   }
 
+  private static void validateInputNameUniqueness(List<BucketedInput<?>> inputs)
+      throws IllegalArgumentException {
+    HashSet<String> inputNames = new HashSet<>();
+    inputs.stream()
+        .forEach(i -> {
+              if (!inputNames.add(i.getTupleTag().getId())) {
```
clairemcginty, line 85: I think I would just change this to:

```
assert(inputNames.add(i.getTupleTag().getId())
```

And not add `IllegalArgumentException` to these signatures

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
```diff
@@ -110,12 +128,14 @@ private CoGbkWithSecondaryBuilder(Class<K1> primaryKeyClass, Class<K2> secondary
      * Returns a new {@link CoGbkWithSecondary} with the given first sorted-bucket source in {@link
      * Read}.
      */
-    public CoGbkWithSecondary<K1, K2> of(Read<?> read) {
```
clairemcginty, line 113: Looks like this method got deleted?

clairemcginty (commented):
Check should also be added here? https://github.com/spotify/scio/blob/aa81b5450490b81a99f1f327b237924e5fb58f45/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java#L107

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
farzad-sedghi: My IDE made a lot of white space changes, I made a bunch of mistakes fixing them back :)

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
farzad-sedghi: oh ok I can do that. Can you please say why that is preferred?

clairemcginty: Sure! It's just a small enough edge case that I don't think it warrants changing a user-facing API signature for. Additionally it's the convention we're using elsewhere in scio-smb (check out at the constructors for all BucketMetadata implementations, which all use `assert`s)

farzad-sedghi: hmm ok I see the signature change is a good reason. But as far as I know the main difference between assert and throwing an exception is the ability to turn assert of in prod (it is a development feature). Isn't that why we use it in other places?

farzad-sedghi: in general it is easy to know where assert should be used and where it should not. e.g. when the input is set by user, you should NOT use assertion. In this case "the user" is the developer which chooses the name of the tuple tags? But that is at development time from their POV, while they are using our prod version of lib. confusing!

farzad-sedghi: also this is a runtime exception so no need to have it in the signature

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -69,7 +69,7 @@ final class SortedBucketScioContext(@transient private val self: ScioContext) ex
     rhs: SortedBucketIO.Read[R],
     targetParallelism: TargetParallelism = TargetParallelism.auto()
   ): SCollection[(K, (L, R))] = {
-    val t = SortedBucketIO.read(keyClass).of(lhs).and(rhs).withTargetParallelism(targetParallelism)
+    val t = SortedBucketIO.read(keyClass).of(lhs, rhs).withTargetParallelism(targetParallelism)
```
shnapz, line 72: Why did you move from approach with `and`? To easier catch all TupleTags in one method?

farzad-sedghi: that was the convention used in the generated smb multi join file. Also it causes a bit less object overhead, not that it matters much

clairemcginty (approved):
nice!

PR #4713, opened 2023-02-16
Force scio coders on all input SCollection

RustedBones:
Some scio IOs were not seting scio coders on the underlying PCollection and relying on the default coder provided by beam. This can cause issues:
- coder options are not respected (Nullable)
- union may fail. Default beam coders are not 'wrapped'. The beam runtime check on coder equality fails.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
@@ -19,9 +19,10 @@ package com.spotify.scio.coders.instances
 
 import java.io.{InputStream, OutputStream}
 import com.spotify.scio.coders.{AvroCoderMacros, Coder}
+import com.spotify.scio.util.ScioUtil
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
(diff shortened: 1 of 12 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
```diff
@@ -124,8 +125,15 @@ trait AvroCoders {
...
-  implicit def genAvro[T <: SpecificRecordBase]: Coder[T] =
+  implicit def avroSpecificRecordCoder[T <: SpecificRecord: ClassTag]: Coder[T] = {
+    val clazz = ScioUtil.classOf[T]
+    val schema = SpecificData.get().getSchema(clazz)
+    val useReflectApi = true // keep this for backward compatibility
+    Coder.beam(AvroCoder.of(clazz, schema, useReflectApi))
+  }
+
+  @deprecated("Use avroSpecificRecordCoder instead", since = "0.12.5")
+  def genAvro[T <: SpecificRecordBase]: Coder[T] =
```
RustedBones, line 136: Drop this as done in #4590. This coder can't be called on a generic type with bounds.

clairemcginty (approved):

PR #4718, opened 2023-02-23
Add skewed job methods with topN and top percent

RustedBones:
Current version of skewedJoins rely on a proper threshold tuning.

In some cases, when the lhs is quite dynamic, it is hard to get control of the hot keys (and the underlying hash join).

by introducing top percentage and top N, we can create a bound on the hot keys and be sure the hot join will fit in memory


RustedBones:
MiMa fails with
```log
[error] scio-core: Failed binary compatibility check against com.spotify:scio-core_2.12:0.12.5! Found 3 potential problems
[error]  * synthetic method skewedJoin$default$2()Long in class com.spotify.scio.values.PairSkewedSCollectionFunctions has a different result type in current version, where it is com.spotify.scio.values.HotKeyMethod rather than Long
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("com.spotify.scio.values.PairSkewedSCollectionFunctions.skewedJoin$default$2")
[error]  * synthetic method skewedLeftOuterJoin$default$2()Long in class com.spotify.scio.values.PairSkewedSCollectionFunctions has a different result type in current version, where it is com.spotify.scio.values.HotKeyMethod rather than Long
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("com.spotify.scio.values.PairSkewedSCollectionFunctions.skewedLeftOuterJoin$default$2")
[error]  * synthetic method skewedFullOuterJoin$default$2()Long in class com.spotify.scio.values.PairSkewedSCollectionFunctions has a different result type in current version, where it is com.spotify.scio.values.HotKeyMethod rather than Long
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("com.spotify.scio.values.PairSkewedSCollectionFunctions.skewedFullOuterJoin$default$2")
```
I've moved the default params to the new overloaded method.

clairemcginty (approved):

scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
```diff
@@ -37,329 +71,425 @@ final private case class Partitions[K, V](hot: SCollection[(K, V)], chill: SColl
...
+      case HotKeyMethod.TopN(value) =>
+        val cms = CMSOperations.aggregate(lhsKeys, TopNCMS.aggregator(eps, delta, seed, value))
+        lhs.skewedJoin(rhs, cms)
     }
   }
 
+  @deprecated("Use skewedJoin with HotKeyMethod.Threshold instead ", "0.13.0")
+  def skewedJoin[W](
+    rhs: SCollection[(K, W)],
+    hotKeyThreshold: Long,
```
clairemcginty, line 134: should we preserve the default value here to make this a non-breaking change?

PR #4719, opened 2023-02-23
Neo4j parametrized from existing SCollection

sumitsu:
Expands the Scio Neo4j API to expose the ability to execute parameterized Cypher queries in parallel based upon inputs from an existing `SCollection`, as supported by the [Beam `Neo4jIO` implementation](https://beam.apache.org/releases/javadoc/current/org/apache/beam/sdk/io/neo4j/Neo4jIO.html).

In the original Scio-Neo4j PR (spotify/scio#4488), it appears that there was some [discussion](https://github.com/spotify/scio/pull/4488/#issuecomment-1211753698) regarding a prospective general approach to sharding / parallel-reading. The implementation in this PR makes no such attempt to be transparent to the user, nor to calculate shard sizes automatically, but merely to expose the functionality already present in Beam. The client is responsible for:

1. determining how to map / group an existing `SCollection` into reasonably-scoped querying shards
2. incorporating `WHERE` clause parameters which limit query scope into the Cypher `String`
3. providing the function which maps input `SCollection` elements to values for those parameters


scio-neo4j/src/main/scala/com/spotify/scio/neo4j/Neo4jIO.scala
```diff
@@ -45,21 +43,14 @@ object Neo4jIO {
   private[neo4j] def neo4jIoId(opts: Neo4jConnectionOptions, cypher: String): String =
     s"${opts.username}:${opts.password}@${opts.url}:$cypher"
 
-  private[neo4j] def dataSourceConfiguration(
```
RustedBones, line 48: I don't think you had to move things from there. It could be imported.

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/ops/Neo4jSCollectionReadOps.scala
```diff
@@ -0,0 +1,149 @@
...
+import scala.jdk.CollectionConverters._
+
+/**
+ * Operations for transforming an existing [[SCollection]] via a Neo4j query (read)
+ * @param self
+ *   existing [[SCollection]] (prior to Neo4j read)
+ * @tparam X
+ *   type parameter for existing [[SCollection]]
+ */
+class Neo4jSCollectionReadOps[X](self: SCollection[X]) {
```
RustedBones, line 38: This is not different from the `Neo4jSCollectionWriteOps`. we should define all extension methods in the them in the same class

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/ops/Neo4jSCollectionReadOps.scala
```diff
@@ -0,0 +1,149 @@
...
+   * @tparam Y
+   *   result [[SCollection]] type
+   * @return
+   *   [[SCollection]] containing the union of query results from a parameterized query invocation
+   *   for each original [[SCollection]] element
+   */
+  def neo4jCypherWithParams[Y](
+    neo4jConf: Neo4jOptions,
+    cypher: String,
+    elementToParamFunction: X => Map[String, AnyRef]
```
RustedBones, line 77: It would look nicer if the function is curried.
In the write API, we actually user the `ValueType` to convert `T` to the params.
I think we can also have such API

RustedBones:
Thanks a lot for the PR. I took the liberty of updating your branch directly to address some comments.

scio-neo4j/src/it/scala/com/spotify/scio/neo4j/Neo4jIOIT.scala
```diff
@@ -63,51 +67,77 @@ class Neo4jIOIT extends PipelineSpec with Eventually with ForAllTestContainer {
...
-    val options = PipelineOptionsFactory.create()
-    options.setRunner(classOf[DirectRunner])
+  val martin = Person("Martin Sheen")
+  val morgan = Person("Morgan Freeman")
+  val michael = Person("Michael Douglas")
+
+  val americanPresident = Movie("American President", 1995)
+
+  val options = PipelineOptionsFactory.create()
+  options.setRunner(classOf[DirectRunner])
```
clairemcginty, line 79: DirectRunner is the default so this shouldn't be needed as far as I know :)

scio-neo4j/src/main/scala/com/spotify/scio/neo4j/syntax/SCollectionSyntax.scala
```diff
@@ -17,15 +17,84 @@
...
+    coder: Coder[U]
+  ): SCollection[U] =
+    self.applyTransform(
+      beam.Neo4jIO
+        .readAll[T, U]()
+        .withDriverConfiguration(dataSourceConfiguration(neo4jConf.connectionOptions))
+        .withSessionConfig(neo4jConf.sessionConfig)
+        .withTransactionConfig(neo4jConf.transactionConfig)
+        .withCypher(cypher)
+        .withParametersFunction(neo4jInType.to(_).asMap())
```
sumitsu, line 72: Ah nice! Yes, agree that this looks much better. 🙂

Just to confirm that my understanding is correct: it would be incumbent upon the user to write the Cypher query String such that the parameter names match the field names used by the (implicit) `ValueType[T]` implementation? (Of course, in most cases, it would probably be easiest to use a case class (with the existing `CaseMapper`-powered `ValueType`) and match the query to the case class field names.)

RustedBones: Yes, that's correct.
The write API works the same way: if the cipher refers to a field which is not in the model (after the case mapper transform), there will be a runtime exception.

clairemcginty (approved):

Issue #3332, opened 2020-09-21
Methods from saveAsBigQuery category don't expose crucial Beam parameters


mkuthan:
I would like to discuss API of methods from saveAsBigQuery category. The methods provide very high level of abstraction for underlaying Beam API. In general this is expected behaviour to keep Scio API much more friendly than original Beam API but sometimes some more sophisticated setup is really needed, e.g:

- to force FILE_LOADS write method for unbounded collection
- to define triggeringFrequency and numFileShards for FILE_LOADS
- to define failedInsertRetryPolicy for STREAMING_INSERTS
- to enable or disable Avro logical types
- perhaps much more other use-cases for every BigQueryIO.writer configuration knob

For sure, there is a nice Scio API extension point "saveAsCustomOutput" but it would be useful to have more specific saveAsBigQuery methods with access to underlaying Beam BigQueryIO. Not easy task, because the BigQueryIO API is let't say far from concise design - you can build writer in 100 different ways but in 95 you will get error during pipeline startup (e.g this parameter is also required if that parameter is used).

I prepared some minimal abstraction for "saveAsCustomOutput" and it has simplified my pipelines a lot. Please loot at the code as a good starting point for further discussion.
https://github.com/mkuthan/beam-examples/blob/master/src/main/scala/org/mkuthan/beam/examples/bigquery/SaveSpecificRecord.scala

Thanks for Scio, it saves a lot of my time!

nevillelyh:
Yeah this is one area the 2 styles don't mix well, i.e. default arguments vs the ever growing Beam builder setters. :shrug:

I don't have a better idea for this. One thing we could possibly do is adding a `f: BigQueryIO => BigQueryIO =>` for every read/write sugar, so that users can set additional params via the lambda, but it'll also make it easy to shoot oneself in the foot. Also there might be IOs with specialized types like `{Generic,Specific}RecordAvroIO`, not sure how well they fit in here.

Thoughts?

mkuthan:
Thanks for quick response!
Any advantages of exposing `f: BigQueryIO => BigQueryIO =>` over `saveAsCustomOutput`? Testability?
Specialized IOs are useful, for different types different formatters / converters  needs to be provided. I also found that insert methods (file_load vs. streaming_insert) requires very different set of options and it might be explicitly mapped to the API.

anish749:
[...]

PR #4722, opened 2023-02-24
Issue 3332

farzad-sedghi:
FIX https://github.com/spotify/scio/issues/3332

This way (advanced) users can provide a lambdas (`configOverride` below) to modify the underlying PTransforms, where they are using BigQueryIO syntax methods. Same approach can possibly be used for all other IOs.
This map function is applied to the ptransform **as the last step before applying the ptransform to the data**.

```
    ...
    implicit def fun: Write[MyTypes.Output] => Write[MyTypes.Output] =
        w => w.withTableDescription("Table description set by overriding PTransform in the job")
     ...
    sc
      .parallelize(List((1,11), (2, 22)))
      .map(kv => MyTypes.Output(kv._1.intValue, kv._2.intValue()))
      .saveAsTypedBigQueryTable(
        table = Table.Spec("scio-playground:farzaddev.test6"),
        writeDisposition = WRITE_TRUNCATE,
        createDisposition = CREATE_IF_NEEDED,
        configOverride = _.withTableDescription("Table description set by overriding PTransform in the job")
      )
    sc.run()
    ()
  }

```


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -51,14 +51,17 @@ final class SCollectionTableRowOps[T <: TableRow](private val self: SCollection[
     createDisposition: CreateDisposition = BigQueryTypedTable.WriteParam.DefaultCreateDisposition,
     tableDescription: String = BigQueryTypedTable.WriteParam.DefaultTableDescription,
     timePartitioning: TimePartitioning = BigQueryTypedTable.WriteParam.DefaultTimePartitioning
+  )(
+    implicit writePTransformOverride: Write[TableRow] => Write[TableRow] = identity
```
RustedBones, line 55: I don't think this should be an implicit parameter

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -215,41 +217,45 @@ object BigQueryTypedTable {
...
-  trait WriteParam {
+  trait WriteParam[T] {
     val schema: TableSchema
     val writeDisposition: WriteDisposition
     val createDisposition: CreateDisposition
     val tableDescription: String
     val timePartitioning: TimePartitioning
     val extendedErrorInfo: ExtendedErrorInfo
     val insertErrorTransform: SCollection[extendedErrorInfo.Info] => Unit
+    val writePTransformOverride: Write[T] => Write[T]
```
RustedBones, line 228: I's suggest `configure` as name

shnapz:
@farzad-sedghi we need to merge it to [v0.13.x](https://github.com/spotify/scio/tree/v0.13.x)

shnapz:
@farzad-sedghi ooops, now we need to merge latest master changes to 0.13.x

RustedBones:
@farzad-sedghi I think this can be closed in favor of #4810

PR #4724, opened 2023-02-27
Distribute CMS hot key set as side input

RustedBones:
Distribute the hot key set from the TopCMS instead of the CMS itself to avoid the extra map operation when partitioning on the hot keys


clairemcginty (approved):
nice!

PR #4727, opened 2023-02-27
Don't re-use Configuration objects in Parquet SMB

clairemcginty:
Minor bugfix for Parquet-SMB that creates new Configuration objects for each Parquet read/write rather than re-using a single default object. We made a similar fix in the scio-parquet module for the 0.12.x release because we encountered a few cases of users doing multiple Parquet reads who found that the configuration they applied to one read was also getting applied to the other.

Thankfully, I don't think this bug has had any actual impact, since `ParquetAvroSortedBucketIO` and `ParquetTypeSortedBucketIO` always set a fresh Configuration object in the builder methods, and users don't typically interact with the *FileOperations classes directly :)

### DIFF
--- a/scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeFileOperations.scala
+++ b/scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeFileOperations.scala
@@ -34,6 +34,7 @@ import java.nio.channels.{ReadableByteChannel, WritableByteChannel}
 
 object ParquetTypeFileOperations {
   val DefaultCompression = CompressionCodecName.GZIP
+  val DefaultConfiguration = null
 
   def apply[T: Coder: ParquetType](): ParquetTypeFileOperations[T] = apply(DefaultCompression)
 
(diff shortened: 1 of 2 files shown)


scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeFileOperations.scala
```diff
@@ -34,14 +34,13 @@ import java.nio.channels.{ReadableByteChannel, WritableByteChannel}
 
 object ParquetTypeFileOperations {
   val DefaultCompression = CompressionCodecName.GZIP
-  val DefaultConfiguration = new Configuration()
```
clairemcginty, line 37: this removal makes the PR fail MiMa. It seems that making these variables public is inconsistent with the behavior in the Java class, where it's package-private; I also don't think it has any usage since, as I mentioned, users don't typically interact with this class directly. So I think it's safe to merge--but I could also make MiMa happy by just setting `val DefaultConfiguration = null` and doing `Option(DefaultConfiguration).getOrElse(new Configuration())` in the calling method.

RustedBones (approved):

scio-smb/src/main/scala/org/apache/beam/sdk/extensions/smb/ParquetTypeFileOperations.scala
clairemcginty: @RustedBones any concern about merging this into 0.12.x when MiMa is failing? the impact of this bug should be zero so it could probably wait for 0.13.x release

kellen: The `def` variant will only be called once anyway so wouldn't resolve the issue if I'm remembering the details correctly.

IMO the `null` variant makes the most sense in terms of being consistent with all the other IOs

kellen (approved):

PR #4734, opened 2023-03-03
Expose cache in scio-grpc

RustedBones:
Allow user to provide a `CacheSupplier` in scio-grpc API to cache responses


clairemcginty (approved):

RustedBones:
@clairemcginty looks that using overload was still breaking binary-compatibility for scala 2.12.
I moved to version with default param and add the mima filter. Is that still fine for master of should we mark for 0.13 ?

clairemcginty:
sorry for the late reply! Even if it's not technically binary compatible, I think this is fine to go in 0.12.6 because of the default value added.

PR #4736, opened 2023-03-06
WIP Filename retaining readFiles

kellen:
Uses some reflection so as not to reinvent (unnecessarily locked-down) beam internals


scio-core/src/main/scala/org/apache/beam/sdk/io/Foobar.scala
```diff
@@ -0,0 +1,125 @@
...
+      val reader: beam.BoundedSource.BoundedReader[T] = use(
+        source
+          .createForSubrangeOfFile(file.getMetadata, range.getFrom, range.getTo)
+          .createReader(c.getPipelineOptions)
+      )
+
+      try {
+        var more = reader.start();
+        while (more) {
+          c.output((filename, reader.getCurrent))
```
kellen, line 114: This is all the same as `ReadFileRangesFn` except scalafied and this line where we also output the filename

scio-extra/src/main/scala/com/spotify/scio/extra/csv/syntax/SCollectionSyntax.scala
```diff
@@ -61,5 +61,11 @@ trait SCollectionSyntax {
       self
         .withName("Read CSV")
         .applyTransform(ParDo.of(CsvIO.ReadDoFn[T](csvConfiguration)))
+
+    def readCsvWithFilename[T: HeaderDecoder: Coder](csvConfiguration: CsvConfiguration)
```
kellen, line 65: CSVs are not splittable due to quoting and multiline values, but also they _may_ have a header in which case reading the whole file is required anyway

scio-core/src/main/scala/org/apache/beam/sdk/io/Foobar.scala
```diff
@@ -0,0 +1,125 @@
...
+  private def yolo(className: String, args: Class[_]*): Constructor[_] = {
+    val clazz = Class.forName(className)
+    val ctor: Constructor[_] = clazz.getDeclaredConstructor(args:_*)
+    ctor.setAccessible(true)
+    ctor
+  }
+
+  def textSource(input: String): FileBasedSource[String] = {
+    yolo(
+      "org.apache.beam.sdk.io.TextSource",
```
kellen, line 32: Insane that this is private

scio-core/src/main/scala/org/apache/beam/sdk/io/Foobar.scala
```diff
@@ -0,0 +1,125 @@
...
+      ValueProvider.StaticValueProvider.of(input),
+      EmptyMatchTreatment.DISALLOW,
+      Array[Byte]('\n')
+    ).asInstanceOf[FileBasedSource[String]]
+  }
+
+  type SplitIntoRangesT = DoFn[beam.FileIO.ReadableFile, KV[beam.FileIO.ReadableFile, OffsetRange]]
+  def splitIntoRangesFn(desiredBundleSizeBytes: Long): SplitIntoRangesT = {
+    yolo(
+      "org.apache.beam.sdk.io.ReadAllViaFileBasedSource$SplitIntoRangesFn",
```
kellen, line 44: Don't reinvent the wheel

scio-core/src/main/scala/org/apache/beam/sdk/io/Foobar.scala
```diff
@@ -0,0 +1,125 @@
...
+import java.lang.reflect.Constructor
+import scala.reflect.ClassTag
+import scala.util.Using.Manager
+
+
+object AccessibleBeam {
+  private def yolo(className: String, args: Class[_]*): Constructor[_] = {
+    val clazz = Class.forName(className)
+    val ctor: Constructor[_] = clazz.getDeclaredConstructor(args:_*)
+    ctor.setAccessible(true)
```
shnapz, line 26: does it work in Java 17?
cc: @clairemcginty

scio-core/src/main/scala/com/spotify/scio/transforms/FilenameRetainingDoFns.scala
```diff
@@ -0,0 +1,165 @@
...
+ * filename in addition to the data contents.
+ */
+class FilenameRetainingReadFileRangesFn[T](
+  createSource: SerializableFunction[String, _ <: beam.FileBasedSource[T]],
+  exceptionHandler: beam.ReadAllViaFileBasedSource.ReadFileRangesFnExceptionHandler
+) extends DoFn[KV[beam.FileIO.ReadableFile, OffsetRange], (String, T)] {
+  type DoFnT = DoFn[KV[beam.FileIO.ReadableFile, OffsetRange], (String, T)]
+
+  @ProcessElement
+  def process(c: DoFnT#ProcessContext): Unit = {
```
shnapz, line 133: There is a newer style of implementing process func without ProcessContext, but `@Element`, `OutputReceiver` etc. See https://github.com/spotify/scio/commit/214981e204d3c289f8d5ec751bafa3ad531875e2

kellen:
Closing in favor of merged beam PR

PR #4737, opened 2023-03-08
Remove test matrix for integration test

RustedBones:
The java matrix was just a waste since it was always on java 11

### DIFF
--- a/.github/workflows/it-tests.yml
+++ b/.github/workflows/it-tests.yml
@@ -18,26 +18,14 @@ jobs:
           credentials_json: ${{ secrets.GCP_CREDENTIALS }}
       - name: cache SBT
         uses: coursier/cache-action@v6
-      - name: Java 11 setup
+      - name: Java 17 setup
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 11
+          java-version: 17
       - name: set JVM opts
         run: scripts/gha_setup.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/it-tests.yml
```diff
@@ -18,26 +18,14 @@ jobs:
           credentials_json: ${{ secrets.GCP_CREDENTIALS }}
       - name: cache SBT
         uses: coursier/cache-action@v6
-      - name: Java 11 setup
+      - name: Java 17 setup
         uses: actions/setup-java@v3
         with:
           distribution: temurin
-          java-version: 11
```
RustedBones, line 25: Should have been
```yaml
java-version: ${{matrix.java}}
```

shnapz: @RustedBones why didn't you apply this change, but set only Java 17?

clairemcginty (approved):

.github/workflows/it-tests.yml
RustedBones: There are come concurrency issue on the BigTable integration test. And IMHO we should be fine by cross testing on unit test.

shnapz: it appears we never had cross-java testing. Maybe cross-scala in unit test, cross-java in IT?

RustedBones: We do have cross-java in unit-testing
image

PR #4738, opened 2023-03-09
Add implicit not found messages for Funnel[T]

kellen:


kellen:
Closing in favor of #4739 as this currently produces a NPE when formatting the message. Appears to be a compiler bug.

Issue #4459, opened 2022-07-14
Planed deprecation for v0.13.0


RustedBones:
Cleanup planned deprecation for `v0.13.0` milestone

- [x] elasticsearch 6 module (end of life)
- [x] widen API from `SpecificRecordBase` class to `SpecificRecord` interface
- [x] remove `genAvro` coder
- [x] remove context from `SCollectionWithFanout` as done [here](https://github.com/spotify/scio/pull/4740/commits/3afb6095ba7728dcc7ee77acd4aa65b3672cccd2)

shnapz:
we can also set `val DefaultSampleWithReplacement: Boolean = false` for skewedJoin

jjengo:
Preparing for a 0.13 Scio release. Still need to plan on when we'll cut the release for OSS.

PR #4740, opened 2023-03-10
Don't forget to fan out

kellen:
Fixes `aggregate` transforms in `withFanout` and `withHotKeyFanout` where the wrapped SCollection was inadvertently converted back to a normal SCollection, thereby dropping the fanout.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
@@ -67,6 +67,7 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
   )(f: KV[K, UI] => (K, UO)): SCollection[(K, UO)] = {
     self.transform(
       _.withName("TupleToKv").toKV
+        .withName(t.getName) // propagate transform name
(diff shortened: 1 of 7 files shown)


clairemcginty (approved):
🤦‍♀️

RustedBones:
That's a good one @kellen 
Here is the mima breakage
```log
[error]  * method this(com.spotify.scio.ScioContext,com.spotify.scio.values.PairSCollectionFunctions,scala.util.Either)Unit in class com.spotify.scio.values.SCollectionWithHotKeyFanout does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("com.spotify.scio.values.SCollectionWithHotKeyFanout.this")
```
Can we avoid the breaking change now and put an item in https://github.com/spotify/scio/issues/4459 to cleanup on next release ?

We can now also use the fan-out in the `sparseIntersectByKey`: I faced some performance issue last time computing the CMS: it is reduced on a single node and created a 10mn delay in some jobs. I tried using fan-out without success due to this bug.

kellen:
@RustedBones Yeah, was reconsidering the breakage over the weekend, will revert that change and open against the 0.13 branch.

shnapz (approved):

PR #4741, opened 2023-03-11
Config override

farzad-sedghi:
replacing https://github.com/spotify/scio/pull/4722


RustedBones:
Added a potential simpler way of testing the override without having to use mocks: https://github.com/farzad-sedghi/scio/pull/1

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -324,12 +334,12 @@ object BigQueryTypedTable {
 
 final case class BigQueryTypedTable[T: Coder](
   reader: beam.BigQueryIO.TypedRead[T],
-  writer: beam.BigQueryIO.Write[T],
+  writer: Write[T],
```
shnapz, line 337: Does it remain of the same type?

PR #4752, opened 2023-03-16
Add fanout for CMS hot key computation in skewed joins

RustedBones:
Add fanout when computing the CMS for the hot keys in skewed joins.

- on deprecated API, use a fanout of 5 (we can't expose without breaking compatibility)
- expose fanout parameter in new API

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
@@ -56,6 +56,72 @@ object HotKeyMethod {
 
 }
 
+object SkewedJoins {
+
+  // some sensible defaults for skewed joins
+  val DefaultHotKeyThreshold = 9000
+  val DefaultHotKeyFanout = 5
+  val DefaultCmsEpsilon = 0.001
+  val DefaultCmsDelta = 1e-10
+  val DefaultCmsSeed = 42
+  val DefaultSampleFraction = 1.0
+  val DefaultSampleWithReplacement = true
+
+  private[scio] def union[T](hot: SCollection[T], chill: SCollection[T]): SCollection[T] =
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
```diff
@@ -56,6 +56,73 @@ object HotKeyMethod {
 
 }
 
+object SkewedJoins {
+
+  // some sensible defaults for skewed joins
+  val DefaultHotKeyThreshold: Int = 9000
+  val DefaultHotKeyMethod: HotKeyMethod.Threshold = HotKeyMethod.Threshold(DefaultHotKeyThreshold)
+  val DefaultHotKeyFanout: Int = 5
```
clairemcginty, line 64: how was 5 chosen? Any change it will degrade performance if not overridden?

RustedBones:
I ran some experiments on a CMS aggregation (last step)
- no fanout
2023-03-17 12 31 10 console cloud google com 41dee663d3b5

- fanout 5
2023-03-17 12 31 00 console cloud google com c8c8011fc3da

- fanout 25
2023-03-17 12 30 48 console cloud google com 5f4e44058886

- fanout 64
2023-03-17 12 30 22 console cloud google com c8141279d4ff

We can see that without fanout, the single node is a bottleneck on the job execution, however, increasing the fanout did not speedup the job as much  as I expected.

I think this setting will be very dependent on the key collection, I've decided to set back default to 1

scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
RustedBones: Set to no fanout. See comment

clairemcginty (approved):

PR #4759, opened 2023-03-28
Add support for DML statements to MockBigQuery

### LINKED ISSUE #4757: Enable MockBigQuery to support DML statements
When using `MockBigQuery` to test DML statements, like so

```
 // Put this in BigQueryIT.scala to run
  it should "support DML on mock data" in {
    val writableTableRef = "data-integration-test.samples_eu.shakespeare" 
    val mbq = MockBigQuery()
    mbq.mockTable(tableRef).withData(MockBQData.inData)
    val insertQuery = s"INSERT INTO `${writableTableRef}` (word, word_count) VALUES ('cat', 2)"
    mbq.queryResult(insertQuery) should contain theSameElementsAs MockBQData.expected ++ Seq(MockBQData.wordCount("cat", 2))
    val updateQuery = s"UPDATE `${writableTableRef}` SET word_count = 4 WHERE word = 'cat'"
    mbq.queryResult(updateQuery) should contain theSameElementsAs MockBQData.expected ++ Seq(MockBQData.wordCount("cat", 4))
    val deleteQuery = s"DELETE FROM `${writableTableRef}` WHERE word = 'cat'"
    mbq.queryResult(deleteQuery) should contain theSameElementsAs MockBQData.expected
  }
```

The code will fail with this error:

```
java.lang.RuntimeException: com.spotify.scio.bigquery.client.QueryJob id:<id redacted> failed with error: {"message":"Cannot set create disposition in jobs with DML statements","reason":"invalidQuery"}
```

fallonchen:
Fixes #4757

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
@@ -123,6 +123,15 @@ class MockBigQuery private (private val bq: BigQuery) {
     queryResult(sqlQuery, flattenResults).map(bqt.fromTableRow)
   }
 
+  /** Run live DML statement against BigQuery service, substituting mocked tables with test data. */
+  def runDML(dmlStatement: String): TableReference = {
+    val isLegacy = bq.query.isLegacySql(dmlStatement)
+    val mockDml = mapping.foldLeft(dmlStatement) { case (q, (src, dst)) =>
+      q.replace(toTableSpec(src, isLegacy), toTableSpec(dst, isLegacy))
+    }
+    bq.query.run(mockDml, writeDisposition = null, createDisposition = null)
+  }
+
   private def toTableSpec(table: TableReference, isLegacy: Boolean) =
     if (isLegacy) {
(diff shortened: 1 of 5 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/MockBigQuery.scala
```diff
@@ -123,6 +123,15 @@ class MockBigQuery private (private val bq: BigQuery) {
     queryResult(sqlQuery, flattenResults).map(bqt.fromTableRow)
   }
 
+  /** Run live DML statement against BigQuery service, substituting mocked tables with test data. */
+  def runDML(dmlStatement: String): TableReference = {
```
RustedBones, line 127: I see we have a `isDML` in `QueryOps`. Maybe this is worth moving it to the `BigQuery` object with `scio` or `bigquery` package visibility  so we can make sure the input here is a DML with
```scala
require(isDML(dmlStatement), s"Expecting DML statement but got '$dmlStatement'")
```

fallonchen: That makes sense to me! I've added this change.

RustedBones (approved):

RustedBones:
Thank you!

PR #4764, opened 2023-03-31
Fix avro coder regression on custom SpecificCoder

### LINKED ISSUE #4763: Avro SpecificRecord codec regression in 0.12.5
After changes (#4713, #4590), Avro SpecificRecord coder doesn't work for avrohugger generated case classes.

Built-in Avro coder gets schema from the static `$Schema` field of generated specific record class.  Such field doesn't exist in case classes generated by avrohugger.

I would ask for reverting changes and use macro for getting Avro schema from specific records. I think that avrohugger is a common way to play with Avro specific records in Scala. 

The macro below works like a charm for case classes generated by avrohugger.
```
_root_.com.spotify.scio.coders.Coder.beam(
        classOf[$companioned],
        new $companioned().getSchema,
        true
      )
    )
``` 

See also: #1702, #1754,  #1755, #1758

I also found that similar method for Avro Codec has been implemented in Flink, see: https://github.com/apache/flink/pull/9565/files

RustedBones:
Fix #4763 

Adding `avrohugger` generated test class with sbt plugin `1.1.0`

### DIFF
--- a/scio-test/src/test/scala/com/spotify/scio/avro/AvroHugger.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/avro/AvroHugger.scala
@@ -16,6 +16,7 @@
 
 package com.spotify.scio.avro
 
+// scalafix:off
 /** MACHINE-GENERATED FROM AVRO SCHEMA. DO NOT EDIT DIRECTLY */
 
 import scala.annotation.switch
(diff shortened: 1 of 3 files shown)


scio-test/src/test/scala/com/spotify/scio/avro/AvroHugger.scala
```diff
@@ -0,0 +1,54 @@
...
+      case _ => new org.apache.avro.AvroRuntimeException("Bad index")
+    }
+    ()
+  }
+
+  def getSchema: org.apache.avro.Schema = AvroHugger.SCHEMA$
+}
+
+object AvroHugger {
+  val SCHEMA$ = new org.apache.avro.Schema.Parser().parse(
```
RustedBones, line 51: this filed is on the `AvroHugger$`, not on `AvroHugger` hence the error

clairemcginty (approved):

PR #4768, opened 2023-04-05
SparkeyUri coder should not fallback

clairemcginty:
`SparkeyUri` extends `Serializable` and its fields are also always Serializable - we can leverage that to avoid falling back to Kryo. The motivation here is to reduce the amount of Kryo usage to make the Java 17 transition smoother for users :)

Also adds `equals`/`hashCode` impls to `RemoteFileUtil`, which is already marked `Serializable` and is intended for use in DoFns -- basically, we can treat it as a no-arg singleton, despite the side-effecting behavior of [RemoteFileUtil.create()](https://github.com/spotify/scio/blob/d750f67cc1a5af87a7c693e0336db7cdb6bfeb8e/scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java#L73). It probably shouldn't have been designed that way to start out with; that method is only intended to be called on from the launcher process, before submitting the job graph to the runner. So, even though we don't attempt to serialize `PipelineOptions` and call `FileSystems.setDefaultPipelineOptions` every time `RemoteFileUtil` is decoded, it doesn't matter--we don't _need_ to invoke that method at any point other than on the initial call to `RemoteFileUtils.create()`. (This is the same behavior that Kryo was already doing.)

MiMa is failing because I've moved the two implicit coders, but this shouldn't affect any users.


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/CoderInstances.scala
```diff
@@ -0,0 +1,27 @@
...
+ * under the License.
+ */
+
+import com.spotify.scio.coders.Coder
+import com.spotify.scio.extra.sparkey.SparkeyUri
+import org.apache.beam.sdk.{coders => bcoders}
+
+trait CoderInstances {
+  implicit def sparkeyUriCoder[T <: SparkeyUri]: Coder[T] =
+    Coder.beam(bcoders.SerializableCoder.of(classOf[SparkeyUri])).asInstanceOf[Coder[T]]
```
clairemcginty, line 26: If `ShardedSparkeyUri` and `SparkeyUri` were in the same file, we could just make them `sealed traits` and let Scio derive coders automatically 👀

clairemcginty:
On closer look, RemoteFileUtil might not be properly Serializable since it's [side-effecting in the constructor](https://github.com/spotify/scio/blob/main/scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java#L73)... investigating more

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/CoderInstances.scala
clairemcginty: It should be ok; added more context to PR description.

shnapz: I wonder what concrete type of Beam coder is resolved here?

clairemcginty:
Looking into Beam internals, SerializableCoder doesn't seem to be recommended for [traits](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/SerializableCoder.java#L105) 🤔  Plus, `SparkeyReader` is not serializable. Maybe we can table this for the Sco 0.12 branch and work on refactoring sparkey Coder usage for 0.13.

clairemcginty:
> Ideally it should be handled using magnolia-based coders as other scala case classes

Agreed @shnapz! :) The current design of SparkeyUri makes it hard to do properly -- I think i'll need to refactor things a bit

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/ShardedSparkeyUri.scala
```diff
@@ -39,14 +38,12 @@ trait ShardedSparkeyUri extends SparkeyUri {
...
 
-  implicit def coderSparkeyURI: Coder[ShardedSparkeyUri] = Coder.kryo[ShardedSparkeyUri]
-
   def basePathForShard(shardIndex: Short, numShards: Short): String =
     f"$basePath/part-$shardIndex%05d-of-$numShards%05d"
 
   def sparkeyUriForShard(shardIndex: Short, numShards: Short): SparkeyUri
 
-  val globExpression: String = s"$basePath/part-*"
+  lazy val globExpression: String = s"$basePath/part-*"
```
RustedBones, line 46: You must add `transient` to avoid serialization (I think that was the goal, right ?)

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/CoderInstances.scala
RustedBones: `SparkeyUri` but we want a `T`

scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
```diff
@@ -98,6 +98,8 @@ trait JavaCoders extends JavaBeanCoders {
   implicit val jDoubleCoder: Coder[java.lang.Double] = fromScalaCoder(Coder.doubleCoder)
 
   implicit val jBooleanCoder: Coder[java.lang.Boolean] = Coder.beam(BooleanCoder.of())
+  implicit val jRemoteFileUtil: Coder[RemoteFileUtil] =
+    Coder.beam(SerializableCoder.of(classOf[RemoteFileUtil]))
```
RustedBones, line 102: I've very suspicious about this.
IMHO `RemoteFileUtil` should not have a coder. It implements serializable to be placed in the graph as part of DoFn closure (And I really think we should not even create an instance and make it pure static).
`RemoteFileUtil` should not be part of any data, thus should not require a coder

RustedBones (changes requested):

scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
clairemcginty: yeah -- I was looking for a quick win on Coders without having to change any signatures so MiMA would be happy, but I think I'll need to do some refactoring on RFU/SparkeyUri so we don't have to hack around them like this

RustedBones: Can you point where the `RemoteFileUtil` might be part of the data ?
Even without refactoring, Looks to me we can live without a `Coder[RemoteFileUtil]`, right ?

clairemcginty: a few places --

- [FileDownloadFn](https://github.com/spotify/scio/blob/main/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java#L41) as instance variable
- [DistCache](https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/values/DistCache.scala#L45) (I need to double-check on this one how DistCache gets serialized; maybe RFU is not part of data)
- [RemoteSparkeyUri](https://github.com/spotify/scio/blob/main/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala#L71) which is the return type of one of the sparkey side-input views

I think we can definitely redesign this as a completely static class (scala Object style) that doesn't get serialized into the job graph -- will try to work on that today/tomorrow.

scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
```diff
@@ -62,15 +62,16 @@ public FileDownloadDoFn(RemoteFileUtil remoteFileUtil, SerializableFunction<Path
...
       SerializableFunction<Path, OutputT> fn,
       int batchSize,
       boolean keep) {
-    this.remoteFileUtil = remoteFileUtil;
     this.fn = fn;
     this.batch = new ArrayList<>();
     this.batchSize = batchSize;
     this.keep = keep;
+
+    RemoteFileUtil.configure(options);
```
RustedBones, line 74: Aren't we able to access the pipeline option in the `@Setup` method to run this code ?
We wouldn't need the options in the constructor

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -69,13 +69,14 @@ public Path load(URI key) throws Exception {
               });
 
   /** Create a new {@link RemoteFileUtil} instance. */
-  public static RemoteFileUtil create(PipelineOptions options) {
+  private RemoteFileUtil() {}
+
+  public static void configure(PipelineOptions options) {
```
RustedBones, line 74: Why not calling `FileSystems.setDefaultPipelineOptions(options)` directly ?
I see we have this already in some places

scio-extra/src/main/scala/com/spotify/scio/extra/annoy/AnnoyUri.scala
```diff
@@ -60,10 +60,12 @@ private class LocalAnnoyUri(val path: String) extends AnnoyUri {
 }
 
 private class RemoteAnnoyUri(val path: String, options: PipelineOptions) extends AnnoyUri {
-  private[this] val rfu: RemoteFileUtil = RemoteFileUtil.create(options)
+  {
+    RemoteFileUtil.configure(options)
+  }
```
RustedBones, line 65: I'm not sure we should do this here but let the user initialize outside the model

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
clairemcginty: like as part of a static initialization? we could do that -- my thought process was that we don't want `FileSystems.setDefaultPipelineOptions(options)` to be called on DoFn deserialization (on the worker), only at graph construction time, right? I guess it doesn't matter too much, though.

RustedBones: Maybe I miss smth, Isn't the `RemoteFileUtil` used in the transform step only ?
At the graph, we only require the file URI + deserialization method.
`RemoteFileUtil` is used on the worker to retrieve those files right ?

PR #4771, opened 2023-04-06
Sync joda-time with beam

RustedBones:
Beam is depending on `joda-time`. Sync version used by scio with beam's one


shnapz (approved):
@RustedBones I am curious if you detected some compatibility issues with this

RustedBones:
No. I was just surprised to see the update, as I was sure beam depends on joda.
I think I missed this one because there wasn't any conflicts. Just being on the safe side here.

clairemcginty (approved):

PR #4772, opened 2023-04-06
Support logical types in SpecificRecord Parquet writes

clairemcginty:
`.saveAsParquetAvroFile(path)` was failing for specific record collections with:

```
java.lang.ClassCastException: class org.joda.time.DateTime cannot be cast to class java.lang.Number (org.joda.time.DateTime is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')
	at org.apache.parquet.avro.AvroWriteSupport.writeValueWithoutConversion(AvroWriteSupport.java:336)
	at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:284)
	at org.apache.parquet.avro.AvroWriteSupport.writeRecordFields(AvroWriteSupport.java:200)
	at org.apache.parquet.avro.AvroWriteSupport.write(AvroWriteSupport.java:174)
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.write(InternalParquetRecordWriter.java:138)
	at org.apache.parquet.hadoop.ParquetWriter.write(ParquetWriter.java:310)
	at com.spotify.scio.parquet.avro.ParquetAvroFileBasedSink$ParquetAvroWriter.write(ParquetAvroFileBasedSink.java:117)
	at org.apache.beam.sdk.io.WriteFiles.writeOrClose(WriteFiles.java:632)
	at org.apache.beam.sdk.io.WriteFiles.access$1100(WriteFiles.java:125)
	at org.apache.beam.sdk.io.WriteFiles$WriteShardsIntoTempFilesFn.processElement(WriteFiles.java:959)
```

This is happening for 2 reasons:

1. The default "model" for Parquet's [AvroWriteSupport](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L285), `SpecificData`, doesn't contain a mapping of logical types --> field converts. (In regular Avro reads/writes, the `SpecificRecord` implementations hold a map of types-->conversions directly so the user doesn't have to do anything; but due to Parquet's Avro abstraction layer, conversions must be manually specified.)

2. Even if you specify a custom `model` in your Parquet configuration (via `AvroReadSupport.AVRO_DATA_SUPPLIER` and `AvroWriteSupport.AVRO_DATA_SUPPLIER` configuration keys) that _does_ contain these mappings, it doesn't actually get applied during writes. Parquet requires us to manually set `.withDataModel` when constructing a `ParquetAvroWriter`, and will [hard-set it to SpecificData](https://github.com/apache/parquet-mr/blob/59e9f78b8b3a30073db202eb6432071ff71df0ec/parquet-avro/src/main/java/org/apache/parquet/avro/AvroParquetWriter.java#L163) if left unspecified; I've filed a [ticket](https://issues.apache.org/jira/browse/PARQUET-2265) with the Parquet JIRA because I think that the data model should be infererred from the data_supplier Configuration key, if present. (Parquet Avro Reader already handles this correctly, so there's nothing we need to do there.)

This PR fixes both issues. It sets a default `model` for _all_ SpecificRecord reads/writes that contains _all_ logical type mappings, and it fixes the `ParquetAvroWriter` builders to actually pass in the configured `model`.


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -230,10 +233,9 @@ object ParquetAvroIO {
       // org.apache.beam.sdk.coders.AvroCoder.
       if (ScioUtil.classOf[A] == classOf[GenericRecord]) {
         job.getConfiguration.setBoolean("parquet.avro.compatible", false)
-        job.getConfiguration.set(
-          "parquet.avro.data.supplier",
-          "org.apache.parquet.avro.GenericDataSupplier"
-        )
+        AvroReadSupport.setAvroDataSupplier(job.getConfiguration, classOf[GenericDataSupplier])
+      } else {
+        AvroReadSupport.setAvroDataSupplier(job.getConfiguration, classOf[LogicalTypeSupplier])
```
clairemcginty, line 238: I'm a bit torn on setting this as the default for all SR parquet-avro operations, regardless if the schema contains a logical type or not; but it should be harmless, just a small in-memory map that never gets used.

The downside of making this an opt-in feature is that the default behavior of `.saveAsAvroFile` w.r.t LogicalTypes will remain different from the default behavior of `.saveAsParquetAvroFile`; users wouldn't be able to migrate from Avro to Parquet-Avro with just the drop-in replacement. 🤔

Maybe we could scan the `schema` param for any logical-type fields, and if we detect any, only *then* do we set the data-supplier to `LogicalTypeSupplier`. If we don't find any, we let it default to just `SpecificDataSupplier`.

farzad-sedghi: maybe we can start with this simpler solution, and optimize it when it's well tested?

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroIOTest.scala
```diff
@@ -119,6 +121,63 @@ class ParquetAvroIOTest extends ScioIOSpec with TapSpec with BeforeAndAfterAll {
...
+      .map(identity) should containInAnyOrder(records)
+
+    sc2.run()
+    ()
+  }
+
+  it should "write and read GenericRecords with logical types" in {
+    val records: Seq[GenericRecord] = (1 to 10).map { _ =>
+      val gr = new GenericData.Record(TestLogicalTypes.SCHEMA$)
+      gr.put("timestamp", DateTime.now().getMillis)
```
farzad-sedghi, line 154: is `.getMillis` added intentionally?

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroIOTest.scala
```diff
@@ -119,6 +121,63 @@ class ParquetAvroIOTest extends ScioIOSpec with TapSpec with BeforeAndAfterAll {
...
+    val sc2 = ScioContext()
+    sc2
+      .parquetAvroFile[TestLogicalTypes](s"$path/*.parquet")
+      .map(identity) should containInAnyOrder(records)
+
+    sc2.run()
+    ()
+  }
+
+  it should "write and read GenericRecords with logical types" in {
```
farzad-sedghi, line 151: do you think it's useful to do a cross test e.g. write with Generic and read with SpecificRcord?

clairemcginty: I think in that regard, that's more in the realm of native Avro functionality than any Scio functionality that we own. When I was validating this change with real data, I did do a run with GenericRecord writes and SpecificRecord reads, which succeeded :) but I don't think we need a devoted unit test for it.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
clairemcginty: probably the simplest solution is making this totally opt in for any users whose schema contains a logical type :) I'm just trying to reason if there's any downside to always including logical-type conversions

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -288,6 +302,17 @@ object ParquetAvroIO {
...
     filenamePolicySupplier: FilenamePolicySupplier = WriteParam.DefaultFilenamePolicySupplier
   )
+
+  private[avro] def getDataSupplier(
+    key: String,
+    conf: Configuration
+  ): Option[Class[_ <: AvroDataSupplier]] = {
+    Option(conf.getClass(key, null)).map {
+      case cls if classOf[AvroDataSupplier].isAssignableFrom(cls) =>
+        cls.asInstanceOf[Class[_ <: AvroDataSupplier]]
```
clairemcginty, line 312: Is there a neater way to do this in Scala?

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -288,6 +302,14 @@ object ParquetAvroIO {
     tempDirectory: String = WriteParam.DefaultTempDirectory,
     filenamePolicySupplier: FilenamePolicySupplier = WriteParam.DefaultFilenamePolicySupplier
   )
+
+  private[avro] def getDataSupplier(
+    key: String,
+    conf: Configuration
+  ): Option[Class[_ <: AvroDataSupplier]] =
+    Option(conf.getClass(key, null)).map { case _ =>
```
clairemcginty, line 310: `Configuration` doesn't have a `.containsKey` type method, thus this awkward workaround 😅

scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroFileBasedSink.java
```diff
@@ -107,9 +111,24 @@ public ParquetAvroWriter(
...
+
+      // Workaround for PARQUET-2265
+      if (configuration.getClass(AvroWriteSupport.AVRO_DATA_SUPPLIER, null) != null) {
+        Class<? extends AvroDataSupplier> dataModelSupplier =
+            configuration.getClass(
+                AvroWriteSupport.AVRO_DATA_SUPPLIER,
+                SpecificDataSupplier.class,
+                AvroDataSupplier.class);
+        builder =
+            builder.withDataModel(
```
clairemcginty, line 127: reference: https://github.com/apache/parquet-mr/blob/59e9f78b8b3a30073db202eb6432071ff71df0ec/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L403-L407

scio-parquet/src/main/java/com/spotify/scio/parquet/avro/LogicalTypeSupplier.java
```diff
@@ -0,0 +1,39 @@
...
+/** A SpecificDataSupplier that supplies built-in conversions for Avro LogicalTypes. */
+public class LogicalTypeSupplier extends SpecificDataSupplier {
+  @Override
+  public GenericData get() {
+    return new SpecificData() {
+      {
+        addLogicalTypeConversion(new TimeConversions.DateConversion());
+        addLogicalTypeConversion(new TimeConversions.TimeConversion());
+        addLogicalTypeConversion(new TimeConversions.TimestampConversion());
+        addLogicalTypeConversion(new Conversions.DecimalConversion());
```
shnapz, line 35: Maybe it's worth to add all remaining conversions in `TimeConversions` and `Conversions`?

clairemcginty: oh, I guess I forgot a few `*Micros` conversions, and UUID.... will add them, good catch!

clairemcginty: added!

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroIOTest.scala
```diff
@@ -119,6 +127,105 @@ class ParquetAvroIOTest extends ScioIOSpec with TapSpec with BeforeAndAfterAll {
...
+      gr.set("timestamp", DateTime.now())
+      gr.set(
+        "decimal",
+        BigDecimal.decimal(1.0).setScale(2).bigDecimal
+      )
+      gr.build()
+    }
+    val path = dir.toPath.resolve("logicalTypesGr").toString
+
+    implicit val coder = {
```
clairemcginty, line 170: Realized that GenericRecord coders by default break on logical types -- I'll file a separate issue for that.

farzad-sedghi (approved):

PR #4774, opened 2023-04-10
Pass dataModel to AvroParquetWriter.Builder

clairemcginty:
Context: `AvroParquetWriter` doesn't load its data-model from the `parquet.avro.write.data.supplier` Configuration param, but instead must be passed in directly to the Builder or else it'll default to `SpecificData.get()`.

Since we don't set `.withDataModel` in Scio's `AvroParquetWriter.Builder`, users effectively cannot override the data supplier used in their Avro write. I filed a [ticket](https://issues.apache.org/jira/browse/PARQUET-2265) on the Parquet JIRA to change this behavior on the Parquet side, but for now we'll have to workaround it by setting `.withDataModel` based on the value of `parquet.avro.write.data.supplier`.

Note: this doesn't replace #4772. Whereas 4772 would address the most common use cases (datetime logical types in Avro 1.8), users with more advanced use cases (i.e., using Avro 1.11+logical types) will still need this PR in order to override data supplier.

### DIFF
--- a/scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroFileBasedSink.java
+++ b/scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroFileBasedSink.java
@@ -27,7 +27,11 @@ import org.apache.beam.sdk.io.hadoop.SerializableConfiguration;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.util.MimeTypes;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.util.ReflectionUtils;
+import org.apache.parquet.avro.AvroDataSupplier;
 import org.apache.parquet.avro.AvroParquetWriter;
+import org.apache.parquet.avro.AvroWriteSupport;
+import org.apache.parquet.avro.SpecificDataSupplier;
 import org.apache.parquet.hadoop.ParquetWriter;
 import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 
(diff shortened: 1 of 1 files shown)


scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroFileBasedSink.java
```diff
@@ -107,9 +111,21 @@ public ParquetAvroWriter(
...
+              AvroWriteSupport.AVRO_DATA_SUPPLIER,
+              SpecificDataSupplier.class,
+              AvroDataSupplier.class);
+
       AvroParquetWriter.Builder<T> builder =
-          AvroParquetWriter.<T>builder(outputFile).withSchema(schema);
-      writer = WriterUtils.build(builder, conf.get(), compression);
+          AvroParquetWriter.<T>builder(outputFile)
+              .withSchema(schema)
+              .withDataModel(ReflectionUtils.newInstance(dataModelSupplier, configuration).get());
```
clairemcginty, line 126: This is directly copied from parquet's [AvroWriteSupplier](https://github.com/apache/parquet-mr/blob/59e9f78b8b3a30073db202eb6432071ff71df0ec/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L403-L407); unfortunately the method is private so I can't just invoke it.

farzad-sedghi: anyway we can add test coverage for this?

clairemcginty:
Actually I think this might conflict with the logic set out in #4772; let me see if I can combine them.

clairemcginty:
Merged into #4772.

PR #4778, opened 2023-04-12
Ensures guava callback to complete chained futures

RustedBones:
- Catch all `Throwable` on Guava callbacks to ensure returned future is completed
- Move from `computeIfAbsent` to `put` or `putIfAbsent` since keys are UUIDs and lazy computation was useless
- Simplify semaphore and cache handling in `BaseAsyncLookupDoFn`
- Reset `inFlightRequests` and `semaphore` permits in `startBundle` to avoid leaks/locks
- Apply same code structure on `BaseAsyncDoFn`

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
@@ -168,6 +168,7 @@ public abstract class BaseAsyncLookupDoFn<A, B, C, F, T>
       } else if (inFlight != null) {
         // pending request for the same element
         futures.put(uuid, handleOutput(inFlight, input, uuid, timestamp, window));
+        requestCount++;
       } else {
         // semaphore release is not performed on exception.
         // let beam retry the bundle. startBundle will reset the semaphore to the
@@ -179,8 +180,8 @@ public abstract class BaseAsyncLookupDoFn<A, B, C, F, T>
         // make sure semaphore are released when waiting for futures in finishBundle
         final F unlockedFuture = handleSemaphore(future);
         futures.put(uuid, handleOutput(unlockedFuture, input, uuid, timestamp, window));
+        requestCount++;
       }
-      requestCount++;
     } catch (InterruptedException e) {
       LOG.error("Failed to acquire semaphore", e);
       throw new RuntimeException("Failed to acquire semaphore", e);
--- a/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
@@ -58,6 +58,7 @@ public abstract class BaseAsyncLookupDoFn<A, B, C, F, T>
   private final boolean deduplicate;
 
   // Data structures for handling async requests
+  private final int maxPendingRequests;
   private final Semaphore semaphore;
   private final ConcurrentMap<UUID, F> futures = new ConcurrentHashMap<>();
   private final ConcurrentMap<A, F> inFlightRequests = new ConcurrentHashMap<>();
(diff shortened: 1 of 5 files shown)


scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -92,7 +99,7 @@ default void waitForFutures(Iterable<CompletableFuture<V>> futures)
         throws InterruptedException, ExecutionException {
       CompletableFuture[] array =
           StreamSupport.stream(futures.spliterator(), false).toArray(CompletableFuture[]::new);
-      CompletableFuture.allOf(array).get();
```
RustedBones, line 95: all implementation do not throw in `waitForFutures` but this one. Change to match same behavior

scio-test/src/test/scala/com/spotify/scio/transforms/AsyncLookupDoFnTest.scala
```diff
@@ -65,8 +64,9 @@ class AsyncLookupDoFnTest extends PipelineSpec {
   )(tryFn: T => Try[String]): Unit = {
     val output = runWithData(1 to 10)(_.parDo(doFn)).map { kv =>
       val r = tryFn(kv.getValue) match {
-        case Success(v) => v
-        case Failure(e) => e.getMessage
+        case Success(v)                      => v
+        case Failure(e: CompletionException) => e.getCause.getMessage
```
RustedBones, line 68: Java future wraps the exception in a `CompletionException`. Unwrap for the test

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -246,6 +203,7 @@ public void finishBundle(FinishBundleContext context) {
         throw new RuntimeException("Failed to process futures", e);
       } catch (ExecutionException e) {
         LOG.error("Failed to process futures", e);
+        throw new RuntimeException("Failed to process futures", e);
```
RustedBones, line 206: This was probably removed due to inconsistent behaviour of the java future, throwing in `waitForFutures` if a any of the future failed.

scio-test/src/test/scala/com/spotify/scio/transforms/FutureHandlersTest.scala
```diff
@@ -0,0 +1,157 @@
...
+        handler.waitForFutures(List[F](f1, f2).asJava)
+      }
+
+      fail(f1)(new Exception("f1 failed"))
+      backgroundTask.isCompleted shouldBe false
+      complete(f2)("f2 done")
+      noException shouldBe thrownBy {
+        Await.result(backgroundTask, 1.second)
+      }
+    }
```
RustedBones, line 52: This test was failing for java

scio-test/src/test/scala/com/spotify/scio/transforms/FutureHandlersTest.scala
```diff
@@ -0,0 +1,157 @@
...
+      )
+      val e = new Exception("failure")
+      fail(f)(e)
+      result shouldBe Failure(e)
+      val ee = the[ExecutionException] thrownBy (access(chainedFuture))
+      val cause = ee.getCause
+      cause.getMessage shouldBe "failure"
+      cause.getSuppressed.headOption.map(_.getMessage) shouldBe Some("callback failure")
+    }
+  }
```
RustedBones, line 134: Those tests were failing for guava, propagating the exception in the callback thread (very probably from client implementation)

RustedBones (commented):
@farzad-sedghi @shnapz  Added unit test as discussed

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -45,7 +45,7 @@ public class FutureHandlers {
     default void waitForFutures(Iterable<ListenableFuture<V>> futures)
         throws InterruptedException, ExecutionException {
       // Futures#allAsList only works if all futures succeed
-      Futures.whenAllComplete(futures).run(() -> {}, MoreExecutors.directExecutor()).get();
+      Futures.successfulAsList(futures).get();
```
clairemcginty, line 48: just wondering, does this use the same executor (`MoreExecutors.directExecutor()`)? 

Also, how does `successfulAsList` handle failed futures? does it await all futures and only return failed ones?

clairemcginty (approved):

PR #4779, opened 2023-04-13
Set provided name on transforms with side outputs

### LINKED ISSUE #4777: SCollectionWithSideOutput doesn't respect name from the input SCollection
For the following pipeline, you don't get proper step name for the latest map:

```
val sideOutput = SideOutput[AnySideOutputType]()

val events: SCollection[AnyType] = ???
events
  .withName("Foo")
  .map(...)
  .withName("Bar")
  .withSideOutput(sideOutput)
  .map(...)
```

The step name "Bar" is lost when `withSideOutput` creates `SCollectionWithSideOutput`. The latest map gets the "Foo" step name. It's annoying if the sideOutput is a part of reusable transformation like unzipping Either into two SCollections:

```
implicit class SCollectionEitherOps[L: Coder, R: Coder](self: SCollection[Either[L, R]]) {
  /**
   * Unzip Scala Either into SCollection of "right" elements, and SCollection of "left/invalid" elements
   */
  def unzip: (SCollection[R], SCollection[L]) = {
    val leftOutput = SideOutput[L]()

    val (rightOutput, sideOutputs) = self
      .withSideOutputs(leftOutput)
      .withName(self.name) // propagate origin name as a prefix for right and left outputs
      .flatMap {
        case (Right(r), _) => Some(r)
        case (Left(l), ctx) =>
          ctx.output(leftOutput, l)
          None
      }

    (rightOutput, sideOutputs(leftOutput))
  }
}
```

And usage:

```
val events: SCollection[Either[L,R]] = ???

events.withName("Foo").map(...).withName("Bar").unzip
```

Now, `flatMap` step inside `unzip` gets the "Foo" name as a prefix instead of "Bar".

RustedBones:
Fix #4777 

Do not run scio context when only asserting on names to speedup testing

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideOutput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideOutput.scala
@@ -32,19 +32,27 @@ import scala.jdk.CollectionConverters._
  * s of the [[SideOutput]] s are accessed via the additional [[SideOutputCollections]] return value.
  */
 class SCollectionWithSideOutput[T] private[values] (
-  val internal: PCollection[T],
-  val context: ScioContext,
+  coll: SCollection[T],
   sides: Iterable[SideOutput[_]]
 ) extends PCollectionWrapper[T] {
+
+  override val internal: PCollection[T] = coll.internal
+  override val context: ScioContext = coll.context
+
   private val sideTags = TupleTagList.of(sides.map(_.tupleTag).toList.asJava)
+
+  override def withName(name: String): this.type = {
+    coll.withName(name)
+    this
+  }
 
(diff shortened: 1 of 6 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideOutput.scala
```diff
@@ -32,19 +32,27 @@ import scala.jdk.CollectionConverters._
  * s of the [[SideOutput]] s are accessed via the additional [[SideOutputCollections]] return value.
  */
 class SCollectionWithSideOutput[T] private[values] (
-  val internal: PCollection[T],
-  val context: ScioContext,
```
RustedBones, line 36: This is a breaking change.
We need to have access to the `Scollecction` to get the name. The new definition is consistent with other `SCollection` wrapper

farzad-sedghi (approved):

PR #4782, opened 2023-04-17
Make Parquet logical type support opt-in

clairemcginty:
Makes Parquet logical type support opt-in, rather than included by default. `scio-parquet` will warn if a logical type is detected in the schema but no logical type provider is configured.

The reasoning for this is that Avro logical type conversion classes are heavily tied to Avro version -- `TimeConversions.TimestampConversion` in Avro 1.8 is renamed to `TimeConversions.TimestampMillisConversion` in Avro 1.11, for example. So by including these converters by default, we're essentially hard-failing any Avro 1.11 users immediately, even if their schema doesn't have any logical types.

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
@@ -28,7 +28,6 @@ import com.spotify.scio.util.{FilenamePolicySupplier, Functions, ScioUtil}
 import com.spotify.scio.values.SCollection
 import com.twitter.chill.ClosureCleaner
 import org.apache.avro.Schema
-import org.apache.avro.generic.GenericRecord
 import org.apache.avro.reflect.ReflectData
 import org.apache.avro.specific.SpecificRecordBase
 import org.apache.beam.sdk.io._
@@ -42,14 +41,17 @@ import org.apache.beam.sdk.values.TypeDescriptor
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.mapreduce.Job
 import org.apache.parquet.avro.{
+  AvroDataSupplier,
   AvroParquetInputFormat,
   AvroParquetReader,
   AvroReadSupport,
+  AvroWriteSupport,
   GenericDataSupplier
 }
 import org.apache.parquet.filter2.predicate.FilterPredicate
 import org.apache.parquet.hadoop.ParquetInputFormat
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
+import org.slf4j.LoggerFactory
 
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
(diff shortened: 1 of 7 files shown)


site/src/main/paradox/io/Parquet.md
```diff
@@ -112,6 +112,66 @@ def yourAvroSchema: org.apache.avro.Schema = ???
...
+case class AvroLogicalTypeSupplier() extends AvroDataSupplier {
+  override def get(): GenericData = {
+    val specificData = SpecificData.get()
+
+    // Add conversions as needed
+    specificData.addLogicalTypeConversion(new TimeConversions.TimestampMillisConversion())
+
+    specificData
+  }
+}
```
RustedBones, line 170: I feel we should be able to do better that this.
With specific records, all conversions are added to a pre-generated model contained in the static `$MODEL` field of the class, and can be accessed with `SpecificData.getForClass` in avro 1.9+. It is still possible to get this work with avro 1.8 with some tricks.
This supplier should only be required for records generated with `ReflectData` or `GeneridData` where conversions must be explicitly given.

clairemcginty: I did see `$MODEL` -- as well as a static field `org.apache.avro.Conversion<?>[] conversions` that has everything we need -- but was hesitant to use them because in Avro 1.8 at least, they're both private so I'd have to use reflection on private/maybe unstable fields. It would be ideal to get the conversions directly from the specific record class though

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -117,6 +119,17 @@ final case class ParquetAvroIO[T: ClassTag: Coder](path: String) extends ScioIO[
   override protected def write(data: SCollection[T], params: WriteP): Tap[T] = {
     val isAssignable = classOf[SpecificRecordBase].isAssignableFrom(cls)
     val writerSchema = if (isAssignable) ReflectData.get().getSchema(cls) else params.schema
+    val conf = Option(params.conf).getOrElse(new Configuration())
```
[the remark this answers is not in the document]
clairemcginty: agreed! I'll clean up config usage in a follow up PR.

site/src/main/paradox/io/Parquet.md
clairemcginty: Looked into this a bit more!

I think you're right, we can definitely go from `T`/`Schema` --> a `SpecificData` model populated with conversion types by re-using Avro's model-cache logic.

But the complex bit is actually wiring this model in. Parquet allows you to specify a model by either (a) setting `.withDataModel()` on the read/write builder, or (b) supplying the Configuration `*DATA_SUPPLIER` param. To use our derived `SpecificData` model, we'd have to use `.withDataModel` directly, I think -- I can't see how we'd make this work with a [generic Supplier class](https://github.com/apache/parquet-mr/blob/master/parquet-avro/src/main/java/org/apache/parquet/avro/AvroDataSupplier.java) that requires a no-arg constructor.

For Parquet writes, we do invoke the write builder directly, so we could take our derived `SpecificData` and set it using [.withDataModel](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/java/com/spotify/scio/parquet/avro/ParquetAvroFileBasedSink.java#L127).

But for reads, we don't get to directly access the `AvroParquetReader` and set `.withDataModel`--there are too many layers of indirection--so the only way we can specify a data model is to configure the `DATA_SUPPLIER` parameter with the name of a Supplier class, which as I mentioned above, I don't know how we could wire our custom data model into.

PR #4791, opened 2023-05-02
Clean up Configuration initialization in scio-parquet

clairemcginty:


scio-parquet/src/main/scala/com/spotify/scio/parquet/package.scala
```diff
@@ -67,5 +70,8 @@ package object parquet {
       }
       conf
     }
+
+    private[parquet] def ofNullable(conf: Configuration): Configuration =
+      Option(conf).getOrElse(ParquetConfiguration.empty())
```
RustedBones, line 75:
```suggestion
      Option(conf).getOrElse(empty())
```

RustedBones (approved):

PR #4801, opened 2023-05-08
Add helper methods for Parquet ReadFiles transforms

clairemcginty:
Adds helper methods to create Parquet ReadFiles transforms.

Motivation: I realized it was really hard for users to construct a correct Parquet ReadFiles for Avro because the required configuration is so specific. So I added helpers, designed to be used in a similar way as Beam's AvroIO.readFiles:

```scala
// beam AvroIO.readFiles
sc.parallelize(files).readFiles(BAvroIO.readFiles(schema))

// Scio Parquet ReadFiles
sc.parallelize(files).readFiles(ParquetRead.readAvroGenericRecordFiles(schema))
```

feedback on the method naming/placement welcome -- I guess `ParquetRead.readAvroGenericRecordFiles` is kind of redundant?


scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
     readSupportFactory: ReadSupportFactory[T],
     conf: SerializableConfiguration,
     projectionFn: SerializableFunction[T, R]
-  ): ParDo.SingleOutput[ReadableFile, R] = {
+  ): PTransform[PCollection[ReadableFile], PCollection[R]] = {
```
clairemcginty, line 61: I think this will technically break MiMA :( The reason is that `ParDo.of(..)` constructs a `PTrasnform[PCollection[_ <: ReadableFile], ...]`. Thus, the pre-existing signature didn't work when used with SCollection#readFiles as the signatures didn't match.

clairemcginty:
on second thought, maybe I can move these methods into `ParquetAvroIO` and `ParquetTypeIO` objects, so they can be accessed i.e. `ParquetAvroIO.readFiles(schema)`... that's more analogous to Beam-style transform constructors, I think

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
shnapz: maybe candidate to 0.13?

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
...
+    predicate: FilterPredicate = null,
+    conf: Configuration = null
+  ): PTransform[PCollection[ReadableFile], PCollection[T]] = readTyped[T, T](
+    identity,
+    predicate,
+    conf
+  )
+
+  def readTyped[T: ClassTag: Coder: ParquetType, R](
+    projectionFn: T => R,
```
RustedBones, line 78: Why 2 APIs and not setting `identity` as default projectionFn ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
...
+    val cleanedFn = Functions.serializableFn(ClosureCleaner.clean(projectionFn))
+    readFiles(
+      ReadSupportFactory.avro,
+      new SerializableConfiguration(configuration),
+      cleanedFn
+    )
+  }
+
+  def readAvro[T <: SpecificRecordBase: ClassTag](
+    projection: Schema = null,
```
RustedBones, line 121: Why do we need a `projection` schema in case of `SpecificRecord` ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
...
+
+    val cleanedFn = Functions.serializableFn(ClosureCleaner.clean(projectionFn))
+    readFiles(
+      ReadSupportFactory.avro,
+      new SerializableConfiguration(configuration),
+      cleanedFn
+    )
+  }
+
+  def readAvro[T <: SpecificRecordBase: ClassTag](
```
RustedBones, line 120: IMHO we should define API with weakest abstraction `SpecificRecord` instead. See #4459

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
...
+  def readAvroGenericRecordFiles(
+    schema: Schema,
+    predicate: FilterPredicate = null,
+    conf: Configuration = null
+  ): PTransform[PCollection[ReadableFile], PCollection[GenericRecord]] =
+    readAvroGenericRecordFiles[GenericRecord](schema, identity, predicate, conf)
+
+  def readAvroGenericRecordFiles[T](
+    schema: Schema,
+    projectionFn: GenericRecord => T,
```
RustedBones, line 98: This does not look to be a projection: We are reading the whole schema data even if the function selects a single field of the GenericRecord.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
...
+  def readTyped[T: ClassTag: Coder: ParquetType, R](
+    projectionFn: T => R,
+    predicate: FilterPredicate,
+    conf: Configuration
+  ): PTransform[PCollection[ReadableFile], PCollection[R]] = {
+    val configuration = ParquetConfiguration.ofNullable(conf)
+    Option(predicate).foreach(p => ParquetInputFormat.setFilterPredicate(configuration, p))
+
+    val cleanedFn = Functions.serializableFn(ClosureCleaner.clean(projectionFn))
+    readFiles(ReadSupportFactory.typed[T], new SerializableConfiguration(configuration), cleanedFn)
```
RustedBones, line 86: I'm not really sure the `projectionFn` is doing what it is supposed to.
I checked in the `ParquetReadFn` [here](https://github.com/spotify/scio/blob/ddc4a397dea226ce0132bf2bbff76be6d3382be4/scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala#L240) and it seems to me it is the equivalent of a chained `map` step.

IMHO we should either:
- simplify the API by removing projection and let user do a `map` post read (I am also surprised here we do not have an implicit `Coder[R]` for the output type)
- ask user to create a model `T` which is a projection of the model used when writing the parquet file

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
...
+  def readAvro[T <: SpecificRecordBase: ClassTag](
+    projection: Schema = null,
+    predicate: FilterPredicate = null,
+    conf: Configuration = null
+  ): PTransform[PCollection[ReadableFile], PCollection[T]] =
+    readAvro[T, T](projection, identity, predicate, conf)
+
+  def readAvro[T <: SpecificRecordBase: ClassTag, R](
+    projection: Schema,
+    projectionFn: T => R,
```
RustedBones, line 129: I don're really like those 2 to be separated as they are strongly coupled.
If we want to offer projection, I think we better have to rework the `Projection` helper so it encapsulate both the mapping function and the schema.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
clairemcginty: Just to make the API friendlier for users who aren't supplying a projection, so they con't have to specify their `T` type twice

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
clairemcginty: Re: implicit `Coder[R]` -- that would be required when the ReadFiles transform is actually applied, here: https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L1396

will circle back on the `projectionFn`, I want to check something...

clairemcginty: ok, so for the projection -- the main advantage of offering the projection, as I understand it, is that it happens pre-Coder#encode.

I think that for GenericRecords, this doesn't offer any advantage, _as long as_ the user correctly specifies an implicit Coder using the projected schema -- see [unit test](https://github.com/spotify/scio/pull/4801/files#diff-8cee8349c982f42099f3e288feab6e0fc8f1031973f33942c34b2df23ec70425R181).

But for SpecificRecords, say you have a schema with two non-nullable fields, but you only want to project one of those fields. Without being able to apply a `projectionFn` pre-Coder#encode, serialization will fail due to the absence of a non-nullable field. I just updated the unit test to use `Account` as a test record, which has non-nullable fields; the SpecificRecord test w/ no projectionFn is currently [failing](https://github.com/spotify/scio/pull/4801/files#diff-8cee8349c982f42099f3e288feab6e0fc8f1031973f33942c34b2df23ec70425R256-R273).

So overall, I think that:

a. ProjectionFn support for GenericRecord reads isn't important, and could be removed; however, I'm worried that users will be confused about setting the correct Coder here.
b. ProjectionFn support should be a requirement for SpecificRecord reads if your projection excludes non-nullable, non-default fields.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
clairemcginty: I'm happy to rework the Parquet-Avro ReadParam to make this more clear! I think that it depends a bit if we'd like to include this change in a 0.12 release (I can revert the `readFiles` signature changes to make MiMa happy), in which case it should be left as-is, or if we'd like to save it for 0.13.

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +58,92 @@ object ParquetRead {
     readSupportFactory: ReadSupportFactory[T],
     conf: SerializableConfiguration,
     projectionFn: SerializableFunction[T, R]
-  ): ParDo.SingleOutput[ReadableFile, R] = {
+  ): PTransform[PCollection[ReadableFile], PCollection[R]] = {
     val sdf = new ParquetReadFn[T, R](readSupportFactory, conf, projectionFn)
```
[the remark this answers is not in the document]
clairemcginty: Yeah, I think it's a little less performant on RunnerV1 (doesn't scale up as well).

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +57,163 @@ object ParquetRead {
...
+   * @param predicate
+   *   a Parquet [[FilterPredicate]] predicate
+   * @param conf
+   *   a Parquet [[Configuration]]
+   */
+  def readAvroGenericRecordFiles[T](
+    schema: Schema,
+    projectionFn: GenericRecord => T,
+    predicate: FilterPredicate,
+    conf: Configuration
```
RustedBones, line 142: Should we create method overload to avoid setting those at null ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +57,163 @@ object ParquetRead {
...
+   * @param predicate
+   *   a Parquet [[FilterPredicate]] predicate
+   * @param conf
+   *   a Parquet [[Configuration]]
+   */
+  def readAvro[T <: SpecificRecord: ClassTag, R](
+    projection: Schema,
+    projectionFn: T => R,
+    predicate: FilterPredicate,
+    conf: Configuration
```
RustedBones, line 201: same here ?

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
```diff
@@ -43,8 +57,163 @@ object ParquetRead {
...
+   *   a function mapping T => R
+   * @param predicate
+   *   a Parquet [[FilterApi]] predicate
+   * @param conf
+   *   a Parquet [[Configuration]]
+   */
+  def readTyped[T: ClassTag: ParquetType, R](
+    projectionFn: T => R,
+    predicate: FilterPredicate,
+    conf: Configuration
```
RustedBones, line 98: and here

RustedBones (approved):

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetRead.scala
clairemcginty: I've been thinking about that -- my concern is that overloading might make things more confusing from user POV. Say  we have:

````scala
def readAvro[T <: SpecificRecord: ClassTag](): PTransform[PCollection[ReadableFile], PCollection[T]] =
    readAvro[T](null, null, null)

def readAvro[T <: SpecificRecord: ClassTag](
    projection: Schema,
    predicate: FilterPredicate,
    conf: Configuration
  ): PTransform[PCollection[ReadableFile], PCollection[T]] = ...
```

In this case, users can either pass 0 arguments, or all 3. But I think it's somewhat common use case to pass in only a projection, or only a predicate -- and that would be more difficult for the user if we went with the overloaded API.

RustedBones: I meant we can create overloads for all possibilities

PR #4803, opened 2023-05-12
Relax avro record type bound

RustedBones:
Part of #4459

Also fixes a bug for 2nd  key in parquet for scala model


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -25,30 +25,29 @@
...
   @JsonProperty private final String keyField;
 
   @JsonProperty
   @JsonInclude(JsonInclude.Include.NON_NULL)
   private final String keyFieldSecondary;
 
-  @JsonIgnore private final String[] keyPath;
-  @JsonIgnore private final String[] keyPathSecondary;
+  @JsonIgnore private final AtomicReference<int[]> keyPath = new AtomicReference<>();
+  @JsonIgnore private final AtomicReference<int[]> keyPathSecondary = new AtomicReference<>();
```
RustedBones, line 50: We now have to delay the construction for those, as the schema is required to get access to the field index instead of the field name

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
```diff
@@ -270,9 +295,9 @@ private <K> K extractAvroKey(Class<K> keyClazz, String[] keyPath, V value) {
   }
 
   // FIXME: what about `Option[T]`
-  private <K> K extractScalaKey(String[] keyPath, V value) {
+  private <K> K extractScalaKey(Method[] keyGetters, V value) {
     Object obj = value;
-    for (Method getter : getOrInitGetters(keyPath, value.getClass())) {
```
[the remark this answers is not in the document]
kellen: Yikes!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroUtils.java
```diff
@@ -37,31 +38,28 @@
 class AvroUtils {
   private AvroUtils() {}
 
-  public static String[] toKeyPath(String keyField) {
-    return keyField.split("\\.");
-  }
-
-  public static String validateKeyField(String keyField, Class<?> keyClass, Schema schema) {
-    final String[] keyPath = toKeyPath(keyField);
+  public static int[] toKeyPath(String keyField, Class<?> keyClass, Schema schema) {
```
kellen, line 41: Can you add a docstring about how the `int[]` should be interpreted? Assuming it's a series of avro field indexes...

kellen (approved):

PR #4808, opened 2023-05-16
Remove old skewed-join API

RustedBones:
Part of #4459


scio-core/src/main/scala/com/spotify/scio/values/PairSkewedSCollectionFunctions.scala
```diff
@@ -200,57 +211,15 @@ class PairSkewedSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
-   *   left side sample fraction. Default is `1.0` - no sampling.
-   * @param withReplacement
-   *   whether to use sampling with replacement, see
-   *   [[SCollection.sample(withReplacement:Boolean,fraction:Double)* SCollection.sample]]. Default
-   *   is true.
-   */
+  @deprecated("Use skewedJoin with hotKeyMethod instead", since = "0.13.0")
   def skewedJoin[W](
     rhs: SCollection[(K, W)],
-    hotKeyThreshold: Long = SkewedJoins.DefaultHotKeyThreshold,
```
kellen, line 248: Won't default removal make this a breaking change rather than just a deprecation?

RustedBones: Yes, maybe it's better to simply break and provide a scalafix

RustedBones:
@kellen added scalafix rule

PR #4809, opened 2023-05-17
Add prefix & suffix param for all IO APIs

RustedBones:
This PR allows all file based IO to define either `prefix` + `shardnameTemplate` or `filenamePolicySupplier` to create files.

To make sure the `tap` returned by the write method is capturing written files, they must be adapted to match on the file `suffix`, only deterministic part of the filename.
To do so, all file based IO read are give a `suffix` param, with `null` as default with the following behavior:
- if set, the file pattern is constructed with `path` as a folder with `$path/*$suffix`
- if null, path will be used as before, either as a pattern or a file.
- if set and `path` s a pattern (contains a `*`), an exception is thrown


RustedBones:
ScioIO `tap` relies on files to be named with fixed `part` prefix.
IMHO `path` in the ScioIO must not be a pattern. (This is also a limitation in the SMB IOs).
Matcher should be constructed form `prefix` + `suffix` provided in the read param

scio-extra/src/main/scala/com/spotify/scio/extra/json/JsonIO.scala
```diff
@@ -67,25 +68,27 @@ final case class JsonIO[T: Encoder: Decoder: Coder](path: String) extends ScioIO
...
     private[scio] val DefaultPrinter = Printer.noSpaces
+    private[scio] val DefaultFilenamePolicySupplier = null
     private[scio] val DefaultShardNameTemplate: String = null
+    private[scio] val DefaultPrefix: String = null
     private[scio] val DefaultTempDirectory = null
-    private[scio] val DefaultFilenamePolicySupplier = null
   }
 
-  final case class WriteParam(
+  final case class WriteParam private (
```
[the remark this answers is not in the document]
RustedBones: This only restrict the `new` constructor, the `apply` stays public. I think it is for API evolution where we only would have to overload apply in the companion object.

scio-avro/src/main/scala/com/spotify/scio/avro/taps.scala
```diff
@@ -163,6 +166,6 @@ final case class AvroTaps(self: Taps) {
     val avroT = AvroType[T]
 
     import scala.concurrent.ExecutionContext.Implicits.global
-    avroGenericFile(path, avroT.schema, params).map(_.map(avroT.fromGenericRecord))
```
shnapz: should we add scalafix rules for `avroGenericFile` to `avroFile` transform?

RustedBones: You're not looking at the full diff. `avroGenericFile` was created but then reverted by using overload. `avroFile` was the API and is kept after the changes.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
```diff
@@ -290,33 +301,44 @@ final case class GenericRecordParseIO[T](path: String, parseFn: GenericRecord =>
...
     numShards: Int = WriteParam.DefaultNumShards,
-    private val _suffix: String = WriteParam.DefaultSuffix,
+    suffix: String = WriteParam.DefaultSuffix,
     codec: CodecFactory = WriteParam.DefaultCodec,
     metadata: Map[String, AnyRef] = WriteParam.DefaultMetadata,
+    filenamePolicySupplier: FilenamePolicySupplier = WriteParam.DefaultFilenamePolicySupplier,
+    prefix: String = WriteParam.DefaultPrefix,
     shardNameTemplate: String = WriteParam.DefaultShardNameTemplate,
-    tempDirectory: String = WriteParam.DefaultTempDirectory,
-    filenamePolicySupplier: FilenamePolicySupplier = WriteParam.DefaultFilenamePolicySupplier
```
RustedBones, line 315: @kellen Should I revert the param ordering in the `WriteParam` to minimize breaking changes ?

scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
```diff
@@ -91,45 +103,44 @@ final private[scio] class FileStorage(protected[scio] val path: String) {
...
-    } else if (nums.nonEmpty) {
-      // found xxxxx-of-yyyyy pattern
-      val parts = nums.map(_._1).sorted
-      val total = nums.map(_._2).toSet
-      metadata.size == nums.size && // all paths matched
-      total.size == 1 && total.head == parts.size && // yyyyy part
-      parts.head == 0 && parts.last + 1 == parts.size // xxxxx part
-    } else {
+    } else if (writtenShards.isEmpty) {
+      // assume progress is complete when shard info is not retrieved and files are present
```
kellen, line 135: Is this all we can do here?

scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
```diff
@@ -42,136 +44,141 @@ final case class TextIO(path: String) extends ScioIO[String] {
...
+
+    private[scio] def apply(params: WriteParam): ReadParam =
+      new ReadParam(
+        compression = params.compression,
+        suffix = params.suffix + params.compression.getSuggestedSuffix
+      )
   }
 
-  final case class ReadParam(
+  final case class ReadParam private (
```
kellen, line 128: General question: what do we gain by locking all these down?

scio-core/src/main/scala/com/spotify/scio/io/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -87,8 +86,14 @@ final class DynamicSpecificRecordSCollectionOps[T <: SpecificRecord](
...
       val write =
-        writeDynamic(path, numShards, suffix, destinationFn, tempDirectory)
-          .via(sink)
+        writeDynamic(
+          path = path,
+          destinationFn = destinationFn,
+          numShards = numShards,
+          prefix = prefix,
+          suffix = suffix,
+          tempDirectory = tempDirectory
```
kellen, line 95: Why the provide the parameter names explicitly here? Did you mix up prefix/suffix?

scio-test/src/test/scala/com/spotify/scio/io/TapsTest.scala
```diff
@@ -57,6 +52,7 @@ class TapsTest extends AnyFlatSpec with Matchers {
   it should "fail missing text file" in {
     TapsSysProps.Algorithm.value = "immediate"
     val f = tmpFile
+    Files.delete(f)
```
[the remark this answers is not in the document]
kellen: Creating a java `File` doesn't actually create the underlying file

site/src/main/paradox/FAQ.md
```diff
@@ -94,9 +94,10 @@ object MyJob {
   def main(cmdlineArgs: Array[String]): Unit = {
     val (sc, args) = ContextAndArgs(cmdlineArgs)
 
-    val collections =
-      Seq("gs://bucket1/data/*.avro", "gs://bucket2/data/*.avro")
-        .map(sc.avroFile[TestRecord](_))
+    val collections = Seq(
+      "gs://bucket1/data/",
+      "gs://bucket2/data/"
+    ).map(path => sc.avroFile[TestRecord](path, suffix=".avro"))
```
kellen, line 100: Is this the way we want to recommend users perform reads going forward? I marginally prefer the old version since it _seems_ more compact & explicit, but that might just be Stockholm syndrome

kellen (approved):

scio-core/src/main/scala/com/spotify/scio/io/TextIO.scala
kellen: Nevermind, seeing that Andrew's question was the same.

scio-avro/src/main/scala/com/spotify/scio/avro/AvroIO.scala
kellen: No, I don't think this is important.

scio-test/src/test/scala/com/spotify/scio/io/TapsTest.scala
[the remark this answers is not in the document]
RustedBones: `Files.createTempFile` does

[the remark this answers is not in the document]
RustedBones: `Files.createTempFile` does

scio-core/src/main/scala/com/spotify/scio/io/dynamic/syntax/SCollectionSyntax.scala
RustedBones: Yes, Since API is not too strongly types, using named parameter helped to make sure I did not mess up with parameter ordering

scio-core/src/main/scala/com/spotify/scio/io/FileStorage.scala
RustedBones: It was the previous behavior. when user give a custom naming policy/shard template, we do not have this info the the `ReadParameter, thus we can't find shard numbers easily.

Issue #3928, opened 2021-07-27
Support clustering in saveAs..BigQueryTable


dmateusp:
I am trying to write to a clustered table using `saveAsTypedBigQueryTable` and getting:

```
<table> is configured to use TableDestinationCoderV2. Set withClustering() on BigQueryIO.write() and,  if you provided a custom DynamicDestinations instance, override getDestinationCoder() to return TableDestinationCoderV3.
```

I only hit this when I am processing many partitions at once. I see that the SCIO API doesn't expose the withClustering option.

https://beam.apache.org/releases/javadoc/2.31.0/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.Write.html#withClustering-com.google.api.services.bigquery.model.Clustering-

laviandra:
+1

PR #4810, opened 2023-05-17
Support missing BQ write parameters

RustedBones:
Fix #4334 
Fix #3928

Part of #3332

PR #4813, opened 2023-05-19
feat: Support clustering in saveAs..BigQueryTable

amo12937:
refers: #3928

I'd like to use `saveAsTypedBigQueryTable` to insert data into a clustered table in my project.


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -291,6 +294,15 @@ case class TimePartitioning(
   }
 }
 
+/** Scala wrapper for [[com.google.api.services.bigquery.model.Clustering]]. */
+case class Clustering(
```
RustedBones, line 298: Probably worth to be `final`. I see this is also missing from the `TimePartitioning` where you copied the design.

RustedBones (commented):
Thanks a lot!
I'll probably group this one with #4810.

PR #4814, opened 2023-05-19
Mark Kryo coders nondeterminisitic

kellen:


scio-test/src/test/scala/com/spotify/scio/coders/KryoAtomicCoderTest.scala
```diff
@@ -134,6 +137,20 @@ class KryoAtomicCoderTest extends PipelineSpec {
     "@KryoRegistrar class FooKryoRegistrar extends Product {}" shouldNot compile
   }
 
+  it should "support BoundedWindow" in {
+    import org.apache.beam.sdk.transforms.windowing.BoundedWindow
+    case class TestBoundedWindow(
+      x: Int,
+      override val maxTimestamp: time.Instant = BoundedWindow.TIMESTAMP_MAX_VALUE
```
RustedBones, line 144: `org.joda.time.Instant` is already in the global import, your could have used `Instant` and remove the `import org.joda.time`
```suggestion
      override val maxTimestamp: Instant = BoundedWindow.TIMESTAMP_MAX_VALUE
```

RustedBones (approved):

clairemcginty (approved):

PR #4815, opened 2023-05-19
Add a better worded exception of unsupported SpecificRecord coder case

shnapz:


scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
```diff
@@ -128,11 +128,26 @@ trait AvroCoders {
...
+      clazz.getDeclaredConstructor().newInstance().getSchema
+    } catch {
+      // when T is abstract class or interface it is not possible to retrieve a schema
+      case cause: Throwable =>
+        val msg =
+          "Failed to create a coder for SpecificRecord because it is impossible to retrieve an " +
+            s"Avro schema by instantiating $clazz. Use only a concrete type implementing " +
+            s"SpecificRecord or use GenericRecord type in your transformations if a concrete " +
+            s"type is not known in compile time."
+        throw new Throwable(msg, cause)
```
RustedBones, line 142: I think this isn't worth chaining the exception here, as the one raised by 
```scala
SpecificData.get().getSchema(clazz)
```
is already suppressed. We're already in a fallback/recover case.

scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
```diff
@@ -128,11 +128,26 @@ trait AvroCoders {
...
+    }
+
     // Try to get the schema with SpecificData.getSchema
     // This relies on private SCHEMA$ field that may not be defined on custom SpecificRecord instance
-    // Otherwise create a default instance and call getSchema
     val schema = Try(SpecificData.get().getSchema(clazz))
-      .getOrElse(clazz.getDeclaredConstructor().newInstance().getSchema)
+      // Otherwise create a default instance and call getSchema
+      .getOrElse(getSchemaFromEmptyInstance)
```
RustedBones, line 149: `orElse` will make this more idiomatic IMHO
```suggestion
      .orElse(Try(clazz.getDeclaredConstructor().newInstance().getSchema))
      .getOrElse {
        val msg = ...
        throw new RuntimeException(msg)
      }
```

scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
shnapz: - my intuition is that this extra info won't hurt
- this would make clear if failure happens for anything other than "constructor not found", like `getSchema` might fail because of some bug in apache avro

RustedBones (approved):

PR #4816, opened 2023-05-23
Added a custom Kryo serializer for Avro schema

shnapz:
In Java17 kryo fails to serialize Avro schema with:
```
Caused by: java.lang.IllegalAccessException: class com.twitter.chill.Instantiators$ cannot access a member of class org.apache.avro.Schema$StringSchema with modifiers "public"
```
This affects our users. Instead of enforcing them to avoid serialization of Avro schema, this easy fix allows them to migrate to java 17

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
@@ -20,7 +20,6 @@ package com.spotify.scio.coders
 import java.io.{EOFException, InputStream, OutputStream}
 import java.nio.file.Path
 import java.util.concurrent.atomic.AtomicInteger
-
 import com.esotericsoftware.kryo.KryoException
 import com.esotericsoftware.kryo.io.{InputChunked, OutputChunked}
 import com.esotericsoftware.kryo.serializers.JavaSerializer
@@ -30,6 +29,7 @@ import com.spotify.scio.options.ScioOptions
 import com.twitter.chill._
 import com.twitter.chill.algebird.AlgebirdRegistrar
 import com.twitter.chill.protobuf.ProtobufSerializer
+import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.specific.SpecificRecordBase
 import org.apache.beam.sdk.coders.{AtomicCoder, CoderException => BCoderException, InstantCoder}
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/instances/kryo/SchemaSerializer.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package com.spotify.scio.coders.instances.kryo
+
+import com.esotericsoftware.kryo.Kryo
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.twitter.chill.KSerializer
+import org.apache.avro.Schema
+
+private[coders] class SchemaSerializer extends KSerializer[Schema] {
+  private lazy val parser = new Schema.Parser()
```
clairemcginty, line 26: I think we can mark this as `@transient` so it's not encoded as part of a `Coder[Schema]`

scio-core/src/main/scala/com/spotify/scio/coders/KryoAtomicCoder.scala
```diff
@@ -123,6 +123,7 @@ final private class ScioKryoRegistrar extends IKryoRegistrar {
     k.forClass[LocalTime](new JodaLocalTimeSerializer)
     k.forClass[LocalDateTime](new JodaLocalDateTimeSerializer)
     k.forClass[DateTime](new JodaDateTimeSerializer)
+    k.forSubclass[Schema](new SchemaSerializer)
```
clairemcginty, line 126: To match the naming convention for other Kryo serializers, maybe let's call this `AvroSchemaSerializer`?

clairemcginty, line 126: One thing to keep in mind is that starting in Avro 1.10, `Schema`s are marked as `Serializable` [in Avro](https://github.com/apache/avro/blob/release-1.10.0/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L87), so (theoretically!) we shouldn't need to do a Kryo override if Avro 1.10+ is used.

We could check for the runtime Avro version, like:

```scala
Try(
  Schema.Parser.class.getPackage().getImplementationVersion()
).foreach { avroVersion =>
  // This could be more elegant, maybe a regex match?
  if (
    avroVersion.startsWith("1.7.") ||
    avroVersion.startsWith("1.8.") ||
    avroVersion.startsWith("1.9.")
  ) {
     k.forSubclass[Schema](new SchemaSerializer)
  }
}
```

@RustedBones, wdyt?

PR #4876, opened 2023-06-13
Fix dependency issues

RustedBones:
- ship direct runner for testing
- move gcs-connector to examples runtime dependency
- fix google api services storage version not in sync with google library_bom

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1184,6 +1184,7 @@ lazy val `scio-examples`: Project = project
       "org.scala-lang.modules" %% "scala-collection-compat" % scalaCollectionCompatVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       // runtime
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -1182,6 +1183,7 @@ lazy val `scio-examples`: Project = project
       "org.scala-lang.modules" %% "scala-collection-compat" % scalaCollectionCompatVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       // runtime
+      "com.google.cloud.bigdataoss" % "gcs-connector" % s"hadoop2-$bigdataossVersion",
```
RustedBones, line 1186: @clairemcginty this used to be shipped with `scio-parquet`. I feel we should stay agnostic to the storage system and this must be manually added by users if needed now.

RustedBones:
See https://github.com/apache/beam/pull/26910

PR #4879, opened 2023-06-13
Group grpc transforms

RustedBones:
Group grpc transform operations so they are nested inside a single block


RustedBones:
Only `snyk` check failing but there was no dependency changes.

clairemcginty (approved):

PR #4880, opened 2023-06-14
[Proposal] Make scio-smb's ADJUSTMENT_FACTOR a PipelineOption

clairemcginty:
Gives users an extra tuning knob for SMB reads. Different inputs types may benefit from more or less source splits (I.e., Parquet reads may run into memory issues with too few splits). By parameterizing it as a pipeline option, users can tweak the source splits to suit their use case.


kellen (approved):
Fix the magic number and 👍

kellen:
Closing per offline discussion

PR #4883, opened 2023-06-14
Move to avro extensions

kellen:


kellen:
@RustedBones do we want this in 0.13?

scio-core/src/main/scala/com/spotify/scio/io/dynamic/syntax/SCollectionSyntax.scala
```diff
@@ -29,6 +29,7 @@ import org.apache.avro.specific.SpecificRecord
 import org.apache.beam.sdk.coders.StringUtf8Coder
 import org.apache.beam.sdk.io.{Compression, FileIO}
 import org.apache.beam.sdk.{io => beam}
+import org.apache.beam.sdk.extensions.avro.{io => avroio}
```
RustedBones, line 32: probably nicer to have `{AvroIO => BAvroIO}`

RustedBones (approved):

RustedBones:
I think the simple move is fine for 0.13. There should not be much code change and we'll get rid of all the deprecation warnings internally.

PR #4886, opened 2023-06-15
Union coder workaround for flink

RustedBones:


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -752,12 +752,19 @@ class ScioContext private[scio] (
   // `T: Coder` context bound is required since `scs` might be empty.
   def unionAll[T: Coder](scs: => Iterable[SCollection[T]]): SCollection[T] = {
     val tfName = this.tfName // evaluate eagerly to avoid overriding `scs` names
-    scs match {
-      case Nil => empty()
```
RustedBones, line 756: This was failing for non-list input

PR #4887, opened 2023-06-15
Safe coder unwrap

RustedBones:
Unwrapping was don on `Ref` and `LazyCoder`. This was required to extract values from tuple coders when `RecordCoder` was used (materialization wraps into a `Ref`/`LazyCoder`)

IMHO, we should expect `TupleCoder` in those cases and fail if a `RecordCoder` was given.

We could then only unwrap `MaterlializedCoder` and `NullableCoder` (double check the option is enabled). implicit scio coders will get stable.

Probably a better fix than #4886

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
@@ -17,6 +17,7 @@
 
 package com.spotify.scio.coders
 
+import com.spotify.scio.coders.CoderMaterializer.CoderOptions
 import com.spotify.scio.coders.{instances => scio}
 import com.spotify.scio.values.{SCollection, SideInput}
 import org.apache.beam.sdk.{coders => beam}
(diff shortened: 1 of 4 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
```diff
@@ -27,100 +28,109 @@ import scala.annotation.tailrec
...
+    val options = CoderOptions(coll.context.options)
     val coder = coll.internal.getCoder
-    val (k, v) = unwrap(coder) match {
+    val (k, v) = unwrap(options, coder) match {
       case c: scio.Tuple2Coder[K, V] =>
         (c.ac, c.bc)
-      case c: RecordCoder[(K, V)] =>
-        val ac = coderElement[K](c)(0)
-        val bc = coderElement[V](c)(1)
-        (ac, bc)
```
RustedBones, line 56: We'll not support users who force non tuple coders for `(K, V)`

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
```diff
@@ -27,100 +28,109 @@ import scala.annotation.tailrec
...
 private[scio] object BeamCoders {
   @tailrec
-  private def unwrap[T](coder: beam.Coder[T]): beam.Coder[T] =
+  private def unwrap[T](options: CoderOptions, coder: beam.Coder[T]): beam.Coder[T] =
     coder match {
-      case c: WrappedCoder[T]       => unwrap(c.bcoder)
-      case c: beam.NullableCoder[T] => c.getValueCoder
-      case _                        => coder
+      case c: MaterializedCoder[T]                            => unwrap(options, c.bcoder)
+      case c: beam.NullableCoder[T] if options.nullableCoders => c.getValueCoder
```
clairemcginty, line 34: what will happen in the case that a user doesn't have the `--nullableCoders` option set, but still uses NullableCoder in their pipeline? (i.e. they have a step like `.map(x => ...)(Coder.beam(NullableCoder.of(...))`

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
RustedBones: changed to accept any `StructuredCoder` of size 2

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
```diff
@@ -17,110 +17,91 @@
...
-      case c: scio.BaseSeqLikeCoder[Iterable, V] @unchecked => c.elemCoder
-      case _ =>
-        throw new IllegalArgumentException(
-          s"Failed to extract value coder from Coder[Iterable[V]]: $coder"
-        )
-    }
-
-  /** Get key-value coders from a `SideInput[Map[K, Iterable[V]]]`. */
-  def getMultiMapKV[K, V](si: SideInput[Map[K, Iterable[V]]]): (Coder[K], Coder[V]) = {
-    val coder = si.view.getPCollection.getCoder
```
RustedBones, line 113: This was very unsafe. The `PCollection` behind a view has no guarantee to be of the same type (int contains a transform function). Drop this helper and expect a coder when we pass a `SideInput`

scio-test/src/test/scala/com/spotify/scio/values/PairHashSCollectionFunctionsTest.scala
```diff
@@ -77,9 +77,22 @@ class PairHashSCollectionFunctionsTest extends PipelineSpec {
...
+    }
+  }
+
+  it should "support hashJoin() with transformed .asMultiMapSideInput" in {
+    runWithContext { sc =>
+      val p1 = sc.parallelize(Seq(("a", 1), ("a", 2), ("b", 3)))
+      val p2 = sc
+        .parallelize(Seq(("a", "11"), ("b", "12"), ("b", "13")))
+        .asMultiMapSideInput
+        .map(_.map { case (k, vs) => k -> vs.map(_.toInt) })
```
RustedBones, line 91: This transformation was failing, because we tried to use a `Coder[String]` to encode the produced `Int` value

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
RustedBones: The null coder will be propagated. This is required by `countByValue` for example, otherwise we would fail on `null` values.

clairemcginty (approved):

PR #4913, opened 2023-07-07
Remove avro from scio-core, implement binary file source

kellen:
Removes avro as a dependency from `scio-core`, moves all avro behaviors to `scio-avro`.

Implements a `sc.binaryFile` method taking a `BinaryFileReader` for parsing the specific binary file format.

Reimplements `materialize` with the existing `saveAsBinaryFile` and new `binaryFile` and the existing strategy of using Coder-encoded bytes. As a consequence, files written by materialize are no longer splittable.


shnapz:
Target `0.14`?

scio-avro/src/main/scala/com/spotify/scio/avro/schemas/package.scala
```diff
@@ -0,0 +1,20 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.avro
+
+package object schemas extends instances.AvroInstances
```
shnapz, line 20: How this impacts the typical usage?

kellen: I don't think there's really any users of this schemas package to speak of.

But the change would be if you are using something from this package, you'd need to add:
```
import com.spotify.scio.avro.schemas._
```
In addition to the existing:
```
```

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/SCollectionSyntax.scala
```diff
@@ -96,7 +96,7 @@ final class ObjectFileSCollectionOps[T](private val self: SCollection[T]) extend
       shardNameTemplate,
       tempDirectory
     )
-    self.write(ObjectFileIO(path))(param)
+    self.write(ObjectFileIO(path)(coder))(param)
```
RustedBones, line 99: wasn't the coder taken implicitly ?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/BinaryInOut.scala
```diff
@@ -0,0 +1,115 @@
...
+      .debug()
+    sc2.run().waitUntilDone()
+  }
+
+  // This BinaryFileReader subclass must match the behavior of the write.
+  case object MyBinaryFileReader extends BinaryFileReader {
+    // This example has no state, but a record header could for example contain a number of
+    // expected records in the entire file or some block, and the state could then be used track
+    // the number of records read and to determine when the reader would need to switch modes from
+    // reading records to reading block metadata or the file footer.
```
RustedBones, line 65: Thanks for those explanations!

scio-avro/src/main/scala/com/spotify/scio/avro/dynamic/syntax/AvroSyntax.scala
```diff
@@ -0,0 +1,195 @@
...
+      ).via(sink)
+
+      self.applyInternal(write)
+    }
+
+    ClosedTap[Nothing](EmptyTap)
+  }
+}
+
+trait AvroSyntax {
```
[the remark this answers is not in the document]
kellen: I do not think any of the AvroSCollectionSyntax maybe. IMO having the repeated names is bad.

scio-avro/src/main/scala/com/spotify/scio/avro/dynamic/syntax/AvroSyntax.scala
```diff
@@ -0,0 +1,195 @@
...
+
+    ClosedTap[Nothing](EmptyTap)
+  }
+}
+
+/**
+ * Enhanced version of [[com.spotify.scio.values.SCollection SCollection]] with dynamic destinations
+ * methods.
+ */
+final class DynamicGenericRecordSCollectionOps[T <: GenericRecord](private val self: SCollection[T])
```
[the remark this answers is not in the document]
kellen: Agreed

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/SCollectionSyntax.scala
kellen: Thanks for the reminder to look into this. `ObjectFileIO` was missing its type param and getting confused by the new coders in scope from the `avro._` import. Fixed.

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
```diff
@@ -42,14 +46,36 @@ import scala.util.Try
...
+    val coder = ByteArrayCoder.of()
+    val srcFn = Functions.serializableFn { path: String =>
+      new BinaryIO.BinarySource(path, params.emptyMatchTreatment, params.reader)
+    }
+
+    sc.withName("Create filepattern")
+      .parallelize(List(filePattern))
+      .applyTransform("Match All", FileIO.matchAll())
+      .applyTransform(
+        "Read Matches",
```
clairemcginty, line 68: do we need to set transform names here? I'd think that Beam would set them under the hood for us.

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
```diff
@@ -206,4 +232,114 @@ object BinaryIO {
...
+  final private[scio] class BinarySource(
+    path: String,
+    emptyMatchTreatment: EmptyMatchTreatment,
+    binaryFileReader: BinaryFileReader
+  ) extends FileBasedSource[Array[Byte]](
+        StaticValueProvider.of(path),
+        emptyMatchTreatment,
+        Long.MaxValue
+      ) {
+    override def isSplittable: Boolean = false
```
clairemcginty, line 332: wouldn't this be `true` because this source implements file subranges?

scio-examples/src/main/scala/com/spotify/scio/examples/extra/BinaryInOut.scala
```diff
@@ -0,0 +1,115 @@
...
+import java.nio.ByteBuffer
+import com.spotify.scio.io.BinaryIO.BinaryFileReader
+import java.io.InputStream
+
+// Example: Binary Input and Output
+// Usage:
+// `sbt "runMain com.spotify.scio.examples.extra.BinaryInOut
+// --project=[PROJECT] --runner=DataflowRunner --region=[REGION NAME] --output=[OUTPUT]"`
+
+object BinaryInOut {
```
clairemcginty, line 30: great example!

Can we also add a test for this? I feel like it's common for users to be confused on how to use `JobTest` with new IOs so it would be great if we had an example test to point them to.

scio-avro/src/main/scala/com/spotify/scio/avro/io/AvroFileStorage.scala
```diff
@@ -0,0 +1,80 @@
...
+}
+
+final private[scio] class AvroFileStorage(path: String, suffix: String) {
+
+  private def getAvroSeekableInput(meta: Metadata): SeekableInput =
+    new SeekableInput {
+      require(meta.isReadSeekEfficient)
+      private val in = {
+        val channel = FileSystems.open(meta.resourceId()).asInstanceOf[SeekableByteChannel]
+        // metadata is lazy loaded on GCS FS and only triggered upon first read
```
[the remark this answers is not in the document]
kellen: 🤷 This code is just moved from `FileStorage`

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
kellen: This replicates the internal usages in beam. There does not appear to be any naming applied from within the transform.

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
kellen: No, the subrange API is called for the entire file, then `isSplittable` is checked to determine if further splits should occur. 

The actual format we write with `saveAsBinaryFile` is not splittable since we have no record count, no index, and no blocks of records.

scio-examples/src/main/scala/com/spotify/scio/examples/extra/BinaryInOut.scala
kellen: I don't think this would be useful; the logic is encapsulated in the reader, so all the test would do is assert that some array values are written and read back, but it wouldn't capture (or test) any of the file parsing or formatting.

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
clairemcginty: ah okay. It's funny that we don't need to implement `createSingleFileReader` then -- I would assume that the runner would invoke `createSingleFileReader` if `iSplittable == false` and `createForSubrangeOfFile` otherwise

clairemcginty (approved):

scio-core/src/main/scala/com/spotify/scio/io/BinaryIO.scala
kellen: It's also because in javaland there's a single class with two constructors; one for the path-created variant and one for the file-variant. Here it's split into `BinarySource` and `BinarySingleFileSource`

RustedBones (commented):
can you also change the `scio-repl` packaging to favor avro classes from the `avro-extension` module instead of the one from the java runner ?

PR #4932, opened 2023-07-21
Added abstract JdbcDoFn to execute lookup

shnapz:
More convenient way to implement lookup to Jdbc source. What can be added more:
syntax sugar for SCollection to integrate lookup lambda?

### DIFF
--- a/scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
+++ b/scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
@@ -72,9 +72,7 @@ public abstract class JdbcDoFn<A, B> extends DoFn<A, B> {
   }
 
   @ProcessElement
-  public void processElement(
-      @Element A input,
-      OutputReceiver<B> out) {
+  public void processElement(@Element A input, OutputReceiver<B> out) {
     B result = lookup(connection, input);
     out.output(result);
   }
--- a/scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
+++ b/scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
@@ -1,0 +1,66 @@
+package com.spotify.scio.jdbc;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import javax.sql.DataSource;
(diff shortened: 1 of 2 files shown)


scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * A {@link DoFn} that performs synchronous lookup using JDBC connection.
+ *
+ * @param <A> input element type.
+ * @param <B> JDBC lookup value type.
+ */
+public abstract class JdbcDoFn<A, B> extends DoFn<A, B> {
+
+  private final SerializableFunction<Void, DataSource> dataSourceProviderFn;
```
RustedBones, line 34: I think `Supplier` is more adapted here

scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
```diff
@@ -0,0 +1,81 @@
...
+  private DataSource dataSource;
+  private Connection connection;
```
RustedBones, line 36: Looks those are interfaces without serialization guarantees. Probably worth to mark them as transient.

scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
shnapz: Neither of available `Supplier` interfaces are serializable

scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
shnapz: Done, also added unit test to verify serializable

scio-jdbc/src/main/java/com/spotify/scio/jdbc/JdbcDoFn.java
RustedBones: ha ok. Maybe we can simply add a 'nicer' API with the Supplier as constructor parameter, and convert it to the serializable function.

RustedBones (approved):

PR #4934, opened 2023-07-24
Fix compiler warnings

RustedBones:
Fix compiler warnings
- remove deprecated compiler option and accept non unit statements
- remove unused method parameters (breaking changes for v0.14)
- silence required unused parameter (eg `IsJavaBean` evidence)
- set explicit types on implicits
- fix implicit type conversion (mostly widening)
- add `@uncheked` annotations in pattern matching when required
- avoid usage of slow avro generic coder (provide schema)

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
@@ -22,7 +22,7 @@ import com.spotify.scio.coders.Coder
 import com.google.common.{hash => g}
 import org.apache.beam.sdk.coders.CustomCoder
 
-class GuavaBloomFilterCoder[T](implicit val funnel: g.Funnel[T])
+class GuavaBloomFilterCoder[T](implicit val funnel: g.Funnel[_ >: T])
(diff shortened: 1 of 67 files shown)


kellen (approved):

kellen:
Seems to have broken some tests ;)

RustedBones:
Yes. will unset draft as soon as those are fixed.

scio-core/src/main/scala/com/spotify/scio/coders/instances/GuavaCoders.scala
```diff
@@ -22,7 +22,7 @@ import com.spotify.scio.coders.Coder
 import com.google.common.{hash => g}
 import org.apache.beam.sdk.coders.CustomCoder
 
-class GuavaBloomFilterCoder[T](implicit val funnel: g.Funnel[T])
+class GuavaBloomFilterCoder[T](implicit val funnel: g.Funnel[_ >: T])
```
RustedBones, line 25: `BLoomFilter[String]` can be created from a `Funnel[CharSequence]` (expected funnel is `<? extends T>`).
Associated coder should also be able to be constructed with such funnel

kellen (approved):
love it

PR #4937, opened 2023-07-26
Fix race in sparkey write

### LINKED ISSUE #4890: Race in sparkey write
The sparkey side inputs write directly to their final destination so will fail in the face of retries, this should be rewritten to have a multiphase 'commit' ala WriteFiles, though I do not think that `FileBasedSink` will work here so we will need to reimplement some functionality

kellen:
Intended to fix #4890

Introduces a write to a temporary location before a finalization step moving temp files to their final destination.

Unifies the various SparkeyUri subclasses for simpler handling, removes kryo coder, moves provision of RemoteFileUtil to transform steps.


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -19,104 +19,138 @@ package com.spotify.scio.extra.sparkey
...
+  private[sparkey] val isLocal = ScioUtil.isLocalUri(new URI(path))
+  private val isSharded = path.endsWith("*")
+  private[sparkey] val basePath =
+    if (!isSharded) path else path.split("/").dropRight(1).mkString("/")
 
-private object RemoteSparkeyUri {
-  def apply(basePath: String, options: PipelineOptions): RemoteSparkeyUri =
-    RemoteSparkeyUri(basePath, RemoteFileUtil.create(options))
-}
+  private[sparkey] def globExpression: String = s"$basePath/part-*"
```
RustedBones, line 47: I think this helper is a bit risky since it should only be used when `isSharded` is `false`

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -19,104 +19,138 @@ package com.spotify.scio.extra.sparkey
...
-    val uris = SparkeyUri.extensions.map(e => new URI(basePath + e))
-    val paths = rfu.download(uris.asJava).asScala
-    new ThreadLocalSparkeyReader(paths.head.toFile)
+  private[sparkey] def sparkeyUriForShard(shard: Short, numShards: Short): SparkeyUri =
+    SparkeyUri(SparkeyUri.basePathForShard(basePath, shard, numShards))
+
+  private[sparkey] def paths: Seq[ResourceId] = {
+    if (isSharded) throw new IllegalStateException("Internal use only for single files.")
+    SparkeyUri.extensions.map { e =>
+      val url = if (isLocal) new File(basePath + e).toString else new URI(basePath + e).toString
```
RustedBones, line 55: Why are we creating a `File` or `URI` and converting it straight to string ? Won't this get back the original string given in the constructor ?

RustedBones: also, can't we use URI for all cases (local & remote) ?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -19,104 +19,138 @@ package com.spotify.scio.extra.sparkey
...
-          val dst = new URI(u.basePath + e)
-          u.rfu.upload(src, dst)
-        }
-      case _ => ()
+private[sparkey] object ShardedSparkeyUri {
+  private[sparkey] def numShardsFromPath(path: String): Short =
+    path.split("-of-").toList.last.split("\\.").head.toShort
+
+  private[sparkey] def shardIndexFromPath(path: String): Short =
+    path.split("part-").toList.last.split("-of-").head.toShort
```
RustedBones, line 112: I think a regex would be more readable to extract those to in a single pass

RustedBones (commented):
shouldn't this target 0.14 as this is breaking mima ?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -19,104 +19,141 @@ package com.spotify.scio.extra.sparkey
...
-import org.apache.beam.sdk.options.PipelineOptions
+import com.spotify.sparkey.SparkeyReader
+import org.apache.beam.sdk.io.FileSystems
+import org.apache.beam.sdk.io.fs.{EmptyMatchTreatment, MatchResult, ResourceId}
 
+import java.nio.file.Path
+import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
+case class InvalidShards(str: String) extends RuntimeException(str)
```
clairemcginty, line 33: Could we make this more specifically named? like `InvalidNumShardsException` ?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyUri.scala
```diff
@@ -19,104 +19,141 @@ package com.spotify.scio.extra.sparkey
...
-    SparkeyUri.extensions.map(e => new File(basePath + e)).exists(_.exists)
-}
+case class SparkeyUri(path: String) {
+  private[sparkey] val isLocal = ScioUtil.isLocalUri(new URI(path))
+  private val isSharded = path.endsWith("*")
+  private[sparkey] val basePath =
+    if (!isSharded) path else path.split("/").dropRight(1).mkString("/")
+
+  private[sparkey] def globExpression: String = {
+    if (!isSharded) throw new IllegalArgumentException("Internal use only for sharded sparkeys.")
```
clairemcginty, line 53:
```suggestion
    assert(!isSharded, "Sharded Sparkey API does not support wildcard lookups")
```

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/SparkeyWriter.scala
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.file.{Files, Paths}
+
+private[sparkey] class SparkeyWriter(
+  val uri: SparkeyUri,
+  rfu: RemoteFileUtil,
+  compressionType: CompressionType,
+  compressionBlockSize: Int,
+  maxMemoryUsage: Long = -1
+) {
+  private val localFile =
```
clairemcginty, line 35: can we make this lazy so it's not executed until the pipeline is run? I'm just concerned as-is it wouldn't work with PTransformOverride, for example

PR #4944, opened 2023-07-31
Tensorflow metadata

RustedBones:
- clean protoc configuration
- download tensorflow metadata from released artifact
- package tensorflow medatada with `scio-tensorflow`
- depend on `scio-tensorflow` as provided in `scio-parquet`
- use tfmd (tensorflow metadata) schema instead of `me.lyh:parquet-tensorflow`
- create tensorflow example-parquet support based on official avro-parquet implementation


build.sbt
```diff
@@ -1105,7 +1105,18 @@ lazy val `scio-tensorflow`: Project = project
...
       "org.slf4j" % "log4j-over-slf4j" % slf4jVersion % Test,
       "org.slf4j" % "slf4j-simple" % slf4jVersion % Test
-    )
+    ),
+    Compile / PB.protoSources += target.value / s"metadata-$tensorFlowMetadataVersion",
+    Compile / PB.unpackDependencies := {
+      val tfMetadata = new URL(
+        s"https://github.com/tensorflow/metadata/archive/refs/tags/v$tensorFlowMetadataVersion.zip"
+      )
+      val protoFiles = IO.unzipURL(tfMetadata, target.value, "*.proto")
```
clairemcginty, line 1114: do we want to future-proof this a little by adding the TF version to the file filter (i.e. `/**/v0/*.proto`)?

scio-parquet/src/main/java/com/spotify/scio/parquet/tensorflow/ExampleWriteSupport.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.parquet.io.api.RecordConsumer;
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.schema.Type;
+import org.tensorflow.metadata.v0.Feature;
+import org.tensorflow.metadata.v0.FeatureType;
+import org.tensorflow.metadata.v0.Schema;
+import org.tensorflow.proto.example.Example;
+
+public class ExampleWriteSupport extends WriteSupport<Example> {
```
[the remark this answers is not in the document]
RustedBones: Added and it works!
It is actually a good thing that the original schema was not stored in the extraMetadata as reusing the key would have created parsing issue (users would have to provide the schema as projection).

RustedBones:
There is something in the official [documentation](https://www.tensorflow.org/api_docs/python/tf/train/Example):
> - If a Feature K doesn't exist in an example, a K-specific default will be used, if configured.
> - If a Feature K exists in an example but contains no items, the intent is considered to be an empty tensor and no default will be used.

Right now, parquet storage doesn't do distinction between feature without items and omitted feature.

scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -74,12 +75,11 @@ final case class ParquetExampleIO(path: String) extends ScioIO[Example] {
     conf: Configuration,
     params: ReadP
   ): SCollection[Example] = {
-    val job = Job.getInstance(conf)
     val filePattern = ScioUtil.filePattern(path, params.suffix)
 
     Option(params.projection).foreach { projection =>
-      ExampleParquetInputFormat.setFields(job, projection.asJava)
```
RustedBones, line 81: This seems to be useless as the `job` is not used in the sdf implementation

build.sbt
clairemcginty: sounds good 👍

clairemcginty (approved):

PR #4946, opened 2023-08-02
Allow tuple coder extraction from wrapped RecordCoder

### LINKED ISSUE #4945: Pre-allocated pair Coders breaks SCollection map api?
Pre-allocated pair coders seem to break (part of?) the map api's of SCollections.
This was working before 0.13 (found during an upgrade from 0.12.8 to 0.13.1).
Example:
```scala
class Example extends AnyFlatSpec with Matchers with PipelineTestUtils {
  it should "work" in {
    val result = runWithData(List("a", "b", "c")) { data =>
      data.map(x => x -> x).sumByKey(_ + _)
    }
    result must contain allElementsOf(List("a" -> "a", "b" -> "b", "c" -> "c"))
  }

  it should "not work" in {
    implicit val coder: Coder[(String, String)] = Coder.gen
    val result = runWithData(List("a", "b", "c")) { data =>
      // Throws IllegalArgumentException: Failed to extract key-value coders from Coder[(K, V)]: RecordCoder[scala.Tuple2](_1 -> StringUtf8Coder, _2 -> StringUtf8Coder)
      data.map(x => x -> x).sumByKey(_ + _)
    }
    result must contain allElementsOf (List("a" -> "a", "b" -> "b", "c" -> "c"))
  }
}
```

RustedBones:
fix #4945 

`SCollection[(K, V)]` should work with both `Tuple2Coder` (prefered) and `RecordCoder`.

Unfortunately, when operations requiring to split the coder, the extraction failed for `RecordCoder` due to improper coder definitions:
- `WrappedCoder` should not mask the underlying coder arguments
- `RecordCoder` should be a `StructuredCoder` and return the list of components.

Modify the tuple unwrapping so that we access the underlying coder arguments and pattern macth on those.
This will work for tuple, record and potentially for user custom  coder if well defined.

This PR also adds more test for `BeamCoders`

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
@@ -19,7 +19,7 @@ package com.spotify.scio.coders
 
 import com.spotify.scio.coders.CoderMaterializer.CoderOptions
 import com.spotify.scio.values.SCollection
-import org.apache.beam.sdk.coders.{Coder => BCoder, NullableCoder, StructuredCoder}
+import org.apache.beam.sdk.coders.{Coder => BCoder, NullableCoder}
 import org.apache.beam.sdk.values.PCollection
 
 import scala.annotation.tailrec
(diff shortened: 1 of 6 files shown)


clairemcginty (approved):
lgtm! Thanks for adding the new test suite too

PR #4951, opened 2023-08-11
Don't re-use ReflectData in AvroCoder

clairemcginty:
Addresses concurrency issue with encoding Avro records at scale -- we suspect the root cause is concurrent access of the shared ReflectData singleton.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
@@ -22,7 +22,7 @@ import com.spotify.scio.util.ScioUtil
 import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.io.{DatumReader, DatumWriter}
-import org.apache.avro.reflect.{ReflectDatumReader, ReflectDatumWriter}
+import org.apache.avro.reflect.{ReflectData, ReflectDatumReader, ReflectDatumWriter}
 import org.apache.avro.specific.{SpecificData, SpecificFixed, SpecificRecord}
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
```diff
@@ -153,7 +153,8 @@ trait AvroCoders {
       override def apply(writer: Schema, reader: Schema): DatumReader[T] = {
         // create the datum writer using the schema api
         // class API might be unsafe. See schemaForClass
-        val datumReader = new ReflectDatumReader[T](schemaForClass(clazz).get);
+        val schema = schemaForClass(clazz).get
+        val datumReader = new ReflectDatumReader[T](schema, schema, new ReflectData())
```
RustedBones, line 157:
```suggestion
        val datumReader = new ReflectDatumReader[T](writer, reader, new ReflectData())
```
And remove `setExpected` and `setSchema`

scio-core/src/main/scala/com/spotify/scio/coders/instances/AvroCoders.scala
```diff
@@ -164,7 +165,7 @@ trait AvroCoders {
       override def apply(writer: Schema): DatumWriter[T] = {
         // create the datum writer using the schema api
         // class API might be unsafe. See schemaForClass
-        val datumWriter = new ReflectDatumWriter[T](schemaForClass(clazz).get)
+        val datumWriter = new ReflectDatumWriter[T](schemaForClass(clazz).get, new ReflectData())
```
RustedBones, line 168:
```suggestion
        val datumWriter = new ReflectDatumWriter[T](writer, new ReflectData())
```
and remove `setSchema`

RustedBones (approved):

PR #4962, opened 2023-08-22
improve testing framework

### LINKED ISSUE #4963: TransformOverride runtime typeValidation in wrong order
The runtime [typeValidation](https://github.com/spotify/scio/blob/main/scio-test/src/main/scala/com/spotify/scio/testing/TransformOverride.scala#L107) verifies `expected instanceof actual" instead of "actual instanceof expected", causing the operation to only work if the types are exactly the same.

Example
```scala
import com.spotify.scio.testing.{PipelineSpec, TransformOverride}

import scala.reflect.ClassTag

class ReplaceTest extends PipelineSpec {
  def string(s: SCollection[String]): SCollection[String] = {
    s
      .map(List(_))
      .transform("transform") { c =>
        c.map(x => f"prefix-${x}")
      }
  }

  it should "run without replacement" in {
    runWithContext { sc =>
      val actual = string(sc.parallelize(Seq("a", "b")))
      actual should containInAnyOrder(Seq("prefix-List(a)", "prefix-List(b)"))
    }
  }

  it should "run with replacement currently does not work" in {
    runWithContext { sc =>
      val actual = string(sc.parallelize(Seq("a", "b")))
      actual should containInAnyOrder(Seq("override-List(a)", "override-List(b)"))
      sc.pipeline.replaceAll(
        List(TransformOverride.of("transform", (x: List[String]) => f"override-${x}")).asJava
      )
    }
  }

  it should "run with replacement works if the type tag is overriden incorrectly" in {
    runWithContext { sc =>
      val actual = string(sc.parallelize(Seq("a", "b")))
      actual should containInAnyOrder(Seq("override-List(a)", "override-List(b)"))

      val actualRuntimeClass: ClassTag[List[String]] = ClassTag(Class.forName("scala.collection.immutable.$colon$colon"))
      sc.pipeline.replaceAll(
        List(TransformOverride.of("transform", (x: List[String]) => f"override-${x}")(actualRuntimeClass)).asJava
      )
    }
  }
}
```

RustedBones:
Fix #4963 

- [x] Add `runWitnOverrides` to run pipeline unit-test, replacing some parts
- [x] Add `JobTest` io support for SMB
- [x] Add `JobTest` API for anonymous jobs (defined from context)

### DIFF
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
@@ -1,0 +1,8 @@
+package com.spotify.scio.smb
+
+import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
+
+final case class SMBIO[K, T](id: String, keyBy: T => K) extends KeyedIO[K, T] with TestIO[(K, T)] {
+  override val tapT: TapT.Aux[(K, T), (K, T)] = TapOf[(K, T)]
+  override def testId: String = s"SMBIO($id)"
+}
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2023 Spotify AB
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.spotify.scio.smb
 
 import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
@@ -16,9 +16,21 @@
 
 package com.spotify.scio.smb
 
+import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
+import org.apache.beam.sdk.extensions.smb.SortedBucketIO
 
-final case class SMBIO[K, T](id: String, keyBy: T => K) extends KeyedIO[K, T] with TestIO[(K, T)] {
-  override val tapT: TapT.Aux[(K, T), (K, T)] = TapOf[(K, T)]
-  override def testId: String = s"SMBIO($id)"
+import scala.jdk.CollectionConverters._
+
+final case class SMBIO[K: Coder, T](id: String, keyBy: T => K) extends TestIO[T] with KeyedIO[T] {
+  override type KeyT = K
+  override val tapT: TapT.Aux[T, T] = TapOf[T]
+  override def testId: String = SMBIO.testId(id)
+  override def keyCoder: Coder[K] = Coder[K]
+}
+
+object SMBIO {
+  private[smb] def testId(read: SortedBucketIO.Read[_]): String =
+    testId(read.getInputDirectories.asScala.mkString(","))
+  private def testId(id: String): String = s"SMBIO($id)"
 }
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/SortedBucketIO.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/SortedBucketIO.scala
@@ -20,8 +20,6 @@ import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
 import com.spotify.scio.util.ScioUtil
 
-import scala.jdk.CollectionConverters._
-
 final class SortedBucketIO[K, T](path: String, override val keyBy: T => K)(implicit
   override val keyCoder: Coder[K]
 ) extends TestIO[T]
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/SortedBucketIO.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/SortedBucketIO.scala
@@ -18,24 +18,25 @@ package com.spotify.scio.smb
 
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
-import org.apache.beam.sdk.extensions.smb.{SortedBucketIO => BIO}
+import com.spotify.scio.util.ScioUtil
 
 import scala.jdk.CollectionConverters._
 
-final class SortedBucketIO[K, T](id: String, override val keyBy: T => K)(implicit
+final class SortedBucketIO[K, T](path: String, override val keyBy: T => K)(implicit
   override val keyCoder: Coder[K]
 ) extends TestIO[T]
     with KeyedIO[T] {
   override type KeyT = K
   override val tapT: TapT.Aux[T, T] = TapOf[T]
-  override def testId: String = SortedBucketIO.testId(id)
+  override def testId: String = SortedBucketIO.testId(path)
 }
 
 object SortedBucketIO {
-  def apply[K: Coder, T](id: String, keyBy: T => K): SortedBucketIO[K, T] =
-    new SortedBucketIO[K, T](id, keyBy)
+  def apply[K: Coder, T](path: String, keyBy: T => K): SortedBucketIO[K, T] =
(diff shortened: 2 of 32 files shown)


scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -36,15 +37,15 @@ trait PipelineTestUtils {
    * }
    * }}}
    */
-  def runWithContext[T](fn: ScioContext => T): ScioExecutionContext = {
```
RustedBones, line 39: This is breaking. type parameter was not really useful here

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
```diff
@@ -115,13 +115,19 @@ object JobTest {
...
+      input(io, IterableInputSource(value.map(x => io.keyBy(x) -> x)))
+
     /**
      * Feed an input in the form of a `PTransform[PBegin, PCollection[T]]` to the pipeline being
      * tested. Note that `PTransform` inputs may not be supported for all `TestIO[T]` types.
      */
     def inputStream[T: Coder](io: ScioIO[T], stream: TestStream[T]): Builder =
       input(io, TestStreamInputSource(stream))
 
+    def keyedInputStream[K: Coder, T: Coder](io: KeyedIO[K, T], stream: Iterable[T]): Builder =
```
shnapz, line 128: Not sure if it is useful. What could be a hypothetical use case?

RustedBones: As we have an API with `TestStream` for normal input, I just tried to keep consistency. 
It looks like `TestStream` is a more beam/java like way of giving some test data

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala
```diff
@@ -68,35 +70,44 @@ final class SortedBucketScioContext(@transient private val self: ScioContext) ex
...
-
-          while (itB.hasNext) {
-            val b = itB.next()
-            val ai = resA.iterator()
-            while (ai.hasNext) {
-              val a = ai.next()
-              out.output((key, (a, b)))
+  ): SCollection[(K, (L, R))] = self.requireNotClosed {
+    if (self.isTest) {
+      val testInput = TestDataManager.getInput(self.testId.get)
```
clairemcginty, line 75: maybe we could add a static helper method to the `SMBIO` companion object to help transform `SortedBucketIO.Read`s into `testInputs` ?

scio-smb/src/test/scala/com/spotify/scio/smb/SortMergeBucketTest.scala
```diff
@@ -0,0 +1,56 @@
...
+      new User(1, "lastname", "firstname", "email@foobar.com", Collections.emptyList(), address)
+
+    JobTest[SmbJob.type]
+      .args(
+        "--users=users",
+        "--accounts=accounts",
+        "--output=output"
+      )
+      .keyedInput(SMBIO[Integer, User]("users", _.getId), Seq(user))
+      // input is also possible but error prone as key must be given manually
```
clairemcginty, line 66: is it possible to do some type-matching in `def input` to automatically add a `.map(i => (io.keyBy(i), i)` to the input Iterable if `ScioIO` matches trait `KeyedIO` ?

scio-test/src/main/scala/com/spotify/scio/testing/JobTest.scala
RustedBones: I get it. I forgot to replace with `TestStream` type

scio-smb/src/test/scala/com/spotify/scio/smb/SortMergeBucketTest.scala
RustedBones: It is possible but then the `KeyedIO` must contain the coder which is not ideal. will give a try

scio-smb/src/main/scala/com/spotify/scio/smb/SMBIO.scala
```diff
@@ -0,0 +1,8 @@
+package com.spotify.scio.smb
+
+import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
+
+final case class SMBIO[K, T](id: String, keyBy: T => K) extends KeyedIO[K, T] with TestIO[(K, T)] {
```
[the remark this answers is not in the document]
RustedBones: I'll try to use it for the writer side too. Will rename if there is a mismatch

RustedBones:
In order to make testing work with `sortmergeTransform`, I had to introduce some breaking API changes.
Rebased the branch on `v0.14.x`

scio-test/src/main/scala/com/spotify/scio/testing/RunEnforcementJobTest.scala
```diff
@@ -48,6 +49,12 @@ trait RunEnforcementJobTest extends AnyFlatSpec { this: PipelineSpec =>
     jt
   }
 
+  def jobTest(job: ScioContext => Any): InnerJobTest.Builder = {
```
RustedBones, line 52: @clairemcginty @kellen makbe this should be kept capitalized, as the version with the type. WDYT ?

```suggestion
  def JobTest(job: ScioContext => Any): InnerJobTest.Builder = {
```

kellen (approved):
Nice work!

scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -155,6 +158,31 @@ trait PipelineTestUtils {
...
+   * {{{
+   * runWithOverrides(TransformOverride.of("operation", (v: Int) => v.toString)) { sc =>
+   *   val result = sc.parallelize(Seq(1, 2, 3))
+   *     .withName("operation")
+   *     .map(operationFn)
+   *   result should contain theSameElementAs("1", "2", "3")
+   * }
+   * }}}
+   */
+  def runWithOverrides(
```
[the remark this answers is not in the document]
RustedBones: I tried but then it's not possible to decide when the pipeline should be run.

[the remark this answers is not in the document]
kellen: Ahh yeah that makes sense. Could we nest it somehow?

```
runWithContext { sc =>
  sc.withOverrides(
    TransformOverride.of(...)
  )
}
```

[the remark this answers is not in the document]
RustedBones: looks nice and should be possible. Will try that.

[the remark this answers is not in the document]
RustedBones: This change is a bit risky because the transform must be defined before the `withOverrides` call or we need to store the override and call  `replaceAll` just before run.

I tried to integrate the `TestDataManager` here but it is inconvenient

kellen (approved):
🙌

scio-smb/src/test/scala/com/spotify/scio/smb/SortMergeBucketTest.scala
```diff
@@ -0,0 +1,182 @@
...
+      .via { case (_, (users, accounts), outputCollector) =>
+        outputCollector.accept(setUserAccounts(users, accounts))
+      }
+
+    sc.run().waitUntilDone()
+  }
+
+}
+
+class SortMergeBucketTest extends PipelineSpec {
```
clairemcginty, line 120: Small nit: maybe rename to `SortedBucketIOTest` to match other `ScioIO` test classes?

clairemcginty (approved):
amazing work!

scio-smb/src/main/scala/com/spotify/scio/smb/SortedBucketIO.scala
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package com.spotify.scio.smb
+
+import com.spotify.scio.coders.Coder
+import com.spotify.scio.io.{KeyedIO, TapOf, TapT, TestIO}
+import com.spotify.scio.util.ScioUtil
+
+final class SortedBucketIO[K, T](path: String, override val keyBy: T => K)(implicit
```
clairemcginty, line 23: now that I think about it, this class name is a bit confusing as it conflicts with the beam packaged SortedBucketIO https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java. I was testing out a custom SortedBucketIO.Read override, but since I had `import com.spotify.scio.smb._` in my job (in order to get the Scala API bindings for reads/writes), I had to rename the `org.apache.beam.sdk.extensions.smb.SortedBucketIO` import to be able to use it

Maybe `SmbIO`? `SortedBucketTestIO`?

RustedBones: Yes I noticed that. I tried to keep consistency with java naming but It's probably to better to avoid conflict with `SmbIO`

clairemcginty: cool. I can make that change today!

PR #4968, opened 2023-08-30
Update beam to v2.50.0

RustedBones:
Testing beam 2.50 (not officially released, but artifact is available)


RustedBones:
pre-release [checks](https://github.com/spotify/scio/actions/runs/6023621025) passed

kellen (approved):

PR #4969, opened 2023-08-30
Exclude logger implementation

### LINKED ISSUE #4966: Exclude conflicting log libraries from Hadoop dependencies
Exclude non-`sl4fj-jdk14` libraries (reload, logback-classic, etc...) from being pulled in transitively via Hadoop dependencies in scio-parquet/scio-smb.

RustedBones:
Fix #4966

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -249,6 +249,9 @@ val commonSettings = formatSettings ++
     Compile / doc / javacOptions := Seq("-source", "1.8"),
     excludeDependencies ++= Seq(
       "org.apache.beam" % "beam-sdks-java-io-kafka",
+      // logger implementation must be given by the runner lib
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -249,6 +249,9 @@ val commonSettings = formatSettings ++
     Compile / doc / javacOptions := Seq("-source", "1.8"),
     excludeDependencies ++= Seq(
       "org.apache.beam" % "beam-sdks-java-io-kafka",
+      // logger implementation must be given by the runner lib
+      "ch.qos.reload4j" % "reload4j",
+      "log4j" % "log4j",
```
RustedBones, line 254: we should probably not exclude this one so we can have it in the test configuration

clairemcginty (approved):

PR #4973, opened 2023-08-31
(fix #4970) Default to Parquet-SplittableDoFn if RunnerV2 is enabled

### LINKED ISSUE #4970: Make SplittableDoFn-based Parquet impl the default if RunnerV2 experiment is set

clairemcginty:


scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadConfiguration.scala
```diff
@@ -36,4 +41,20 @@ object ParquetReadConfiguration {
   // SplittableDoFn
   val UseSplittableDoFn = "scio.parquet.read.useSplittableDoFn"
   private[scio] val UseSplittableDoFnDefault = false
+
+  private[scio] def getUseSplittableDoFn(conf: Configuration, opts: PipelineOptions): Boolean = {
+    Option(conf.get(UseSplittableDoFn)) match {
+      case Some(v) => v.toBoolean
+      case None
+          if Option(opts.as(classOf[ExperimentalOptions]).getExperiments)
+            .exists(_.contains("use_runner_v2")) =>
```
RustedBones, line 50: Nit maybe worth to have a small `useRunnerV2(opts)` helper here

RustedBones (approved):

PR #4975, opened 2023-09-01
Patch datum factory for specific data in IOs

RustedBones:
Quick patch for https://github.com/apache/beam/issues/28279

Unfortunately, contains some code duplication:
- avro coder still in scio-core
- SMB avro not in scio package + not streamlined with beam AvroIO

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
@@ -31,6 +31,7 @@ import org.apache.avro.generic.IndexedRecord;
 import org.apache.avro.io.DatumReader;
 import org.apache.avro.reflect.ReflectData;
 import org.apache.avro.reflect.ReflectDatumReader;
+import org.apache.avro.reflect.ReflectDatumWriter;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.extensions.avro.coders.AvroCoder;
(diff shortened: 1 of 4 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
```diff
@@ -193,10 +202,15 @@ private static class AvroReader<ValueT> extends FileOperations.Reader<ValueT> {
     public void prepareRead(ReadableByteChannel channel) throws IOException {
       final Schema schema = schemaSupplier.get();
 
-      DatumReader<ValueT> datumReader =
-          recordClass == null
-              ? new GenericDatumReader<>(schema)
-              : new ReflectDatumReader<>(recordClass);
```
RustedBones, line 199: SMB avro read was not symmetric with writer and did not have any logical type conversion applied.

shnapz (approved):

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
shnapz: We were lucky nobody used logical types in SMB?

RustedBones: Looks like it yes

clairemcginty (approved):

PR #4977, opened 2023-09-04
Adds GrpcBatchDoFn

senegalo:
## Description

Adds a all the classes required for a GrpcDoFn that can be batch on the bundle level with the ability to supply a cache that would operate on the single elements instead of on the whole batch.


scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -0,0 +1,325 @@
...
+      @Element Input input,
+      @Timestamp Instant timestamp,
+      OutputReceiver<KV<Input, TryWrapper>> out,
+      BoundedWindow window) {
+    flush(r -> out.output(KV.of(r.input, r.output)));
+    final Cache<String, Output> cache = getResourceCache();
+
+    try {
+
+      final Output cached = cache.getIfPresent(input);
```
RustedBones, line 136: I think this is missing the Id extraction here

RustedBones, line 136: Looks this needs to access from the id instead

senegalo: yeah .. working on it .. i switched from `CacheSupplier<Input, Output>` to `CacheSupplier<String,Output>` where string is the inputId as it's much easier to handle. I apprently forgot to change it there.. pushing a commit right away

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -0,0 +1,325 @@
...
+
+      if (cached != null) {
+        // found in cache
+        out.output(KV.of(input, success(cached)));
+      } else {
+        final UUID uuid = UUID.randomUUID();
+        inputElements.offer(Pair.of(uuid, input));
+        inputsMap.put(this.idExtractorFn.apply(input), Triple.of(input, timestamp, window));
```
RustedBones, line 144: I think we can have issue here when 2 same ids are in the same batch. We'll loose the timestamp and window info from the 1st element.

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -0,0 +1,329 @@
...
+          return null;
+        });
+  }
+
+  // Flush pending elements errors and results
+  private void flush(Consumer<Result> outputFn) {
+    Result r = results.poll();
+    while (r != null) {
+      outputFn.accept(r);
+      resultCount++;
```
RustedBones, line 306: `requestCount` is incremented on batch level but `resultCount` is on the element

senegalo: Nice catch .. fixing

senegalo: Ohh .. forget what i said .. these are not counters and only health check to make sure the number of requests match the number of results .. will just make them match.

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -0,0 +1,330 @@
...
+  private final ConcurrentLinkedQueue<Pair<UUID, List<Result>>> results =
+      new ConcurrentLinkedQueue<>();
+  private long inputCount;
+  private long outputCount;
+
+  public BatchedBaseAsyncLookupDoFn(
+      int batchSize,
+      SerializableFunction<List<Input>, BatchRequest> batchRequestFn,
+      SerializableFunction<BatchResponse, List<Pair<String, Output>>> batchResponseFn,
+      SerializableFunction<Input, String> idExtractorFn,
```
RustedBones, line 89: grouped batch parameters together.

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -0,0 +1,330 @@
...
+  private final SerializableFunction<Input, String> idExtractorFn;
+  private final int maxPendingRequests;
+  private final CacheSupplier<String, Output> cacheSupplier;
+
+  private final Semaphore semaphore;
+  private final ConcurrentMap<UUID, FutureType> futures = new ConcurrentHashMap<>();
+  private final ConcurrentMap<String, List<Triple<Input, Instant, BoundedWindow>>> inputs =
+      new ConcurrentHashMap<>();
+
+  private final Queue<Input> batch = new ArrayDeque<>();
```
RustedBones, line 79: No need of a concurrent collection here as it is only accessed in `processFunction` and `finishBundle` which are thread safe.

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -0,0 +1,330 @@
...
+
+  // Flush pending elements errors and results
+  private void flush(Consumer<Result> outputFn) {
+    Pair<UUID, List<Result>> r = results.poll();
+    while (r != null) {
+      final UUID key = r.getKey();
+      final List<Result> batchResult = r.getValue();
+      batchResult.forEach(outputFn);
+      outputCount += batchResult.size();
+      futures.remove(key);
```
RustedBones, line 312: result here is removed only once for every batch

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
```diff
@@ -92,6 +92,62 @@ class GrpcSCollectionOps[Request](private val self: SCollection[Request]) extend
...
+    channelSupplier: () => Channel,
+    clientFactory: Channel => Client,
+    maxPendingRequests: Int,
+    batchSize: Int,
+    cacheSupplier: CacheSupplier[String, Output] = new NoOpCacheSupplier[String, Output]()
+  )(
+    lookupFn: Client => BatchRequest => ListenableFuture[BatchResponse],
+    batchRequestFn: Seq[Request] => BatchRequest,
+    batchResponseFn: BatchResponse => Seq[(String, Output)],
+    idExtractorFn: Request => String
```
RustedBones, line 110: I'd move this to the 1st parameter group with `batchSize`

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
senegalo: didn't know that :)

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
```diff
@@ -92,6 +92,65 @@ class GrpcSCollectionOps[Request](private val self: SCollection[Request]) extend
       .mapValues(_.asScala.map(_.asScala))
   }
 
+  def batchGrpcLookup[
+    BatchRequest: Coder,
```
RustedBones, line 96: I don;t think we need this coder since the type is internal to the transform

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
```diff
@@ -92,6 +92,65 @@ class GrpcSCollectionOps[Request](private val self: SCollection[Request]) extend
...
+    BatchResponse: Coder,
```
RustedBones, line 97: same here

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
```diff
@@ -92,6 +92,65 @@ class GrpcSCollectionOps[Request](private val self: SCollection[Request]) extend
...
+    Output: Coder,
```
RustedBones, line 98: For consistency with non-batched, I think we should name it `Response`

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
```diff
@@ -92,6 +92,65 @@ class GrpcSCollectionOps[Request](private val self: SCollection[Request]) extend
...
+    batchSize: Int,
+    batchRequestFn: Seq[Request] => BatchRequest,
+    batchResponseFn: BatchResponse => Seq[(String, Output)],
+    idExtractorFn: Request => String,
+    maxPendingRequests: Int,
+    cacheSupplier: CacheSupplier[String, Output] = new NoOpCacheSupplier[String, Output]()
+  )(
+    lookupFn: Client => BatchRequest => ListenableFuture[BatchResponse]
+  )(implicit
+    requestCoder: Coder[Request],
```
RustedBones, line 112: this coder should be available from the `self` SCollection

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
```diff
@@ -92,6 +92,65 @@ class GrpcSCollectionOps[Request](private val self: SCollection[Request]) extend
...
+    tryCoder: Coder[Try[Output]]
```
RustedBones, line 113: prefer expecting non wrapped coder.

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
senegalo: Makes sense .. fixed

scio-grpc/src/main/scala/com/spotify/scio/grpc/SCollectionSyntax.scala
senegalo: Ohh .. didn't know that. 
I unfortunately have a very shallow understanding of Coders. I understand what they do (help serialize types) but i don't understand how they propagate. Had the same the issue in Flink when i used it before as well :( 
if you have a good resource would be awesome (i'll definitely look up as well as it's time to get educated about the topic)

anyway .. i added `import self.coder` which fixed the issue.

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
```diff
@@ -237,6 +237,15 @@ private FutureType handleOutput(FutureType future, List<Input> batchInput, UUID
                   pair -> {
                     final String id = pair.getLeft();
                     final Output output = pair.getRight();
+
+                    if(!inputs.containsKey(id)) {
```
senegalo, line 241: It took me a bit of time to realize my mistake. But basically if there is a miss match between the ids extracted from the input and the ids returned by the gRPC endpoint we only get an `IllegalStateException` with the following message `Expected requestCount == responseCount, but %s != %s`
throwing an exception inside the handleOutput is useless since we really don't operate on the futures after the `handleOutput` callbacks are added so i opted for adding and error log instead. 

If you have any better suggestions let me know :)

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -44,7 +44,6 @@ public interface Guava<V> extends Base<ListenableFuture<V>, V> {
     @Override
     default void waitForFutures(Iterable<ListenableFuture<V>> futures)
         throws InterruptedException, ExecutionException {
-      // Futures#allAsList only works if all futures succeed
```
senegalo, line 47: i think this comment was on an old code that used `Futures#allAsList` instead of the current `Futures.successfulAsList`. As the comment mentions `allAsList` fails if any future fails while `successfulAsList` returns a null value for any failed future.

RustedBones: This is not old if you check git blame. This comment was just a hit why we do NOT use `Futures#allAsList`. maybe should have been turned into
```
use Future#successfulAsList instead of Futures#allAsList which only works if all futures succeed
```

scio-core/src/main/java/com/spotify/scio/transforms/BatchedBaseAsyncLookupDoFn.java
senegalo: makes sense .. i added the remainder part of the code in the else so it will only be executed if the inputs contain the key

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncBatchLookupDoFn.java
```diff
@@ -0,0 +1,356 @@
...
+                    final Output output = pair.getRight();
+
+                    if (!inputs.containsKey(id)) {
+                      LOG.error(
+                          "The ID '{}' received in the gRPC batch response does not "
+                              + "match any IDs extracted via the idExtractorFn for the requested  "
+                              + "batch sent to the gRPC endpoint. Please ensure that the IDs returned "
+                              + "from the gRPC endpoints match the IDs extracted using the provided"
+                              + "idExtractorFn for the same input.",
+                          id);
```
RustedBones, line 263: if this is a programming error, we should rather fail instead of logging

senegalo: I totally agree and that was actually my initial approach. But because we never act on this future after this point (no more callbacks to check if it's completed with a failure or success) i was unable to stop the execution here (or maybe i missed how to) So i opted here for the log. 
With that said if that log is emitted that means one `Result` will be missing from the bundle and the number or requests will not match the number of responses and the bundle will fail. 
If that's not good enough then maybe we could restructure the future handling fail immediately if the any of the output callbacks was not able to find the input.

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcBatchDoFn.java
```diff
@@ -128,7 +128,7 @@ public static class Builder<
     private SerializableFunction<List<Input>, BatchRequest> batchRequestFn;
     private SerializableFunction<BatchResponse, List<Pair<String, Output>>> batchResponseFn;
     private SerializableFunction<Input, String> idExtractorFn;
-    private Integer maxPendingRequests;
+    private int maxPendingRequests = GrpcDoFn.DEFAULT_MAX_PENDING_REQUESTS;
     private Integer batchSize;
```
RustedBones, line 132: We have a default for `maxPendingRequests` in the non batched.
Shoulw we also add a default for the batch size or require the user to provide one ?

RustedBones:
I think this is now ready for review by the whole @spotify/flatmap team

scio-grpc/src/main/java/com/spotify/scio/grpc/GrpcBatchDoFn.java
RustedBones: IMHO, when knowing the target system, I'm not super keen on setting defaults. Even for the max pending requests. Just added the value for consistency.

senegalo: Ohh .. i completely missunderstood your question !! 

I thought we needed a default for the `batchSize` and that was my answer to that !! 

but of course if we have a default in one we should totally have in the other. That makes perfect sense !

scio-grpc/src/test/scala/com/spotify/scio/grpc/GrpcBatchDoFnTest.scala
```diff
@@ -194,7 +194,7 @@ class GrpcBatchDoFnTest extends PipelineSpec with BeforeAndAfterAll {
     val input = (0 to 10).map { i =>
       ConcatRequestWithID
         .newBuilder()
-        .setRequestId("1")
+        .setRequestId(i.toString)
```
senegalo, line 197: But i think that shouldn't be the case .. 
The first test tests that given that all the inputs in the bundle return unique ids they have matched outputs in numbers
This test checks that if all inputs have the same exact id it would still match outputs in numbers 

The reasoning behind this check is that we do send a single request for the first element we find and use the response for all the matching inputs and i wanted to be sure that this behavior is maintained here.

senegalo: i would even go to say that this test is broken (the original one) 
If you follow the test thoroughly you would find that i have inputs with different values sharing the same request ids and when those go to the gRPC server and come back they have different output values ! so it makes perfect sense that they fail when they come back as the expectation i had was that for each input object i sent i would get a "different" output which doesn't make sense. 

i think we should change the test to just duplicate the inputs list and pass it to the scio pipline instead of manipulating the request ID

also .. i am insanely puzzled how this NEVER failed locally for me !

senegalo: Actually .. never mind .. i managed to force it to use bigger bundles and i think it has something to do with the number of CPUs on my machine "somehow" 

i created a LOT (2000) of duplicated input with a big batch size to properly test that scenario. 

just pushed the changes :)

RustedBones (approved):

PR #4980, opened 2023-09-05
Propagate ParquetOutputFormat options to ParquetWriter

clairemcginty:
We were not properly propagating all of the `ParquetOutputFormat` [options](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/README.md#properties) to `ParquetWriter`: as is shown in some of the parquet-mr [examples](https://github.com/apache/parquet-mr/blob/9b5a962df3007009a227ef421600197531f970a5/parquet-avro/src/test/java/org/apache/parquet/avro/TestReflectReadWrite.java#L143-L148), they are _not_ automatically parsed from the `Configuration`; the expectation is that these options are passed directly to `ParquetWriter`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1365,6 +1365,7 @@ lazy val `scio-smb`: Project = project
       "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % Provided,
       "org.apache.avro" % "avro" % avroVersion % Provided,
       "org.apache.hadoop" % "hadoop-common" % hadoopVersion % Provided,
+      "org.apache.hadoop" % "hadoop-mapreduce-client-core" % hadoopVersion % Provided,
       "org.apache.parquet" % "parquet-avro" % parquetVersion % Provided excludeAll ("org.apache.avro" % "avro"),
       "org.apache.parquet" % "parquet-column" % parquetVersion % Provided,
       "org.apache.parquet" % "parquet-common" % parquetVersion % Provided,
--- a/build.sbt
+++ b/build.sbt
@@ -1365,7 +1365,7 @@ lazy val `scio-smb`: Project = project
       "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % Provided,
       "org.apache.avro" % "avro" % avroVersion % Provided,
       "org.apache.hadoop" % "hadoop-common" % hadoopVersion % Provided,
-      "org.apache.hadoop" % "hadoop-mapreduce-client-core" % hadoopVersion % Provided,
+      "org.apache.hadoop" % "hadoop-mapreduce-client-core" % hadoopVersion % Test,
       "org.apache.parquet" % "parquet-avro" % parquetVersion % Provided excludeAll ("org.apache.avro" % "avro"),
       "org.apache.parquet" % "parquet-column" % parquetVersion % Provided,
       "org.apache.parquet" % "parquet-common" % parquetVersion % Provided,
--- a/build.sbt
+++ b/build.sbt
@@ -1336,6 +1336,10 @@ lazy val `scio-smb`: Project = project
   .settings(beamRunnerSettings)
   .settings(
     description := "Sort Merge Bucket source/sink implementations for Apache Beam",
+    unusedCompileDependenciesFilter -= Seq(
+      // ParquetUtils calls functions defined in parent class from hadoop-mapreduce-client-core
+      moduleFilter("org.apache.hadoop", "hadoop-mapreduce-client-core")
+    ).reduce(_ | _),
     libraryDependencies ++= Seq(
       // compile
       "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
@@ -1365,7 +1369,7 @@ lazy val `scio-smb`: Project = project
       "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % Provided,
       "org.apache.avro" % "avro" % avroVersion % Provided,
       "org.apache.hadoop" % "hadoop-common" % hadoopVersion % Provided,
-      "org.apache.hadoop" % "hadoop-mapreduce-client-core" % hadoopVersion % Test,
+      "org.apache.hadoop" % "hadoop-mapreduce-client-core" % hadoopVersion % Provided,
       "org.apache.parquet" % "parquet-avro" % parquetVersion % Provided excludeAll ("org.apache.avro" % "avro"),
       "org.apache.parquet" % "parquet-column" % parquetVersion % Provided,
       "org.apache.parquet" % "parquet-common" % parquetVersion % Provided,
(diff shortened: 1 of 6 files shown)


scio-parquet/src/main/java/com/spotify/scio/parquet/WriterUtils.java
```diff
@@ -28,12 +31,47 @@ public static <T, SELF extends ParquetWriter.Builder<T, SELF>> ParquetWriter<T>
       ParquetWriter.Builder<T, SELF> builder, Configuration conf, CompressionCodecName compression)
       throws IOException {
     // https://github.com/apache/parquet-mr/tree/master/parquet-hadoop#class-parquetoutputformat
-    int rowGroupSize =
-        conf.getInt(ParquetOutputFormat.BLOCK_SIZE, ParquetWriter.DEFAULT_BLOCK_SIZE);
+    long rowGroupSize =
```
clairemcginty, line 34: `.withRowGroupSize(int rowGroupSize)` was deprecated; we should use `long`.

scio-parquet/src/main/java/com/spotify/scio/parquet/WriterUtils.java
```diff
@@ -28,12 +31,47 @@ public static <T, SELF extends ParquetWriter.Builder<T, SELF>> ParquetWriter<T>
...
+        .withBloomFilterEnabled(ParquetOutputFormat.getBloomFilterEnabled(conf))
+        .withDictionaryPageSize(ParquetOutputFormat.getDictionaryPageSize(conf))
+        .withMaxRowCountForPageSizeCheck(ParquetOutputFormat.getMaxRowCountForPageSizeCheck(conf))
+        .withMinRowCountForPageSizeCheck(ParquetOutputFormat.getMinRowCountForPageSizeCheck(conf))
+        .withValidation(ParquetOutputFormat.getValidation(conf))
         .withRowGroupSize(rowGroupSize)
         .build();
   }
+
+  private static <T> Map<String, T> getColumnarConfig(
```
clairemcginty, line 68: Unfortunately there isn't a public API available in parquet-mr to do this per-column logic for us :(

build.sbt
```diff
@@ -1365,6 +1365,7 @@ lazy val `scio-smb`: Project = project
       "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % Provided,
       "org.apache.avro" % "avro" % avroVersion % Provided,
       "org.apache.hadoop" % "hadoop-common" % hadoopVersion % Provided,
+      "org.apache.hadoop" % "hadoop-mapreduce-client-core" % hadoopVersion % Provided,
```
clairemcginty, line 1368: without this, was getting a bunch of errors from my `ParquetUtils` class, i.e.:

```shell
[error] /workdir/scio/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetUtils.java:63:1: cannot access org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
[error]   class file for org.apache.hadoop.mapreduce.lib.output.FileOutputFormat not found
```

scio-parquet/src/test/scala/com/spotify/scio/parquet/TestWriterUtils.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.parquet.hadoop.ParquetFileReader
+import org.apache.parquet.hadoop.metadata.ColumnChunkMetaData
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import java.nio.file.{Files, Path}
+import scala.jdk.CollectionConverters._
+
+class TestWriterUtils extends AnyFlatSpec with Matchers with BeforeAndAfterAll {
```
RustedBones, line 35: great

RustedBones (commented):
I'm wondering if we should mark `scio-parquet` as provided dependency for `scio-smb` to remove some duplication

clairemcginty:
> I'm wondering if we should mark `scio-parquet` as provided dependency for `scio-smb` to remove some duplication

I was thinking the same thing, we’re already duplicating a lot to support logical types across both modules. Maybe we could do a 0.13.4 release with this fix as-is, then refactor in 0.14? Since ParquetUtils is package private there shouldn’t be any mima issues removing it later

build.sbt
RustedBones: The build complains about this scoping: it is not used by compilation.
The lib looks to be needed for runtime, so in our case this should be scoped to `Test` (we do not want to ship transitively)

I think we should revisit the build to mark `scio-parquet`, `scio-avro` or any IO module having an smb impl as provided instead of the storage format libs.

RustedBones: See #5004

RustedBones:
> ParquetUtils is package private there shouldn’t be any mima issues removing it later

Sounds good. having smb in the `org.apache.beam.sdk.extensions.smb` is indeed blocking a lot of code sharing

clairemcginty:
I noticed there's one more ParquetWriter config not exposed by Scio -- [byteStreamSplitEncoding](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L521). However, Parquet currently doesn't expose this via a Configuration key -- you can only enable it by setting it directly on the writer. From a Scio design perspective, we can either offer it as a param in the Parquet Write configuration case class, or create our own Configuration key for it (like [these](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadConfiguration.scala#L27)).

It's a fairly niche option -- from what I've read, it slightly improves [compression of numeric types](https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9) -- so I think I'm *slightly* in favor of creating our own Configuration key for this. We can look into contributing a dedicated key for it to parquet-mr, too, then deprecate our key in favor of the "official" one once released. Thoughts?

build.sbt
clairemcginty: hm, looks like that scope change [brings back](https://github.com/spotify/scio/actions/runs/6225615484/job/16896427165?pr=4980) the compilation errors:

```
/home/runner/work/scio/scio/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetUtils.java:71:1: cannot access org.apache.hadoop.mapreduce.JobContext
```

maybe should be Runtime?

RustedBones:
> I think I'm slightly in favor of creating our own Configuration key for this
i'd rather create an issue and wait for fix in upstream

build.sbt
RustedBones: added to the filter with comment. I think the tool is not smart enough to find this kind of transitive usage.

RustedBones (approved):

PR #5003, opened 2023-09-13
Update dev docs

RustedBones:
Changes driven after discussion with @senegalo


senegalo (approved):
Looks perfect to me :+1:  will try to follow the new guide and see if i can spot something already

PR #5010, opened 2023-09-21
Move to hbase bigtable

RustedBones:
Missing hbase magnolify in the example


kellen:
Related to magnolify#841 ... pick up work on magnolify#841 PR when this become more pressing

PR #5011, opened 2023-09-21
Remove TF model from the resource cache before closing

jrglee:
During `PredictDoFn` teardown, the model gets closed without being removed from the cache. This can create some illegal states if the PredictDoFn subclass gets used more than once. There are valid use cases to use the same prediction computation in different parts of the DAG. It is safer to remove the model from the cache before closing it.

This solution is still not perfect because there is a potential for race conditions if setup and teardown happens at the same time with two instances sharing a JVM. This is possible when running tests in parallel in a direct runner, for example.

TODO: I could not run tests locally because I have a M1, may need to rely on the github action to see what else needs to change

### DIFF
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
@@ -112,9 +112,14 @@ sealed trait PredictDoFn[T, V, M <: Model[_]]
   @Teardown
   override def teardown(): Unit = {
     Log.info(s"Tearing down predict DoFn $this")
-    val (running, m) = getResource().get(modelId)
-    if (running.decrementAndGet() == 0) {
-      m.close()
+    Option(getResource.get(modelId)) match {
+      case Some((running, m)) =>
+        if (running.decrementAndGet() == 0) {
+          getResource.remove(modelId)
+          m.close()
+        }
+      case _ =>
+        Log.warn("No model to close while tearing down predict DoFn")
     }
   }
 }
--- a/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
+++ b/scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
@@ -18,8 +18,7 @@
 package com.spotify.scio.tensorflow
 
 import java.time.Duration
-import java.util.concurrent.{ConcurrentHashMap, ConcurrentMap}
-import java.util.function.Function
+import java.util.concurrent.ConcurrentHashMap
 import com.spotify.zoltar.tf.{TensorFlowLoader, TensorFlowModel}
 import com.spotify.zoltar.Model
 import org.apache.beam.sdk.transforms.DoFn.{
@@ -40,15 +39,13 @@ import com.spotify.scio.transforms.DoFnWithResource
 import com.spotify.scio.transforms.DoFnWithResource.ResourceType
 import com.spotify.zoltar.Model.Id
 
-import java.util.concurrent.atomic.AtomicInteger
-
 sealed trait PredictDoFn[T, V, M <: Model[_]]
     extends DoFnWithResource[T, V, PredictDoFn.Resource[M]] {
   import PredictDoFn._
 
   def modelId: String
 
-  def loadModel: M
+  def loadModel(): M
 
   def model: M = getResource.get(modelId)._2
 
@@ -60,22 +57,20 @@ sealed trait PredictDoFn[T, V, M <: Model[_]]
 
   def outputTensorNames: Seq[String]
 
-  override def createResource(): Resource[M] = new ConcurrentHashMap[String, (AtomicInteger, M)]()
+  override def createResource(): Resource[M] = new ConcurrentHashMap[String, (Int, M)]()
 
   override def getResourceType: DoFnWithResource.ResourceType = ResourceType.PER_CLASS
 
   @Setup
   override def setup(): Unit = {
     super.setup()
-    val (a, _) = getResource.computeIfAbsent(
+    getResource.compute(
       modelId,
-      new Function[String, (AtomicInteger, M)] {
-        override def apply(v1: String): (AtomicInteger, M) =
-          new AtomicInteger(0) -> loadModel
+      {
+        case (_, null)              => 1 -> loadModel()
+        case (_, (refCount, model)) => (refCount + 1) -> model
       }
     )
-    a.incrementAndGet()
-    ()
   }
 
   /** Process an element asynchronously. */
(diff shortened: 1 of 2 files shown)


scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -112,9 +112,14 @@ sealed trait PredictDoFn[T, V, M <: Model[_]]
   @Teardown
   override def teardown(): Unit = {
     Log.info(s"Tearing down predict DoFn $this")
-    val (running, m) = getResource().get(modelId)
-    if (running.decrementAndGet() == 0) {
-      m.close()
+    Option(getResource.get(modelId)) match {
+      case Some((running, m)) =>
+        if (running.decrementAndGet() == 0) {
+          getResource.remove(modelId)
```
RustedBones, line 118: we could factorize calls to `getResource`

RustedBones (commented):
Thanks for the PR. There's indeed a race condition when model is in the cache:
1. `setup` accesses the cached model
2. `teardown` accesses the cached model
3. `teardown` decrements and close
4. `setup` increments

Since we use a `ConcurrentHashMap`, we should handle the reference count directly in the `compute` methods instead of returning an atomic counter.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -40,15 +39,13 @@ import com.spotify.scio.transforms.DoFnWithResource
...
-import java.util.concurrent.atomic.AtomicInteger
-
 sealed trait PredictDoFn[T, V, M <: Model[_]]
     extends DoFnWithResource[T, V, PredictDoFn.Resource[M]] {
   import PredictDoFn._
 
   def modelId: String
 
-  def loadModel: M
+  def loadModel(): M
```
RustedBones, line 48: Calling this will trigger a side effect. Reflect it by adding `()`

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -111,16 +106,25 @@ sealed trait PredictDoFn[T, V, M <: Model[_]]
...
+          null
+        case (_, (refCount, model)) =>
+          (refCount - 1) -> model
+      }
+    )
   }
 }
 
 object PredictDoFn {
-  type Resource[M <: Model[_]] = ConcurrentMap[String, (AtomicInteger, M)]
```
RustedBones, line 123: `ConcurrentMap` does not ensure `compute` to be atomic. Replace with `ConcurrentHashMap` which does.

RustedBones: Changing type here fails the binary compatibility check. Will revert to lower abstraction

jkiley66: `override def createResource(): Resource[M] = new ConcurrentHashMap[String, (AtomicInteger, M)]()`

Won't the above code already take care of the right implementation to use?
I thought getResource will use resource created by `createResource`, of course except the AtomicInteger to Int change that has been introduced?

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -60,22 +57,20 @@ sealed trait PredictDoFn[T, V, M <: Model[_]]
...
   @Setup
   override def setup(): Unit = {
     super.setup()
-    val (a, _) = getResource.computeIfAbsent(
+    getResource.compute(
       modelId,
-      new Function[String, (AtomicInteger, M)] {
-        override def apply(v1: String): (AtomicInteger, M) =
-          new AtomicInteger(0) -> loadModel
+      {
```
jkiley66, line 69: This doesn't hurt to move the two ops in one.

I think this fundamentally looks same as `computeIfAbsent` and `a.incrementAndGet` called in sequence.
In the previous implementation as well only one thread out of many could have created `computeIfAbsent` and rest of the parallel threads would have incremented the count on line 77.

Do we have any example of a failed test in parallel execution? Has any bug been filed for this.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
RustedBones: That's correct. It would have worked because we make the **assumption** the implementation given by `createResource` has an atomic `compute`. Inspecting the resource javadoc on `compute` in `setup` and `teardown` tells this call is not guaranteed to be atomic. By changing the type, we explicitly require the resource to have an atomic `compute`.

In this case this is not really important because `PredictDoFn` is sealed and `createResource` is never overridden.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
RustedBones: > Do we have any example

See example in this comment: https://github.com/spotify/scio/pull/5011#pullrequestreview-1639305224

The `setup` thead can access an element which would be then be deleted by `teardown` if there is a race in between the `computeIfAbsent` and  `incrementAndGet`.
this is very unlikely and we've never had any failure nor report.

jkiley66: Ok we moved the `loadModel()` and `close()` to a critical section. Makes sense.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
jrglee: Thanks for the reasoning. I really wondered why `ConcurrentMap` in a sealed trait. In a regular trait it would enable other concurrent maps like a Caffeine cache.

scio-tensorflow/src/main/scala/com/spotify/scio/tensorflow/TensorFlowDoFns.scala
```diff
@@ -111,16 +106,25 @@ sealed trait PredictDoFn[T, V, M <: Model[_]]
...
-    }
+    Log.info("Tearing down predict DoFn {}", this)
+    getResource.compute(
+      modelId,
+      {
+        case (_, null) =>
+          Log.warn("No model to close while tearing down predict DoFn")
+          null
+        case (_, (1, model)) =>
+          model.close()
```
jrglee, line 117: This `close()` is the real side effect, and in a map with a non-locking `compute` it would be a problem because it would lead to a retry trying to close an already closed model. But in this case it works because it is a `ConcurrentHashMap`. It is good to have some comments to explain this implementation detail since it's hard to enforce with unit tests.

I can add it since it is technically my PR.

jrglee:
Thanks @RustedBones for making my small attempt of fix a real fix now. What are the next steps?

RustedBones:
Sorry for the late response. I've marked it for next milestone `0.13.4` we are preparing. This should be ready soon

clairemcginty (approved):

PR #5014, opened 2023-09-24
add a runWithMocks function that's more flexible than JobTest

AndersonReyes:
# TODO
add tests

# Description
Adds function similar to runWithContext that allows job test mocking


scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -42,6 +43,17 @@ trait PipelineTestUtils {
     sc.run()
   }
 
+  /** Similar to JobTest[T] flow but less dependent on having a static main method */
+  def runWithMocks[T](setupMocks: JobTest.Builder => Unit)(fn: ScioContext => T): T = {
+    val jt: JobTest.Builder = JobTest("test")(BeamOptions(List()))
```
AndersonReyes, line 48: Maybe move beam options as default argument

clairemcginty (commented):
oh interesting idea!

So as I understand it, what this solution offers that existing built-ins like `runWithContext`/`runWithData` don't is the easy assertions on outputs/counters. Makes sense -- my concern is that the the API of `JobTest.Builder => Unit` is potentially unclear to users (and exposes certain implementation details to them -- they have to call `.run()` on the Builder still right?)

I wonder if we could just make it easier to do Counter assertions from `runWith{Context,Data}`... maybe an API like this:

```scala
runWithContext { sc =>
   sc
     .parallelize(Seq(1, 2, 3))
     .tap(i => SomeCounter.inc()) 
     .sum should containSingleValue (6) // assert on outputs
} { case (result: ScioResult) =>
  result should haveCounterValue(SomeCounter, 3) // assert on counter values
}
```

would that handle the use case you had in mind? 🤔

RustedBones:
Looks similar to the [runWithOverrides](https://github.com/spotify/scio/blob/1841eaa14112f0649c242ee8c21b90ebe03dfbb2/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala#L177) planned for 0.14. Does this fit your use-case @AndersonReyes ?

AndersonReyes:
> Looks similar to the [runWithOverrides](https://github.com/spotify/scio/blob/1841eaa14112f0649c242ee8c21b90ebe03dfbb2/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala#L177) planned for 0.14. Does this fit your use-case @AndersonReyes ?

I think so, in this case we can just give unique names to all IO and then override them. The use case is having  a flyte scio pipeline but not needing to define a main function just for a job test. In this case we can run the task and pass the sc with the overrides as well. I’ll see if I can checkout out the 0.14 branch give it a try.

AndersonReyes:
> Looks similar to the [runWithOverrides](https://github.com/spotify/scio/blob/1841eaa14112f0649c242ee8c21b90ebe03dfbb2/scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala#L177) planned for 0.14. Does this fit your use-case @AndersonReyes ?

@RustedBones  this only works for reading right? I tested a fn that also saved as well and it didn't work because i got the error of saving outside JobTest.

RustedBones:
Ok I see. I'd be more in favo modify the job test instead of the `runWithContext` if it needs to mock some IOs. I added this [change](https://github.com/spotify/scio/pull/4962/commits/a53adfc3d5b133d8d5ccdeb8e46f289d25b14794)  to the test improvement PR for 0.14.

Would it be more suitable for you ?

AndersonReyes:
thank you both @clairemcginty @RustedBones

Issue #5015, opened 2023-09-26
Parquet reads should use ParquetFileReader#readNextFilteredRowGroup(), not readNextRowGroup()


clairemcginty:
This should speed up reads by filtering at the page level.

PR #5025, opened 2023-10-10
Use readNextFilteredRowGroup instead of readNextRowGroup

RustedBones:
See #5015


clairemcginty:
I updated the branch to make this option configurable -- via `Configuration` key, rather than a typed `ReadParam` key, which is probably suboptimal, but it does follow the pattern of how we're setting other read options in `ParquetReadFn` (i.e split granularity). @RustedBones lmk what you think

scio-parquet/src/main/scala/com/spotify/scio/parquet/read/ParquetReadFn.scala
```diff
@@ -54,10 +55,35 @@ class ParquetReadFn[T, R](
   private val logger = LoggerFactory.getLogger(this.getClass)
 
   private lazy val granularity =
```
RustedBones, line 57: since we now have 2 granularity settings (split and filter). I think we should disambiguate here

clairemcginty (approved):

PR #5026, opened 2023-10-12
Expose bigtable read maxBufferElementCount option

RustedBones:


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigtable/BigTableIO.scala
```diff
@@ -89,6 +84,7 @@ object BigtableRead {
   object ReadParam {
     val DefaultKeyRanges: Seq[ByteKeyRange] = Seq.empty[ByteKeyRange]
     val DefaultRowFilter: RowFilter = null
+    val DefaultMaxBufferElementCount: Option[Int] = None
```
RustedBones, line 87: As this setting is nullable, I prefer the scala option instead of the `Integer = null`. WDYT ?

build.sbt
```diff
@@ -204,6 +204,22 @@ lazy val mimaSettings = Def.settings(
...
+    ),
+    ProblemFilters.exclude[DirectMissingMethodProblem](
+      "com.spotify.scio.bigtable.BigtableRead#ReadParam.this"
+    ),
+    ProblemFilters.exclude[IncompatibleMethTypeProblem](
+      "com.spotify.scio.bigtable.syntax.ScioContextOps.bigtable"
+    ),
+    ProblemFilters.exclude[IncompatibleMethTypeProblem](
+      "com.spotify.scio.bigtable.syntax.ScioContextOps.bigtable$extension"
     )
```
RustedBones, line 223: This is binary breaking, Should we target 0.14 instead ?

clairemcginty (approved):

PR #5032, opened 2023-10-17
Pass avro reader schema for SMB operation

RustedBones:
Fix regression introduced in #4975. Schema is not given to the reflect datum reader.


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
```diff
@@ -210,7 +203,7 @@ public void prepareRead(ReadableByteChannel channel) throws IOException {
         // same as SpecificRecordDatumFactory in scio-avro
         ReflectData data = new ReflectData(recordClass.getClassLoader());
         org.apache.beam.sdk.extensions.avro.schemas.utils.AvroUtils.addLogicalTypeConversions(data);
-        datumReader = new ReflectDatumReader<>(data);
```
RustedBones, line 213: schema was not passed to the reader

clairemcginty (approved):

kellen (approved):

PR #5033, opened 2023-10-17
Make macro print debug logs if `bigquery.types.debug` enabled

RustedBones:
Right now, logging goes through `slf4j`, but there isn't any logger impl at compile time, giving
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementational 1s
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

use the macro front end to print debug statements

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
@@ -46,10 +46,7 @@ private[types] object SchemaProvider {
           case t if isCaseClass(t) => toFields(t)
           case t                   => throw new RuntimeException(s"Unsupported type $t")
         }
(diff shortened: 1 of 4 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/SchemaProvider.scala
```diff
@@ -46,10 +46,7 @@ private[types] object SchemaProvider {
           case t if isCaseClass(t) => toFields(t)
           case t                   => throw new RuntimeException(s"Unsupported type $t")
         }
-        val r = new TableSchema().setFields(fields.toList.asJava)
-        debug(s"SchemaProvider.schemaOf[${typeOf[T]}]:")
-        debug(r)
```
kellen, line 51: Is this now covered elsewhere?

kellen (approved):

PR #5034, opened 2023-10-17
Fix issues with voyager

RustedBones:


scio-extra/src/main/scala/com/spotify/scio/extra/voyager/syntax/SCollectionSyntax.scala
```diff
@@ -100,8 +100,12 @@ class VoyagerPairSCollectionOps(
     require(!uri.exists, s"Voyager URI ${uri.value} already exists")
 
     self.transform { in =>
-      in.reifyAsIterableInGlobalWindow
-        .map { xs =>
+      // Do not use reifyAsIterableInGlobalWindow
+      // as iterableCoder may materialize the full collection
+      self.context
+        .parallelize(Seq((): Unit))
```
clairemcginty, line 106: out of curiosity, how does Dataflow render this job graph? I guess it adds another root-level transform for `.parallelize()` that feeds into the Voyager SCollection?

clairemcginty (approved):

scio-extra/src/main/scala/com/spotify/scio/extra/voyager/syntax/SCollectionSyntax.scala
RustedBones: yes: 
image
I'm just getting rid on the extra `map` from `reifyAsIterableInGlobalWindow` that can trigger the `iterableCoder` forcing the side input collection to fit in memory. Here we directly access the view that should be able to stream side input records from the view

PR #5035, opened 2023-10-17
Support a transform() API for ScioContext

clairemcginty:
The `SCollection#transform` exists to wrap multiple SColl-to-SColl transforms into a single, named composite transform, but we don't have an equivalent for root-level transforms invoked on `ScioContext`.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -700,6 +700,24 @@ class ScioContext private[scio] (
   ): SCollection[U] =
     applyTransform(Option(name), root)
 
+  def transform[U](f: ScioContext => SCollection[U]): SCollection[U] = transform(this.tfName)(f)
+
+  def transform[U](name: String)(f: ScioContext => SCollection[U]): SCollection[U] =
+    wrap(transform_(name)(f(_).internal))
+
+  private[scio] def transform_[U <: POutput](f: ScioContext => U): U =
+    transform_(tfName)(f)
+
+  private[scio] def transform_[U <: POutput](name: String)(f: ScioContext => U): U = {
+    val sc: ScioContext = this
+    applyInternal(
+      name,
+      new PTransform[PBegin, U]() {
+        override def expand(pBegin: PBegin): U = f(sc)
+      }
+    )
+  }
+
   /**
    * Get an SCollection for a text file.
    * @group input
--- a/scio-test/src/test/scala/com/spotify/scio/ScioContextTest.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/ScioContextTest.scala
@@ -227,6 +227,20 @@ class ScioContextTest extends PipelineSpec {
     actualCommitedCounterValue shouldBe Some(0)
   }
 
+  it should "support wrapped root-level transforms" in {
+    val sc = ScioContext()
+    val scioCounter = ScioMetrics.counter(name = "all-map-ops-counter")
+    sc.initCounter(scioCounter)
+
+    sc.transform("Transform1")(_.parallelize(1 to 10).tap(_ => scioCounter.inc()))
+    sc.transform("Transform2")(_.parallelize(11 to 20).tap(_ => scioCounter.inc()))
+
+    sc.run()
+      .waitUntilDone()
+      .counter(scioCounter)
+      .committed shouldBe Some(20)
+  }
+
   "PipelineOptions" should "propagate" in {
     trait Options extends DataflowPipelineOptions {
       @Required


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -700,6 +700,24 @@ class ScioContext private[scio] (
...
+
+  private[scio] def transform_[U <: POutput](f: ScioContext => U): U =
+    transform_(tfName)(f)
+
+  private[scio] def transform_[U <: POutput](name: String)(f: ScioContext => U): U = {
+    val sc: ScioContext = this
+    applyInternal(
+      name,
+      new PTransform[PBegin, U]() {
+        override def expand(pBegin: PBegin): U = f(sc)
```
clairemcginty, line 716: This line is kind of sketchy in that it *technically* throws away the supplied `pBegin` -- ideally we'd have a stable way to go from a `PInput` to its singleton enclosing `ScioContext`, but that's non-trivial...

@RustedBones , any ideas?

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -700,6 +699,30 @@ class ScioContext private[scio] (
...
+    val sc: ScioContext = this
+    applyInternal(
+      name,
+      new PTransform[PBegin, U]() {
+        override def expand(pBegin: PBegin): U = {
+          assert(
+            _pipeline.equals(pBegin.getPipeline),
+            "transform() was invoked with a different Pipeline than the current ScioContext's Pipeline"
+          )
+          f(sc)
```
clairemcginty, line 720: These lines is kind of sketchy in that they technically throw away the supplied pBegin -- ideally we'd have a stable way to go from a PInput to its singleton enclosing ScioContext, but that's non-trivial...

@RustedBones , any ideas?

RustedBones:
I think [customInput](https://spotify.github.io/scio//api/com/spotify/scio/ScioContext.html#customInput%5BT,I%3E:org.apache.beam.sdk.values.PBegin%3C:org.apache.beam.sdk.values.PInput%5D(name:String,transform:org.apache.beam.sdk.transforms.PTransform%5BI,org.apache.beam.sdk.values.PCollection%5BT%5D%5D):com.spotify.scio.values.SCollection%5BT%5D) was doing exactly that and allows for mocking the IOs

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -700,6 +700,30 @@ class ScioContext private[scio] (
...
+  def transform[U](f: ScioContext => SCollection[U]): SCollection[U] = transform(this.tfName)(f)
+
+  def transform[U](name: String)(f: ScioContext => SCollection[U]): SCollection[U] =
+    wrap(transform_(name)(f(_).internal))
+
+  private[scio] def transform_[U <: POutput](f: ScioContext => U): U =
+    transform_(tfName)(f)
+
+  private[scio] def transform_[U <: POutput](name: String)(f: ScioContext => U): U = {
+    val sc: ScioContext = this
```
RustedBones, line 712: are you sure the context can be accessed as a closure ? IIRC it can throw `NotSerializableException`

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
RustedBones: We so not check in the `SCollection.transfrom` that the resulting `SCollection's` context is the same as the original. IMHO it's not worth. But if we want the extra check, we should add it to the other one too

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -700,6 +700,30 @@ class ScioContext private[scio] (
...
+    applyInternal(
+      name,
+      new PTransform[PBegin, U]() {
+        override def expand(pBegin: PBegin): U = {
+          assert(
+            _pipeline.equals(pBegin.getPipeline),
```
RustedBones, line 718:
```suggestion
            _pipeline == pBegin.getPipeline,
```

scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -700,6 +700,30 @@ class ScioContext private[scio] (
...
+            "transform() was invoked with a different Pipeline than the current ScioContext's Pipeline"
+          )
+          f(sc)
```
RustedBones, line 721:
```suggestion
          f(ScioContext.this)
```

RustedBones (approved):
Nice!

PR #5040, opened 2023-10-20
IT test with null and empty values

RustedBones:


scio-google-cloud-platform/src/it/scala/com/spotify/scio/bigquery/types/StorageIT.scala
```diff
@@ -110,62 +65,43 @@ class StorageIT extends AnyFlatSpec with Matchers {
...
+    val p = sc.typedBigQuery[Nested]().internal
     PAssert.that(p).containsInAnyOrder(expected)
     sc.run()
   }
 
-  it should "work with rowRestriction" in {
-    val expected =
-      (0 until 5).map(i => (i.toLong, s"s$i", i.toLong, s"s$i", i.toLong, s"s$i")).asJava
+  it should "work with selectedFields" in {
+    val expected = (0 until 10).map(newNestedWithFields).asJava
```
RustedBones, line 79: This test now fails with
```Expected: iterable with items [
<NestedWithFields(Required$2(0,0),None)>, <NestedWithFields(Required$2(1,1),Some(Optional$2(1)))><NestedWithFields(Required$2(2,2),Some(Optional$2(2)))>, ...] in any order
     but: not matched: <NestedWithFields(Required$2(0,0),Some(Optional$2(0)))>
```

Tried to debug but it seems it comes from the bigquery session. avro byte sent tells the `optional.int` field is set to `0`. There is probably an issue with projection in the library.

RustedBones:
See https://github.com/googleapis/java-bigquerystorage/issues/2284

PR #5062, opened 2023-11-09
Add a ShardRangeString for sqlserver's uniqueidentifier type

avandel:
To sort (order by, min/max, etc) a unique identifier(guid) SQL server has its own algo which is very well described in this thread https://stackoverflow.com/questions/7810602/sql-server-guid-sort-algorithm-why
If we used the uuidShardRangeStringCodec for UUIDs, it wouldn't work because it considers that the underlying database is doing a lexicographical sort.
I therefore implemented a new shard string for SQL server uuids which takes into account the SQL server sorting method.


scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardString.scala
```diff
@@ -96,4 +100,85 @@ object RangeShardStringCodec {
...
+          bytes(0),
+          bytes(1),
+          bytes(2),
+          bytes(3),
+          bytes(4),
+          bytes(5),
+          bytes(6),
+          bytes(7),
+          bytes(9),
+          bytes(8),
```
RustedBones, line 157: According to the post, there are should be many 'groups' here. shouldn't the order reverse be
`3, 2, 1, 0, 5, 4, 7, 6, 9, 8, 15, 14, 13, 12, 11, 10` ?

RustedBones (commented):
Thanks for the PR. I'm not sure the algo proposed match the thread algorithm. I tried to find some doc, but the [official one](https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/compare-guid-uniqueidentifier-values?view=sql-server-ver16) only mentions 

> the last six bytes of a value are most significant.

But nothing on the lsb ordering

avandel:
> Thanks for the PR. I'm not sure the algo proposed match the thread algorithm. I tried to find some doc, but the [official one](https://learn.microsoft.com/en-us/sql/connect/ado-net/sql/compare-guid-uniqueidentifier-values?view=sql-server-ver16) only mentions
> 
> > the last six bytes of a value are most significant.
> 
> But nothing on the lsb ordering

Hello,

I think the algo described in the thread is OK. For example, with this query

```
SELECT * FROM (
  SELECT cast('00000000-7053-4BC3-0001-000034F5C77F' AS uniqueidentifier) as col1 
  UNION SELECT cast('00000000-7053-4BC3-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000000-7053-4BC3-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000000-7053-4BC4-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000000-7053-5BC4-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000000-7053-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000000-7054-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000010-7055-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000001-7055-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000001-7055-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('10000011-7055-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1
  UNION SELECT cast('00000111-7055-4BC5-0002-000034F5C78F' AS uniqueidentifier) as col1

) AS T ORDER BY col1;
```

I obtain this result 
```00000000-7053-4BC3-0001-000034F5C77F
00000000-7053-4BC3-0002-000034F5C78F
00000000-7053-4BC4-0002-000034F5C78F
00000000-7053-5BC4-0002-000034F5C78F
00000000-7053-4BC5-0002-000034F5C78F
00000000-7054-4BC5-0002-000034F5C78F
00000001-7055-4BC5-0002-000034F5C78F
00000010-7055-4BC5-0002-000034F5C78F
10000011-7055-4BC5-0002-000034F5C78F
00000111-7055-4BC5-0002-000034F5C78F
```
which is expected according the algo mentioned.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardString.scala
avandel: I take `0, 1, 2, 3, 4, 5, 6, 7, 9, 8, 15, 14, 13, 12, 11, 10` because they say the groups 10-15 and 8-9 are stored as big-endian and the others as little-endian. After I reverse the array and I build a Bigint giving more weight to the bytes on the left.

RustedBones:
The thread mentions
> The comparison is made by looking at byte "groups" right-to-left, and left-to-right within a byte "group". A byte group is what is delimited by the '-' character. More technically, we look at bytes {10 to 15} first, then {8-9}, then {6-7}, then {4-5}, and lastly {0 to 3}.

Using this ordering, sorting the given UUID above gives the same result

RustedBones:
WDYT of this: https://github.com/avandel/scio/compare/main..spotify:scio:sqlguid-impl2

kellen:
Failing test is fixed in a PR I made elsewhere

scio-jdbc/src/it/scala/com/spotify/scio/jdbc/PopulateTestData.scala
```diff
@@ -0,0 +1,41 @@
+package com.spotify.scio.jdbc
+
+import com.spotify.scio.jdbc.sharded.JdbcUtils
+
+import java.sql.DriverManager
```
avandel, line 5: You don't use this import here I think

RustedBones: Thanks. Looks our scalafix is not running in the `IntegrationTest` config since this has not been spotted by our CI.

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/sharded/ShardString.scala
```diff
@@ -87,13 +92,50 @@ object RangeShardStringCodec {
   implicit val uuidUpperShardRangeStringCodec: RangeShardStringCodec[UuidUpperString] =
     uuidShardRangeStringCodec(str => UuidUpperString(str.toUpperCase))
 
+  def sqlServerUuidShardRangeStringCodec[S <: SqlServerUuidString](
+    uuidStringBuilder: String => S
+  ): RangeShardStringCodec[S] =
+    new RangeShardStringCodec[S] {
+
+      def encode(bigInt: BigInt): S = {
```
kellen, line 100: Can we add a link to the backing SQL Server docs here?

kellen (approved):

PR #5064, opened 2023-11-10
Support mixed FileOperations per source in SortedBucketSource

clairemcginty:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -434,11 +434,36 @@ public BucketedInput(
         String filenameSuffix,
```
clairemcginty, line 434: Hmm, writing a test for this now, we'll need a way to couple `filenameSuffix` to the specific directory as well

scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/SortedBucketSourceTest.java
```diff
@@ -692,6 +684,85 @@ public void testUniqueTupleTagIdSecondaryKey() {
...
+                                    KV.of(".avro", AvroFileOperations.of(AvroGeneratedUser.class)),
+                                TestUtils.fromFolder(rhsFolder).toString(),
+                                    KV.of(
+                                        ".parquet",
+                                        ParquetAvroFileOperations.of(AvroGeneratedUser.class))),
+                            null)),
+                    TargetParallelism.max(),
+                    "metrics-key")));
+
+    PAssert.thatMap(output)
```
clairemcginty, line 751: So this test fails with an assertion error:

```
java.lang.IllegalArgumentException: All FileOperations Coders must use the same encoding type; found: [org.apache.beam.sdk.io.AvroGeneratedUser, org.apache.avro.generic.GenericRecord]
```

I'd forgotten that `ParquetAvroFileOperations` treats record classes as GenericRecords by default ... hmm..... will have to think on this one :(

clairemcginty: Maybe we don't need the Coder check at all

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -453,20 +484,19 @@ public Predicate<V> getPredicate() {
     }
 
     public Coder<V> getCoder() {
-      return fileOperations.getCoder();
+      // Todo what if two input directories use a different Coder for the same type?
```
clairemcginty, line 487: For the same SpecificRecord type `T`, `ParquetFileOperations` returns a `Coder<GenericRecord>`, compared to `AvroFileOperations`, which returns a `Coder<T>`... while my unit test passed, I'm not totally comfortable with the ramifications of picking a coder at random 😬

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroBucketMetadata.java
```diff
@@ -178,25 +178,45 @@ public void populateDisplayData(Builder builder) {
 
   @Override
   public boolean isPartitionCompatibleForPrimaryKey(BucketMetadata o) {
-    if (o == null || getClass() != o.getClass()) return false;
-    AvroBucketMetadata<?, ?, ?> that = (AvroBucketMetadata<?, ?, ?>) o;
-    return getKeyClass() == that.getKeyClass() && keyField.equals(that.keyField);
+    if (o == null) return false;
```
clairemcginty, line 181: todo - make this generic somehow -- i.e. some kind of method `public boolean Set<Class<? extends BucketMetadata>> isCompatibleWith()` ?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -578,22 +615,60 @@ public String toString() {
...
-      outStream.writeObject(fileOperations);
       outStream.writeObject(predicate);
       outStream.writeObject(keying);
+
+      long numDistinctFileSuffixes =
+          directories.values().stream().map(KV::getKey).distinct().count();
+      long numDistinctFileOperations =
+          directories.values().stream().map(kv -> kv.getValue().getClass()).distinct().count();
+
+      // If all partitions use the same file operations type, don't keep re-encoding it
```
clairemcginty, line 626: In the past serialized transform size has been an issue (if hundreds+ of partitions are being read), but this solution is definitely a bit convoluted 😅

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -453,20 +495,18 @@ public Predicate<V> getPredicate() {
     }
 
     public Coder<V> getCoder() {
-      return fileOperations.getCoder();
+      return directories.entrySet().iterator().next().getValue().getValue().getCoder();
```
kellen, line 498: Add a variable name here for clarity?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
kellen: You could potentially add a header to write out the distinct file operations with indicies and have a reference to those indicies in each of the output objects. Slight increase in overhead if all file ops are different, but would decrease payload size even more if there are some (but not exactly 1) shared operations

kellen:
How might a user use this change?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -290,14 +290,47 @@ <K> byte[] encodeKeyBytes(K key, Coder<K> coder) {
...
+    return new HashSet<>();
+  }
 
   public abstract K1 extractKeyPrimary(V value);
 
   public abstract K2 extractKeySecondary(V value);
 
+  public abstract int hashPrimaryKeyMetadata();
+
+  public abstract int hashSecondaryKeyMetadata();
```
RustedBones, line 332: can we reduce visibility ?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/JsonBucketMetadata.java
```diff
@@ -148,32 +148,12 @@ public void populateDisplayData(Builder builder) {
   }
 
   @Override
-  public boolean isPartitionCompatibleForPrimaryKey(final BucketMetadata o) {
-    if (o == null || getClass() != o.getClass()) return false;
-    JsonBucketMetadata<?, ?> that = (JsonBucketMetadata<?, ?>) o;
-    return getKeyClass() == that.getKeyClass()
-        && keyField.equals(that.keyField)
-        && Arrays.equals(keyPath, that.keyPath);
```
RustedBones, line 156: This is strange on this one we were checking the path. I suppose this is fine to drop

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
```diff
@@ -493,11 +493,8 @@ public abstract static class TransformOutput<K1, K2, V> implements Serializable
 
   /** Represents a single sorted-bucket source written using {@link SortedBucketSink}. */
   public abstract static class Read<V> implements Serializable {
-
     @Nullable
-    abstract ImmutableList<String> getInputDirectories();
-
-    abstract String getFilenameSuffix();
```
RustedBones, line 500: why has this been removed ?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
clairemcginty: that's a good idea - will give it a try

clairemcginty:
> How might a user use this change?

You can use it if you're instantiating a SortedBucketSource transform directly (i.e. creating your own `BucketedInput` instances), or extend a new `SortedBucketIO.Read` implementation that accepts mixed file types

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
clairemcginty: It's no longer a guarantee that a `SortedBucketIO.Read` implementation has a single `filenameSuffix` -- i.e. it could be both Parquet + Avro

I guess this could cause issues if we want to support Taps for SMB reads later down the line -- maybe the Builders could support an `ImmutableMap<String, String>` mapping input directories --> filename suffixes

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/JsonBucketMetadata.java
clairemcginty: It should be redundant since `keyPath` is [derived deterministically](https://github.com/spotify/scio/blob/main/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/JsonBucketMetadata.java#L117) from `keyField` !

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperations.java
```diff
@@ -141,7 +146,9 @@ protected FileIO.Sink<ValueT> createSink() {
   @SuppressWarnings("unchecked")
   @Override
   public Coder<ValueT> getCoder() {
-    return (AvroCoder<ValueT>) AvroCoder.of(getSchema());
+    return recordClass == null
+        ? (AvroCoder<ValueT>) AvroCoder.of(getSchema())
+        : AvroCoder.of(recordClass, true);
```
shnapz, line 151: How did this work before? Codec for GenericRecord was used for SpecificRecords?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
```diff
@@ -189,27 +192,18 @@ public void populateDisplayData(DisplayData.Builder builder) {
   }
 
   @Override
-  public boolean isPartitionCompatibleForPrimaryKey(BucketMetadata o) {
-    if (o == null || getClass() != o.getClass()) return false;
-    ParquetBucketMetadata<?, ?, ?> that = (ParquetBucketMetadata<?, ?, ?>) o;
-    return getKeyClass() == that.getKeyClass() && keyField.equals(that.keyField);
+  int hashPrimaryKeyMetadata() {
```
shnapz, line 195: those `hash` functions seem to be the same in all metadata implementations, aren't they? maybe push them down to the base class?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperations.java
RustedBones: I made a bug in beam https://github.com/apache/beam/pull/29518. In current version it will use `SpecificData`. Prefer explicit `AvroCoder.reflect`

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
clairemcginty: They are the same in practice, but in theory, a `BucketMetadata` doesn't _have_ to have a `keyField`/`keyPath` and those aren't part of the base class, so we can't generalize it unfortunately :(

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
clairemcginty: yeah 😅 I ended up adding a public getter for directories/FileOperations on `BucketedInput`, which should allow us to implement SMB taps later on

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetBucketMetadata.java
shnapz: Oh Ok!

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
```diff
@@ -493,15 +493,9 @@ public abstract static class TransformOutput<K1, K2, V> implements Serializable
...
-
-    @Nullable
-    abstract ImmutableList<String> getInputDirectories();
-
-    abstract String getFilenameSuffix();
-
     public abstract TupleTag<V> getTupleTag();
 
-    protected abstract BucketedInput<V> toBucketedInput(SortedBucketSource.Keying keying);
+    public abstract BucketedInput<V> toBucketedInput(SortedBucketSource.Keying keying);
```
clairemcginty, line 498: We can use this public method to construct SMB taps from the Scala API bindings, which take in `SortedBucketIO.Read` objects 👍

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -290,14 +290,47 @@ <K> byte[] encodeKeyBytes(K key, Coder<K> coder) {
...
+    return new HashSet<>();
+  }
 
   public abstract K1 extractKeyPrimary(V value);
 
   public abstract K2 extractKeySecondary(V value);
 
+  abstract int hashPrimaryKeyMetadata();
+
+  abstract int hashSecondaryKeyMetadata();
```
farzad-sedghi, line 332: the secondary key impl. (lexicographic) we have today is one way of implementing it, which in some cases like analytics is less efficient as opposed to sth like z-ordering. Might be nice to think a bit more to find out with a more future proof API e.g. support N keys?

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -578,22 +631,59 @@ public String toString() {
...
+          fileOperationsEncoding.put(i, KV.of(fileOps.getKey(), fileOps.getValue()));
+          i++;
+        }
+        directoriesEncoding.put(entry.getKey(), fileOperationsMetadata.get(metadataKey));
+      }
+
+      final Coder<FileOperations<V>> fileOperationsCoder =
+          new FileOperations.FileOperationsCoder<>();
+
+      MapCoder.of(VarIntCoder.of(), KvCoder.of(StringUtf8Coder.of(), fileOperationsCoder))
```
RustedBones, line 659: This coder can probably be a member

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSource.java
```diff
@@ -578,22 +631,59 @@ public String toString() {
...
+          .encode(fileOperationsEncoding, outStream);
+
+      MapCoder.of(ResourceIdCoder.of(), VarIntCoder.of()).encode(directoriesEncoding, outStream);
```
RustedBones, line 662: This one too

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/FileOperations.java
```diff
@@ -249,4 +255,24 @@ private ReadableFile toReadableFile(ResourceId resourceId) {
...
+    @Override
+    public FileOperations<T> decode(InputStream inStream) throws CoderException, IOException {
+      final ObjectInputStream ois = new ObjectInputStream(inStream);
+      try {
+        return (FileOperations<T>) ois.readObject();
+      } catch (ClassNotFoundException e) {
+        throw new CoderException(e);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
clairemcginty: when i was working on this in December I remember having type erasure issues trying to get SerializableCoder working with `FileOperations<V>` type -- but I just tried again and got it working, so 🤷‍♀️ updated the PR!

RustedBones (approved):

PR #5069, opened 2023-11-13
Drop collection compat shim in favor of official scala-collection-compat

RustedBones:


clairemcginty (approved):

scio-core/src/main/scala-2.12/com/spotify/scio/coders/instances/JavaCollectionWrappers.scala
```diff
@@ -0,0 +1,42 @@
...
+    Class.forName("scala.collection.convert.Wrappers$JIterableWrapper")
+  val JCollectionWrapperClass: Class[_] =
+    Class.forName("scala.collection.convert.Wrappers$JCollectionWrapper")
+  val JListWrapperClass: Class[_] =
+    Class.forName("scala.collection.convert.Wrappers$JListWrapper")
+
+  object JIterableWrapper {
+    def unapply(arg: Any): Option[JIterable[_]] = arg match {
+      case arg if arg.getClass == JListWrapperClass =>
+        val underlying = JListWrapperClass
```
RustedBones, line 34: Checking against the wrong class. This must be `JIterableWrapperClass`

PR #5106, opened 2023-12-05
Add location to BigQuery LoadOps

f-loris:
This PR adds the option to set the `location` for a BigQuery load job. This is needed that load works correctly when loading data to a table located in a single region (non EU/US). When for example loading data to a table in `me-central2`, the load itself works, but the `waitForJobs` never completes as BigQuery API always returns a 404 when querying the job infos: 
```
BigQuery request failed: id: <project>-2519647d-c737-4e12-acc2-d7cc6861f2b8, error: com.google.api.client.googleapis.json.GoogleJsonResponseException: 404 Not Found
GET https://bigquery.googleapis.com/bigquery/v2/projects/<project>/jobs/<project>-2519647d-c737-4e12-acc2-d7cc6861f2b8
{
  "code": 404,
  "errors": [
    {
      "domain": "global",
      "message": "Not found: Job <project><project>-2519647d-c737-4e12-acc2-d7cc6861f2b8",
      "reason": "notFound"
    }
  ],
  "message": "Not found: Job <project>:<project>-2519647d-c737-4e12-acc2-d7cc6861f2b8",
  "status": "NOT_FOUND"
}
```

When the location ist set correctly, `waitForJobs` will be able the get the infos approriately as the location will be set here: https://github.com/spotify/scio/blob/11e0012ef784851be82f0e96d857e89a6196ce30/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/JobOps.scala#L97

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
@@ -49,7 +49,8 @@ final private[client] class LoadOps(client: Client, jobService: JobOps) {
     skipLeadingRows: Int = 0,
     fieldDelimiter: Option[String] = None,
     ignoreUnknownValues: Boolean = false,
-    encoding: Option[String] = None
+    encoding: Option[String] = None,
+    location: Option[String] = None
   ): Try[TableReference] =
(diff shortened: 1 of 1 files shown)


RustedBones (approved):
Makes sense! Thanks

PR #5119, opened 2023-12-14
Use avro builder API

RustedBones:


scio-avro/src/main/scala/com/spotify/scio/avro/dynamic/syntax/AvroDynamicSCollectionSyntax.scala
```diff
@@ -142,22 +142,19 @@ final class DynamicProtobufSCollectionOps[T <: Message](private val self: SColle
...
+      nm.putAll((metadata ++ ProtobufUtil.schemaMetadataOf(ct)).asJava)
+
       val sink = BAvroIO
-        .sinkViaGenericRecords(
-          avroSchema,
-          (element: T, _: Schema) => AvroBytesUtil.encode(elemCoder, element)
-        )
+        .sink(AvroBytesUtil.schema)
+        .asInstanceOf[BAvroIO.Sink[T]]
+        .withDatumWriterFactory(AvroBytesUtil.datumWriterFactory)
```
RustedBones, line 157: Small rewrite here to move away from deprecated `sinkViaGenericRecords` API

scio-avro/src/main/scala/com/spotify/scio/coders/AvroBytesUtil.scala
```diff
@@ -17,41 +17,47 @@
...
+  val schema: Schema = SchemaBuilder
+    .record("AvroBytesRecord")
+    .fields()
+    .requiredBytes("bytes")
+    .endRecord()
+
+  private val byteField = schema.getField("bytes")
+
+  def datumWriterFactory[T: Coder]: DatumWriterFactory[T] = {
+    val bCoder = CoderMaterializer.beamWithDefault(Coder[T])
```
RustedBones, line 39: Logic change here since we do not use the coder options for byte serialization. I think this cased of an issue when consuming data from PubSub, and tunring one Nullable, records could not be read.

clairemcginty (approved):
🙌

Issue #5115, opened 2023-12-13
ParquetAvroFileOperations does not set dataSupplier


clairemcginty:
Parquet-Avro will default to using a SpecificData model, which breaks GenericRecord reads if there's an available class on the classpath with the same name as the Avro schema due to SpecificData's [class lookup implementation](https://github.com/apache/avro/blob/release-1.8.2/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L368). In contrast, ParquetAvroIO in scio-parquet [always sets dataSupplier for generic reads](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala#L206-L211).

I put up a branch with a [sample repro](https://github.com/spotify/scio/compare/main...smb-parquet-avro-datasupplier)

Issue #5114, opened 2023-12-13
ParquetAvroTap does not set projection or dataSupplier


clairemcginty:
ParquetAvroTap does not use the value of `params.suffix`: https://github.com/spotify/scio/blob/46e72af/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala#L321-L343

repro in ParquetAvroIOTest:

```scala
  "ParquetAvroTap" should "work with projections" in {
    val schema = new Schema.Parser().parse(
      """
        |{
        |"type":"record",
        |"name":"TestRecordProjection",
        |"fields":[{"name":"int_field","type":["null", "int"]}]}
        |""".stripMargin
    )

    implicit val coder = avroGenericRecordCoder(schema)

    ParquetAvroTap(
      s"${testDir.toPath.resolve("*.parquet")}",
      ParquetAvroIO.ReadParam(identity[GenericRecord], schema)
    ).value.foreach { gr =>
      gr.get("int_field") should not be (null)
      gr.get("string_field") should be (null) // this assertion fails
    }
  }
```

Not setting dataSupplier can also be problematic for GenericRecord reads -- Parquet-Avro defaults to using a `SpecificData` model for reads, which will fail if a class matching the schema name [exists on the classpath but doesn't implement a no-arg constructor](https://github.com/apache/avro/blob/release-1.8.2/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L368).

(Note: for anyone impacted by this issue, you can work around it by setting projection/data supplier via Configuration:

```scala
val conf = ParquetConfiguration.of(
  AvroReadSupport.AVRO_DATA_SUPPLIER -> classOf[GenericDataSupplier] // Set data supplier
)

val projection: Schema = ...
AvroReadSupport.setRequestedProjection(conf, projection) // Set projection

// Use configuration in tap
val tap = ParquetAvroTap(path, ParquetAvroIO.ReadParam(identity[GenericRecord], conf = conf))
```

shnapz:
@clairemcginty is the following a typo? `ParquetAvroTap .value` does use a suffix to create a filePattern
> ParquetAvroTap does not use the value of `params.suffix`

clairemcginty:
> @clairemcginty probably you mean `params.projection` here:
> 
> > ParquetAvroTap does not use the value of `params.suffix`

Ooops, yes, I meant projection 🤦‍♀️

PR #5121, opened 2023-12-14
In SMB and ParquetAvroIOTap set GenericDataSupplier and read schemas

shnapz:
Addressing:
- https://github.com/spotify/scio/issues/5115
- https://github.com/spotify/scio/issues/5114


scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroSink.scala
```diff
@@ -0,0 +1,38 @@
+package com.spotify.scio.parquet.avro
```
shnapz, line 1: moved this class out from package object

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroTap.scala
```diff
@@ -0,0 +1,45 @@
...
+
+final case class ParquetAvroTap[A, T: ClassTag: Coder](
+  path: String,
+  params: ParquetAvroIO.ReadParam[A, T]
+) extends Tap[T] {
+  override def value: Iterator[T] = {
+    val filePattern = ScioUtil.filePattern(path, params.suffix)
+    val conf = ParquetConfiguration
+      .ofNullable(params.conf)
+    conf.setReadSchemas(params)
```
shnapz, line 23: Except of these two lines, the rest of the class remains the same, just moved

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroSortedBucketIO.java
```diff
@@ -241,21 +241,24 @@ public Read<T> withPredicate(Predicate<T> predicate) {
       return toBuilder().setPredicate(predicate).build();
     }
 
+    @SuppressWarnings("unchecked")
+    @Override
+    FileOperations<T> getFileOperations() {
```
shnapz, line 246: Added this to reflect the same style as in Write and TransformOutput

scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperationsTest.java
```diff
@@ -95,10 +102,26 @@ public void testGenericRecord() throws Exception {
...
+    fileOperations.iterator(file).forEachRemaining(actual::add);
+
+    Assert.assertEquals(USER_RECORDS, actual);
+  }
+
   @Test
   public void testSpecificRecord() throws Exception {
     final ParquetAvroFileOperations<AvroGeneratedUser> fileOperations =
-        ParquetAvroFileOperations.of(AvroGeneratedUser.getClassSchema());
+        ParquetAvroFileOperations.of(AvroGeneratedUser.class);
```
shnapz, line 124: In Prod we use the overload with Schema for GenericRecords only

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/package.scala
```diff
@@ -42,27 +40,30 @@ package object avro extends Syntax {
...
-        .withRowGroupSize(rowGroupSize)
-        .build
+  def classIsSpecificRecord[T: ClassTag]: Boolean = {
+    val avroClass: Class[T] = ScioUtil.classOf[T]
+    classOf[SpecificRecord] isAssignableFrom avroClass
+  }
+
+  implicit class ConfigurationSyntax(conf: Configuration) {
+
+    def setReadSchemas[A, T](params: ParquetAvroIO.ReadParam[A, T]): Unit = {
```
clairemcginty, line 50: What do you think of broadening the scope of this method and moving into a class method of `ParquetAvroIO.ReadParam`? Like:

```scala
 final case class ReadParam[A: ClassTag, T: ClassTag] private (
    projectionFn: A => T,
    projection: Schema = ReadParam.DefaultProjection,
    predicate: FilterPredicate = ReadParam.DefaultPredicate,
    conf: Configuration = ReadParam.DefaultConfiguration,
    suffix: String = null
  ) {
   val avroClass: Class[A] = ScioUtil.classOf[A]
   val isSpecific = classOf[SpecificRecord] ScioUtil.classOf[T] 
   
   private[avro] def getReadConfiguration(): Configuration = {
       val c = ParquetConfiguration.ofNullable(conf)
       // Set schema/projection
       val readSchema: Schema = if (isSpecific) ReflectData.get().getSchema(avroClass) else params.projection

       AvroReadSupport.setAvroReadSchema(readSchema)
       if (params.projection != null) { /* set projection... */ }
       if (params.predicate != null) { /* set predicate */ }
   
       // Set data supplier
       if (!isSpecific && conf.get(AvroReadSupport.AVRO_DATA_SUPPLIER) == null) {
         ....
      }
  }
```

That way it's easy to access from both `read` and `tap` methods 👍

shnapz: This is great! I missed it, the great place for common functions on input conf!

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -219,13 +212,27 @@ object ParquetAvroIO {
       }
     }
 
+    def setReadSchemas[A, T](params: ParquetAvroIO.ReadParam[A, T]): Unit = {
```
RustedBones, line 215: `ReadParam` is already parameterized with `[A, T]`. I don't think we want type shadowing here

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroTap.scala
```diff
@@ -0,0 +1,64 @@
...
+
+final case class ParquetAvroTap[A, T: ClassTag: Coder](
+  path: String,
+  params: ParquetAvroIO.ReadParam[A, T]
+) extends Tap[T] {
+  override def value: Iterator[T] = {
+    val filePattern = ScioUtil.filePattern(path, params.suffix)
+    val conf = ParquetConfiguration
+      .ofNullable(params.conf)
+    params.setReadSchemas(params)
```
RustedBones, line 40: do we want to do this ?

shnapz: probably was tired when committed this, I am changing the approach

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
shnapz: Removing this function

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -159,7 +145,7 @@ object ParquetAvroIO {
   object ReadParam {
     val DefaultProjection: Schema = null
     val DefaultPredicate: FilterPredicate = null
-    val DefaultConfiguration: Configuration = null
+    def DefaultConfiguration: Configuration = ParquetConfiguration.empty()
```
RustedBones, line 148: TBH, I'd prefer keeping this as null and handle it internally

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -198,67 +190,57 @@ object ParquetAvroIO {
...
       sc.applyTransform(
         ParquetRead.read[A, T](
           ReadSupportFactory.avro,
           new SerializableConfiguration(conf),
-          path,
-          Functions.serializableFn(cleanedProjectionFn)
+          filePattern,
+          Functions.serializableFn { x =>
+            cleanedProjectionFn(x)
+          }
```
RustedBones, line 230: Nit. can we revert to previous syntax?

RustedBones (commented):
I see that using an empty config as parameter allows some simplification, but I'd still prefer to pass a null there for consistency.

This is a personal preference, not a blocker. @clairemcginty  might also have an opinion about it.

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
clairemcginty: Agree, I think, partly because it matches other ScioIO param conventions and partly because I'm worried that over time we'll forget why this was made a `def` and refactor it back into a `val` :)

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -315,32 +297,25 @@ object ParquetAvroIO {
...
-        .build()
-      new Iterator[T] {
-        private var current: A = reader.read()
-        override def hasNext: Boolean = current != null
-        override def next(): T = {
-          val r = params.projectionFn(current)
-          current = reader.read()
-          r
-        }
+    private[avro] def setupConfigAndGetSchema[T: ClassTag](): Schema = {
```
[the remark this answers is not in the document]
shnapz: Agreed. In case of write it is used only once, but I aimed to have similarity with read param handling. I will inline

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperations.java
```diff
@@ -94,7 +96,24 @@ public static <V extends IndexedRecord> ParquetAvroFileOperations<V> of(
...
+    return of(schema, DEFAULT_COMPRESSION, predicate, conf);
+  }
+
+  /**
+   * Designated for GenericRecords only
+   * */
+  public static <V extends IndexedRecord> ParquetAvroFileOperations<V> of(
+      Schema schema, CompressionCodecName compression, FilterPredicate predicate,
+      Configuration conf) {
+    if (conf.get(AvroReadSupport.AVRO_DATA_SUPPLIER) == null) {
```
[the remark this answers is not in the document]
shnapz: Oh right, with the help of newly introduced `private final Class<ValueT> recordClass;` we can do that

scio-parquet/src/test/scala/com/spotify/scio/parquet/avro/ParquetAvroIOTest.scala
```diff
@@ -79,208 +82,278 @@ class ParquetAvroIOTest extends ScioIOSpec with TapSpec with BeforeAndAfterAll {
...
+    )
+
+  it should "work with specific records" in {
     val xs = (1 to 100).map(AvroUtils.newSpecificRecord)
-    testTap(xs)(_.saveAsParquetAvroFile(_))(".parquet")
-    testJobTest(xs)(ParquetAvroIO(_))(
-      _.parquetAvroFile[TestRecord](_).map(identity)
-    )(_.saveAsParquetAvroFile(_))
+
+    forAllCases(readConfigs) { case (c, _) =>
```
shnapz, line 102: the main idea of changes in this file is to test both SDR and Legacy in the same test cases

PR #5127, opened 2023-12-19
Move class definitions outside package objects

RustedBones:
Class defined in package object create the following warning:

```
[warn] ...it is not recommended to define classes/objects inside of package objects.
[warn] If possible, define class xxx in package types instead
```

Moved the class definition outside the package object.

implicit value classes were not affected by warning but to increase consistency across all modules, I created a syntax package and a deprecated type alias in case the value class was used explicitly


kellen:
@RustedBones did you end up doing this in another PR or am I hallucinating? if not, should we merge this for 0.14?

RustedBones:
No this has never been approved, and it's not required for 0.14.
I'm closing this one. Will do another pass to fix warnings later

PR #5128, opened 2023-12-20
Make Sparkey testable

### LINKED ISSUE #3173: Support sparkey in JobTest
When writing a sparkey in one job via e.g. `asSparkey` and reading in another job via e.g. `sparkeySideInput`, testing the second job is clunky. Ideally a `SparkeyIO` class would be available to be used via `JobTest`

### LINKED ISSUE #4819: `map` on SideInput is a terrible idea and should be deprecated.
Currently it is possible to do something along the line of:

```scala
coll.asSetSingletonSideInput
      .map(doSmth)
```

When using a Global window (always the case for batch workflows), the result of mapping over the `SideInput` value is not cached. Every time the side input is accessed, `doSmth` is applied to it. When `doSmth` is not a trivial operation, it can make the job _very_ slow.

I think mapping on a `SideInput` should just not be supported. If the values in that SideInput are to be modified, it should be done _before_ the `SCollection` is turned into a view. For example:

```scala
coll.map(doSmth).asSetSingletonSideInput
```  

I implemented this exact fix on an expensive production workflow and it made the runtime go from ~1h30min to ~30min.

kellen:
Adds `SparkeyIO`, a `TestIO` (rather than full-fledged `ScioIO`). Deprecates sideinput `map`.

Suggestions around naming the output Sparkey IO, currently `SparkeyIO.output[K, V]`, are welcome as this feels awkward. An alternative is to use `SparkeyIOOutput[K, V]` directly, but this is also ugly.

Instead of #3172

Fixes #3173
Fixes #4819

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -130,7 +130,7 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K, 
 
 @deprecated(since = "0.14.0")
 private[values] class DelegatingSideInput[A, B](val si: SideInput[A], val mapper: A => B)
-  extends SideInput[B] {
+    extends SideInput[B] {
 
   // Only update the cached value (and re-run the mapper) if the underlying SI does the same.
   override protected[values] def updateCacheOnGlobalWindow: Boolean = si.updateCacheOnGlobalWindow
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -52,6 +52,7 @@ trait SideInput[T] extends Serializable {
   /**
    * Create a new [[SideInput]] by applying a function on the elements wrapped in this SideInput.
    */
+  @deprecated(since = "0.14.0")
   def map[B](f: T => B): SideInput[B] = new DelegatingSideInput[T, B](this, f)
 
   private[scio] val view: PCollectionView[_]
@@ -127,6 +128,7 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K, 
     JMapWrapper.ofMultiMap(context.sideInput(view))
 }
 
+@deprecated(since = "0.14.0")
 private[values] class DelegatingSideInput[A, B](val si: SideInput[A], val mapper: A => B)
   extends SideInput[B] {
 
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -49,6 +49,11 @@ trait SideInput[T] extends Serializable {
     cache
   }
 
+  /**
+   * Create a new [[SideInput]] by applying a function on the elements wrapped in this SideInput.
+   */
+  def map[B](f: T => B): SideInput[B] = new DelegatingSideInput[T, B](this, f)
+
   private[scio] val view: PCollectionView[_]
 }
 
@@ -120,6 +125,17 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K, 
     extends SideInput[Map[K, Iterable[V]]] {
   override def get[I, O](context: DoFn[I, O]#ProcessContext): Map[K, Iterable[V]] =
     JMapWrapper.ofMultiMap(context.sideInput(view))
+}
+
+private[values] class DelegatingSideInput[A, B](val si: SideInput[A], val mapper: A => B)
+  extends SideInput[B] {
+
+  // Only update the cached value (and re-run the mapper) if the underlying SI does the same.
+  override protected[values] def updateCacheOnGlobalWindow: Boolean = si.updateCacheOnGlobalWindow
+
+  override def get[I, O](context: DoFn[I, O]#ProcessContext): B = mapper(si.get(context))
+
+  private[scio] val view: PCollectionView[_] = si.view
 }
 
 /** Encapsulate context of one or more [[SideInput]]s in an [[SCollectionWithSideInput]]. */
--- a/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
@@ -49,11 +49,6 @@ trait SideInput[T] extends Serializable {
     cache
   }
 
-  /**
-   * Create a new [[SideInput]] by applying a function on the elements wrapped in this SideInput.
-   */
-  def map[B](f: T => B): SideInput[B] = new DelegatingSideInput[T, B](this, f)
-
   private[scio] val view: PCollectionView[_]
 }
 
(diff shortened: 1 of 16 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
```diff
@@ -127,6 +128,7 @@ private[values] class MultiMapSideInput[K, V](val view: PCollectionView[JMap[K,
     JMapWrapper.ofMultiMap(context.sideInput(view))
 }
 
+@deprecated(since = "0.14.0")
```
clairemcginty, line 131: can we include a deprecation message advising users what methods to use instead?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,203 @@
...
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+
+/** Special version of [[com.spotify.scio.io.ScioIO]] for use with sparkey read methods. */
+case class SparkeyIO(path: String) extends TestIO[SparkeyReader] {
+  override val tapT: TapT.Aux[SparkeyReader, SparkeyReader] = TapOf[SparkeyReader]
+}
+
+/** Special version of [[com.spotify.scio.io.ScioIO]] for use with sparkey write methods. */
+case class SparkeyIOOutput[K, V](path: String) extends TestIO[(K, V)] {
```
clairemcginty, line 42: As a naming suggestion, maybe `SparkeyWriteIO`/`SparkeyReadIO`? just a thought

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,203 @@
...
+                  if (isUnsharded) baseUri else baseUri.sparkeyUriForShard(shard, numShards)
+
+                val srcResources = srcs ++ uri.paths
+                val dstResources = dsts ++ dstUri.paths
+
+                (srcResources, dstResources)
+            }
+
+          // rename source files to dest files
+          logger.info(s"Copying ${items.size} files.")
```
clairemcginty, line 175:
```suggestion
          logger.info(s"Copying ${items.size} files from temp to final GCS destination.")
```

scio-core/src/main/scala/com/spotify/scio/values/SideInput.scala
kellen: No this is just a pure "don't do this"

scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
```diff
@@ -182,7 +224,9 @@ class SparkeyTest extends PipelineSpec {
   it should "support .asSparkey with specified local file" in {
     val tmpDir = Files.createTempDirectory("sparkey-test-")
     val basePath = tmpDir.resolve("sparkey").toString
-    runWithContext(sc => sc.parallelize(sideData).asSparkey(basePath))
+    val sc = ScioContext()
+    sc.parallelize(sideData).asSparkey(basePath)
+    sc.run()
```
RustedBones, line 229:
```suggestion
    runWithRealContext()(_..parallelize(sideData).asSparkey(basePath))
```

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,205 @@
...
+object SparkeyIO {
+  @transient private lazy val logger = LoggerFactory.getLogger(this.getClass)
+
+  val DefaultNumShards: Short = 1
+  val DefaultSideInputNumShards: Short = 64
+  val DefaultCompressionType: CompressionType = CompressionType.NONE
+  val DefaultCompressionBlockSize: Int = 0
+
+  def apply(path: String): SparkeyTestIO[SparkeyReader] = SparkeyTestIO[SparkeyReader](path)
+  def output[K, V](path: String): SparkeyTestIO[(K, V)] = SparkeyTestIO[(K, V)](path)
```
RustedBones, line 50: If we rename class to `SparkeyIO`, we can get rid of those 2

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/MockSparkeyReader.scala
```diff
@@ -0,0 +1,78 @@
...
+    extends MockEntry[Array[Byte], Array[Byte]] {
+  override def getKey: Array[Byte] = k
+  override def getKeyAsString: String = new String(k, StandardCharsets.UTF_8)
+  override def getValue: Array[Byte] = v
+  override def getValueAsString: String = new String(v, StandardCharsets.UTF_8)
+}
+
+case class MockByteArraySparkeyReader(data: Map[Array[Byte], Array[Byte]])
+    extends MockSparkeyReader {
+  override def getAsByteArray(key: Array[Byte]): Array[Byte] = data(key)
```
RustedBones, line 77: shouldn't iterator be implemented here as well ?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/MockSparkeyReader.scala
```diff
@@ -0,0 +1,78 @@
...
+  override def getKeyAsString: String = k
+  override def getValue: Array[Byte] = v.getBytes(StandardCharsets.UTF_8)
+  override def getValueAsString: String = v
+}
+
+case class MockStringSparkeyReader(data: Map[String, String]) extends MockSparkeyReader {
+  override def getAsString(key: String): String = data(key)
+  override def iterator(): java.util.Iterator[Entry] = data.iterator.map { case (k, v) =>
+    MockStringEntry(k, v).asInstanceOf[Entry]
+  }.asJava
```
RustedBones, line 64: Prefer explicit type instead of casting
```suggestion
  override def iterator(): java.util.Iterator[Entry] = data.iterator.map[Entry] { case (k, v) =>
    MockStringEntry(k, v)
  }.asJava
```

scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
```diff
@@ -68,11 +102,19 @@ class SparkeyTest extends PipelineSpec {
   val bigSideData: IndexedSeq[(String, String)] =
     (0 until 100).map(i => (('a' + i).toString, i.toString))
 
+  "JobTest" should "support mocking sparkey" in {
+    val input = Map("a" -> "b", "c" -> "d")
+    JobTest[SparkeyJob.type]
+      .args("--input=foo", "--output1=bar", "--output2=baz")
+      .input(SparkeyIO("foo"), Seq(MockStringSparkeyReader(input)))
```
RustedBones, line 109: Can we instead have `SparkeyIO` expect a `Map[K, V]` (or `Seq[(K, V)]`) instead of a reader ? We would create the reader in the IO after `isTest` is checked.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.beam.sdk.io.FileSystems
+import org.apache.beam.sdk.io.fs.MoveOptions.StandardMoveOptions
+import org.apache.beam.sdk.io.fs.{EmptyMatchTreatment, ResourceId}
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+
+/** Special version of [[com.spotify.scio.io.ScioIO]] for use with sparkey methods. */
+case class SparkeyTestIO[T] private (path: String) extends TestIO[T] {
```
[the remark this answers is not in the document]
kellen: If we want a single class, which needs to be a case class, then this prevents convenient usage of the constructors and would require users to do `SparkeyIO[SparkeyReader](path)` which I was trying to avoid

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
kellen: See previous comment

scio-extra/src/test/scala/com/spotify/scio/extra/sparkey/SparkeyTest.scala
kellen: I tried this initially; the type parameter on `ScioIO` has to match the type of the mock data iterable passed to `input`, then in the internals of `JobTest` the SCollection returned must also be of the same type, so without changing the contract of ScioIO I don't think this is possible

RustedBones (approved):

PR #5130, opened 2024-01-03
feat: add save dynamic csv

klDen:
## Description
Add save dynamic CSV files.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
@@ -23,7 +23,7 @@ import com.spotify.scio.io._
 import com.spotify.scio.util.{FilenamePolicySupplier, ScioUtil}
 import com.spotify.scio.values.SCollection
 
-import kantan.csv._
+import kantan.csv.{CsvSink => _, _}
 import kantan.codecs.compat._
 import kantan.csv.CsvConfiguration.{Header, QuotePolicy}
 import kantan.csv.engine.ReaderEngine
@@ -196,7 +196,7 @@ object CsvIO {
       .withSuffix(params.suffix)
       .withNumShards(params.numShards)
       .withCompression(params.compression)
-      .via(new com.spotify.scio.extra.csv.CsvSink(params.csvConfiguration))
+      .via(new CsvSink(params.csvConfiguration))
 
   private def read[T: HeaderDecoder: Coder](sc: ScioContext, path: String, params: ReadParam) = {
     val filePattern = ScioUtil.filePattern(path, params.suffix)
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
@@ -17,27 +17,29 @@
 
 package com.spotify.scio.extra.csv
 
-import java.io.{Reader, Writer}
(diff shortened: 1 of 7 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
```diff
@@ -17,27 +17,26 @@
...
+import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.io._
-import com.spotify.scio.ScioContext
-import com.spotify.scio.util.ScioUtil
-import com.spotify.scio.util.FilenamePolicySupplier
+import com.spotify.scio.util.{FilenamePolicySupplier, ScioUtil}
 import com.spotify.scio.values.SCollection
-import kantan.csv._
-import kantan.codecs.compat._
```
RustedBones, line 30: IntelliJ probably tells this import is unused for 2.13, but it is required for +compile with 2.12

scio-extra/src/main/scala/com/spotify/scio/extra/csv/package.scala
```diff
@@ -18,5 +18,38 @@
...
+
+/**
+ * Main package for CSV type-safe APIs. Import all.
+ *
+ * {{{
+ * import com.spotify.scio.extra.csv._
+ * }}}
+ */
+package object csv extends AllSyntax {
+  final private[scio] class CsvSink[T: HeaderEncoder](csvConfig: CsvConfiguration)
```
RustedBones, line 37: Defining classes in package object has been discouraged. This creates the following warning:
>it is not recommended to define classes/objects inside of package objects.
If possible, define class CsvSink in package csv instead.

It would be nicer to move this to another compile unit inside the csv package.

scio-extra/src/main/scala/com/spotify/scio/extra/csv/CsvIO.scala
klDen: Sounds good. Re-adding it

scio-extra/src/main/scala/com/spotify/scio/extra/csv/package.scala
klDen: Got it. Is there a plan to migrate other classes that are defined in package objects? e.g. https://github.com/klDen/scio/blob/d97f0387393315d89fd0bbd6684306bd5f37a376/scio-parquet/src/main/scala/com/spotify/scio/parquet/package.scala

RustedBones: Yes, this is a work in progress: https://github.com/spotify/scio/pull/5127

RustedBones (approved):
LGTM. I took the freedom to refactor the test a bit.
Thanks for the contribution!

scio-extra/src/test/scala/com/spotify/scio/extra/csv/dynamic/CsvDynamicTest.scala
```diff
@@ -0,0 +1,160 @@
...
+  int: Int = 0,
+  long: Long = 0L,
+  float: Float = 0f,
+  bool: Boolean = false,
+  string: String = ""
+)
+
+object TypedRecord {
+
+  val cscHeaderCodec: HeaderCodec[TypedRecord] = HeaderCodec.caseCodec(
```
RustedBones, line 42: typo

PR #5132, opened 2024-01-03
fs.gs.inputstream.fadvise defaults to SEQUENTIAL

farzad-sedghi:
Based on [this article](https://cloud.google.com/blog/products/data-analytics/new-release-of-cloud-storage-connector-for-hadoop-improving-performance-throughput-and-more) linked in our docs, Scio jobs should have `fs.gs.inputstream.fadvise` set to `SEQUENTIAL`:

```
Ideal use cases for fadvise in SEQUENTIAL mode include:
 - Traditional MapReduce jobs that scan entire files sequentially
 - DistCp file transfers
```

This makes sense as the batch jobs tend to read the records sequentially as opposed to random look ups, and Scio jobs fit that category very well. So to me seems like the documentation has been misinterpreted when the default was set to `RANDOM`.
Furthermore in my benchmarks I always saw a slight better performance for fadvise=SEQUENTIAL and rarely the opposite.

### DIFF
--- a/scio-parquet/src/main/resources/core-site.xml
+++ b/scio-parquet/src/main/resources/core-site.xml
@@ -20,6 +20,6 @@
   <property>
     <!-- This defaults to SEQUENTIAL in GoogleCloudStorageReadOptions -->
     <name>fs.gs.inputstream.fadvise</name>
-    <value>RANDOM</value>
+    <value>SEQUENTIAL</value>
   </property>
 </configuration>
--- a/scio-smb/src/main/resources/core-site.xml
+++ b/scio-smb/src/main/resources/core-site.xml
@@ -20,6 +20,6 @@
   <property>
     <!-- This defaults to SEQUENTIAL in GoogleCloudStorageReadOptions -->
     <name>fs.gs.inputstream.fadvise</name>
-    <value>RANDOM</value>
+    <value>SEQUENTIAL</value>
   </property>
 </configuration>
--- a/site/src/main/paradox/io/Parquet.md
+++ b/site/src/main/paradox/io/Parquet.md
@@ -282,7 +282,7 @@ val parquetConf: Configuration = {
 ### Common Configuration Options
 
 - `parquet.block.size` - This determines block size for HDFS and row group size. 1 GiB is recommended over the default 128 MiB, although you'll have to weigh the tradeoffs: a larger block size means fewer seek operations on blob storage, at the cost of having to load a larger row group into memory.
-- `fs.gs.inputstream.fadvise` - Parquet relies heavily on random seeks so this GCS connector setting should be set to `RANDOM`. See this [blog post](https://cloud.google.com/blog/products/data-analytics/new-release-of-cloud-storage-connector-for-hadoop-improving-performance-throughput-and-more) for more.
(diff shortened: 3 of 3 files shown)


clairemcginty (approved):
cool! IIRC, at the time our understanding was that random access was optimal for Parquet specifically (rather than for MapReduce jobs generally) -- in order to construct a multi-field projection from columnar storage you'd be jumping around quite a lot in the row group to fetch all the required fields.

Out of curiosity, I did a quick Dataflow benchmark using different projection %s/fadvise settings:

| Projection Size | Fadvise | VCPU Hours |
| --------------- | --------|------------- |
| 2%  | Random | 45.846 vCPU hr |
| 2%  | Sequential | 28.45 vCPU hr |
| 100%  | Random | 47.814 vCPU hr |
| 100%  | Sequential | 40.909 vCPU hr |

So it looks like Sequential helps a LOT with partial projection (which require less random seeking required to construct record classes), and slightly less with full schema projections (but still better than Random!). Nice!

farzad-sedghi:
> cool! IIRC, at the time our understanding was that random access was optimal for Parquet specifically (rather than for MapReduce jobs generally) -- in order to construct a multi-field projection from columnar storage you'd be jumping around quite a lot in the row group to fetch all the required fields.
> 
> Out of curiosity, I did a quick Dataflow benchmark using different projection %s/fadvise settings:
> 
> Projection Size	Fadvise	VCPU Hours
> 2%	Random	45.846 vCPU hr
> 2%	Sequential	28.45 vCPU hr
> 100%	Random	47.814 vCPU hr
> 100%	Sequential	40.909 vCPU hr
> So it looks like Sequential helps a LOT with partial projection (which require less random seeking required to construct record classes), and slightly less with full schema projections (but still better than Random!). Nice!
> 
> (note: I didn't test SplittableDoFn reads where the file has multiple row groups -- that might be a case where Random wins -- maybe we can note that in the docs?)

Thanks for the tests. FYI all my tests were based on SplittableDoFn.

RustedBones (approved):
Looks build is failing making the site because the PR is triggered from fork. Ok to merge

farzad-sedghi:
@clairemcginty @RustedBones what do you guys think about `AUTO` mode? it's the default mode and it acts smart as it starts in `SEQUENTIAL` mode until there is a backward or (conditional) forward seek when it switches to `RANDOM`. thinking maybe best is to adopt that behavior as a default.

clairemcginty:
> @clairemcginty @RustedBones what do you guys think about `AUTO` mode? it's the default mode and it acts smart as it starts in `SEQUENTIAL` mode until there is a backward or (conditional) forward seek when it switches to `RANDOM`. thinking maybe best is to adopt that behavior as a default.

I tested out AUTO in my quick benchmark yesterday and it seemed to be slightly worse for partial projections (31.3 VCPU hours compared to SEQUENTIAL's 28.4) and much worse for full projections (it got to 160.9 vCPU hr compared to SEQUENTIAL's 40.9 before I gave up and killed the job...)

farzad-sedghi:
> > @clairemcginty @RustedBones what do you guys think about `AUTO` mode? it's the default mode and it acts smart as it starts in `SEQUENTIAL` mode until there is a backward or (conditional) forward seek when it switches to `RANDOM`. thinking maybe best is to adopt that behavior as a default.
> 
> I tested out AUTO in my quick benchmark yesterday and it seemed to be slightly worse for partial projections (31.3 VCPU hours compared to SEQUENTIAL's 28.4) and much worse for full projections (it got to 160.9 vCPU hr compared to SEQUENTIAL's 40.9 before I gave up and killed the job...)

wow that doesn't sound good at all! Was it worst than `RANDOM` or the same? especially for the failing case...

clairemcginty:
> wow that doesn't sound good at all! Was it worst than `RANDOM` or the same? especially for the failing case...

AUTO outperformed RANDOM for partial projection, but was much worse for full projection

PR #5133, opened 2024-01-04
Update beam to v2.53

RustedBones:
Updating beam to 2.53

Some changes in bigtable forced to create some more method in the beam package because package visibility

### DIFF
--- a/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBulkWriter.java
+++ b/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBulkWriter.java
@@ -115,11 +115,9 @@ public class BigtableBulkWriter
 
     @StartBundle
     public void startBundle(StartBundleContext c) throws IOException {
-      if (bigtableWriter == null) {
-        bigtableWriter =
-            new BigtableServiceHelper(bigtableOptions, c.getPipelineOptions())
-                .openForWriting(tableName);
-      }
+      bigtableWriter =
+          new BigtableServiceHelper(bigtableOptions, c.getPipelineOptions())
+              .openForWriting(tableName);
       recordsWritten = 0;
     }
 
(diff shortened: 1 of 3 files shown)


scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBulkWriter.java
```diff
@@ -142,17 +142,13 @@ public void processElement(@Element Iterable<KV<ByteString, Iterable<Mutation>>>
 
     @FinishBundle
     public void finishBundle() throws Exception {
-      bigtableWriter.flush();
```
clairemcginty, line 145: is the `flush` call no longer needed?

RustedBones: flush is not part of the API anymore. See this commit: https://github.com/apache/beam/commit/b566480cec6d413efa0d62711b3f3f9ca7a0f6fd

scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBulkWriter.java
```diff
@@ -142,17 +142,13 @@ public void processElement(@Element Iterable<KV<ByteString, Iterable<Mutation>>>
...
-      checkForFailures(failures);
-      LOG.debug("Wrote {} records", recordsWritten);
-    }
-
-    @Teardown
-    public void tearDown() throws Exception {
```
shnapz, line 151: Isn't there an overhead when creating a new bigbableWriter on each bundle?

RustedBones: probably yes, but I don't know how to workaround this since `close` is the only way to flush now.
I'll simplify the code to remove the null check in the `startBundle` as we'll always have to create a new writer

RustedBones: checked in the beam implementation, they also recreate writer on every bundle

clairemcginty (approved):

PR #5138, opened 2024-01-08
Move populate test data to compile scope

RustedBones:
- Leverage compile scope in integration project to populate test data and remove hack in GHA
- Update comment helper in test with good command
- Split `PopulateTestData` for jdbc


clairemcginty (approved):
nice 🙏 this is much cleaner!

Issue #5080, opened 2023-11-17
Support Taps for SMB writes/transforms


clairemcginty:
The Scala bindings for SMB writes return a `ClosedTap[Nothing]` ([1](https://github.com/spotify/scio/blob/95e5b6778676aac9502ec87b0e0028f1cad2efcd/scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala#L50), [2](https://github.com/spotify/scio/blob/19e3b2d55cee9c69a0881c48858afd016d731440/scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketScioContextSyntax.scala#L821)). For Scio 0.14, it could be nice to finally support this -- maybe the tap type could be `Map[BucketShardId, Seq[T]]`?

clairemcginty:
We can easily implement this as a tap in `SortedBucketIO` once https://github.com/spotify/scio/pull/4962 is merged

PR #5144, opened 2024-01-08
Support Tap for SMB writes (addresses #5080)

clairemcginty:
addresses #5080

Simple implementation of Tap for SMB writes

todo:

- [x] Tap for SMB writes
- [x] Tap for SMB pre-keyed writes
- [x] Tap for SMB transforms

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
@@ -334,6 +334,10 @@ public class SortedBucketIO {
       return FileSystems.matchNewResource(tempLocationOpt, true);
     }
 
+    public FileOperations<V> getFileOperations() {
+      return fileOperations;
+    }
+
     @Override
     public WriteResult expand(PBegin input) {
       Preconditions.checkNotNull(outputDirectory, "outputDirectory is not set");
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
@@ -600,7 +600,7 @@ public class SortedBucketIO {
 
     abstract int getSorterMemoryMb();
 
-    abstract FileOperations<V> getFileOperations();
+    public abstract FileOperations<V> getFileOperations();
 
     abstract BucketMetadata<K1, K2, V> getBucketMetadata();
 
--- a/scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
+++ b/scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
@@ -19,12 +19,15 @@ package com.spotify.scio.smb.syntax
 
 import com.spotify.scio.annotations.experimental
 import com.spotify.scio.coders.Coder
-import com.spotify.scio.io.{ClosedTap, EmptyTap}
+import com.spotify.scio.io.{ClosedTap, EmptyTap, TapOf}
 import com.spotify.scio.testing.TestDataManager
 import com.spotify.scio.values._
 import org.apache.beam.sdk.coders.KvCoder
-import org.apache.beam.sdk.extensions.smb.{SortedBucketIO, SortedBucketIOUtil}
-import org.apache.beam.sdk.values.KV
+import org.apache.beam.sdk.extensions.smb.{BucketShardId, SortedBucketIO, SortedBucketIOUtil}
+import org.apache.beam.sdk.io.fs.ResourceId
+import org.apache.beam.sdk.values.{KV, PCollection, TupleTag}
+
+import scala.jdk.CollectionConverters._
 
 trait SortMergeBucketSCollectionSyntax {
   implicit def toSortMergeBucketKeyedSCollection[K, V](
@@ -49,16 +52,30 @@ final class SortedBucketSCollection[T](private val self: SCollection[T]) {
    *   contains information about key function, bucket and shard size, etc.
    */
   @experimental
-  def saveAsSortedBucket(write: SortedBucketIO.Write[_, _, T]): ClosedTap[Nothing] = {
+  def saveAsSortedBucket(
+    write: SortedBucketIO.Write[_, _, T]
+  )(implicit coder: Coder[T]): ClosedTap[T] = {
     if (self.context.isTest) {
-      val testOutput = TestDataManager.getOutput(self.context.testId.get)
-      testOutput[T](SortedBucketIOUtil.testId(write))(self)
+      TestDataManager.getOutput(self.context.testId.get)(SortedBucketIOUtil.testId(write))
+      ClosedTap(TapOf[T].saveForTest(self))
     } else {
-      self.applyInternal(write)
-    }
+      val writeResult = self.applyInternal(write).expand()
 
-    // @Todo: Implement taps for metadata/bucket elements
-    ClosedTap[Nothing](EmptyTap)
+      val bucketFiles = self.context
+        .wrap(
+          writeResult
+            .get(new TupleTag("WrittenFiles"))
+            .asInstanceOf[PCollection[KV[BucketShardId, ResourceId]]]
+        )
+        .materialize
+
+      val fileOps = write.getFileOperations
+
+      ClosedTap(
+        bucketFiles.underlying
+          .flatMap(kv => fileOps.iterator(kv.getValue).asScala)
+      )
+    }
   }
 }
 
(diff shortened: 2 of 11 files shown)


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
```diff
@@ -600,7 +600,7 @@ public abstract static class Write<K1, K2, V> extends PTransform<PCollection<V>,
 
     abstract int getSorterMemoryMb();
 
-    abstract FileOperations<V> getFileOperations();
+    public abstract FileOperations<V> getFileOperations();
```
clairemcginty, line 603: We could get around the signature change by adding a new method to SortedBucketIO like:

```java
public Iterator<T> tapBucketFile(ResourceId resourceId) {
   return getFileOperations().open(resourceId);
}
```

but this seems overtly hacky to me - plus, we re-create a FileOperations instance for every bucket/shard combo

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -49,16 +52,30 @@ final class SortedBucketSCollection[T](private val self: SCollection[T]) {
...
-      self.applyInternal(write)
-    }
+      val writeResult = self.applyInternal(write).expand()
 
-    // @Todo: Implement taps for metadata/bucket elements
-    ClosedTap[Nothing](EmptyTap)
+      val bucketFiles = self.context
+        .wrap(
+          writeResult
+            .get(new TupleTag("WrittenFiles"))
```
clairemcginty, line 67: this is set here: https://github.com/spotify/scio/blob/v0.13.6/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketSink.java#L512

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -49,16 +52,30 @@ final class SortedBucketSCollection[T](private val self: SCollection[T]) {
...
+            .asInstanceOf[PCollection[KV[BucketShardId, ResourceId]]]
+        )
+        .materialize
+
+      val fileOps = write.getFileOperations
+
+      ClosedTap(
+        bucketFiles.underlying
+          .flatMap(kv => fileOps.iterator(kv.getValue).asScala)
```
clairemcginty, line 76: We could keep the `BucketShardId` so the tap returns a mapping from Bucket ID --> records in that bucket - although we'd have to also find a way to implement this in test context (probably by making another `SortedBucketIO.Write#getBucketMetadata` public too and mapping the bucketing function over each elements)

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/SortedBucketIO.java
clairemcginty: tbh it might be simplest to just make all these getters public 🤷‍♀️

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -49,16 +53,20 @@ final class SortedBucketSCollection[T](private val self: SCollection[T]) {
    *   contains information about key function, bucket and shard size, etc.
    */
   @experimental
-  def saveAsSortedBucket(write: SortedBucketIO.Write[_, _, T]): ClosedTap[Nothing] = {
+  def saveAsSortedBucket(
+    write: SortedBucketIO.Write[_, _, T]
+  ): ClosedTap[T] = {
+    val beamValueCoder = self.internal.getCoder
+    implicit val valueCoder: Coder[T] = Coder.beam(beamValueCoder)
```
RustedBones, line 60: There should normally be a scio coder available directly in the SCollection
```suggestion
    import self.coder
```

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -49,16 +53,20 @@ final class SortedBucketSCollection[T](private val self: SCollection[T]) {
...
+      TestDataManager.getOutput(self.context.testId.get)(SortedBucketIOUtil.testId(write))
+      ClosedTap(TapOf[T].saveForTest(self))
     } else {
-      self.applyInternal(write)
-    }
+      val writeResult = self.applyInternal(write)
 
-    // @Todo: Implement taps for metadata/bucket elements
-    ClosedTap[Nothing](EmptyTap)
+      ClosedTap(SmbIO.tap(write.getFileOperations, writeResult).apply(self.context))
```
RustedBones, line 68: Skip apply in scala
```suggestion
      ClosedTap(SmbIO.tap(write.getFileOperations, writeResult)(self.context))
```

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
```diff
@@ -82,17 +90,18 @@ final class SortedBucketPairSCollection[K, V](private val self: SCollection[KV[K
...
+      ClosedTap(TapOf[V].saveForTest(self.map(_.getValue)))
     } else {
-      self.applyInternal(write.onKeyedCollection(beamValueCoder, verifyKeyExtraction))
-    }
+      val writeResult =
+        self.applyInternal(write.onKeyedCollection(beamValueCoder, verifyKeyExtraction))
 
-    // @Todo: Implement taps for metadata/bucket elements
-    ClosedTap[Nothing](EmptyTap)
+      ClosedTap(SmbIO.tap(write.getFileOperations, writeResult).apply(self.context))
```
RustedBones, line 104:
```suggestion
      ClosedTap(SmbIO.tap(write.getFileOperations, writeResult)(self.context))
```

scio-smb/src/test/scala/com/spotify/scio/smb/SmbIOTest.scala
```diff
@@ -179,4 +192,106 @@ class SmbIOTest extends PipelineSpec {
...
+      .saveAsSortedBucket(
+        AvroSortedBucketIO
+          .write(classOf[String], "name", classOf[Account])
+          .to(tempFolder.getAbsolutePath)
+          .withNumBuckets(4)
+          .withNumShards(2)
+          .withFilenamePrefix("custom-prefix")
+      )
+
+    val result = sc.run().waitUntilDone()
```
RustedBones, line 211:
```suggestion
    val result = runWithRealContext() { sc =>
      sc.parallelize(accountsIterable)
         .saveAsSortedBucket(
        AvroSortedBucketIO
          .write(classOf[String], "name", classOf[Account])
          .to(tempFolder.getAbsolutePath)
          .withNumBuckets(4)
          .withNumShards(2)
          .withFilenamePrefix("custom-prefix")
    }.waitUntilDone()
```

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
clairemcginty: unfortunately this doesn't work because `SmbIO::tap` requires an implicit coder (required by `bucketFiles.underlying.flatMap`)... so if a curried param is supplied, the compiler expects it to be that coder. We'd have to do:

```scala
ClosedTap(
   SmbIO.tap(write.getFileOperations, writeResult)(Coder[V])(self.context)
)
```

scio-smb/src/main/scala/com/spotify/scio/smb/syntax/SortMergeBucketSCollectionSyntax.scala
clairemcginty: awesome!

scio-smb/src/test/scala/com/spotify/scio/smb/SmbIOTest.scala
clairemcginty: I don't know if this would work because we need to preserve `tap` (the result of `.saveAsSortedBucket`), but `runWithRealContext` only returns a `ScioExecutionContext`

clairemcginty: Maybe we can create a new testing helper method that returns a ClosedTap? like:

```scala
def runWithRealContext[T](
    options: PipelineOptions = PipelineOptionsFactory.create()
)(fn: ScioContext => Tap[T]): ClosedTap = {
    val sc = ScioContext(options)
    val tap = fn(sc)
    val result = sc.run().waitUntilDone()
    tap.get(result)
}

RustedBones: Ha missed that too. Let's stick with current structure

PR #4928, opened 2023-07-19
Improve avro support 

RustedBones:
- Leverage avro datum factory so that `GenericRecords` and `SpecificRecords` in scio uses `String` as `CharSequence` implementation instead of `Utf8`
- Allow to pass `AvroDatumfactory` to all avro IOs in oder to increase support of logical types
- Refactor parquet-avro to respect schema and projection

PR #5152, opened 2024-01-11
Integrate avro datum factory in scio-avro

RustedBones:
subset of #4928

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
@@ -62,12 +62,10 @@ private[scio] class SpecificRecordDatumFactory[T <: SpecificRecord](recordType: 
     }
   }
 
-  @transient private lazy val avroVersion = runtimeAvroVersion()
-
   override def apply(writer: Schema): DatumWriter[T] = {
     val datumWriter = new SpecificDatumWriter[T]()
     // avro 1.8 generated code does not add conversions to the data
-    if (avroVersion.exists(_.startsWith("1.8."))) {
+    if (runtimeAvroVersion.exists(_.startsWith("1.8."))) {
       val data = datumWriter.getData
       val conversions = specificRecordConversions(recordType)
       conversions.foreach(data.addLogicalTypeConversion)
@@ -79,7 +77,7 @@ private[scio] class SpecificRecordDatumFactory[T <: SpecificRecord](recordType: 
   override def apply(writer: Schema, reader: Schema): DatumReader[T] = {
     val datumReader = new ScioSpecificDatumReader()
     // avro 1.8 generated code does not add conversions to the data
-    if (avroVersion.exists(_.startsWith("1.8"))) {
+    if (runtimeAvroVersion.exists(_.startsWith("1.8."))) {
       val data = datumReader.getData
       val conversions = specificRecordConversions(recordType)
       conversions.foreach(data.addLogicalTypeConversion)
@@ -92,7 +90,8 @@ private[scio] class SpecificRecordDatumFactory[T <: SpecificRecord](recordType: 
 }
 
 private[scio] object SpecificRecordDatumFactory {
-  private def runtimeAvroVersion(): Option[String] =
+
+  @transient private lazy val runtimeAvroVersion: Option[String] =
     Option(classOf[Schema].getPackage.getImplementationVersion)
 
   private def specificRecordConversions[T <: SpecificRecord](
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
@@ -16,27 +16,83 @@
 
 package com.spotify.scio.avro
 
-import org.apache.avro.Schema
-import org.apache.avro.io.{DatumReader, DatumWriter}
-import org.apache.avro.reflect.{ReflectData, ReflectDatumReader, ReflectDatumWriter}
+import org.apache.avro.{Conversion, Schema}
+import org.apache.avro.generic.{GenericDatumReader, GenericRecord}
+import org.apache.avro.io.DatumReader
+import org.apache.avro.specific.{SpecificDatumReader, SpecificRecord}
 import org.apache.beam.sdk.extensions.avro.io.AvroDatumFactory
-import org.apache.beam.sdk.extensions.avro.schemas.utils.AvroUtils
 
 /**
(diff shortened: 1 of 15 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
```diff
@@ -16,27 +16,97 @@
...
+/**
+ * AvroDatumFactory for [[SpecificRecord]] forcing underlying [[CharSequence]] implementation to
+ * [[String]]
+ *
+ * Avro default [[CharSequence]] implementation is [[org.apache.avro.util.Utf8]] which can't be used
+ * when joining or as SMB keys as it doest not implement equals
+ */
+private[scio] class SpecificRecordDatumFactory[T <: SpecificRecord](recordType: Class[T])
+    extends AvroDatumFactory.SpecificDatumFactory[T](recordType) {
+  import SpecificRecordDatumFactory._
```
clairemcginty, line 57: Maybe I'm just reading the diff wrong, but are these imports actually needed in the `SpecificRecordDatumFactory` class?

scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
```diff
@@ -16,27 +16,97 @@
...
+      conversions.foreach(data.addLogicalTypeConversion)
+    }
+    datumWriter.setSchema(writer)
+    datumWriter
   }
+
+  override def apply(writer: Schema, reader: Schema): DatumReader[T] = {
+    val datumReader = new ScioSpecificDatumReader()
+    // avro 1.8 generated code does not add conversions to the data
+    if (avroVersion.exists(_.startsWith("1.8"))) {
```
clairemcginty, line 82:
```suggestion
    if (avroVersion.exists(_.startsWith("1.8."))) {
```

scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
```diff
@@ -16,27 +16,97 @@
...
 
+  @transient private lazy val avroVersion = runtimeAvroVersion()
+
   override def apply(writer: Schema): DatumWriter[T] = {
-    val data = new ReflectData(recordType.getClassLoader)
-    AvroUtils.addLogicalTypeConversions(data)
-    new ReflectDatumWriter[T](writer, data)
+    val datumWriter = new SpecificDatumWriter[T]()
+    // avro 1.8 generated code does not add conversions to the data
+    if (avroVersion.exists(_.startsWith("1.8"))) {
```
clairemcginty, line 70:
```suggestion
    if (avroVersion.exists(_.startsWith("1.8."))) {
```

scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
RustedBones: Just getting access to `runtimeAvroVersion` and `specificRecordConversions` methods from the companion object.

clairemcginty (approved):
lgtm -- thanks for this long overdue refactor!

scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
```diff
@@ -16,27 +16,97 @@
...
+    extends AvroDatumFactory.SpecificDatumFactory[T](recordType) {
+  import SpecificRecordDatumFactory._
+  private class ScioSpecificDatumReader extends SpecificDatumReader[T](recordType) {
+    override def findStringClass(schema: Schema): Class[_] = super.findStringClass(schema) match {
+      case cls if cls == classOf[CharSequence] => classOf[String]
+      case cls                                 => cls
+    }
   }
 
+  @transient private lazy val avroVersion = runtimeAvroVersion()
```
RustedBones, line 65: This is a constant. Can be moved to the companion object

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
```diff
@@ -47,11 +48,34 @@ final class ScioContextOps(private val self: ScioContext) extends AnyVal {
     self.read(ObjectFileIO[T](path))(ObjectFileIO.ReadParam(suffix))
 
   def avroFile(path: String, schema: Schema): SCollection[GenericRecord] =
-    self.read(GenericRecordIO(path, schema))(AvroIO.ReadParam())
+    avroFile(path, schema, GenericRecordIO.ReadParam.DefaultSuffix)
 
   // overloaded API. We can't use default params
-  def avroFile(path: String, schema: Schema, suffix: String): SCollection[GenericRecord] =
-    self.read(GenericRecordIO(path, schema))(AvroIO.ReadParam(suffix))
+  def avroFile(
```
RustedBones, line 54: In the ideal situation we should have `avroGenericFile` and `avroSpecificFile` to avoid those method overload

scio-avro/src/main/scala/com/spotify/scio/coders/avro/AvroCoders.scala
```diff
@@ -132,42 +123,22 @@ trait AvroCoders {
...
     Coder.beam(new SlowGenericRecordCoder)
 
   implicit def avroSpecificRecordCoder[T <: SpecificRecord: ClassTag]: Coder[T] = {
-    val clazz = ScioUtil.classOf[T]
-    val schema = AvroCoders.schemaForClass(clazz).getOrElse {
-      val msg =
-        "Failed to create a coder for SpecificRecord because it is impossible to retrieve an " +
-          s"Avro schema by instantiating $clazz. Use only a concrete type implementing " +
-          s"SpecificRecord or use GenericRecord type in your transformations if a concrete " +
-          s"type is not known in compile time."
```
RustedBones, line 150: I dropped this. Non standard avro classes should be handled using custom datum facory and coder

scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala
clairemcginty: Agree that the overloading in these methods has always been confusing -- particularly with the `schema` param since it's only required for Generic reads. On the other hand, it's probably the most commonly used Scio API and migration pain would be high, I think... maybe we can think about this for 0.15?

RustedBones: sure. nothing urgent. Just wanted to raise this since adding another new parameter to the API will be a pain.

Issue #5103, opened 2023-12-01
Expose extendedErrorInfo in all BQ APIs


RustedBones:
`extendedErrorInfo` is only exposed in the `dynamic` BQ write, for other APIs, we need to go through the `configOverride`.

This is required when combining when the `failedInsertRetryPolicy = InsertRetryPolicy.retryTransientErrors`.
However this does not give possibility to use a custom transform for errors.

Issue #4995, opened 2023-09-09
Enhance saveAsCustomOutput and provide access to POutput


mkuthan:
Function `saveAsCustomOutput` is an excellent extension point for implementing my own IO. It gives me also an abstraction for testing the whole pipeline using `JobTest`. I would like thanks all Scio authors for that functionality.

But for IOs like BigQuery, POutput is not a terminal step, it delivers information about errors. The API for accessing such errors is very different for each writing to BigQuery method. 

* `getFailedInserts` or `getFailedInsertsWithErr` for Streaming Inserts
* `getFailedStorageApiInserts` for Storage Write API
* for Batch Loads there is no such method currently

You can also access successful rows for Streaming Inserts, Storage Write API and Batch Loads with another 3 methods. 

I know that I could use `internal.apply(name, io)` method and extract errors like this:

```
val io = BigQueryIO.writeTableRows().to(tableId)

val errors = scollection.transform(name) { in =>
  val results = in.internal.apply("Write to BQ", io)

  scollection.context.wrap(results.getFailedStorageApiInserts)
    .map(failedRow => (failedRow.getRow, failedRow.getErrorMessage))
}
errors
```

But I can't use `output(CustomIO[Out](id)) { results => ...}` from `JobTest` anymore. I could hack this limitation with TransformOverride but it is not so easy to write assertion then:

```
transformOverride(TransformOverride.ofIter[In, Out](
        id,
        (r: In) =>
          // how to make an assertion here?
          Option.empty[Out].toList
      ))
```

I would love to see the following enhancement for `saveAsCustomOutput`. This is only a short showcase of the overall idea, but I'm glad to hear from you better API or implementation: 

```
def saveAsCustomOutput[O <: POutput](
    name: String,
    transform: PTransform[PCollection[T], O]
)(outputFn: O => Unit): ClosedTap[Nothing] = {
  if (self.context.isTest) {
    TestDataManager.getOutput(self.context.testId.get)(CustomIO[T](name))(self)
  } else {
    val pOutput = self.internal.apply(name, transform)
    outputFn(pOutput)
  }

  ClosedTap[Nothing](EmptyTap)
}
```

With such extension I'm able to do anything I need with the writeResult, for example:

```
val io = BigQueryIO.writeTableRows().to(tableId)
var errors = scol.context.empty[(TableRow, String)] // I don't like var but didn't find better way to access output from the closure below

scol.saveAsCustomOutput(id, io) { writeResult =>
[...]

PR #5172, opened 2024-01-16
Handle BQ write result as ClosedTap side output

RustedBones:
Proposal to address #5103

This potentially could also be an alternative solution for #4995

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
@@ -174,9 +174,15 @@ trait WriteResultIO[T] { self: ScioIO[T] =>
     data: SCollection[T],
     params: WriteP
   ): ClosedTap[tapT.T] = {
-    val (tap, outputs) =
-      if (data.context.isTest) writeWithResult(data, params) else writeWithResult(data, params)
-    ClosedTap(tap, Some(outputs))
+    if (data.context.isTest) {
+      val tap = writeTest(data, params)
+      // TODO: add possibility to fill this for testing
+      val output = SideOutputCollections.empty(data.context)
+      ClosedTap(tap, Some(output))
+    } else {
+      val (tap, outputs) = writeWithResult(data, params)
+      ClosedTap(tap, Some(outputs))
+    }
   }
 
   final override def write(data: SCollection[T], params: self.WriteP): Tap[self.tapT.T] =
--- a/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
@@ -20,7 +20,7 @@ package com.spotify.scio.io
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.testing.TestDataManager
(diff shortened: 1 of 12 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -402,10 +401,42 @@ final case class BigQueryTypedTable[T: Coder](
...
       .pipe(w => Option(params.failedInsertRetryPolicy).fold(w)(w.withFailedInsertRetryPolicy))
-      .pipe(params.configOverride)
-      .pipe(w => if (params.extendedErrorInfo == Disabled) w else w.withExtendedErrorInfo())
+      .pipe(w => Option(params.configOverride).fold(w)(_(w)))
+      .pipe(w => if (params.extendedErrorInfo) w.withExtendedErrorInfo() else w)
+
+    val wr = data.applyInternal(transform)
+
+    var tuple = PCollectionTuple.empty(data.context.pipeline)
+    if (
```
clairemcginty, line 410: 🎨  might be more readable as a `match` statement on `(params.method, data.internal.isBounded` - wdyt?

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -402,10 +401,42 @@ final case class BigQueryTypedTable[T: Coder](
...
+      (params.method == WriteMethod.DEFAULT && data.internal.isBounded == PCollection.IsBounded.UNBOUNDED)
+    ) {
+      tuple = tuple.and(BigQueryIO.SuccessfulInserts.tupleTag, wr.getSuccessfulInserts)
+    } else if (BigQueryIO.storageWriteMethod.contains(params.method)) {
+      tuple = tuple.and(
+        BigQueryIO.SuccessfulStorageApiInserts.tupleTag,
+        wr.getSuccessfulStorageApiInserts
+      )
+    } else {
+      throw new IllegalArgumentException(s"Unknown write method: ${params.method}")
```
clairemcginty, line 426: imo, this seems like more of a case of an `UnsupportedOperationException` than illegal/unknown argument exception

clairemcginty (commented):
I really like the idea of enclosing the PCollectionTuple in `Tap`! Makes sense and it simplifies our BQ config a lot 👍

scio-core/src/main/scala/com/spotify/scio/io/ScioIO.scala
```diff
@@ -167,6 +169,27 @@ trait KeyedIO[T] { self: ScioIO[T] =>
   def keyCoder: Coder[KeyT]
 }
 
+trait WriteResultIO[T] { self: ScioIO[T] =>
+  override private[scio] def writeWithContext(
+    data: SCollection[T],
+    params: WriteP
+  ): ClosedTap[tapT.T] = {
+    val (tap, outputs) =
+      if (data.context.isTest) writeWithResult(data, params) else writeWithResult(data, params)
```
RustedBones, line 178: TODO handle testing

clairemcginty (approved):

PR #5173, opened 2024-01-16
Support projection in ParquetAvroSortedBucketIO

clairemcginty:
Support a `Schema` projection as a builder arg to `ParquetAvroSortedBucketIO.Read`.

Adding a new constructor arg to `ParquetAvroFileOperations` made the constructor overloading too complex, so I reworked it into a simple Builder pattern. I tried out AutoValue, but it's a pretty heavyweight solution... if anyone has a strong opinion on the Builder style i'm very open to it :)

### DIFF
--- a/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperationsTest.java
+++ b/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperationsTest.java
@@ -175,7 +175,9 @@ public class ParquetAvroFileOperationsTest {
             .endRecord();
 
     final ParquetAvroFileOperations<GenericRecord> fileOperations =
-        ParquetAvroFileOperations.of(USER_SCHEMA).withProjection(projection);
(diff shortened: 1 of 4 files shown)


scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/ParquetAvroFileOperationsTest.java
```diff
@@ -175,11 +174,10 @@ public void testGenericProjection() throws Exception {
...
             .endRecord();
 
-    final Configuration configuration = new Configuration();
-    AvroReadSupport.setRequestedProjection(configuration, projection);
-
     final ParquetAvroFileOperations<GenericRecord> fileOperations =
-        ParquetAvroFileOperations.of(USER_SCHEMA, CompressionCodecName.ZSTD, configuration);
+        ParquetAvroFileOperations.of(USER_SCHEMA)
+            .withCompression(CompressionCodecName.ZSTD)
+            .withProjection(projection);
```
RustedBones, line 180: Much nicer with the builder pattern

RustedBones (approved):

PR #5178, opened 2024-01-17
Add Scalafix rule for LogicalTypeSupplier removal

clairemcginty:


scalafix/input-0_14/src/main/scala/fix/v0_14_0/FixLogicalTypeSuppliers.scala
```diff
@@ -0,0 +1,70 @@
...
+      "foo" -> "bar"
+    )
+  )
+
+  val conf = new Configuration()
+  conf.setClass(AvroReadSupport.AVRO_DATA_SUPPLIER, classOf[LogicalTypeSupplier], classOf[AvroDataSupplier])
+  conf.setClass(AvroWriteSupport.AVRO_DATA_SUPPLIER, classOf[LogicalTypeSupplier], classOf[LogicalTypeSupplier])
+  conf.setClass(AvroReadSupport.AVRO_DATA_SUPPLIER, classOf[AvroLogicalTypeSupplier], classOf[AvroDataSupplier])
+  conf.setClass(AvroWriteSupport.AVRO_DATA_SUPPLIER, classOf[AvroLogicalTypeSupplier], classOf[LogicalTypeSupplier])
+  conf.setClass("someClass", classOf[String], classOf[CharSequence])
```
clairemcginty, line 69: For some reason these tests are failing Scalafix unit tests; Scalafix removes the linen contents but then inserts an emdot:

```scala
   val conf = new Configuration()
- ∙
- ∙
- ∙
- ∙
   conf.setClass("someClass", classOf[String], classOf[CharSequence])
```

@RustedBones have you seen this before?

scalafix/rules/src/main/scala/fix/v0_14_0/FixLogicalTypeSupplier.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val ParquetAvroWriteMatcher = SymbolMatcher.normalized(s"$ParquetAvroPrefix/SCollectionOps#saveAsParquetAvroFile")
+}
+
+class FixLogicalTypeSupplier extends SemanticRule("FixLogicalTypeSupplier") {
+  import FixLogicalTypeSupplier._
+
+  private def updateIOArgs(fnArgs: Seq[Term])(implicit doc: SemanticDocument): Seq[Term] = {
+    def removeTypeSupplier(confArgs: Seq[Term]): Option[Term] = {
+      val newConfiguration = confArgs.filterNot {
+        case q"$l -> classOf[LogicalTypeSupplier]" => true
```
clairemcginty, line 26: I couldn't figure out how to grab the symbol or signature of a `classOf` TypeRef, so left it as a string literal rather than a matched Symbol, but it feels a bit brittle...

RustedBones: do this also match normal tuple ? (without the arrow syntax ?)

scalafix/input-0_14/src/main/scala/fix/v0_14_0/FixLogicalTypeSuppliers.scala
RustedBones: Your intelliJ trims lines

scalafix/rules/src/main/scala/fix/v0_14_0/FixLogicalTypeSupplier.scala
```diff
@@ -0,0 +1,87 @@
...
+  private def isLogicalTypeSupplier(term: Term)(implicit doc: SemanticDocument): Boolean =
+    term match {
+      case q"classOf[$tpe]" => LogicalTypeSupplierMatcher.matches(tpe.symbol)
+      case _ =>
+        term.symbol.info
+          .map(_.signature)
+          .collect { case MethodSignature(_, _, returnedType) => returnedType }
+          .collect { case TypeRef(_, sym, tpe :: Nil) if JavaClassMatcher.matches(sym) => tpe }
+          .collect { case TypeRef(_, sym, _) => sym }
+          .exists(LogicalTypeSupplierMatcher.matches)
```
RustedBones, line 40: This works when class is stored in a val.
Unfortunately for this shift, since the class is removed, the val needs to also go away but that's kind of hard to do.

clairemcginty: yeah, I didn't try to implement that due to the complexity

PR #5181, opened 2024-01-18
Integrate datum-factory in smb-avro

RustedBones:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
```diff
@@ -103,50 +82,40 @@ public void populateDisplayData(Builder builder) {
...
-                      org.apache.beam.sdk.extensions.avro.schemas.utils.AvroUtils
-                          .addLogicalTypeConversions(data);
-                      return new ReflectDatumWriter<>(writer, data);
-                    });
+        ((AvroIO.Sink<ValueT>) AvroIO.sink(getSchema()))
+                .withDatumWriterFactory(datumFactory)
+                .withCodec(codec.getCodec());
 
     if (metadata != null) {
       return sink.withMetadata(metadata);
```
RustedBones, line 97: I think there was a bug here: codec was not set if metadata was not null

clairemcginty: ack! good thing there wasn't much adoption of the the metadata feature here... 😬

clairemcginty (approved):

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
RustedBones: BTW, metadata is not exposed on the IO level. Is this expected ?

clairemcginty: it's not part of AvroSortedBucketIO? It should be 😬

PR #5194, opened 2024-01-22
(fixes #5193) Serialize BucketMetadata#hashType as String

### LINKED ISSUE #5193: Serialize BucketMetadata#HashType as String?
Currently, BucketMetadata is written to JSON including the HashFunction enum value of its HashType, i.e.:

```json
{
  "numBuckets": 2,
  "numShards": 1,
  "keyClass": "java.lang.String",
  "hashType": "MURMUR3_32",
  "keyField":" user_id"
}
```

However, this requires that all SMB reads have the writer's HashFunction type available on the classpath in order to deserialize `metadata.json` back into a `BucketMetadata`. As a result, if we ever evolve the available HashFunctions and an SMB writer wants to use a new one, all of its consumers will have to upgrade to the latest Scio to be able to deserialize its `BucketMetadata`.

The reader only needs to know about HashFunction in order to assert that it's the same across all partitions/sources -- and we could use a String for that purpose. Should we future-proof the SMB API by serializing HashFunction explicitly as a String?

i.e.

```diff
public abstract class BucketMetadata<K1, K2, V> implements Serializable, HasDisplayData {
-   @JsonProperty private final HashType hashType;
+   @JsonProperty private final String hashType;

+ // Used by writer code
  public HashType getHashType() {
-   return hashType;
+   return HashType.valueOf(hashType);
  }

+ // Works mostly as-is
  boolean isCompatibleWith(BucketMetadata other) {
    return other != null
        // version 1 is backwards compatible with version 0
        && (this.version <= 1 && other.version <= 1)
-        && this.hashType == other.hashType
+        && this.hashType.equals(other.hashType)
        // This check should be redundant since power of two is checked in BucketMetadata
        // constructor, but it's cheap to double-check.
        && (Math.max(numBuckets, other.numBuckets) % Math.min(numBuckets, other.numBuckets) == 0);
  }
}
```

Since HashType is already written using its String representation there should be no cross-version compatibility issues associated with this change -- you'd still be able to read old metadata.json files

clairemcginty:
see #5193 for motivation/details

Note that this change should have no user-facing impact--all `public` BucketMetadata constructors still use type `HashType`, not `String`.

I did standardize on the amount of overloaded constructors per BucketMetadata... some implementations had specializations for Void secondary keys and some did not. In the interest of simplicity (and since for the most part, users should not be constructing these directly), I removed the specialized overloads for Void secondary keys.


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -371,6 +393,19 @@ int rehashBucket(byte[] keyBytes, int newNumBuckets) {
   // Serialization
   ////////////////////////////////////////
 
+  private static class HashFunctionSupplier implements Supplier<HashFunction>, Serializable {
```
clairemcginty, line 396: this class/signature is required for Guava's Suppliers::memoize to construct a Serializable memoized supplier: https://github.com/google/guava/blob/v32.1.2/guava/src/com/google/common/base/Suppliers.java#L112-L115

RustedBones (commented):
If I read this correctly, this only improves case where one of the read partition hashType is not supported. We'll read metadata but fail during comparison with a message saying there is a metadata mismatch.
However if all hashType are equal but not supported, we're still failing at a later stage, correct ?

RustedBones:
Note: Hash function is only used on write. Reader can work even if hash type is not known as long as they match

clairemcginty:
> Note: Hash function is only used on write. Reader can work even if hash type is not known as long as they match

Yup, reader only needs to access HashFunction as something Comparable across other metadata.json files 👍 After that, it doesn't need to invoke it at all.

RustedBones (approved):

PR #5195, opened 2024-01-22
Use new vendored Guava version

clairemcginty:
Beam 2.53.0 uses vendored Guava 32.1.2 ([ex](https://github.com/apache/beam/blob/v2.53.0/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaTranslation.java#L57)). This PR updates Scio to match


regadas (approved):
Nice!

RustedBones (approved):

PR #5198, opened 2024-01-23
Add kryo serializer for GAX api exceptions

RustedBones:
Seen some exception

```
Error message from worker: generic::unknown: 
org.apache.beam.sdk.util.UserCodeException: 
java.lang.IllegalArgumentException: 

Unable to encode element 'Failure(com.google.api.gax.rpc.InternalException: io.grpc.StatusRuntimeException: INTERNAL: xxx.)' 
with coder 'DisjunctionCoder[scala.util.Try](false -> RecordCoder[scala.util.Failure](exception -> KryoAtomicCoder), true -> RecordCoder[scala.util.Success](value -> xxx))'.
```


scio-core/src/main/scala/com/spotify/scio/coders/instances/kryo/GrpcSerializer.scala
```diff
@@ -60,3 +63,51 @@ private[coders] class StatusRuntimeExceptionSerializer extends KSerializer[Statu
...
+  override def write(kryo: Kryo, output: Output, e: ApiException): Unit = {
+    kryo.writeObject(output, e.getMessage)
+    kryo.writeClassAndObject(output, e.getCause)
+    e.getStatusCode match {
+      case grpc: GrpcStatusCode =>
+        kryo.writeClass(output, classOf[GrpcStatusCode])
+        kryo.writeObject(output, grpc.getTransportCode.toStatus, statusSer)
+      case http: HttpJsonStatusCode =>
+        kryo.writeClass(output, classOf[HttpJsonStatusCode])
+        kryo.writeObject(output, http.getTransportCode)
```
RustedBones, line 78: not really part of gRPC but since `gax-httpjson` is in the classpath, handling this type too

scio-core/src/main/scala/com/spotify/scio/coders/instances/kryo/GrpcSerializer.scala
```diff
@@ -60,3 +63,51 @@ private[coders] class StatusRuntimeExceptionSerializer extends KSerializer[Statu
...
+      case statusCode =>
+        kryo.writeClass(output, statusCode.getClass)
+    }
+    kryo.writeObject(output, e.isRetryable)
+    // kryo.writeObjectOrNull(output, e.getErrorDetails, classOf[ErrorDetails])
```
RustedBones, line 83: leave out error details from serialization.

PR #5199, opened 2024-01-24
Require import for kryo implicit fallback coder

RustedBones:
- Only define coders in dedicated trait and avoid forward coder usage (trait mixed in `Coder` calling `Coder.xxx` method) 
- prefer `lazy val` for non-parametrized coders
- Add new Sparkey coders required for testing
- Move implicit coder fallback into `com.spotify.scio.coders.kyro`


scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
```diff
@@ -193,6 +193,11 @@ object SortMergeBucketTransformExample {
   def pipeline(cmdLineArgs: Array[String]): ScioContext = {
     val (sc, args) = ContextAndArgs(cmdLineArgs)
 
+    implicit val coderUserData: Coder[GenericRecord] =
```
clairemcginty, line 196: it is too bad that the SMB read/transform APIs now require Coders just to execute the `isTest` path :( But I can't see a way around that.

PR #5201, opened 2024-01-24
Allow timeout for async DoFn

RustedBones:


kellen (approved):
Is this a speculated fix to the hanging pipelines?

RustedBones:
Yes. One user reported that setting the timeout will at least retry the bundle instead of hanging forever.
The root cause for the future that never completes is however still unknown.

It would be also nice to set the timeout as an option for the GrpcDoFn

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -87,7 +110,7 @@ public void onFailure(Throwable t) {
               }
             }
           },
-          MoreExecutors.directExecutor());
```
RustedBones, line 90: MoreExecutors.directExecutor() doc says callback may not be executed if the exception terminates the executing thread

> In other cases, no code will catch the exception, and it may terminate whichever thread happens to trigger the execution.

I've changed to use another executor that also completes the future when if it refuses to execute the callback. This is the strategy used by guava when transforming futures.

RustedBones:
Closing for now in favor of #5209 will reconsider if not solved

PR #5205, opened 2024-01-25
Initial Iceberg bucket support

regadas:


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -230,9 +272,33 @@ public HashFunction create() {
...
+      }
     };
 
     public abstract HashFunction create();
+
+    public BucketIdFn bucketIdFn() {
+      return BucketIdFn.defaultFn();
+    }
+
+    public Encoder encoder() {
```
clairemcginty, line 299: nit: Maybe rename to `KeyEncoder` so the usage is more self-explanatory?

clairemcginty (commented):
I think this change makes sense! It would be nice to add an additional test checking the Icerberg key encoding/bucket fn to [BucketMetadataTest](https://github.com/spotify/scio/blob/main/scio-smb/src/test/java/org/apache/beam/sdk/extensions/smb/BucketMetadataTest.java) 👍

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -217,6 +229,38 @@ public void populateDisplayData(Builder builder) {
...
+
+    static <T> KeyEncoder<T> defaultKeyEncoder(Coder<T> coder) {
+      return value -> {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+          coder.encode(value, baos);
+        } catch (Exception e) {
+          throw new RuntimeException("Could not encode key " + value, e);
+        }
+        return baos.toByteArray();
```
narape, line 255:
```suggestion
        try {
          final ByteArrayOutputStream baos = new ByteArrayOutputStream();
          coder.encode(value, baos);
          return baos.toByteArray();
        } catch (Exception e) {
          throw new RuntimeException("Could not encode key " + value, e);
        }
```
Nitpick: in general, it is better to reduce the lifespan of variables

clairemcginty (approved):

PR #5212, opened 2024-01-29
Scio 0.14 migration guide

RustedBones:


site/src/main/paradox/releases/migrations/v0.14.0-Migration-Guide.md
```diff
@@ -1,5 +1,21 @@
...
+
+Some coders have moved away from the default implicit scope. By updating to scio 0.14, you may encounter the following
+error:
+
+```
+Cannot find an implicit Coder instance for type:
+...
+```
+
+If the type is or contains an avro class, you can import the `com.spotify.scio.avro._` package to get the implicit avro coders
```
clairemcginty, line 13: I think it's worth explicitly calling out that this includes scio-smb reads

clairemcginty, line 13:
```suggestion
If the type is or contains an Avro class (either `GenericRecord` or a `SpecificRecord` implementation), you can import the `com.spotify.scio.avro._` package to get the implicit avro coders
```

PR #5213, opened 2024-01-29
Add scala enumeration implicit coder

RustedBones:
Coder do not support scala enumeration out of the box.

If a model contains the following enum:
```scala
object ScalaColor extends Enumeration {
  val Red, Green, Blue = Value
}
```
Code will fail to compile with
```
Cannot find an implicit Coder instance for type:

  >> com.spotify.scio.coders.ScalaColor.Value
```

Add default coder support for scala `Enumeration`


clairemcginty (approved):
nice! the new Coder assertions look great btw

RustedBones:
Agreed. Kudos to @shnapz

PR #5216, opened 2024-01-30
Add PipelineTestUtils helper for Taps

clairemcginty:


scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -53,6 +54,14 @@ trait PipelineTestUtils {
     sc.run()
   }
 
+  def tapWithRealContext[T](
```
RustedBones, line 57: In terms of design, I think we should stick with the `runWith` prefix for all APIs.
IMHO `runWithOutput` would be better.

We can also leverage this API to define the `runWithLocalOutput`

RustedBones (approved):

PR #5218, opened 2024-01-30
Added uploadTypedRows to BigQuery client

shnapz:
`writeTypedRows` is not always suitable. It is using GCP's `insertAll`. Some GCP APIs do not reflect recent loads this way, like:
`table.getNumRows()` returns 0
[Old post](https://stackoverflow.com/questions/23612225/google-apps-script-and-big-query-tabledate-insertall).
Data loaded as upload from file does not cause the same problem.

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
@@ -166,13 +166,17 @@ final private[client] class LoadOps(client: Client, jobService: JobOps) {
 
       blobId
     }.flatMap { blobId =>
-      avro(
-        List(blobId.toGsUtilUri),
-        tableSpec,
-        schema = Some(bqt.schema),
-        createDisposition = createDisposition,
-        writeDisposition = writeDisposition
-      )
+      try {
+        avro(
+          List(blobId.toGsUtilUri),
+          tableSpec,
+          schema = Some(bqt.schema),
+          createDisposition = createDisposition,
+          writeDisposition = writeDisposition
+        )
+      } finally {
+        client.blobStorage.delete(blobId)
+      }
     }
   }
 


build.sbt
```diff
@@ -83,7 +83,7 @@ val googleCloudDatastoreVersion = "0.108.6"
 val googleCloudMonitoringVersion = "3.32.0"
 val googleCloudPubSubVersion = "1.107.13"
 val googleCloudSpannerVersion = "6.55.0"
-val googleCloudStorageVersion = "2.30.1"
+val googleCloudStorageVersion = "2.26.0"
```
shnapz, line 86: in sync with [Beam](https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/26.22.0/index.html)

RustedBones: Your link points to GCP libraries-bom 26.22.0 but beam 2.53 uses 26.28.0 [here](https://github.com/apache/beam/blob/260b554a007725c9277c98f5ec8ef704570a238a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L757C87-L757C94).

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
```diff
@@ -119,6 +126,56 @@ final private[client] class LoadOps(client: Client, jobService: JobOps) {
...
+      val blobId =
+        BlobId.fromGsUtilUri(
+          s"${tempLocation.stripSuffix("/")}/upload_${RandomStringUtils.randomAlphanumeric(10)}.avro"
+        )
+      val blobInfo = BlobInfo.newBuilder(blobId).build
+      client.blobStorage.createFrom(
+        blobInfo,
+        new ByteArrayInputStream(out.toByteArray),
+        Storage.BlobWriteOption.doesNotExist(),
+        Storage.BlobWriteOption.crc32cMatch()
```
clairemcginty, line 164: what's this option needed for? 👀

shnapz: validate the checksum after upload using CRC32 :) why not?

build.sbt
shnapz: Right, reverting this

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
```diff
@@ -119,6 +126,56 @@ final private[client] class LoadOps(client: Client, jobService: JobOps) {
...
     )
 
+  /**
+   * Upload List of rows to Cloud Storage as Avro file and load to BigQuery table. Note that element
+   * type `T` must be annotated with [[BigQueryType]].
+   */
+  def uploadTypedRows[T <: HasAnnotation: TypeTag](
+    tableSpec: String,
+    rows: List[T],
+    tempLocation: String,
```
RustedBones, line 136: Since this is temp, shouldn't we clean if afterward ?

shnapz: on the other hand if I do:
```
  avro(
        List(blobId.toGsUtilUri),
        tableSpec,
        schema = Some(bqt.schema),
        createDisposition = createDisposition,
        writeDisposition = writeDisposition
  )

  storage.delete(blobId)
```
I am not confident that `avro` is fully synchronous and BQ doesn't read that file on the background

clairemcginty (approved):

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/client/LoadOps.scala
RustedBones: `execute` contains `jobService.waitForJobs(loadJob)` so I think this is fine.

I'm wondering also if we should create a `SaveOps` to allow saving some file formats (avro/json)

PR #5219, opened 2024-01-31
runWithOutput

RustedBones:


scio-test/src/main/scala/com/spotify/scio/testing/PipelineTestUtils.scala
```diff
@@ -54,14 +54,6 @@ trait PipelineTestUtils {
     sc.run()
   }
 
-  def tapWithRealContext[T](
-    options: PipelineOptions = PipelineOptionsFactory.create()
```
RustedBones, line 58: as `runWithLocalOutput` does not take custom options, removed also from the `runWithOutput` API

clairemcginty (approved):

PR #5221, opened 2024-01-31
Add counter for SMB Predicate filtering

clairemcginty:
Screenshot 2024-01-31 at 3 44 14 PM

### DIFF
--- a/scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
+++ b/scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
@@ -178,7 +178,7 @@ object SortMergeBucketJoinExample {
     sc.sortMergeJoin(
       classOf[Integer],
       ParquetAvroSortedBucketIO
-        .read(new TupleTag[GenericRecord](), SortMergeBucketExample.UserDataSchema)
+        .read(new TupleTag[GenericRecord]("users"), SortMergeBucketExample.UserDataSchema)
         .withProjection(
           SchemaBuilder
             .record("UserProjection")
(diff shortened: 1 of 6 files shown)


scio-examples/src/main/scala/com/spotify/scio/examples/extra/SortMergeBucketExample.scala
```diff
@@ -178,7 +178,7 @@ object SortMergeBucketJoinExample {
     sc.sortMergeJoin(
       classOf[Integer],
       ParquetAvroSortedBucketIO
-        .read(new TupleTag[GenericRecord](), SortMergeBucketExample.UserDataSchema)
+        .read(new TupleTag[GenericRecord]("users"), SortMergeBucketExample.UserDataSchema)
```
clairemcginty, line 181: unfortunately, if you don't name your TupleTag it gets a randomly generated ID and you end up with a counter called `SortedBucketSource{1}-PredicateFilteredRecordsCount_com.spotify.scio.examples.extra.SortMergeBucketJoinExample$.pipeline:181#a43f085462b77df0` 😬 

another option would be to use the TypeDescriptor of the TupleTag (in this case, GenericRecord, so you'd end up with `SortedBucketSource{1}-PredicateFilteredRecordsCount_GenericRecord` ... but I think that's worse since two sources might have the same parameterized type.

clairemcginty: that said, it's common practice to name the TupleTag used in an SMB read, as far as I know, so this shouldn't be much of an issue

RustedBones (approved):

PR #5225, opened 2024-02-02
Add support of BIGNUMERIC in BigQueryIO

shnapz:


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/StorageUtil.scala
```diff
@@ -79,10 +79,18 @@ object StorageUtil {
...
-            "NUMERIC"
-          case t => s"Unsupported logical type: $t"
+            val precision = schema.getObjectProp("precision").asInstanceOf[Int]
+            val scale = schema.getObjectProp("scale").asInstanceOf[Int]
+            (precision, scale) match {
+              case (38, 9)  => "NUMERIC"
+              case (77, 38) => "BIGNUMERIC"
+              case _ =>
+                throw new IllegalStateException(
+                  s"Unsupported decimal precision and scale: ($precision, $scale)"
```
shnapz, line 89: FYI custom scale and precision is definition of BQ types:
```
NUMERIC(37, 1)
BIGNUMERIC(76, 37)
```
As an alternative we can treat all _custom_ ones as BIGNUMERIC, but then we will recognize
`NUMERIC(37, 1)` as BIGNUMERIC, which is not correct

clairemcginty (commented):
looks good to me - if you can add a small unit test to TypeProviderTest before merging, that would be great!

clairemcginty (approved):
Nice!

PR #5227, opened 2024-02-05
Streamline parquet-avro with avro API

RustedBones:


kellen:
Summarizing: this PR is attempting to align the parquet-avro behavior with the other APIs, but is "very breaking" and now needs additional updating

PR #5234, opened 2024-02-09
Always set Parquet-Avro projection

clairemcginty:


RustedBones (commented):
Can we test the opposite case, when we have an avro schema with default field missing from the parquet data ?

I fear in this case the projection will fail because the optional field is missing in parquet. Projection may have to be the perfect intersection of schema

clairemcginty:
> Can we test the opposite case, when we have an avro schema with default field missing from the parquet data ?
> 
> I fear in this case the projection will fail because the optional field is missing in parquet. Projection may have to be the perfect intersection of schema

Added a new default field to the projection schema -- test still passes: https://github.com/spotify/scio/pull/5234/commits/f88d458a56b26b7f322a9889d0cc941e876695b0

scio-smb/src/test/avro/schema.avsc
```diff
@@ -21,5 +21,22 @@
...
+    "name": "AvroGeneratedUserProjection",
+    "namespace": "com.spotify.scio.smb",
+    "doc": "Record for testing trimmed projections",
+    "fields": [
+        {
+          "name": "name",
+          "type": "string"
+        },
+        {
+          "name": "optionalFieldWithDefault", /* Not present in original AvroGeneratedUser schema */
```
RustedBones, line 36: I don't think comments are allowed. you should use prob use `doc` field

RustedBones:
Can `tensorflow` package follow the same logic for consistency ?

clairemcginty:
> Can `tensorflow` package follow the same logic for consistency ?

I think tensorflow package is a bit different, because there isn't a readSchema to fall back on AFAIK... https://github.com/spotify/scio/blob/v0.14.0/scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala#L76-L79

I can look into it a bit more/add a unit test, because I don't understand 100% how this works

RustedBones (approved):

PR #5237, opened 2024-02-12
Scalafix: Add Avro coder import for SpecificRecord JobTest IOs

clairemcginty:


scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
```diff
@@ -128,6 +133,19 @@ class FixAvroCoder extends SemanticRule("FixAvroCoder") {
           // Coder[T] is a variable type where T is an avro type
           findMatchingValTypeParams(t, CoderMatcher)
             .exists(isAvroType)
+        case q"$jobTestBuilder(..$args)" if JobTestBuilderMatcher.matches(jobTestBuilder) =>
```
clairemcginty, line 136: I spent some time trying to make this generic, so that it would match any function with an implicit coder evidence `T`, by inspecting each function argument for one that matched type `T`, but it got pretty complex/edge-casey. Not sure it's worth it

scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
```diff
@@ -140,6 +158,8 @@ class FixAvroCoder extends SemanticRule("FixAvroCoder") {
               importee"avroSpecificFixedCoder") =>
             Patch.removeImportee(i) + Patch.addGlobalImport(avroImport)
         }.asPatch
+      case importer"com.spotify.scio.avro.{..$imps}" =>
+        imps.map(i => Patch.removeImportee(i) + Patch.addGlobalImport(avroImport)).asPatch
```
RustedBones, line 162: This may have an impact on import alias, but I think this is still worth it

RustedBones (approved):

PR #5246, opened 2024-02-19
Cache TensorFlow Metadata proto files

Duhemm:
Previously, the build would download over and over the Tensorflow metadata proto sources. On almost every build, close to a second (depending, among others, on the connection speed) would be spent re-downloading and unzipping the same bytes. If the server where the artifact is hosted became unavailable, developers would not be able to build the project locally.

With this patch, this is no longer the case, as the proto files will be cached locally. Cleaning the project, deleting or modifying a protobuf source file will invalidate the cache and the sources will be downloaded again.

-------------------

The issue with `protocUnpackDependencies` re-downloading the same zip on (nearly) every run was discovered using the [Develocity](https://gradle.com/gradle-enterprise-solutions/sbt-build-system/) Timeline feature, which displays task executions on different swimlanes.

Before this fix, the task `protocUnpackDependencies` is clearly shown as being a bottleneck in the build process
Screenshot 2024-02-19 at 10 40 27

After this patch is applied, the bottleneck is gone:
Screenshot 2024-02-19 at 10 41 41

After the bottleneck is gone, common tasks (eg. `scio-tensorflow / test`) start faster: Gathering all tests’ names goes from 0.9s to 0.2s (-75%).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,6 +16,7 @@
  */
 
 import sbt.*
+import sbt.util.CacheImplicits.*
 import Keys.*
 import explicitdeps.ExplicitDepsPlugin.autoImport.moduleFilterRemoveValue
 import sbtassembly.AssemblyPlugin.autoImport.*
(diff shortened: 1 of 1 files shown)


RustedBones (approved):
Thanks a lot @Duhemm.
I was aware of this but didn't push the extra mile to get it cached.

Really impressed you've discovered and fixed that.
I'll shamelessly reuse this in `spotify/magnolify` if you don't mind.

PR #5262, opened 2024-02-20
Fix byte[] equality issue for MockByteArraySparkeyReader

clairemcginty:
MockByteArraySparkeyReader was not working properly since two `Array[Byte]`s containing the same bytes produce a different hashCode every time they're deserialized -- thus, the `Map#get` operation was throwing a `NoSuchElementException` even if it contained an Array[Byte] key matching the requested key bytes.

This solution wraps map keys in a ByteBuffer, which is safe to use as a Map key. It also returns `null` if the key doesn't exist in the map rather than throwing a `NoSuchElementException`, which matches the prior Sparkey behavior

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
@@ -33,7 +33,7 @@ import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 /** Special version of [[com.spotify.scio.io.ScioIO]] for use with sparkey methods. */
-case class SparkeyTestIO[T] private (path: String) extends TestIO[T] {
+private[sparkey] case class SparkeyTestIO[T](path: String) extends TestIO[T] {
   override val tapT: TapT.Aux[T, T] = TapOf[T]
 }
 
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/MockSparkeyReader.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/MockSparkeyReader.scala
@@ -21,6 +21,7 @@ import com.spotify.sparkey.SparkeyReader.Entry
 import com.spotify.sparkey.{IndexHeader, LogHeader, SparkeyReader}
 
 import java.io.InputStream
+import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import scala.jdk.CollectionConverters._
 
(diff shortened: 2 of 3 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
```diff
@@ -33,7 +33,7 @@ import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 /** Special version of [[com.spotify.scio.io.ScioIO]] for use with sparkey methods. */
-case class SparkeyTestIO[T] private (path: String) extends TestIO[T] {
+private[sparkey] case class SparkeyTestIO[T](path: String) extends TestIO[T] {
```
clairemcginty, line 36: It seemed confusing to support both this constructor as well as SparkeyIO#apply, [below](https://github.com/spotify/scio/blob/v0.14.0/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala#L48).

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/MockSparkeyReader.scala
```diff
@@ -74,7 +75,10 @@ case class MockByteArrayEntry(k: Array[Byte], v: Array[Byte])
 
 case class MockByteArraySparkeyReader(data: Map[Array[Byte], Array[Byte]])
     extends MockSparkeyReader {
-  override def getAsByteArray(key: Array[Byte]): Array[Byte] = data(key)
+  private lazy val internal = data.map { case (k, v) => (ByteBuffer.wrap(k), v) }
```
kellen, line 78: Using ByteString might be better here.

kellen (approved):
D'oh. Thanks for fixing.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/instances/MockSparkeyReader.scala
clairemcginty: true, although that bring in a protobuf dependency to scio-sparkey where there was none before

probably not a big deal

RustedBones: why not using scala `Seq[Byte]` or `Vector[Byte]` ?

clairemcginty: good idea! updated

RustedBones (approved):

PR #5265, opened 2024-02-21
Implement hashCode() and equals() for FileOperations subclasses

clairemcginty:
Non-consistent equals() seems to be an issue for FlinkRunner.

### DIFF
--- a/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
+++ b/scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
@@ -22,6 +22,7 @@ import java.io.Serializable;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


shnapz (approved):

scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/AvroFileOperations.java
```diff
@@ -110,6 +111,25 @@ Schema getSchema() {
...
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof AvroFileOperations)) {
+      return false;
+    }
+    final AvroFileOperations<ValueT> that = (AvroFileOperations<ValueT>) obj;
+    return that.getSchema().equals(this.getSchema())
+        && that.codec.getCodec().toString().equals(this.codec.getCodec().toString())
```
clairemcginty, line 127: Context: CodecFactory has a specialized override of toString that prints the codec type.

PR #5267, opened 2024-02-21
Broaden use cases to FixAvroCoder 

clairemcginty:
adds two more use cases for AvroCoder:

1. `sc.parallelize(elements)` where `elements` is a collection of Avro types
2. invocation of a parameterized method with bound `T <: SpecificRecord: Coder`

### DIFF
--- a/scalafix/input-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder13.scala
+++ b/scalafix/input-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder13.scala
@@ -1,0 +1,14 @@
+/*
+rule = FixAvroCoder
+ */
+package fix.v0_14_0
+
+import com.spotify.scio.ScioContext
+
+object FixAvroCoder13 {
+  val sc = ScioContext()
+
+  val a: A = ???
+
+  sc.parallelize(Seq(a))
+}
--- a/scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
+++ b/scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
@@ -134,7 +134,7 @@ object FixAvroCoder {
   def isAvroType(sym: Symbol)(implicit sd: SemanticDocument): Boolean =
     AvroMatcher.matches(sym) || hasParentClass(sym, AvroMatcher)
 
-  def hasAvroTypeSignature(tree: Tree, checkLiftedType: Boolean)(implicit doc: SemanticDocument): Boolean = {
+  def hasAvroTypeSignature(tree: Tree, checkLiftedType: Boolean)(implicit doc: SemanticDocument): Boolean =
     tree.symbol.info.map(_.signature).exists {
       case MethodSignature(_, _, TypeRef(_, returnType, _)) if !checkLiftedType && isAvroType(returnType) =>
         true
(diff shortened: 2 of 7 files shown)


scalafix/input-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder13.scala
```diff
@@ -0,0 +1,14 @@
...
+package fix.v0_14_0
+
+import com.spotify.scio.ScioContext
+
+object FixAvroCoder13 {
+  val sc = ScioContext()
+
+  val a: A = ???
+
+  sc.parallelize(Seq(a))
```
clairemcginty, line 13: Note: I didn't want to add a separate file for each case, but i manually checked all permutations of `sc.parallelize(Seq(a))`, `sc.parallelize(Seq(new A))`, `val seqA = Seq(a); sc.parallelize(seqA)`, `val seqA = Seq(new A); sc.parallelize(seqA)`, etc...

scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
```diff
@@ -17,6 +17,7 @@ object FixAvroCoder {
   )
   val AvroMatcher: SymbolMatcher = SymbolMatcher.normalized(
     "org/apache/avro/specific/SpecificRecord",
+    "org/apache/avro/specific/SpecificRecordBase",
```
clairemcginty, line 20: For context, I've seen several user repos that use type bound `T <: SpecificRecordBase`

(maybe we should also refactor that type bound from `T <: SpecificRecordBase` => `T <: SpecificRecord`?)

shnapz: > maybe we should also refactor
if it's not a ton of scalafix code :) generally a more relaxed type constraint is better, but is there any other argument?

scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
```diff
@@ -145,11 +176,18 @@ class FixAvroCoder extends SemanticRule("FixAvroCoder") {
           // Coder[T] is a variable type where T is an avro type
           findMatchingValTypeParams(t, CoderMatcher)
             .exists(isAvroType)
-        case q"$jobTestBuilder(..$args)" if JobTestBuilderMatcher.matches(jobTestBuilder) =>
+        case q"$fn(..$args)" if JobTestBuilderMatcher.matches(fn) =>
           args.headOption match {
             case Some(q"$io[$tpe](..$args)") if isAvroType(tpe.symbol) => true
             case _                                                     => false
           }
+        case q"$fn(..$args)" if ParallelizeMatcher.matches(fn) =>
```
shnapz, line 184: maybe we shouldn't match class/methods, but set `usesAvroCoders = true` if any there is any implicit Coder param?

clairemcginty: this one is difficult to implement in a generic way 😢 Say you have a method like `def someMethod[T: Coder])(foo: String, bar: Int, baz: T)` that's invoked as `someMethod("foo", 1, newT)`... you basically have to examine every argument to search for one matching T, then analyze its signature

But the number of cases here is growing so much that the overall complexity might not be that much higher, lol. I'll try it out...

scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
```diff
@@ -129,6 +134,32 @@ object FixAvroCoder {
...
+            coderT.info.map(_.signature).exists {
+              case TypeSignature(_, _, TypeRef(_, maybeAvroType, _)) if AvroMatcher.matches(maybeAvroType) =>
+                true
+              case _ =>
+                false
+            }
+          case _ => false
+        }
+      case _ => false
+    }
```
RustedBones, line 162: Nit, you can remove nesting by converting the option to List and using collect

```suggestion
    expr.symbol.info.map(_.signature)
      .toList
      .collect { case MethodSignature(_, parameterLists, _) => parameterLists }
      .flatMap(_.flatten)
      .flatMap(_.symbol.info.map(_.signature).toList)
      .collect { case ValueSignature(TypeRef(_, symbol, List(TypeRef(_, coderT, _)))) if CoderMatcher.matches(symbol) => coderT }
      .flatMap(_.info.map(_.signature).toList)
      .exists { case TypeSignature(_, _, TypeRef(_, maybeAvroType, _)) => AvroMatcher.matches(maybeAvroType) }
```

RustedBones (approved):

PR #5281, opened 2024-02-28
Improve implicit coder not found message

RustedBones:
Implicit not found message can give more hints

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
@@ -34,24 +34,25 @@ Cannot find an implicit Coder instance for type:
   This can happen for a few reasons, but the most common case is that a data
   member somewhere within this type doesn't have an implicit Coder instance in scope.
 
-  Here are some debugging hints:
-    - For Option types, ensure that a Coder instance is in scope for the non-Option version.
-    - For List and Seq types, ensure that a Coder instance is in scope for a single element.
+  Here are some hints:
+    - For module specific types, you may need to explicitly import the coders, eg avro:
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -34,24 +34,25 @@ Cannot find an implicit Coder instance for type:
...
 
       may become:
         def foo[T](coll: SCollection[SomeClass],
                    param: String)(implicit c: Coder[T]): SCollection[T]
-                                                ^
+                                  ^
       Alternatively, you can use a context bound instead of an implicit parameter:
         def foo[T: Coder](coll: SCollection[SomeClass], param: String): SCollection[T]
-                    ^
-      read more here: https://spotify.github.io/scio/migrations/v0.7.0-Migration-Guide#add-missing-context-bounds
```
RustedBones, line 50: This link is broken

scio-core/src/main/scala/com/spotify/scio/coders/Coder.scala
```diff
@@ -34,24 +34,25 @@ Cannot find an implicit Coder instance for type:
...
+                 ^
+    - For sealed traits and case classes, you can identify the missing member's coder in the REPL:
+        scala> com.spotify.scio.coders.Coder.gen[Foo]
 
-    - You can check that an instance exists for Coder in the REPL or in your code:
-        scala> com.spotify.scio.coders.Coder[Foo]
```
RustedBones, line 53: This does not help and will print the exact same message

clairemcginty (approved):

PR #5282, opened 2024-02-28
Make checkVersion more dynamic

RustedBones:
Scio is not able to link to future migration resources.

Make `checkVersion` more dynamic by testing if a migration page exists in case of breaking version upgrade


clairemcginty (approved):
nice idea! Glad to remove the manual mapping entires

RustedBones:
The checks were useless, we were not able to display a 'forward' migration as we can't be aware of the future migration pages.

PR #5293, opened 2024-03-07
(fix #5290) Support empty input in TrasnformOverride.ofSource

### LINKED ISSUE #5290: TransformOverride#ofSource fails with empty Seq input
Looks like the underlying Create transform requires an explicit coder set for empty collection types:

```shell
Can not determine a default Coder for a ‘Create’ PTransform that has no elements. 
Either add elements, call Create.empty(Coder), Create.empty(TypeDescriptor), or call ‘withCoder(Coder)’ or ‘withType(TypeDescriptor)’ on the PTransform.
```

clairemcginty:
unfortunately this requires changing the signature to require a Coder.

### DIFF
--- a/scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
+++ b/scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
@@ -949,7 +949,7 @@ class JobTestTest extends PipelineSpec {
     e.getMessage should endWith(" was not greater than or equal to 100")
   }
 
-  "transformOverride" should "pass with a source override" in {
+  "transformOverride.ofSource" should "support non-empty input" in {
     JobTest[SourceTransformOverrideJob.type]
       .args("--input=in.txt", "--output=out.txt")
       .input(TextIO("in.txt"), Seq.empty[String]) // still required for pipeline construction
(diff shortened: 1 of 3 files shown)


scio-test/src/test/scala/com/spotify/scio/testing/JobTestTest.scala
```diff
@@ -962,21 +962,18 @@ class JobTestTest extends PipelineSpec {
...
-        )
+        TransformOverride.ofSource[String]("ReadInput", List())
       )
-      .output(TextIO("out.txt"))(_ should containInAnyOrder(List("10", "20")))
+      .output(TextIO("out.txt"))(_ should beEmpty)
       .run()
   }
 
-  it should "pass with a 1-to-n override" in {
+  "TransformOverride.ofIter" should "support 1-to-n inputs" in {
```
clairemcginty, line 976: The rest of this diff is just rearranging the tests to be grouped by API:

```
[info] JobTest
...
[info] transformOverride.ofSource
[info] - should support non-empty input (69 milliseconds)
[info] - should support empty input (78 milliseconds)
[info] TransformOverride.ofIter
[info] - should support 1-to-n inputs (147 milliseconds)
[info] - should support 1-to-n function inputs (56 milliseconds)
[info] TransformOverride.of
[info] - should support non-empty input (66 milliseconds)
[info] - should support function input (56 milliseconds)
[info] - should fail with an incorrect override input type (27 milliseconds)
[info] - should fail with an incorrect function override input type (40 milliseconds)
[info] - should fail differently with an incorrect override generic input type (69 milliseconds)
[info] - should fail with an incorrect override output type (30 milliseconds)
[info] TransformOverride.ofKV
[info] - should support non-empty input (56 milliseconds)
[info] - should suppport a KV function override (50 milliseconds)
[info] TransformOverride.ofIterKV
[info] - should support non-empty input (60 milliseconds)
[info] - should support a 1-to-n KV function input (57 milliseconds)
[info] TransformOverride.ofAsyncLookup
[info] - should support non-empty input (55 milliseconds)
[info] - should pass with an AsyncLookup function override (55 milliseconds)
[info] TransformOverride.ofIterAsyncLookup
[info] - should support a 1-to-n AsyncLookup function override (53 milliseconds)
[info] - should support a 1-to-n AsyncLookup override (55 milliseconds)
```

RustedBones:
I think this is still worth adding this breaking change in the next 0.14 patch. Let's add a filtering to the mima

RustedBones (approved):

PR #5297, opened 2024-03-12
Allow String key type to transform SMB sources with CharSequence key

clairemcginty:
extension of #5291 for transforms.

\+ adds support for sources with multiple partitions, where some partitions have CharSequence key and some have String key

( 🤦‍♀️ )


scio-smb/src/main/java/org/apache/beam/sdk/extensions/smb/BucketMetadata.java
```diff
@@ -403,11 +403,9 @@ private <MetadataT extends BucketMetadata> boolean isIntraPartitionCompatibleWit
     final Class<? extends BucketMetadata> otherClass = other.getClass();
     final Set<Class<? extends BucketMetadata>> compatibleTypes = compatibleMetadataTypes();
 
-    if (compatibleTypes.isEmpty() && other.getClass() != this.getClass()) {
-      return false;
-    } else if (this.getKeyClass() != other.getKeyClass()
```
clairemcginty, line 408: this check was not correct; it should have been comparing the metadata class itself.

kellen (approved):

PR #5321, opened 2024-04-02
Add support for Zstd coders

kellen:
Adds 
* `saveAsZstdDictionary` to train a Zstd dictionary on some arbitrary `SCollection[T]`. Estimates the average size of elements `T`, collects `n` elements based on a target training set size, then trains and saves the Zstd dictionary.
* A scala `ZstdCoder` object with transform Coders for the simple `T` or for each side of a `(K, V)`
* command line argument to map from a type to a dictionary, causing instances of `MyClass` to get Zstd compression automagically. Probably fails if the type is parameterized. `--zstdDictionary=com.spotify.scio.MyClass:gs://bucket/path/dict.bin`


scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,91 @@
...
+                    // ensure class exists
+                    Class.forName(className)
+                  } catch {
+                    case e: ClassNotFoundException =>
+                      throw new IllegalArgumentException(
+                        s"Class for zstdDictionary argument ${s} not found.",
+                        e
+                      )
+                  }
+                  className.replaceAll("\\$", ".") -> path
```
kellen, line 71: Maybe missing something cases here? `Class.forName` requires `OuterClass$InnerClass` but the typename in the coders is `OuterClass.InnerClass`

scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala
```diff
@@ -101,6 +102,22 @@ class PairSCollectionFunctions[K, V](val self: SCollection[(K, V)]) {
...
     new SCollectionWithHotKeyFanout(this, Right(hotKeyFanout))
 
+  /**
+   * Apply Zstd compression to the _value side only_ of this SCollection, using the provided Zstd
+   * dictionary. Dictionary must have been trained on exactly type `V`.
+   *
+   * @param dict
+   *   The Zstd dictionary trained on type `V`.
+   */
+  def setZstdDictionaryForValue(dict: Array[Byte]): SCollection[(K, V)] = {
```
kellen, line 112: Going to reconsider this API. Using it in conjunction with setCoder is going to be confusing since they both mutate the underlying PCollection

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,91 @@
...
   import com.spotify.scio.ScioContext
 
-  private[scio] case class CoderOptions(nullableCoders: Boolean, kryo: KryoOptions)
+  private[scio] case class CoderOptions(
+    nullableCoders: Boolean,
+    kryo: KryoOptions,
+    zstdDictMapping: Map[String, Array[Byte]]
+  )
   private[scio] object CoderOptions {
+    private val cache: ConcurrentHashMap[PipelineOptions, CoderOptions] = new ConcurrentHashMap()
```
clairemcginty, line 46: Are PipelineOptions implementations always guaranteed to be stable map keys? In Beam I don't see that any of them implement hashCode/equals. Maybe we could use the [appName](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/options/ApplicationNameOptions.java#L31) as a key instead? That should help keep the serialzed Coder byte size down too

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,91 @@
...
+        o,
+        { o =>
+          val scioOpts = o.as(classOf[com.spotify.scio.options.ScioOptions])
+          val nullableCoder = scioOpts.getNullableCoders
+          val zstdDictPaths = Option(scioOpts.getZstdDictionary)
+            .map(_.asScala)
+            .getOrElse(List.empty)
+            .map { s =>
+              // worst api?
+              s.split(":", 2).toList match {
```
clairemcginty, line 59: could just do 

```scala
s.split(":") match {
  case Array(className, path) => ...
}
```
?

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
kellen: Ugh no, will revisit this

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
kellen: A blacklist based on package name may also make sense

scio-core/src/main/scala/com/spotify/scio/coders/instances/ZstdCoder.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import scala.reflect.ClassTag
+
+object ZstdCoder {
+  def apply[T: Coder: ClassTag](dict: Array[Byte]): Coder[T] =
+    Coder.transform(Coder[T])(tCoder => Coder.beam(BZstdCoder.of(tCoder, dict)))
+
+  def kv[K: Coder, V: Coder](
+    keyDict: Array[Byte] = null,
+    valueDict: Array[Byte] = null
```
RustedBones, line 31: I'm not a fan of null defaults.
Also, we used `kv` for `beam.KV` in many placed, not tuples. This signature is a bit confusing.

Do we actually need this API since this is a regular `Tuple2Coder` with implicit (un)compressed key and value coders values ?

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  override protected def write(
+    data: SCollection[T],
+    params: ZstdDictIO.WriteParam
+  ): Tap[Nothing] = {
+    val ZstdDictIO.WriteParam(
+      zstdDictSizeBytes,
+      numElementsForSizeEstimation,
+      trainingBytesTarget
+    ) = params
```
RustedBones, line 48: This matching is quite unconventional

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
```diff
@@ -0,0 +1,139 @@
...
+    }
+    if (numElementsForSizeEstimation <= 0) {
+      throw new IllegalArgumentException(
+        s"numElementsForSizeEstimation must be positive, found $numElementsForSizeEstimation"
+      )
+    }
+    // training bytes may not exceed 2GiB a.k.a. the max value of an Int
+    val trainingBytesTargetActual: Int = Option(trainingBytesTarget).getOrElse {
+      val computed =
+        Try(Math.multiplyExact(zstdDictSizeBytes, 100)).toOption.getOrElse(Int.MaxValue)
```
RustedBones, line 64: do we want to catch this or let it fail ?

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
```diff
@@ -0,0 +1,139 @@
...
+          .map { case (_, elements) =>
+            val zstdSampleSize = {
+              val sum = elements.map(_.length.toLong).sum
+              if (sum > Int.MaxValue.toLong) Int.MaxValue else sum.toInt
+            }
+            logger.info(s"Training set size for for Zstd dictionary: ${zstdSampleSize}")
+            val trainer = new ZstdDictTrainer(zstdSampleSize, zstdDictSizeBytes)
+            elements.foreach(trainer.addSample)
+            trainer.trainSamples()
+          }
```
RustedBones, line 115: can't we turn this into an aggregator + present to get the final dictionary ?

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
```diff
@@ -0,0 +1,139 @@
...
+  final case class WriteParam(
+    zstdDictSizeBytes: Int = WriteParam.DefaultZstdDictSizeBytes,
+    numElementsForSizeEstimation: Long = WriteParam.DefaultNumElementsForSizeEstimation,
+    trainingBytesTarget: Int = WriteParam.DefaultTrainingBytesTarget
+  )
+
+  object WriteParam {
+    val DefaultZstdDictSizeBytes: Int = 110 * 1024
+    val DefaultNumElementsForSizeEstimation: Long = 100L
+    val DefaultTrainingBytesTarget: Int = null.asInstanceOf[Int]
```
RustedBones, line 137: This cast gives zero. Tried in `scala-cli`:
```
scala> val i: Int = null.asInstanceOf[Int]
val i: Int = 0
```
For primitive you MUST use option or box in `java.lang.Integer`

RustedBones (changes requested):

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,91 @@
...
+                case className :: path :: Nil =>
```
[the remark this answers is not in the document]
kellen: The path part can contain a `:`, ala `gs://`. 

I suppose we could use a variety of other characters, e.g. `,`, but `:` makes the most sense as a 'mapping' IMO

scio-core/src/main/scala/com/spotify/scio/coders/BeamCoders.scala
```diff
@@ -35,6 +35,14 @@ private[scio] object BeamCoders {
       case _                                             => coder
     }
 
+  private def unwrapZstd[T](options: CoderOptions, coder: BCoder[T]): BCoder[T] =
+    coder match {
+      case c: BZstdCoder[T] =>
+        val underlying = c.getCoderArguments.get(0).asInstanceOf[BCoder[T]]
+        unwrap(options, underlying)
+      case _ => coder
+    }
```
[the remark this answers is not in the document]
kellen: In e.g. `getTupleCoders`, we unwrap only the top-level Zstd coder for `Tuple2[K, V]` but want to retain any Zstd coder on `K`. Putting the zstd unwrapping in `unwrap` leads to the dict being discarded on many/every unwrap call.

scio-core/src/main/scala/com/spotify/scio/coders/instances/ZstdCoder.scala
kellen: This is for user ergonomics if setting the coder manually. 
```
sc.parallelize[(String, String)](
  List(...)
)(ZstdCoder.kv(valueDict = dictBytes))
```

Not providing this means users need to manually lift their dict into a Beam coder into a Scio coder. And using optionals here makes the API unpleasant IMO even though I in general agree with you.

Is `ZstdCoder.tuple2` less annoying/confusing?

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
kellen: Yikes! Moved to an Option

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
kellen: hmm actually I am unsure of what happens in this case, like if the training size is related to the elements added or the dict size itself. 2gb / 100 implies that the max zdict dictionary size should be 20ish megs, so maybe we should throw on that

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
kellen: It's just destructuring ala `case`? I use this frequently but maybe I am an outlier.

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
kellen: I don't think you can merge `ZstdDictTrainer`s

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
kellen: Added blacklist

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
RustedBones: It's fine. Just noted we've not used this pattern very often

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
RustedBones: Ha yes, and you also need the whole sample for the size.

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
```diff
@@ -0,0 +1,141 @@
...
+          .withSideInputs(sampleRateSI)
+          .flatMap {
+            case (s, ctx) if new Random().nextDouble() <= ctx(sampleRateSI) =>
+              Some(toBytes(s))
+            case _ => None
+          }
+          .toSCollection
+          .keyBy(_ => ())
+          .groupByKey
+          .map { case (_, elements) =>
```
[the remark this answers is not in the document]
kellen: This is so users don't need to know the average size of elements in the pipeline; we estimate it based on `numElementsForSizeEstimation`. Taking only those elements is not useful, since you may need many more elements to actually train.

scio-core/src/main/scala/com/spotify/scio/coders/instances/ZstdCoder.scala
clairemcginty: I'm lightly in favor of `tuple2` (unfortunately) -- I think `kv` is already semantically "taken" by Beam `KV` class.

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
clairemcginty: +1 on throwing an exception

kellen: Re-reading, ~100x is just a recommendation, so using fewer items would probably decrease its effectiveness but wouldn't be catastrophic. Gonna throw anyway since recommendations argue against large dictionaries.

scio-core/src/main/scala/com/spotify/scio/io/ZstdDictIO.scala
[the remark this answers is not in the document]
RustedBones: Maybe smth like this can help https://github.com/spotify/scio/pull/5352 ?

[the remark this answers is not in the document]
kellen: Yeah. I imagine doing a bunch of priority queue merges is somewhat less efficient but perhaps doesn't matter in this case since we have small amounts of data in the end.

scio-core/src/test/scala/com/spotify/scio/coders/CoderTestUtils.scala
```diff
@@ -30,4 +37,31 @@ object CoderTestUtils {
...
+    } finally {
+      fos.close()
+    }
+    tmp
+  }
+
+  def zstdOpts(
+    className: String,
+    path: String,
+    packageName: String = "com.spotify.scio.coders.CoderTestUtils$",
```
RustedBones, line 58: I'd rather have `className` only being the full one and possibly an overload with `clazz: Class[_]`

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,109 @@
...
-      val nullableCoder = o.as(classOf[com.spotify.scio.options.ScioOptions]).getNullableCoders
-      new CoderOptions(nullableCoder, KryoOptions(o))
+      val isTest = Try(o.as(classOf[ApplicationNameOptions])).toOption
+        .exists(o => TestUtil.isTestId(o.getAppName))
+
+      cache.computeIfAbsent(
+        ObjectUtils.identityToString(o),
+        { _ =>
+          val scioOpts = o.as(classOf[com.spotify.scio.options.ScioOptions])
+          scioOpts.getAppArguments
```
RustedBones, line 62: seems we discard the returned value. Can this be dropped ?

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,109 @@
...
+          val nullableCoder = scioOpts.getNullableCoders
+          val zstdDictPaths = Option(scioOpts.getZstdDictionary)
+            .map(_.asScala)
+            .getOrElse(List.empty)
+            .map {
+              case s @ ZstdArgRegex(className, path) =>
+                if (!isTest) {
```
RustedBones, line 69: why do we skip this for test ?

scio-core/src/main/scala/com/spotify/scio/coders/CoderMaterializer.scala
```diff
@@ -17,20 +17,109 @@
...
+                )
+            }
+            .groupBy(_._1)
+            .map { case (className, values) => className -> values.map(_._2).toSet }
+
+          val dupes = zstdDictPaths
+            .collect {
+              case (className, values) if values.size > 1 =>
+                s"Class $className -> [${values.mkString(", ")}]"
+            }
```
RustedBones, line 104: maybe a `partitionMap` can be nicer so we have a sanitized return type and list of errors.

RustedBones (approved):

PR #5330, opened 2024-04-09
Move test code in respective projects

RustedBones:
Splitting `scio-test` in sub modules
- `scio-test-core`: depending on `scio-core` + some avro deps for pretty printing
- `scio-test-google-cloud-platform`: depending on `scio-test-core` + GCP deps
- `scio-test-parquet`: for future helpers

In order to use `scio-test-core` in `scio-core/test` scope, I created a symlink to the `com.spotify.scio.testing` package.

As most scio module test scope require test data, set the dependency on `scio-core % "test->test"` to get access to test helpers from `scio-test-core`


scio-avro/src/test/avro/schema.avsc
```diff
@@ -52,20 +52,6 @@
...
-    "doc": "Record for testing",
-    "fields": [
-        {"name": "int_field", "type": ["null", "int"], "default": null},
-        {"name": "long_field", "type": ["null", "long"], "default": null},
-        {"name": "float_field", "type": ["null", "float"], "default": null},
-        {"name": "double_field", "type": ["null", "double"], "default": null},
-        {"name": "boolean_field", "type": ["null", "boolean"], "default": null},
-        {"name": "string_field", "type": ["null", "string"], "default": null},
-        {"name": "array_field", "type": {"type": "array", "items": "string"}, "default": null}
-    ]
```
RustedBones, line 68: This one stayed in `scio-core/test` to test kryo for avro

scio-core/src/test/scala/com/spotify/scio/ArgsTest.scala
```diff
@@ -116,29 +116,29 @@ class ArgsTest extends AnyFlatSpec with Matchers {
 
   trait Options extends PipelineOptions {
     @Required
-    def getInput: String
-    def setInput(input: String): Unit
+    def getTestInput: String
+    def setTestInput(input: String): Unit
     @Required
-    def getOutput: String
-    def setOutput(output: String): Unit
```
RustedBones, line 123: Since `PipelineOptions` are dynamically loaded, Some tests were failing because the `--input` and `--output` parameters were transformed into options instead of staying in the args. Renamed the values to avoid conflicts wit other tests

scio-core/src/test/scala/com/spotify/scio/avro/AvroUtils.scala
```diff
@@ -38,7 +39,7 @@ object AvroUtils {
     .array()
     .items()
     .stringType()
-    .noDefault()
+    .arrayDefault(Collections.emptyList())
```
RustedBones, line 42: to match with the specific record

scio-core/src/test/scala/com/spotify/scio/coders/KryoAtomicCoderTest.scala
```diff
@@ -35,11 +36,25 @@ import org.joda.time.Instant
...
+case class CaseClassWithSpecificRecord(name: String, size: Int, record: TestRecord)
+
+// additional kryo registrar for avro schema null value
+// deserializes null to the singleton instance for schema equality
+@KryoRegistrar
+class TestKryoRegistrar extends IKryoRegistrar {
+
+  override def apply(k: Kryo): Unit =
+    k.forClass[NullNode](new SingletonSerializer(NullNode.getInstance()))
+}
```
RustedBones, line 56: Apparently, this test suite was not run before.
This registrar is needed for the set so schema equality passes. Without this, kry creates a new `NullNode` instance failing equality checks

scio-core/src/test/scala/com/spotify/scio/coders/KryoAtomicCoderTest.scala
```diff
@@ -65,8 +80,7 @@ class KryoAtomicCoderTest extends PipelineSpec {
     Pair("record", 10) kryoCoderShould roundtrip()
   }
 
-  // Enable once https://github.com/scala/scala/pull/10425 is release
```
RustedBones, line 68: fix is now live

build.sbt
```diff
@@ -674,62 +698,81 @@ lazy val `scio-core` = project
...
-      moduleFilter("com.spotify", "magnolify-guava"),
-      // junit is required by beam but marked as provided
-      moduleFilter("junit", "junit")
-    ).reduce(_ | _),
+    undeclaredCompileDependenciesFilter -= moduleFilter("org.scalatest"),
+    unusedCompileDependenciesFilter -= moduleFilter("org.scalatest", "scalatest"),
     libraryDependencies ++= Seq(
-      "com.google.api.grpc" % "proto-google-cloud-bigtable-v2" % googleCloudProtoBigTableVersion,
-      "com.google.http-client" % "google-http-client" % googleHttpClientVersion,
+      "com.google.http-client" % "google-http-client" % googleHttpClientVersion, // TODO should we have this here ?
```
clairemcginty, line 722: Possible to move to `google-cloud-platform` test module?

scio-avro/src/test/scala/com/spotify/scio/avro/AvroJobTestTest.scala
```diff
@@ -0,0 +1,163 @@
+/*
+ * Copyright 2019 Spotify AB.
```
clairemcginty, line 2:
```suggestion
 * Copyright 2024 Spotify AB.
```

scio-avro/src/test/scala/com/spotify/scio/avro/AvroTapTest.scala
```diff
@@ -0,0 +1,150 @@
...
+import com.spotify.scio.proto.SimpleV2.{SimplePB => SimplePBV2}
+import com.spotify.scio.proto.SimpleV3.{SimplePB => SimplePBV3}
+
+import java.nio.ByteBuffer
+
+class AvroTapTest extends TapSpec {
+  val schema: Schema = AvroUtils.schema
+  implicit def coder: Coder[GenericRecord] = avroGenericRecordCoder(schema)
+
+  "AvroTap" should "support saveAsAvroFile with SpecificRecord" in withTempDir { dir =>
```
clairemcginty, line 35: Wdyt about moving the `withTempDir` functionality into `scio-test-core`? It seems to have a lot of usage both within Scio and within many pipeline projects. (doesn't have to be in this PR!)

clairemcginty (approved):
this is amazing! thanks for your hard work doing this refactor

build.sbt
RustedBones: no. This is required by the `Pretty` module. Instead of using `toString` in the scio SCollection matchers, we rely on this with custom formatting for
- avro
- gson

Ideally we should use smth like the cat's `Show` typeclass so any etension can give its formatting. Kept as is to avoid breaking the test API.

kellen:
Should we retarget this for 0.15?

RustedBones:
I think @clairemcginty wanted this to ship some test helpers sooner.
Here the mima sais is breaking because classes moved from one artifact to another. There should in practice be no breaking changes as `scio-test` depends on `scio-test-core` and `scio-test-google-cloud-service` containing the same classes as before.

RustedBones:
I can actually try to run mima on `scio-test-core` with `scio-test` artifact localy to make sure if I didn't overlooked something

kellen:
Was thinking more for the ease of merging 0.15 later, but if there is an urgent need we can get it out sooner and just deal with the merge issues.

RustedBones:
This is the result of the binary compatibility check
```
[error] scio-test-core: Failed binary compatibility check against com.spotify:scio-test_2.13:0.14.3! Found 1 potential problems
[error]  * interface com.spotify.scio.testing.BigtableMatchers does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("com.spotify.scio.testing.BigtableMatchers")
```
which is expected because moved to `scio-test-google-cloud-platform`

scio-avro/src/test/scala/com/spotify/scio/avro/AvroTapTest.scala
RustedBones: Will expose this in a separate PR

PR #5332, opened 2024-04-10
[scio-avro] fix: allow conversions field in record

RustedBones:
We relied on 'conversions' being a reserved field in avro. However, when records do not contain logical types, 'conversions' can be a valid field name.

Make avro 1.8 logical-type patching more lenient

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
@@ -28,6 +28,8 @@ import org.apache.avro.specific.{
 import org.apache.beam.sdk.extensions.avro.io.AvroDatumFactory
 
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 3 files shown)


scio-avro/src/main/scala/com/spotify/scio/avro/AvroDatumFactory.scala
```diff
@@ -108,21 +110,20 @@ private[scio] object SpecificRecordDatumFactory {
 
     schema.getType match {
       case Schema.Type.RECORD =>
-        Option(data.getClass(schema)).foreach { clazz =>
-          try {
-            val field = clazz.getDeclaredField("conversions")
-            field.setAccessible(true)
-            val conversions = field.get(null).asInstanceOf[Array[Conversion[_]]]
```
RustedBones, line 115: Here a `NullPointerException` was thrown because a record instance was expected.

kellen (approved):

PR #5346, opened 2024-04-24
Update beam to 2.56

RustedBones:


scio-smb/src/test/avro/user.avsc
```diff
@@ -0,0 +1,10 @@
+{
+  "namespace": "org.apache.beam.sdk.extensions.avro.io",
+  "type": "record",
+  "name": "AvroGeneratedUser",
```
clairemcginty, line 4: did the Beam Avro class get deleted?

clairemcginty (approved):

scio-smb/src/test/avro/user.avsc
RustedBones: The artifact is published with a class compiled with another, incompatible avro version

clairemcginty (approved):

PR #5350, opened 2024-04-30
[scio-core](feature) Add readFiles and readFilesWithPath apis

RustedBones:
- Add new `readFiles` api based on splittable `FileBasedSource`
- Add new `readFilesWithPath` propagating source file path along with element emitted by the source

- rename `readFiles` to `readTextFiles` for lines api
- add `readAvroGenericFiles` and `readAvroSpecificFiles` apis + `WithPath` api variants

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -1444,10 +1444,10 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     desiredBundleSizeBytes: Long,
     directoryTreatment: DirectoryTreatment,
     compression: Compression
-  )(filesSource: String => FileBasedSource[A])(implicit ev: T <:< String): SCollection[A] = {
-    val fn = Functions.serializableFn(filesSource)
+  )(f: String => FileBasedSource[A])(implicit ev: T <:< String): SCollection[A] = {
+    val createSource = Functions.serializableFn(f)
     val bcoder = CoderMaterializer.beam(context, Coder[A])
-    val fileTransform = new ReadAllViaFileBasedSource(desiredBundleSizeBytes, fn, bcoder)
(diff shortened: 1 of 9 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1424,6 +1429,28 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   *   Desired size of bundles read by the sources.
+   * @param directoryTreatment
+   *   Controls how to handle directories in the input.
+   * @param compression
+   *   Reads files using the given [[org.apache.beam.sdk.io.Compression]].
+   */
+  def readFiles[A: Coder](
+    desiredBundleSizeBytes: Long,
+    directoryTreatment: DirectoryTreatment,
+    compression: Compression
```
RustedBones, line 1446: Cannot give default since we have overload.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1460,6 +1487,48 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+   * Produced elements are paired with the file name they were read from.
+   *
+   * @param desiredBundleSizeBytes
+   *   Desired size of bundles read by the sources.
+   * @param directoryTreatment
+   *   Controls how to handle directories in the input.
+   * @param compression
+   *   Reads files using the given [[org.apache.beam.sdk.io.Compression]].
+   */
+  def readFilesWithPath[A: Coder](
```
kellen, line 1503: It'd be nice to provide variants for this API for the various IO types to alleviate the user burden of knowing about the underlying beam api and needing to construct the various sources.

RustedBones: Added to Text and Avro.

I think we should also add the `readFiles` variant for other file IOs.

kellen (approved):

PR #5351, opened 2024-05-02
Update FixAvroCoder to match transformed Avro SCollections

clairemcginty:
Our Scalafix rule was missing utility-type classes that contain helper methods for SCollections, like:

```
  def someMethod(data: SCollection[A]): SCollection[(String, A)] = {
    data.map(r => ("foo", r))
  }
```

This updates the rule to check for method that returns `SCollection[T <: SpecificRecord]`. There's a tiny chance of a false positive if the method doesn't actually invoke any SCollection mapper functions but I don't think that usage is very common. I could update the method to inspect the function body for usage of functions with a Coder bound but I think that would be very error-prone

### DIFF
--- a/scalafix/output-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder16.scala
+++ b/scalafix/output-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder16.scala
@@ -1,0 +1,11 @@
+package fix.v0_14_0
+
+import com.spotify.scio.values.SCollection
+import com.spotify.scio.avro._
+
+object FixAvroCoder16 {
+
+  def someMethod(data: SCollection[A]): SCollection[(String, A)] = {
+    data.map(r => ("foo", r))
+  }
+}
--- a/scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
+++ b/scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
@@ -42,6 +42,9 @@ object FixAvroCoder {
   val ParallelizeMatcher = SymbolMatcher.normalized(
     "com/spotify/scio/ScioContext#parallelize()."
   )
+  val SCollectionMatcher = SymbolMatcher.normalized(
+    "com/spotify/scio/values/SCollection#"
+  )
 
   /** @return true if `sym` is a class whose parents include a type matching `parentMatcher` */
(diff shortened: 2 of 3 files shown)


scalafix/rules/src/main/scala/fix/v0_14_0/FixAvroCoder.scala
```diff
@@ -162,9 +165,23 @@ object FixAvroCoder {
           coderT
       }
       .flatMap(_.info.map(_.signature).toList)
-      .exists { case TypeSignature(_, _, TypeRef(_, maybeAvroType, _)) =>
-        AvroMatcher.matches(maybeAvroType)
+      .exists {
+        case TypeSignature(_, _, TypeRef(_, maybeAvroType, _)) =>
+          AvroMatcher.matches(maybeAvroType)
+        case _ => false
```
clairemcginty, line 171: I was getting a compiler warning about this line, so added the exhaustive matcher

scalafix/output-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder16.scala
```diff
@@ -0,0 +1,11 @@
+package fix.v0_14_0
+
+import com.spotify.scio.values.SCollection
+import com.spotify.scio.avro._
```
RustedBones, line 4: why do we add the import here ? It is a false positive ?

clairemcginty: It is needed in this case -- without it, `data.map(r => ("foo", r))` throws a Compilation error:

```
[error] /Users/clairem/scio/scalafix/output-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder16.scala:9:13: 
[error] Cannot find an implicit Coder instance for type:
[error] 
[error]   >> (String, fix.v0_14_0.A)
```

RustedBones: I forgot `A` is not a generic type but a `SpecificRecord` :man_facepalming:

RustedBones (approved):

scalafix/output-0_14/src/main/scala/fix/v0_14_0/FixAvroCoder16.scala
clairemcginty: haha, we probably could have given it a better class name 😅

PR #5352, opened 2024-05-02
[scio-core](feature) Sample SCollection with max weight

RustedBones:
Create a sampling API with weighted limit and and byte size limit

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -54,7 +54,6 @@ import scala.collection.immutable.TreeMap
 import scala.reflect.ClassTag
 import scala.util.Try
 import com.twitter.chill.ClosureCleaner
-import org.apache.beam.sdk.util.common.ElementByteSizeObserver
 import org.typelevel.scalaccompat.annotation.{nowarn, unused}
 
 /** Convenience functions for creating SCollections. */
@@ -488,21 +487,8 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   def batchByteSized(
     batchByteSize: Long,
     maxLiveWindows: Int = BatchDoFn.DEFAULT_MAX_LIVE_WINDOWS
-  ): SCollection[Iterable[T]] = {
-    val bCoder = CoderMaterializer.beam(context, coder)
-    val weigher = Functions.serializableFn[T, java.lang.Long] { e =>
-      var size: Long = 0L
-      val observer = new ElementByteSizeObserver {
-        override def reportElementSize(elementByteSize: Long): Unit = size += elementByteSize
-      }
-      bCoder.registerByteSizeObserver(e, observer)
-      observer.advance()
-      size
-    }
-    this
-      .parDo(new BatchDoFn[T](batchByteSize, weigher, maxLiveWindows))(Coder.aggregate)
-      .map(_.asScala)
-  }
+  ): SCollection[Iterable[T]] =
+    batchWeighted(batchByteSize, ScioUtil.elementByteSize(context), maxLiveWindows)
 
   /**
    * Batches elements for amortized processing. Elements are batched per-window and batches emitted
@@ -524,7 +510,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     cost: T => Long,
     maxLiveWindows: Int = BatchDoFn.DEFAULT_MAX_LIVE_WINDOWS
   ): SCollection[Iterable[T]] = {
-    val weigher = Functions.serializableFn(cost.andThen(_.asInstanceOf[java.lang.Long]))
+    val weigher = Functions.serializableFn(cost.andThen(Long.box))
     this
       .parDo(new BatchDoFn[T](batchWeight, weigher, maxLiveWindows))(Coder.aggregate)
       .map(_.asScala)
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -970,7 +970,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     this.aggregate(new WeightedHeapAggregator(totalWeight, cost))
   }
 
-  def sampleBySize(totalByteSize: Long): SCollection[Iterable[T]] = {
+  def sampleByteSized(totalByteSize: Long): SCollection[Iterable[T]] = {
     val bCoder = CoderMaterializer.beam(context, coder)
     val weigher = { (e: T) =>
       var size: Long = 0L
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
@@ -33,7 +33,6 @@ import com.spotify.scio.testing.TestDataManager
 import com.spotify.scio.transforms.BatchDoFn
 import com.spotify.scio.util.FilenamePolicySupplier
 import com.spotify.scio.util._
-import com.spotify.scio.util.random.{BernoulliSampler, PoissonSampler}
 import com.twitter.algebird.{Aggregator, Monoid, MonoidAggregator, Semigroup}
 import org.apache.beam.sdk.coders.{ByteArrayCoder, Coder => BCoder}
 import org.apache.beam.sdk.schemas.SchemaCoder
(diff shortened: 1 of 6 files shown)


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+
+    override def present(reduction: mutable.PriorityQueue[(Long, T)]): Iterable[T] =
+      reduction.iterator.map(_._2).toSeq
+  }
+
+  def sampleWeighted(
+    totalWeight: Long,
+    cost: T => Long
+  ): SCollection[Iterable[T]] = {
+    implicit val ordering: Ordering[(Long, T)] = Ordering.by(_._1)
```
RustedBones, line 969: implicit has to be declared here for the coder

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+  }
+
+  def sampleBySize(totalByteSize: Long): SCollection[Iterable[T]] = {
+    val bCoder = CoderMaterializer.beam(context, coder)
+    val weigher = { (e: T) =>
+      var size: Long = 0L
+      val observer = new ElementByteSizeObserver {
+        override def reportElementSize(elementByteSize: Long): Unit = size += elementByteSize
+      }
+      bCoder.registerByteSizeObserver(e, observer)
```
RustedBones, line 980: this is better than `CoderUtils.encodeToByteArray(beamCoder, v).length` because it uses a ` new CountingOutputStream(ByteStreams.nullOutputStream()))` behind the scene.

The code can be factorized with `batchByteSized` moving out of draft state

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+
+  def sampleWeighted(
+    totalWeight: Long,
+    cost: T => Long
+  ): SCollection[Iterable[T]] = {
+    implicit val ordering: Ordering[(Long, T)] = Ordering.by(_._1)
+    this.aggregate(new WeightedHeapAggregator(totalWeight, cost))
+  }
+
+  def sampleBySize(totalByteSize: Long): SCollection[Iterable[T]] = {
```
RustedBones, line 973: Should be `sampleByteSized` for consistency with `batchByteSized`

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     _.pApply(Sample.fixedSizeGlobally(sampleSize)).map(_.asScala)
   }
 
+  final private class WeightedHeapAggregator(totalWeight: Long, cost: T => Long)(implicit
+    ordering: Ordering[(Long, T)]
+  ) extends Aggregator[T, mutable.PriorityQueue[(Long, T)], Iterable[T]] {
```
RustedBones, line 927: Probably have to move this one to another file

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+              (cost(rvalue), right.dequeue())
+            case (Some((_, lvalue)), _) =>
+              (cost(lvalue), left.dequeue())
+            case (_, Some((_, rvalue))) =>
+              (cost(rvalue), right.dequeue())
+          }
+          weight += w
+          if (weight <= totalWeight) {
+            queue.enqueue(v)
+          }
```
RustedBones, line 956: In `batchByteSized`, we exceed the size and send the batch. Here we do not add the last element. WDYT is best? We should align the 2 behaviors.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
...
+      if (cost(value) <= totalWeight) {
+        queue.enqueue((random.nextLong(), value))
+      }
+      queue
+    }
+
+    override def semigroup: Semigroup[mutable.PriorityQueue[(Long, T)]] = Semigroup.from {
+      (left, right) =>
+        var weight = 0L
+        val queue = mutable.PriorityQueue[(Long, T)]()
```
kellen, line 941: This feels overcomplicated; shouldn't we just start w/ the most-full of the queues and add all the items from the other one, letting the queue evict the lesser-weighted elements?

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
kellen: Nice

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -920,6 +922,68 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     _.pApply(Sample.fixedSizeGlobally(sampleSize)).map(_.asScala)
   }
 
+  final private class WeightedHeapAggregator(totalWeight: Long, cost: T => Long)(implicit
```
kellen, line 925: This is potentially a good idea, but for #5321 we explicitly want random elements, not the largest ones. i.e. we'd want `Aggregator[T, (Long, List[T]), Iterable[T]]` where the `Long` is just the total byte count added so far, where any additional elements over the target would just be dropped. 

For the Zstd use-case getting close to the target is good enough (thus the random sampling) but you could argue that for an aggregator you should keep trying to add elements until the List is as full as possible. :shrug:

RustedBones: > where the Long is just the total byte count added so far

The `Long` in the priority queue is the result of `Random.nextLong()`, not the weight. This is the same strategy used for the `sample(sampleSize: Int)`  to select elements uniformly at random.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
RustedBones: The implementation is based on a binary tree heap, where removal of lowest priority is costly. To achieve good result we should go with a [fibonnacci heap](https://en.wikipedia.org/wiki/Fibonacci_heap) priority queue, where merge and min removals are cheap.

I'm also annoyed on how algebird forces us to `prepare` each element in a mergable structure. I'm thinking of changing for `combine`, so we ca reuse the priority queue

RustedBones: not the `lesser-weighted` but the lower priority yes

RustedBones: Actually we can achieve this with a normal priority queue, with min heap.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
kellen: Ahhh indeed, sorry, missed this

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
kellen: IMO it is good enough to see "adding next item would exceed batch size" and sending the batch then. I guess the only edge case is when there is a single element larger than the batch size in which case :shrug: drop it? send it alone?

Looking for "best fit" seems like an overoptimization for the cases this api would be useful

RustedBones: agreed. In batch all elements needs to be sent, even when a single element is bigger than the batch size.

In the sample, reduce drops elements until we're bellow the size, so it feels a consistent behavior to not even add it if it exceeds the size.

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -916,9 +901,17 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
    *   a new SCollection whose single value is an `Iterable` of the samples
    * @group transform
    */
-  def sample(sampleSize: Int): SCollection[Iterable[T]] = this.transform {
-    _.pApply(Sample.fixedSizeGlobally(sampleSize)).map(_.asScala)
-  }
+  // TODO move to implicit
```
kellen, line 904: Do you want to resolve this in this PR?

kellen (approved):

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
RustedBones: This would be binary breaking. it is a reminder for later. maybe worth an issue with milestone

Issue #5300, opened 2024-03-13
Improve error message when sparkey hits array-size limits


kellen:
When using `saveAsSparkey`, if any shard is ~>2gb then you will get a coder exception and something like
```
Error message from worker: org.apache.beam.sdk.util.UserCodeException: java.lang.OutOfMemoryError: Required array length 2147483639 + 15534 is too large
```

which is not easily interpretable.

See if we can preemptively capture serialized sizes so that we can issue a better error message like "Increase number of sparkey shards"

PR #5402, opened 2024-06-24
Use hash join when writing sparkey

aslotnick:
When writing to sparkey, `allShards` represents every expected shard even if there is no corresponding data in `shards` for that shard number.

`shards.rightOuterJoin(allShards)` (added in https://github.com/spotify/scio/pull/5208) fails when a shard contains large amounts of data, leading to the error described in https://github.com/spotify/scio/issues/5300: `java.lang.OutOfMemoryError: Required array length 2147483639 + 15534 is too large`. 

This PR replaces `rightOuterJoin` with `hashFullOuterJoin` (note that there is no `hashRightOuterJoin` implementation). A hash join is a good fit because the right-hand side contains very little data (only the keys of the shards) and it doesn't need to use an array to represent the large left-hand side's values. As a result, some failing workflows that succeeded in Scio 0.13.* will run successfully again.

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/SparkeyIO.scala
@@ -124,7 +124,7 @@ object SparkeyIO {
 
       // write files to temporary locations
       val tempShardUris = shards
-        .rightOuterJoin(allShards)
+        .hashFullOuterJoin(allShards)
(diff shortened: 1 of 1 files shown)


shnapz (approved):
`allShards` should fit into memory given numShards is `Short`

kellen (approved):
Hey thanks for looking into this, was about to start investigating myself. This seems like it should work and still sidestep the issues fixed in #5208

PR #5411, opened 2024-06-27
Support writing extra metadata in scio-parquet

clairemcginty:
Parquet 0.14 supports setting extra file metadata via ParquetWriter: https://github.com/apache/parquet-java/pull/1241

I added a new `metadata` output param to the scio-parquet Avro/Magnolify/Tensorflow bindings, which matches the naming convention we use in the scio-avro write APIs.


clairemcginty:
This has some breaking API changes and is not urgent, can probably wait for 0.15 release.

PR #5414, opened 2024-07-03
Use sbt-bom

RustedBones:
Leverage `sbt-bom` [plugin](https://github.com/heremaps/here-sbt-bom) and override versions for
- `beam`
- `gcp`
- `jackson`
- `netty`


build.sbt
```diff
@@ -445,66 +432,78 @@ lazy val unusedCompileDependenciesTestSkipped = Def.task {
...
+    resolvers ++= Resolver.sonatypeOssRepos("public"),
+    excludeDependencies += Exclude.beamKafka,
+    excludeDependencies ++= Exclude.loggerImplementations,
+    dependencyOverrides ++= beamBom.key.value.bomDependencies ++
+      gcpBom.key.value.bomDependencies ++
+      jacksonBom.key.value.bomDependencies ++
+      nettyBom.key.value.bomDependencies ++
+      Seq(
+        // slf4j-bom only available for v2
+        "org.slf4j" % "slf4j-api" % slf4jVersion
```
RustedBones, line 480: we absolutely need to downgrade to slf4j v1 as some deps are pulling v2

build.sbt
```diff
@@ -694,8 +687,8 @@ lazy val `scio-core` = project
       "com.github.ben-manes.caffeine" % "caffeine" % caffeineVersion % Provided,
       "com.google.apis" % "google-api-services-dataflow" % googleApiServicesDataflowVersion % Provided,
       "org.apache.beam" % s"beam-runners-flink-$flinkMinorVersion" % beamVersion % Provided,
-      "org.apache.beam" % "beam-runners-google-cloud-dataflow-java" % beamVersion % Provided,
-      "org.apache.beam" % s"beam-runners-spark-$sparkMajorVersion" % beamVersion % Provided,
+      "org.apache.beam" % "beam-runners-google-cloud-dataflow-java" % beamVersion % Provided excludeAll (Exclude.avro),
```
shnapz, line 690: why do you excluding avro in this PR?

RustedBones: I'm not using `dependencyOverrides` for avro anymore to check the pulled version is the expected one. This means we need to exclude also in non compile scope.

I'm actually wondering if we should go with the flow an let OSS scio upgrade to latest avro WDYT ?

RustedBones: Since #5415, no need to evict

PR #5416, opened 2024-07-03
Support BigQuery JSON column type

turb:
Adds support for the JSON column type on BigQuery.

This mimics what has been done for `GEOGRAPHY`: use a simple `case class Json(wkt: String)` container. There are a couple of changes I have copied without knowing their use, so they may have none (like in `StorageUtil`).

An alternative may be to store it into a Json parser implementation model (eg Jackson), however it would tie it to more complex things.

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
@@ -204,31 +204,14 @@ private[types] object TypeProvider {
           tq"${TypeName(s"Function${fields.size}")}[..${fields.map(_.children.head)}, $cName]"
         val traits = (if (fields.size <= 22) Seq(fnTrait) else Seq()) ++ defTblDesc
           .map(_ => tq"${p(c, SType)}.HasTableDescription")
-        val taggedFields = fields.map {
-          case q"$m val $n: _root_.com.spotify.scio.bigquery.types.Geography = $rhs" =>
-            provider.initializeToTable(c)(m, n, tq"_root_.java.lang.String")
-            c.universe.ValDef(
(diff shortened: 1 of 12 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -213,6 +213,14 @@ private[types] object TypeProvider {
               tq"_root_.java.lang.String @${typeOf[BigQueryTag]}",
               q"{$rhs}.wkt"
             )
+          case q"$m val $n: _root_.com.spotify.scio.bigquery.types.Json = $rhs" => // Could not find how to mutualize
```
RustedBones, line 216: I ran some test. This branch is not matched (nor is the Geography one).
I'm wondering if `toTable` works in that case

turb: I only tested in real life `fromTable` (that I needed) TBH. But I may need `toTable` in the distant future.

RustedBones: I'm trying to adapt the integration test to make sure those types work as expected

RustedBones (approved):
I added the `Geography` and `Json` types in the integration test and table creation + insertion are working.

RustedBones:
Thank you!

turb:
It seems something does not work as expected with REPEATED JSON: I am getting two lines for the same entry, the JSON repeated data being splited between the two...

RustedBones:
I'm not sure I get the end state, can you give an example ?

turb:
Nevermind, I did some tests and could not reproduce it. I'll open a bug if I find it back, it may not be related.

PR #5428, opened 2024-07-11
[gcp] Fix BQ BIGNUMERIC type support

RustedBones:
`BIGNUMERIC` is currently not properly supported.
- `toTable` cannot infer a schema of type `BIGNUMERIC`
- reading from a `BIGNUMERIC` table returns badly converted `BigDecimal` (using wrong scale/precision)

Introduce new type `BigNumeric` to solve this issue

### DIFF
--- a/integration/src/test/scala/com/spotify/scio/bigquery/types/BigQueryStorageIT.scala
+++ b/integration/src/test/scala/com/spotify/scio/bigquery/types/BigQueryStorageIT.scala
@@ -50,8 +50,9 @@ class BigQueryStorageIT extends AnyFlatSpec with Matchers {
         dt.toLocalDate.plusDays(i),
         dt.toLocalTime.plusMillis(i),
         dt.toLocalDateTime.plusMillis(i),
-        s"POINT($i $i)", // geography is not an avro logical type
-        s"""{"value":$i}""" // json is not an avro logical type
+        Geography(s"POINT($i $i)"),
+        Json(s"""{"value":$i}"""),
+        BigNumeric(BigDecimal(i))
       )
     }.asJava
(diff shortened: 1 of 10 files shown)


integration/src/test/scala/com/spotify/scio/bigquery/types/BigQueryStorageIT.scala
```diff
@@ -50,8 +50,9 @@ class BigQueryStorageIT extends AnyFlatSpec with Matchers {
         dt.toLocalDate.plusDays(i),
         dt.toLocalTime.plusMillis(i),
         dt.toLocalDateTime.plusMillis(i),
-        s"POINT($i $i)", // geography is not an avro logical type
-        s"""{"value":$i}""" // json is not an avro logical type
```
RustedBones, line 54: I found out that BQ adds custom `sqlType` property to the schema. We can reconstruct the original type

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -148,11 +153,9 @@ private[types] object ConverterProvider {
     // =======================================================================
     // Entry point
     // =======================================================================
-
-    val tn = TermName("r")
     q"""(r: _root_.org.apache.avro.generic.GenericRecord) => {
           import _root_.scala.jdk.javaapi.CollectionConverters._
-          ${constructor(tpe, tn)}
```
RustedBones, line 155: This is technically an `Ident(TermName("r"))`, adapt macro

kellen (approved):
nice!

PR #5431, opened 2024-07-19
Update to BigQuery read API v1 interface

kmjung:
This change updates Scio's usage of the [BigQuery storage read API](https://cloud.google.com/bigquery/docs/reference/storage) to use the `v1` interface rather than `v1beta1`, which will eventually be deprecated. The two interfaces are very similar, to the extent that this change is mostly string substitution.

I haven't yet been able to run end-to-end tests with this change. If there is no automated test coverage as part of the PR approval process, then I will figure out how to do so.


RustedBones:
Thanks for the heads up! We completely missed this upgrade.
CI fails with some expected binary compatibility issue. We'll plan this upgrade in next `0.15` release

kmjung:
> Thanks for the heads up! We completely missed this upgrade.

To be clear, you didn't miss anything per se -- the official deprecation notice hasn't been sent out yet (although this is coming). However, Scio is one of the last read API use cases that I'm aware of which is in wide use and has yet to migrate to v1. If we can start the adoption process for the newer API surface now, the deprecation process will hopefully be less impactful for users.

kmjung:
Hey folks, do you need anything from me here? If not, then is there a timeline for this to be merged?

clairemcginty (approved):

PR #5438, opened 2024-07-24
Improve accuracy of FixQuery rule

clairemcginty:
The accuracy of FixQuery was not good when used on `@BigQueryType.from` macros. By matching on the `<macro>.query.from` expression, we can pick up on the macro class's ClassSignature, which includes `HasQuery` as a parent.

### DIFF
--- a/scalafix/rules/src/main/scala/fix/v0_14_0/FixQuery.scala
+++ b/scalafix/rules/src/main/scala/fix/v0_14_0/FixQuery.scala
@@ -6,17 +6,10 @@ import scala.meta._
 object FixQuery {
   val HasQueryMatcher: SymbolMatcher =
     SymbolMatcher.normalized("com/spotify/scio/bigquery/types/BigQueryType/HasQuery")
-
-  val HasQueryMethodMatcher: SymbolMatcher =
-    SymbolMatcher.normalized("com/spotify/scio/bigquery/types/BigQueryType/HasQuery#query")
 }
 
 class FixQuery extends SemanticRule("FixQuery") {
-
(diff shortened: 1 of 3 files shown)


kellen (approved):

RustedBones (approved):

scalafix/rules/src/main/scala/fix/v0_14_0/FixQuery.scala
```diff
@@ -13,14 +16,26 @@ class FixQuery extends SemanticRule("FixQuery") {
...
 
   override def fix(implicit doc: SemanticDocument): Patch = {
     doc.tree.collect {
-      case t @ q"$qual.$fn" if HasQueryMatcher.matches(fn) =>
+      case t @ q"$qual.$fn" if HasQueryMethodMatcher.matches(fn) =>
         Patch.replaceTree(t, q"$qual.queryRaw".syntax)
       case t @ q"$qual.query" if isQuery(t) =>
         Patch.replaceTree(t, q"$qual.queryRaw".syntax)
+      case t @ q"$qual.query.format" if isQueryFormat(qual) =>
+        Patch.replaceTree(t, q"$qual.queryRaw.format".syntax)
```
RustedBones, line 38: Shouldn't we try to directly use `query(..)` in that case ?
```suggestion
        Patch.replaceTree(t, q"$qual.query".syntax)
```

clairemcginty: I simplified it to single `case` matching `query` directly 👍

PR #5441, opened 2024-08-07
Fix AsyncBatchLookup to emit element with restored metadata

RustedBones:
`flush` call from the `processElement` was not propagating the original element's window and timestamp.


scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncBatchLookupDoFn.java
```diff
@@ -158,10 +160,14 @@ public void startBundle(StartBundleContext context) {
   public void processElement(
       @Element Input input,
       @Timestamp Instant timestamp,
+      PaneInfo pane,
       OutputReceiver<KV<Input, TryWrapper>> out,
       BoundedWindow window) {
     inputCount++;
-    flush(r -> out.output(KV.of(r.input, r.output)));
```
RustedBones, line 164: Looks we were ouputing using the current element's metadata

clairemcginty (approved):

clairemcginty (changes requested):
lgtm after rebasing the bigtable changes :)

RustedBones:
@clairemcginty updated the PR. All async DoFn are affected

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
```diff
@@ -155,40 +159,58 @@ public void startBundle(StartBundleContext context) {
     semaphore.release(maxPendingRequests);
   }
 
+  // kept for binary compatibility. Must not be used
+  // TODO: remove in 0.15.0
+  @Deprecated
```
clairemcginty, line 164: Will this deprecation warning get logged on the user's side? I guess only if they're pulling in an older Beam version that uses the old API, right?

scio-core/src/main/scala/com/spotify/scio/transforms/ScalaAsyncLookupDoFn.scala
```diff
@@ -25,33 +25,33 @@ import scala.util.{Failure, Success, Try}
 /**
  * A [[org.apache.beam.sdk.transforms.DoFn DoFn]] that performs asynchronous lookup using the
  * provided client for Scala [[Future]].
- * @tparam A
+ * @tparam Input
```
clairemcginty, line 28: Nice refactor here 👍 much clearer

scio-core/src/test/scala/com/spotify/scio/transforms/AsyncBatchLookupDoFnTest.scala
```diff
@@ -0,0 +1,364 @@
+/*
+ * Copyright 2019 Spotify AB.
```
clairemcginty, line 2:
```suggestion
 * Copyright 2024 Spotify AB.
```

clairemcginty (approved):

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncLookupDoFn.java
RustedBones: Only if extending the class and overriding this method which should never happen.
However a library compiled with the old API will call this method, so I've left it for now.

PR #5443, opened 2024-08-08
add scala.collection.immutable.SortedMap coder

pgoggijr:
Adds coder support for `scala.collection.immutable.SortedMap`

Implementation is almost identical to that of `Map` - not sure if there's an easy way to reuse the code there since it needs `K: Ordering`

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
@@ -256,23 +256,22 @@ private class MutablePriorityQueueCoder[T: Ordering](bc: BCoder[T])
 }
 
 private[coders] class BitSetCoder extends AtomicCoder[BitSet] {
-  private[this] val lc = VarIntCoder.of()
 
   def decode(in: InputStream): BitSet = {
-    val l = lc.decode(in)
+    val l = VarInt.decodeInt(in)
     val builder = BitSet.newBuilder
     builder.sizeHint(l)
-    (1 to l).foreach(_ => builder += lc.decode(in))
+    (1 to l).foreach(_ => builder += VarInt.decodeInt(in))
 
     builder.result()
   }
 
   def encode(ts: BitSet, out: OutputStream): Unit = {
-    lc.encode(ts.size, out)
-    ts.foreach(v => lc.encode(v, out))
+    VarInt.encode(ts.size, out)
+    ts.foreach(v => VarInt.encode(v, out))
   }
 
-  override def consistentWithEquals(): Boolean = lc.consistentWithEquals()
+  override def consistentWithEquals(): Boolean = true
 
   override def toString: String = "BitSetCoder"
 }
--- a/scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
@@ -362,6 +362,14 @@ private class MutableMapCoder[K, V](kc: BCoder[K], vc: BCoder[V])
 private[coders] class SortedMapCoder[K: Ordering, V](kc: BCoder[K], vc: BCoder[V])
     extends MapLikeCoder[K, V, SortedMap](kc, vc) {
 
+  override def encode(value: SortedMap[K, V], os: OutputStream): Unit = {
+    require(
+      value.ordering == Ordering[K],
+      "SortedMap ordering does not match SortedMapCoder ordering"
+    )
+    super.encode(value, os)
+  }
+
   override def decode(is: InputStream): SortedMap[K, V] =
     decode(is, SortedMap.newBuilder[K, V])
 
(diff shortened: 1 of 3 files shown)


scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -335,6 +336,75 @@ private[coders] class MapCoder[K, V](val kc: BCoder[K], val vc: BCoder[V])
...
+      i = i + 1
+    }
+    builder.result()
+  }
+
+  // delegate methods for determinism and equality checks
+  override def verifyDeterministic(): Unit =
+    throw new NonDeterministicException(
+      this,
+      "Ordering of entries in a SortedMap may be non-deterministic."
```
pgoggijr, line 373: This is probably still true if two elements are equal according to `Ordering[K]`?

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -335,6 +336,75 @@ private[coders] class MapCoder[K, V](val kc: BCoder[K], val vc: BCoder[V])
...
+    )
+  override def consistentWithEquals(): Boolean =
+    kc.consistentWithEquals() && vc.consistentWithEquals()
+  override def structuralValue(value: SortedMap[K, V]): AnyRef =
+    if (consistentWithEquals()) {
+      value
+    } else {
+      val b = Map.newBuilder[Any, Any]
```
pgoggijr, line 381: I'm not sure of how this should be implemented since Ordering will need to be compared

pgoggijr: Although the java implementation seems to do this same thing so maybe it's okay

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -335,6 +336,75 @@ private[coders] class MapCoder[K, V](val kc: BCoder[K], val vc: BCoder[V])
...
+  ): Unit = {
+    lc.registerByteSizeObserver(value.size, observer)
+    value.foreach { case (k, v) =>
+      kc.registerByteSizeObserver(k, observer)
+      vc.registerByteSizeObserver(v, observer)
+    }
+  }
+
+  override def toString: String =
+    s"MapCoder($kc, $vc)"
```
RustedBones, line 403:
```suggestion
    s"SortedMapCoder($kc, $vc)"
```

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -335,6 +336,75 @@ private[coders] class MapCoder[K, V](val kc: BCoder[K], val vc: BCoder[V])
   override def getCoderArguments: JList[_ <: BCoder[_]] = List(kc, vc).asJava
 }
 
+private[coders] class SortedMapCoder[K: Ordering, V](
```
RustedBones, line 339: Now we have several `MapCoder`s, let's factorize some code in a `MapLikeCoder` abstract class

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -335,6 +336,75 @@ private[coders] class MapCoder[K, V](val kc: BCoder[K], val vc: BCoder[V])
...
+      b.result()
+    }
+
+  // delegate methods for byte size estimation
+  override def isRegisterByteSizeObserverCheap(value: SortedMap[K, V]): Boolean = false
+  override def registerByteSizeObserver(
+    value: SortedMap[K, V],
+    observer: ElementByteSizeObserver
+  ): Unit = {
+    lc.registerByteSizeObserver(value.size, observer)
```
RustedBones, line 395: This looks wrong. Here wn need give the size of the encoded length, (from the `VarIntCoder`), not the length of the map.
I see you copied from `MapCoder`, so we have a big there

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
RustedBones: Map can't have 2 keys equal (`get` must return a single value). This can be set as `verifyDeterministic`

pgoggijr: ah cool yeah, I tried this:
```
scala> SortedMap((5, 2), (3, 8))((_, _) => 0)
val res0: scala.collection.immutable.SortedMap[Int,Int] = TreeMap(5 -> 8)
```

RustedBones:
Thank you @pgoggijr !

clairemcginty (approved):

scio-core/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -123,7 +125,7 @@ final class CoderTest extends AnyFlatSpec with Matchers {
       materializeTo[MapCoder[_, _]] and
       beFullyCompliantNonDeterministic()
 
-    SortedMap.from(m) coderShould roundtrip() and
```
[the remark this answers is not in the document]
pgoggijr: oh nice!

PR #5456, opened 2024-08-20
Refactor BQ to expose all beam's configurations

### LINKED ISSUE #5530: Add ErrorHandling in BigQuery
Error handling is exposed in beam Bigquery IO. Expose this in scio's API

RustedBones:
Here are the main changes:

- the BQ `Table` source has a single normalized definition, with multiple constrictors (form string spec or `TableReference`). It nows includes an optional `Table.Filter` that can be used is the storage read API to project and filter.

- read API changes with

| API                   | method | returned type |
|-----------------------|--------|---------------|
| bigQuerySelect        | export | TableRow      |
| bigQuerySelectFormat  | export | T             |
| bigQueryTable         | export | TableRow      |
| bigQueryTableFormat   | export | T             |
| bigQueryStorage       | direct | TableRow      |
| bigQueryStorageFormat | direct | T             |
| typedBigQuery         | export | T             |
| typedBigQueryStorage  | direct | T             |

`Format` API take a `BigqueryIO.Format` object allowing to convert either from `GenericRecord` (this should be prefered) or `TableRow`

- The `Storage` Api allow to pass an `ErrorHandler` (Fix #5530). In order to preserve a flat structure `ScioContext.errorSink(): ErrorSink` has been added. This allow to do the following

```scala
val errorSink = sc.errorSink()
sc.bigQueryStorageFormat[MyType](
  table,
  format,
  errorHandler = errorSink.handler
)
val errors: SCollection[BadRecord] = errorSink.sink()
```
The `handler` can be passed to multiple IOs before `sink` is materialized. The sink will flatten the errors from the IOs.


scio-core/src/main/scala/com/spotify/scio/values/ErrorSink.scala
```diff
@@ -0,0 +1,56 @@
...
+
+package com.spotify.scio.values
+
+import com.spotify.scio.ScioContext
+import org.apache.beam.sdk.transforms.PTransform
+import org.apache.beam.sdk.transforms.errorhandling.{BadRecord, ErrorHandler}
+import org.apache.beam.sdk.values.{PCollection, PCollectionTuple, TupleTag}
+
+/**
+ * A sink for error records.
```
clairemcginty, line 25: bit more explanation on error records could be helpful, maybe:

```suggestion
 * A sink for error records.
 *
 * An error record is produced by certain PTransforms that catch processing exceptions and transform the resulting (element, exception) pair into a [[BadRecord]] instance.
 * When an ErrorSink is configured (via ScioContext#errorSink), these BadRecords can be accessed as an SCollection by invoking the ErrorSink#sink method.
 * An ErrorSink is useful if you'd like to set up special handling of exceptions (incrementing Counters, logging the exceptions in a database, etc).
```

scio-core/src/main/scala/com/spotify/scio/values/ErrorSink.scala
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Once the [[sink]] is materialized, the [[handler]] must not be used anymore.
+ */
+sealed trait ErrorSink {
+  def handler: ErrorHandler[BadRecord, _]
```
clairemcginty, line 30: could `def handler` be `private[scio]`? not sure when a user would need to access this

RustedBones: This is the API exposed by beam. As mentioned in the description we do not pass the `ErrorSink` directly.
```scala
sc.bigQueryStorageFormat[MyType](
  table,
  format,
  errorHandler = errorSink.handler
)
```

I was thinking of adding to the `ScioContext` a beam java like API too

```scala
def registerBadRecordErrorHandler[T](handler: PTransform[PCollection[BadRecord], T] sinkTransform): BadRecordErrorHandler[OutputT]
```

kellen:
Intended to fix #5530

PR #5469, opened 2024-09-03
use noOp cache instead of null when instantiating typedSparkeySideInput

pgoggijr:
When attempting to access an element in a side input instantiated by `typedSparkeySideInput` where `cache` isn't specified, the following error is thrown since the cache is explicitly set to null:

```
Error message from worker: java.lang.NullPointerException: Cannot invoke "com.spotify.scio.util.Cache.get(Object, scala.Function0)" because the return value of "com.spotify.scio.extra.sparkey.instances.TypedSparkeyReader.cache()" is null
	com.spotify.scio.extra.sparkey.instances.TypedSparkeyReader.get(TypedSparkeyReader.scala:42)
	com.spotify.scio.extra.sparkey.instances.TypedSparkeyReader.get(TypedSparkeyReader.scala:31)
```

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
@@ -150,9 +150,17 @@ package object sparkey extends SparkeyReaderInstances with SparkeyCoders {
     def typedSparkeySideInput[T](
       basePath: String,
       decoder: Array[Byte] => T,
-      cache: Cache[String, T] = Cache.noOp[String, T]
+      cache: Cache[String, T] = null
     ): SideInput[TypedSparkeyReader[T]] =
-      sparkeySideInput(basePath, reader => new TypedSparkeyReader[T](reader, decoder, cache))
+      sparkeySideInput(
+        basePath,
+        reader =>
+          new TypedSparkeyReader[T](
+            reader,
+            decoder,
+            if (cache == null) Cache.noOp[String, T] else cache
+          )
+      )
 
     /**
      * Create a SideInput of `CachedStringSparkeyReader` from a [[SparkeyUri]] base path, to be used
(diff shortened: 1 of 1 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -150,7 +150,7 @@ package object sparkey extends SparkeyReaderInstances with SparkeyCoders {
     def typedSparkeySideInput[T](
       basePath: String,
       decoder: Array[Byte] => T,
-      cache: Cache[String, T] = null
+      cache: Cache[String, T] = Cache.noOp[String, T]
```
RustedBones, line 153: To be more in line with beam and other IOs, we've almost always used `null` as default, handled internally later on.
Can you update the PR to stick to this style ?

pgoggijr: just pushed a change to add a null check within the method instead - is that in-line with that style? or would it be preferrable to handle a null `cache` in the `TypedSparkeyReader` class itself?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -152,7 +152,15 @@ package object sparkey extends SparkeyReaderInstances with SparkeyCoders {
...
       cache: Cache[String, T] = null
     ): SideInput[TypedSparkeyReader[T]] =
-      sparkeySideInput(basePath, reader => new TypedSparkeyReader[T](reader, decoder, cache))
+      sparkeySideInput(
+        basePath,
+        reader =>
+          new TypedSparkeyReader[T](
+            reader,
+            decoder,
+            if (cache == null) Cache.noOp[String, T] else cache
```
RustedBones, line 161:
```suggestion
           Option(cache).getOrElse(Cache.noOp[String, T])
```

PR #5494, opened 2024-09-17
Managed + Iceberg IO

kellen:
Adds support for Beam's managed transforms and for Iceberg, which is implemented as a managed transform.

Note this is on a snapshot of magnolify, will need a release before merge.


scio-managed/src/main/scala/com/spotify/scio/managed/ManagedIO.scala
```diff
@@ -0,0 +1,71 @@
...
+    def _convert(a: Object): Object = {
+      a match {
+        case m: Map[_, _] =>
+          m.asInstanceOf[Map[_, Object]].map { case (k, v) => k -> _convert(v) }.asJava
+        case i: Iterable[_] => i.map(o => _convert(o.asInstanceOf[Object])).asJava
+        case _              => a
+      }
+    }
+    config.map { case (k, v) => k -> _convert(v) }.asJava
+  }
```
RustedBones, line 45: I'm wondering if we should introduce either a config AST to ensure what is passed is YAML compatible, or maybe use lightbend [config](https://github.com/lightbend/config/blob/d279841c425e70645feacbdc03e41b3b022ad1f9/config/src/main/java/com/typesafe/config/ConfigObject.java#L81)

scio-managed/src/main/scala/com/spotify/scio/managed/ManagedIO.scala
```diff
@@ -0,0 +1,71 @@
...
+    data.applyInternal(Managed.write(ioName).withConfig(_config))
+    EmptyTap
+  }
+
+  override def tap(read: ManagedIO.ReadParam): Tap[tapT.T] = EmptyTap
+}
+
+object ManagedIO {
+  final case class ReadParam(schema: Schema)
+  final case class WriteParam()
```
RustedBones, line 70: So far, when there was no parameters we've use `Unit`

scio-managed/src/main/scala/com/spotify/scio/iceberg/IcebergIO.scala
```diff
@@ -0,0 +1,79 @@
...
+      "catalog_name" -> catalogName.orNull,
+      "catalog_properties" -> catalogProperties,
+      "config_properties" -> configProperties
+    ).filter(_._2 != null)
+  }
+
+  override protected def read(sc: ScioContext, params: IcebergIO.ReadParam): SCollection[T] =
+    ManagedIO(Managed.ICEBERG, config(params.catalogProperties, params.configProperties))
+      .readWithContext(sc, ManagedIO.ReadParam(rowType.schema))
+      .map(rowType.from)
```
RustedBones, line 56: We've almost never used this pattern.
```suggestion
    val io = ManagedIO(Managed.ICEBERG, config(params.catalogProperties, params.configProperties))
    val params = ManagedIO.ReadParam(rowType.schema)
    sc.transform(_.read(io)(params).map(rowType.from))
```

scio-managed/src/main/scala/com/spotify/scio/iceberg/IcebergIO.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  override protected def write(data: SCollection[T], params: IcebergIO.WriteParam): Tap[tapT.T] = {
+    val tx = data.transform {
+      _.map(rowType.to).setCoder(beamRowCoder)
+    }
+    ManagedIO(Managed.ICEBERG, config(params.catalogProperties, params.configProperties))
+      .writeWithContext(tx, ManagedIO.WriteParam())
+      .underlying
```
RustedBones, line 64: When `ManagedIO.WriteParam` is `Unit`
```suggestion
    val io =  ManagedIO(Managed.ICEBERG, config(params.catalogProperties, params.configProperties))
    data.map(rowType.to).write(io).underlying
```

RustedBones (commented):
Can you add the new module in the `README` ?

RustedBones:
Converted to draft until `magnolify` gets released

scio-managed/src/main/scala/com/spotify/scio/iceberg/IcebergIO.scala
```diff
@@ -0,0 +1,79 @@
...
+
+final case class IcebergIO[T: RowType: Coder](table: String, catalogName: Option[String])
+    extends ScioIO[T] {
+  override type ReadP = IcebergIO.ReadParam
+  override type WriteP = IcebergIO.WriteParam
+  override val tapT: TapT.Aux[T, Nothing] = EmptyTapOf[T]
+
+  private lazy val rowType: RowType[T] = implicitly
+  private lazy val beamRowCoder: RowCoder = RowCoder.of(rowType.schema)
+  implicit private lazy val rowCoder: Coder[Row] = Coder.beam(beamRowCoder)
```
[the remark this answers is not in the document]
kellen: There is unfortunately an internal check in the IcebergIO that the coder is exactly an instance of `RowCoder` so we need to explicitly set the beam coder.

integration/src/test/scala/com/spotify/scio/iceberg/IcebergIOIT.scala
```diff
@@ -0,0 +1,96 @@
...
+    cat.initialize(CatalogName, Map("uri" -> uri).asJava)
+
+    import org.apache.iceberg.Schema
+    cat.createNamespace(Namespace.of(NamespaceName))
+    cat.createTable(
+      TableIdentifier.parse(TableName),
+      new Schema(
+        NestedField.required(0, "a", IntegerType.get()),
+        NestedField.required(1, "b", StringType.get())
+      ),
```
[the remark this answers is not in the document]
kellen: No, this is an Iceberg schema rather than the Beam schema. The lack of create-on-write does raise the question of whether we _also_ need to derive the iceberg schemas

scio-managed/src/main/scala/com/spotify/scio/managed/ManagedIO.scala
kellen: The API can aslo take a yaml file location, e.g. `classpath://foo.yaml` if we wanted to support that.

integration/src/test/scala/com/spotify/scio/iceberg/IcebergIOIT.scala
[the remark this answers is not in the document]
clairemcginty: Maybe `RowType` could also offer a `def icebergSchema`? (Similar to how magnolify-parquet has both `def schema` and `def avroSchema`...)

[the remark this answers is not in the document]
kellen: That would mean pulling in iceberg deps into the beam module fyi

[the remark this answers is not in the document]
clairemcginty: could make it provided, I guess, but point taken

Seems like Beam should have a utility function for converting between Beam/Icerberg Schemas. They have similar stuff for BQ/Avro/BeamSchema interop. Maybe we could contribute there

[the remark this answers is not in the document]
kellen: Beam does have this but it introduces a dep on the iceberg part of the sdk that in theory should be managed.

I could use it in the integration test directly but that wouldn't help users at all.
```
IcebergUtils.beamSchemaToIcebergSchema(rowType.schema)
```

scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -1737,15 +1737,15 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
   }
 
   /**
-   * Generic write method for all `ScioIO[T]` implementations, if it is test pipeline this will
+   * Generic write method for all `ScioIO[T]` implementations, if it is a test pipeline this will
    * evaluate pre-registered output IO implementation which match for the passing `ScioIO[T]`
-   * implementation. if not this will invoke [[com.spotify.scio.io.ScioIO[T]#write]] method along
-   * with write configurations passed by.
+   * implementation. If not, this will invoke [[com.spotify.scio.io.ScioIO[T]#write]] with the
+   * passed write configuration.
```
clairemcginty, line 1743:
```suggestion
   * provided write configuration.
```

scio-managed/src/main/scala/com/spotify/scio/managed/ManagedIO.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import com.spotify.scio.ScioContext
+import com.spotify.scio.io.{EmptyTap, EmptyTapOf, ScioIO, Tap, TapT}
+import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.managed.Managed
+import org.apache.beam.sdk.schemas.Schema
+import org.apache.beam.sdk.values.{PCollectionRowTuple, Row}
+import scala.jdk.CollectionConverters._
+
+final case class ManagedIO(ioName: String, config: Map[String, Object]) extends ScioIO[Row] {
```
clairemcginty, line 27: i guess once this is merged, we can add a `ManagedTypedIO` to the 0.15.x branch? maybe let's file a ticket to track the Magnolify API work...

clairemcginty (commented):
overall looks good to me! the amount of testing+examples are impressive

scio-managed/src/main/scala/com/spotify/scio/managed/ManagedIO.scala
kellen: I don't think we need to support managed very extensively, so I decided to just skip any typed variant for `Managed` in favor of doing it downstream in our io-specific APIs

PR #5495, opened 2024-09-18
Add latest API to SCollection

RustedBones:


clairemcginty (approved):
👀 Add `Latest.perKey()` to [PairSCollectionFunctions](https://github.com/spotify/scio/blob/main/scio-core/src/main/scala/com/spotify/scio/values/PairSCollectionFunctions.scala) as well?

clairemcginty (approved):

PR #5498, opened 2024-09-19
Consistent implementation for reduce operation

RustedBones:


scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala
```diff
@@ -793,7 +793,7 @@ sealed trait SCollection[T] extends PCollectionWrapper[T] {
     val e = ev // defeat closure
     in.map(e.toDouble)
       .asInstanceOf[SCollection[JDouble]]
-      .pApply(Mean.globally())
+      .pApply(Mean.globally().withoutDefaults())
```
RustedBones, line 796: without that, `mean` returns `NaN` for empty collection instead

PR #5502, opened 2024-09-25
Support Snowflake (#5500)

### LINKED ISSUE #5500: Snowflake Support
Hello here,

Apache Beam has Snowflake support, so it's possible to use it with:

```scala
  import import org.apache.beam.sdk.io.snowflake.SnowflakeIO

   val thingMapper = new SnowflakeIO.CsvMapper[Thing] {
       override def mapRow(parts: Array[String]): Thing = Thing(Typed1(parts(0)), Typed2(parts(1)), ...)
    }

    val datasource = SnowflakeIO.DataSourceConfiguration.create()
      .withUsernamePasswordAuth(
        args("snowflake-user"), args("snowflake-password"))
      .withServerName("redacted.snowflakecomputing.com")
      .withDatabase("redacted_db")
      .withRole("redacted_role")
      .withWarehouse("redacted_warehouse")
      .withSchema("redacted_schema")

    val read = SnowflakeIO.read()
      .withDataSourceConfiguration(datasource)
      .fromQuery("SELECT COLUM1, COLUMN2... FROM THINGS")
      .withStagingBucketName("gs://paths") // here for Google Cloud
      .withStorageIntegrationName("redacted_integration")
      .withCsvMapper(thingMapper)
      .withCoder(CoderMaterializer.beam(sc, Thing.coder))

    val things: SCollection[Thing] = sc.customInput("things", read)

```

However a proper scio integration would be great. I suppose derivating `Thing` to `SnowflakeIO.CsvMapper` would need some first a PR in magnolify?

I can work on it.

turb:
Fix #5500

First draft of SnowflakeIO for scio, using the one from Beam.

It has:
- SnowflakeSelect (read only), allowing to read from a select
- SnowflakeTable (rw), allowing to read or write from/to a table
- `ScioContext` and `SCollection` implicits

Notes:
- I wanted to add some unit tests, but was unsure what could be done, since the only docker env I could find is [LocalStack](https://snowflake.localstack.cloud/introduction/) which requires signup
- I did test real-world scenarios for reads, but I ran into https://github.com/apache/beam/issues/24374 on write

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -671,6 +671,7 @@ lazy val scio = project
     `scio-redis`,
     `scio-repl`,
     `scio-smb`,
+    `scio-snowflake`,
     `scio-tensorflow`,
     `scio-test-core`,
     `scio-test-google-cloud-platform`,
@@ -1263,6 +1264,24 @@ lazy val `scio-parquet` = project
       // test
       "org.slf4j" % "slf4j-simple" % slf4jVersion % Test
     )
+  )
+
+lazy val `scio-snowflake` = project
+  .in(file("scio-snowflake"))
+  .dependsOn(
+    `scio-core` % "compile;test->test"
+  )
+  .settings(commonSettings)
+  .settings(
+    description := "Scio add-on for Neo4J",
+    libraryDependencies ++= Seq(
+      // compile
+      "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
+      "org.apache.beam" % "beam-sdks-java-io-snowflake" % beamVersion,
+      "com.nrinaudo" %% "kantan.csv" % kantanCsvVersion,
+      "com.nrinaudo" %% "kantan.codecs" % kantanCodecsVersion
+    ),
+    tlMimaPreviousVersions := Set.empty // TODO: remove once released
   )
 
 val tensorFlowMetadataSourcesDir =
--- a/build.sbt
+++ b/build.sbt
@@ -1273,7 +1273,7 @@ lazy val `scio-snowflake` = project
   )
   .settings(commonSettings)
   .settings(
-    description := "Scio add-on for Neo4J",
+    description := "Scio add-on for Snowflake",
     libraryDependencies ++= Seq(
       // compile
       "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
--- a/build.sbt
+++ b/build.sbt
@@ -1276,10 +1276,11 @@ lazy val `scio-snowflake` = project
     description := "Scio add-on for Snowflake",
     libraryDependencies ++= Seq(
       // compile
-      "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
-      "org.apache.beam" % "beam-sdks-java-io-snowflake" % beamVersion,
+      "com.nrinaudo" %% "kantan.codecs" % kantanCodecsVersion,
       "com.nrinaudo" %% "kantan.csv" % kantanCsvVersion,
-      "com.nrinaudo" %% "kantan.codecs" % kantanCodecsVersion
+      "joda-time" % "joda-time" % jodaTimeVersion,
+      "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
+      "org.apache.beam" % "beam-sdks-java-io-snowflake" % beamVersion
     ),
     tlMimaPreviousVersions := Set.empty // TODO: remove once released
   )
--- a/scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
+++ b/scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2024 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.snowflake
+
+import com.spotify.scio.ScioContext
+import com.spotify.scio.coders.{Coder, CoderMaterializer}
+import com.spotify.scio.io.{EmptyTap, EmptyTapOf, ScioIO, Tap, TapT}
+import com.spotify.scio.values.SCollection
+import kantan.csv.{RowDecoder, RowEncoder}
+import org.apache.beam.sdk.io.snowflake.SnowflakeIO.{CsvMapper, UserDataMapper}
+import org.apache.beam.sdk.io.{snowflake => beam}
+
+object SnowflakeIO {
+
+  private[snowflake] def dataSourceConfiguration(connectionOptions: SnowflakeConnectionOptions) = {
+
+    val datasourceInitial = beam.SnowflakeIO.DataSourceConfiguration
+      .create()
+
+    val datasourceWithAuthent = connectionOptions.authenticationOptions match {
+      case SnowflakeUsernamePasswordAuthenticationOptions(username, password) =>
+        datasourceInitial.withUsernamePasswordAuth(username, password)
+      case SnowflakeKeyPairAuthenticationOptions(username, privateKeyPath, None) =>
+        datasourceInitial.withKeyPairPathAuth(username, privateKeyPath)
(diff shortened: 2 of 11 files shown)


turb:
The build wants me to have `"com.nrinaudo" %% "kantan.csv" % kantanCsvVersion` as a runtime dep, but then it cannot compile. I thought that a compile dep was also supposed to be in runtime (at least in the Maven model).

turb:
@RustedBones I finally had the time to fix this build :)

RustedBones:
Sorry for the delay. Will look at it this week !

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeOptions.scala
```diff
@@ -0,0 +1,100 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.snowflake
+
+trait SnowflakeAuthenticationOptions
```
RustedBones, line 20:
```suggestion
sealed trait SnowflakeAuthenticationOptions
```
To avoid the `match may not be exhaustive` warning

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
```diff
@@ -0,0 +1,134 @@
...
+
+    val datasourceBeforeSchema = datasourceWithAuthent
+      .withServerName(connectionOptions.serverName)
+      .withDatabase(connectionOptions.database)
+      .withRole(connectionOptions.role)
+      .withWarehouse(connectionOptions.warehouse)
+
+    connectionOptions.schema
+      .map(schema => datasourceBeforeSchema.withSchema(schema))
+      .getOrElse(datasourceBeforeSchema)
```
RustedBones, line 54: Nit: to avoid using intermediate variable, you can use syntactic sugar from `scala.util.chaining._`.
(can be done above too)

```suggestion
      .withWarehouse(connectionOptions.warehouse)
      .pipe(ds => connectionOptions.schema.fold(ds)(ds.withSchema))
```

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  private[snowflake] def buildCsvMapper[T](rowDecoder: RowDecoder[T]): CsvMapper[T] =
+    new CsvMapper[T] {
+      override def mapRow(parts: Array[String]): T = {
+        val unsnowedParts = parts.map {
+          case "\\N" => "" // needs to be mapped to an Option
```
RustedBones, line 61: I never used Snowflake. Can you give some context for this case ?

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeOptions.scala
```diff
@@ -0,0 +1,100 @@
...
+ * @param stagingBucketName
+ *   Snowflake staging bucket name where CSV files will be stored
+ * @param storageIntegrationName
+ *   Storage integration name as created in Snowflake
+ */
+final case class SnowflakeOptions(
+  connectionOptions: SnowflakeConnectionOptions,
+  stagingBucketName: String,
+  storageIntegrationName: String
+)
```
RustedBones, line 100: We need to make a distinction between `Options` and `Params`:
- `Options`, passed in the IO constructor, are used to iodentify distinct sources
- `Params` of an IO are some tuning parameters to consume the IO.

We need to split the models here to reflect that

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
```diff
@@ -0,0 +1,134 @@
...
+          SnowflakeIO.dataSourceConfiguration(snowflakeOptions.connectionOptions)
+        )
+        .to(table)
+        .withStagingBucketName(snowflakeOptions.stagingBucketName)
+        .withStorageIntegrationName(snowflakeOptions.storageIntegrationName)
+        .withUserDataMapper(new UserDataMapper[T] {
+          override def mapRow(element: T): Array[AnyRef] = rowEncoder.encode(element).toArray
+        })
+    )
+    EmptyTap
```
RustedBones, line 130: Isn't it possible to create a tap ?

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
turb: `SnowflakeIO` from Apache Beam uses `COPY` from Snowflake, that simply exports to CSV, that is then read from storage. And this copy uses `\N` the represent `null`.

Thing is, kantan maps empty string with `None`, so we provide it.

build.sbt
```diff
@@ -1265,6 +1266,24 @@ lazy val `scio-parquet` = project
...
   )
 
+lazy val `scio-snowflake` = project
+  .in(file("scio-snowflake"))
+  .dependsOn(
+    `scio-core` % "compile;test->test"
+  )
+  .settings(commonSettings)
+  .settings(
+    description := "Scio add-on for Neo4J",
```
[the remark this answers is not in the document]
turb: fixup 42a46745f75bdd8f55fd6f83a0bdac6253135dc4

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeOptions.scala
turb: TIL
fixup 25c5a76e3d4e203b7054ab4a45a92be6d901e940

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
turb: Beam `SnowflakeIO` does not give the path of the temp file, and it's just an intermediary file. So my impression is a tap make little sense here (as much as for `JdbcIo` and `Neo4JIO`).

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
turb: With `fold`: a9b56c4333a33dfad326a71441634ac23c37c0e9

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeOptions.scala
turb: fixup 9ede3f34fda4f1f2c35f387aee9dffc8a4bce3b7

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeOptions.scala
RustedBones: Updated the PR to support the params (esp on the write side)

RustedBones:
Hey @turb, can you test the latest change on a real Snowflake DB to make sure everything still works ?

turb:
> Hey @turb, can you test the latest change on a real Snowflake DB to make sure everything still works ?

Still works! (on read)

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
```diff
@@ -0,0 +1,222 @@
...
+  final def apply[T](opts: SnowflakeConnectionOptions, query: String): SnowflakeIO[T] =
+    new SnowflakeIO[T] with TestIO[T] {
+      final override val tapT = EmptyTapOf[T]
+      override def testId: String = s"SnowflakeIO(${snowflakeIoId(opts, query)})"
+    }
+
+  private[snowflake] def snowflakeIoId(opts: SnowflakeConnectionOptions, target: String): String = {
+    // source params
+    val params = Option(opts.database).map(db => s"db=$db") ++
+      Option(opts.warehouse).map(db => s"warehouse=$db")
```
clairemcginty, line 44: Comparing this to other complex `testId` implementations in our Scio IOs (example [1](https://github.com/spotify/scio/blob/v0.14.9/scio-google-cloud-platform/src/main/scala/com/spotify/scio/pubsub/PubsubIO.scala#L174), [2](https://github.com/spotify/scio/blob/v0.14.9/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala#L254)), I think a format like this would fit better:

```scala
SnowflakeIO(url, target, warehouse?, db?)
```

wdyt @RustedBones

RustedBones: Correct! we don't use named params. Will change that

PR #5515, opened 2024-10-18
Fix RemoteFileUtil to download in parallel as expected.

psobot:
The docs for `RemoteFileUtil::download` say:
> Download a batch of remote `URI`s in parallel.

...however, no parallel downloading happens. The concurrency arguments passed to `LoadingCache` control how many concurrent writes to the cache are allowed concurrently from callers on different threads, but calling [`LoadingCache::get`](https://guava.dev/releases/19.0/api/docs/com/google/common/cache/LoadingCache.html) triggers a load to happen _serially_.

This PR fixes this by creating a new `ExecutorService` to download the provided URIs in parallel within the `download` function. Single-file downloads (i.e.: `download(URI)` instead of `download(List<URI>)`) are not affected.

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
+++ b/scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
@@ -29,7 +29,12 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.List;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.stream.Collectors;
+
 import org.apache.beam.sdk.io.FileSystems;
 import org.apache.beam.sdk.io.fs.MatchResult.Metadata;
 import org.apache.beam.sdk.io.fs.ResourceId;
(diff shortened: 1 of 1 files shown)


scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -103,10 +107,37 @@ public Path download(URI src) {
    * @return {@link Path}s to the downloaded local files.
    */
   public List<Path> download(List<URI> srcs) {
+    return download(srcs, CONCURRENCY_LEVEL);
+  }
+
+  /**
+   * Download a batch of remote {@link URI}s in parallel, using at most numThreads to do so.
+   * `numThreads` may not be larger than the number of available processors * 4.
```
kellen, line 115:
```suggestion
   * Specifying a `numThreads` greater than number of available processors * 4 will have the same effect as specifying `numThreads` equal to the available processors * 4.
```

Or maybe 
```suggestion
   * `numThreads` should at maximum be equal to the number of available processors * 4.
```
and then actually clamp the size of the thread pool

kellen (approved):

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -103,10 +107,37 @@ public Path download(URI src) {
...
+   *
+   * @return {@link Path}s to the downloaded local files.
+   */
+  public List<Path> download(List<URI> srcs, int numThreads) {
+    final ExecutorService executor = Executors.newFixedThreadPool(numThreads);
```
RustedBones, line 120: We should probably move this as transient member, initializing it in the constructor.
The class can implement `AutoClosable` to shutdown the thread pool.

This is mainly relevant for the `FileDownloadDoFn` where we don't want to create a new thread pool for every url batch.

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -44,17 +50,16 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * A utility class for handling remote file systems designed to be used in a {@link
- * org.apache.beam.sdk.transforms.DoFn}.
- */
+/** A utility class for handling remote file systems designed to be used in a {@link DoFn}. */
 public class RemoteFileUtil implements Serializable {
```
kellen, line 54: @RustedBones Did you forget `AutoClosable` per your previous comment?

RustedBones:
Thanks @psobot !

PR #5519, opened 2024-10-24
Fix file batch operation

RustedBones:
RemoteFileUtil did not preserve element ordering wich was assumed in the FileDownloadDoFn

Pane information was not propagated

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
@@ -20,14 +20,15 @@ package com.spotify.scio.transforms;
 import com.spotify.scio.util.RemoteFileUtil;
 import java.net.URI;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.beam.sdk.transforms.display.DisplayData;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
+import org.apache.beam.sdk.transforms.windowing.PaneInfo;
+import org.apache.beam.sdk.values.ValueInSingleWindow;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
```diff
@@ -103,51 +123,31 @@ public void populateDisplayData(DisplayData.Builder builder) {
...
-      return;
-    }
-    LOG.info("Processing batch of {}", batch.size());
-    List<URI> uris = batch.stream().map(e -> e.uri).collect(Collectors.toList());
-    List<OutputT> outputs =
-        remoteFileUtil.download(uris).stream().map(fn::apply).collect(Collectors.toList());
-    // .forEach(c::output);
-    Iterator<OutputT> i1 = outputs.iterator();
-    Iterator<Element> i2 = batch.iterator();
-    while (i1.hasNext() && i2.hasNext()) {
```
RustedBones, line 131: This parts assumes that `List<Output>` is 1:1 mapping of  `List<URI>`

scio-core/src/main/java/com/spotify/scio/util/RemoteFileUtil.java
```diff
@@ -104,7 +106,12 @@ public Path download(URI src) {
    */
   public List<Path> download(List<URI> srcs) {
     try {
-      return paths.getAll(srcs).values().asList();
```
RustedBones, line 107: `getAll(srcs)` returns a `Map<URI, Path>`. Using `values` give no guarantee that returned `Paths` are ordered.

scio-core/src/main/java/com/spotify/scio/transforms/FileDownloadDoFn.java
```diff
@@ -103,51 +123,31 @@ public void populateDisplayData(DisplayData.Builder builder) {
...
   }
 
-  private void processBatch(OutputReceiver<OutputT> outputReceiver) {
+  private void flush(Consumer<ValueInSingleWindow<OutputT>> outputFn) {
     if (batch.isEmpty()) {
       return;
     }
     LOG.info("Processing batch of {}", batch.size());
-    List<URI> uris = batch.stream().map(e -> e.uri).collect(Collectors.toList());
-    remoteFileUtil.download(uris).stream().map(fn::apply).forEach(outputReceiver::output);
```
RustedBones, line 112: original element metadata are not passed back

scio-core/src/test/scala/com/spotify/scio/transforms/FileDownloadDoFnTest.scala
```diff
@@ -42,9 +47,23 @@ class FileDownloadDoFnTest extends PipelineSpec {
...
+        .timestampBy { case (_, i) => new Instant(i + 1) }
+        .keys
+        .flatMapFile(fn, 10, false)
+        .withTimestamp
+
+      val contentAndTimestamp = p.map { case ((i, _), ts) => (i, ts.getMillis) }
+      val paths =  p.map { case ((_, f), _) => f }.distinct
+
+      val expected = (1L to 100L).map(i => (i.toString, i))
+      contentAndTimestamp should containInAnyOrder(expected)
```
RustedBones, line 65: Ran on the original code, got non expected item `(1, 10)` -> (1st element was flushed while processing the 10th)

kellen (approved):
Nice catch!

PR #5520, opened 2024-10-26
Add java priority queue, set, deque, collection coders

kellen:
When using e.g. algebird `PriorityQueueMonoid`, scio needs to serialize java PriorityQueues. If Kryo is used, especially for small queues, there can be significant overhead.

Instead, this PR provides a PriorityQueue coder backed by a scala `Ordering` (as indeed the monoid is).

The coder must be created explicitly by the user so that they are asserting that the comparator of the original PriorityQueues and the reconstituted ones are the same. e.g.

```
implicit val pqCoder = Coders.jPriorityQueueCoder[T](ord)
```

@RustedBones Is there some trick I can use to get an error message when `Coder[java.util.PriorityQueue[T]]` fails to be derived (that's not just the generic implicitNotFound on `Coder`)?


scio-core/src/main/scala/com/spotify/scio/coders/instances/JavaCoders.scala
```diff
@@ -69,6 +68,29 @@ trait JavaCoders extends CoderGrammar with JavaBeanCoders {
...
+   * coder construction time.
+   *
+   * @param ord
+   *   Ordering used as PriorityQueue comparator
+   */
+  def jPriorityQueueCoder[T: Coder: ClassTag](
+    ord: Ordering[T]
+  ): Coder[java.util.PriorityQueue[T]] = {
+    // neither arrays nor PriorityQueues are consistentWithEquals
+    Coder.xmap(ScalaCoders.arrayCoder[T])(
```
RustedBones, line 84: I think it would be better to create a coder instance, explicitly overriding `consistentWithEquals` instead of depending on a specific type for that reason

RustedBones (commented):
I see we're also missing coders for `java.util.Collection[T]`, `java.util.Set[T]` and `java.util.Deque[T]` which are available by default in beam

RustedBones:
I recall we've been comparing underlying coder and value ordering [here](https://github.com/spotify/scio/blob/489bd7a34b95b140ed53484c4fab0cb28bad69dc/scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala#L366), but we've not done that for scala `SortedSet` nor `PriorityQueue`.

The drawback of such check is that we must give an ordering implementation that gives a stable equal after serialization (mostly by using object ordering).

scio-core/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -344,6 +364,23 @@ final class CoderTest extends AnyFlatSpec with Matchers {
...
+    implicit val pqEq: Equality[java.util.PriorityQueue[String]] = {
+      case (a: jPriorityQueue[String], b: jPriorityQueue[_]) =>
+        ArraySeq.unsafeWrapArray(a.toArray) == ArraySeq.unsafeWrapArray(b.toArray)
+      case _ => false
+    }
+
+    pq coderShould roundtrip() and
+      beOfType[Transform[_, _]] and
+      materializeToTransformOf[ArrayCoder[_]] and
+      beFullyCompliantNotConsistentWithEquals()
```
RustedBones, line 383: shouldn't this be non deterministic too ?

scio-core/src/test/scala/com/spotify/scio/coders/CoderTest.scala
```diff
@@ -154,6 +156,23 @@ final class CoderTest extends AnyFlatSpec with Matchers {
       beOfType[CoderTransform[_, _]] and
       materializeTo[ArrayCoder[_]] and
       beFullyCompliantNotConsistentWithEquals()
+
+    {
+      // custom ordering must have stable equal after serialization
+      implicit val pqOrd: Ordering[String] = FlippedStringOrdering
```
RustedBones, line 162: Not supper happy about this

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -244,15 +244,26 @@ private[coders] class MutableSetCoder[T](bc: BCoder[T]) extends SeqLikeCoder[m.S
 }
 
 private class SortedSetCoder[T: Ordering](bc: BCoder[T]) extends SeqLikeCoder[SortedSet, T](bc) {
+
+  override def encode(value: SortedSet[T], os: OutputStream): Unit =
+    super.encode(value, os)
```
clairemcginty, line 249: why is this needed?

clairemcginty (approved):

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
kellen: Removed

scio-core/src/main/scala/com/spotify/scio/coders/instances/ScalaCoders.scala
```diff
@@ -259,6 +259,11 @@ private[coders] class MutablePriorityQueueCoder[T: Ordering](bc: BCoder[T])
     super.encode(value, os)
   override def decode(inStream: InputStream): m.PriorityQueue[T] =
     decode(inStream, m.PriorityQueue.newBuilder[T])
+  override def verifyDeterministic(): Unit =
+    throw new NonDeterministicException(
+      this,
+      "Ordering of elements in a priority queue may be non-deterministic."
+    )
```
RustedBones, line 260: Good catch. I didn't check that `PriorityQueue` iterator is not ordered
> Returns an iterator over the elements in this queue. The iterator does not return the elements in any particular order.

PR #5444, opened 2024-08-13
Refactor bigtable API to use v2 client

RustedBones:
Update all bigtable APIs to use cloud v2 client.

Drop `BigtableBulkWriter`. This is TBD, but it looks to me can manually shard/groupBy and flatten to normal IO if they require this feature. 

Change bigtable `ScioContext` api to use `Admin.Table` and `Admin.Instance` objects that cache clients instances.

PR #5521, opened 2024-11-04
Add Batched lookups for streaming GRPC endpoints and BigTable

lofifnc:
This aims to extend the usability for AsyncBatchLookupDoFn it adds:
* `grpcLookupBatchStream` for batched GRPC endpoints with streaming response.
* `BigTableBatchDoFn` for batch calls to BigTable including a cache.
To make this happen the AsyncBatchLookupDoFn had to be changed to gracefully handle partial batched responses.

### DIFF
--- a/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBatchDoFn.java
+++ b/scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBatchDoFn.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2024 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigtable;
+
+import com.google.cloud.bigtable.config.BigtableOptions;
+import com.google.cloud.bigtable.grpc.BigtableSession;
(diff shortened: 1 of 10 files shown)


RustedBones:
> `grpcSafeBatchLookup` for calling endpoints which don't include missing values into the batch response.

I think we can keep the same API for that case. When receiving a batch response missing some entries form the batch request, we should fail fast and report those. It is then up to the user to filter out those exceptions if this is an expected behavior.  See https://github.com/spotify/scio/pull/5532

lofifnc:
> > `grpcSafeBatchLookup` for calling endpoints which don't include missing values into the batch response.
> 
> I think we can keep the same API for that case. When receiving a batch response missing some entries form the batch request, we should fail fast and report those. It is then up to the user to filter out those exceptions if this is an expected behavior. See #5532

The rule of thump I tried to apply here is fail, if the client you're using is throwing an exception. BigTable not having an entry for a key is kind of something expected. 
If we just return a failure I know, that I would make the mistake not adding handling for that and then on the first job run discovering that I'm getting a bunch of exceptions.

lofifnc:
@RustedBones I've gone ahead and rebased this brach based on your PR #5532 and dropped the `.safeGrpcLookup` for now. IMO this PR looks much more contained now.

scio-google-cloud-platform/src/main/java/com/spotify/scio/bigtable/BigtableBatchDoFn.java
```diff
@@ -0,0 +1,126 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.scio.bigtable;
+
+import com.google.cloud.bigtable.config.BigtableOptions;
+import com.google.cloud.bigtable.grpc.BigtableSession;
```
RustedBones, line 21: As a heads-up, we will probably move to the `BigtableDataClient` in some next minor release, see [here](https://github.com/spotify/scio/pull/5444/). Expect some breaking changes in that part.

RustedBones (approved):
LGTM. Thanks for the contribution. It makes a lot of sense to handle missing response in batch in a lenient way instead of failing the job.

RustedBones:
Thanks!

PR #5523, opened 2024-11-14
BigQuery JSON column: encode as Jackson JsonNode

turb:
On write, BigQuery JSON columns must not be provided as `String`, as it will be escaped. Apache Beam warns:

`Make sure the TableRow value is a parsed JSON to ensure the read as a JSON type. Otherwise it will read as a raw (escaped) string.`

Reverse-eng. shows that is means "as a `JsonNode` from Jackson".

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
@@ -17,7 +17,6 @@
 
 package com.spotify.scio.bigquery.types
 
-import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
 import com.google.api.services.bigquery.model.TableRow
 import com.google.protobuf.ByteString
 import com.spotify.scio.bigquery.types.MacroUtil._
@@ -413,8 +412,8 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Geography] =>
           q"$tree.wkt"
         case t if t =:= typeOf[Json] =>
-          // for BigQuery, we need to provide parsed JSON to prevent escaping
-          q"_root_.com.spotify.scio.bigquery.types.ConverterUtil.readJsonTree($tree)"
+          // for TableRow/json, we need to provide parsed JSON to prevent escaping
+          q"_root_.com.spotify.scio.bigquery.types.Json.parse($tree)"
         case t if t =:= typeOf[BigNumeric] =>
           q"$tree.wkt"
 
@@ -476,9 +475,6 @@ private[types] object ConverterProvider {
 }
 
 object ConverterUtil {
-  private val mapper = new ObjectMapper()
-  def readJsonTree(json: Json): JsonNode = mapper.readTree(json.wkt)
-
   @inline final def notNull[@specialized(Boolean, Int, Long, Float, Double) T](x: T): Boolean =
     x != null
 }
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
@@ -171,13 +171,14 @@ private[types] object ConverterProvider {
       val provider: OverrideTypeProvider =
         OverrideTypeProviderFinder.getProvider
       tpe match {
-        case t if provider.shouldOverrideType(c)(t) => q"$tree.toString"
-        case t if t =:= typeOf[Boolean]             => tree
-        case t if t =:= typeOf[Int]                 => q"$tree.toLong"
-        case t if t =:= typeOf[Long]                => tree
-        case t if t =:= typeOf[Float]               => q"$tree.toDouble"
-        case t if t =:= typeOf[Double]              => tree
-        case t if t =:= typeOf[String]              => tree
+        case t if provider.shouldOverrideType(c)(t)                     => q"$tree.toString"
+        case t if t =:= typeOf[Boolean]                                 => tree
+        case t if t =:= typeOf[Int]                                     => q"$tree.toLong"
+        case t if t =:= typeOf[Long]                                    => tree
+        case t if t =:= typeOf[Float]                                   => q"$tree.toDouble"
+        case t if t =:= typeOf[Double]                                  => tree
+        case t if t =:= typeOf[String]                                  => tree
+        case t if t =:= typeOf[com.fasterxml.jackson.databind.JsonNode] => tree
 
         case t if t =:= typeOf[BigDecimal] =>
           q"_root_.com.spotify.scio.bigquery.Numeric($tree).toString"
@@ -198,7 +199,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Geography] =>
           q"$tree.wkt"
         case t if t =:= typeOf[Json] =>
-          q"$tree.wkt"
+          q"$tree.asJackson"
         case t if t =:= typeOf[BigNumeric] =>
           q"_root_.com.spotify.scio.bigquery.types.BigNumeric($tree.wkt).toString"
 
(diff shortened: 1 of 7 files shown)


turb:
@RustedBones what remains here is binary compatibility: newly compiled `Json` fails if executed with previous lib.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -171,13 +171,14 @@ private[types] object ConverterProvider {
...
-        case t if t =:= typeOf[Double]              => tree
-        case t if t =:= typeOf[String]              => tree
+        case t if provider.shouldOverrideType(c)(t)                     => q"$tree.toString"
+        case t if t =:= typeOf[Boolean]                                 => tree
+        case t if t =:= typeOf[Int]                                     => q"$tree.toLong"
+        case t if t =:= typeOf[Long]                                    => tree
+        case t if t =:= typeOf[Float]                                   => q"$tree.toDouble"
+        case t if t =:= typeOf[Double]                                  => tree
+        case t if t =:= typeOf[String]                                  => tree
+        case t if t =:= typeOf[com.fasterxml.jackson.databind.JsonNode] => tree
```
RustedBones, line 181: I don't think we should accept `JsonNode` here

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -198,7 +199,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Geography] =>
           q"$tree.wkt"
         case t if t =:= typeOf[Json] =>
-          q"$tree.wkt"
+          q"$tree.asJackson"
```
RustedBones, line 202: According to the doc [here](https://github.com/apache/beam/blob/0cafb4644e31ada2929ab897297cc78f76fe006c/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L3755), loading json in avro expects a string type, annotated with the `JSON` sqlType property

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -276,12 +277,13 @@ private[types] object ConverterProvider {
...
-        case t if t =:= typeOf[Float]   => q"$s.toFloat"
-        case t if t =:= typeOf[Double]  => q"$s.toDouble"
-        case t if t =:= typeOf[String]  => q"$s"
+        case t if t =:= typeOf[Boolean]                                 => q"$s.toBoolean"
+        case t if t =:= typeOf[Int]                                     => q"$s.toInt"
+        case t if t =:= typeOf[Long]                                    => q"$s.toLong"
+        case t if t =:= typeOf[Float]                                   => q"$s.toFloat"
+        case t if t =:= typeOf[Double]                                  => q"$s.toDouble"
+        case t if t =:= typeOf[String]                                  => q"$s"
+        case t if t =:= typeOf[com.fasterxml.jackson.databind.JsonNode] => q"$s"
```
RustedBones, line 286: I don't think we should accept `JsonNode` here

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -412,7 +414,7 @@ private[types] object ConverterProvider {
         case t if t =:= typeOf[Geography] =>
           q"$tree.wkt"
         case t if t =:= typeOf[Json] =>
-          q"$tree.wkt"
+          q"$tree.asJackson"
```
RustedBones, line 417: :+1:

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
```diff
@@ -53,14 +54,20 @@ package object types {
...
    * @param wkt
    *   Well Known Text formatted string that BigQuery displays for Json
    */
-  case class Json(wkt: String)
+  case class Json(wkt: String) {
+    def asJackson: JsonNode = Json.mapper.readTree(wkt)
+  }
+  object Json {
+    private val mapper = new ObjectMapper()
+  }
```
RustedBones, line 70: I'd prefer to have this in the conversion part and leave the data as dumb as possible

scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigquery/types/ConverterProviderSpec.scala
```diff
@@ -51,6 +51,12 @@ final class ConverterProviderSpec
       .retryUntil(_.precision <= Numeric.MaxNumericPrecision)
       .map(Numeric.apply)
   }
+  implicit val arbJson: Arbitrary[Json] = Arbitrary(
+    for {
+      key <- Gen.alphaStr
+      value <- Gen.alphaStr
+    } yield Json("{\"" + key + "\":\"" + value + "\"}")
```
RustedBones, line 58: Nit: triple quote is nicer when writing json
```suggestion
    } yield Json(s"""{"$key":"$value"}""")
```

RustedBones (commented):
We need an integration test for that to make sure both avro and json formats are ok with those types

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
turb: There is the same comment for BigQuery, but after a reverse-eng. they meant JsonNode. If opened a PR just to change the warning: https://github.com/apache/beam/pull/33121

However I can keep `asJackson` on line 417, and `.wkt` on line 202.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
turb: I don't actually need it, but I supposed one may want to provide an already-valid JSON. I will remove it.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
turb: Where would be the appropriate place to store the `ObjectMapper` instance, in order to not instantiate it for each row? Directly as a `private val` in `ConverterProvider`?

scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigquery/types/ConverterProviderSpec.scala
turb: TIL. I always supposed triple quote was only for multi-line.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
turb: c0303bbbe64b81d29d76659a10b205332631c67b

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
turb: c0303bbbe64b81d29d76659a10b205332631c67b

RustedBones:
Wrote some integration tests on top of your PR.  Revealed that `BigNumeric`  also has a bug!
Needs some polishing, will add that on top of this work if it's fine. 

Thanks again for the contrib, and nice catch!

RustedBones:
This change causes a regression when using the storage API. Seems the Jackson node is not well supported.

turb:
> This change causes a regression when using the storage API. Seems the Jackson node is not well supported.

That's strange, I'm reading entities with JSON columns using `sc.typedBigQueryStorage` and it works.

RustedBones:
The issue is with the writing side.
It also looks that using `TableRow` instead of `JacskonNode` is better: The coder for `TableRow` transforms `JacskonNode` into a `TableRow` when an element gets serialized.

I'd try to fix those problems in #5529

turb:
FWIW, I tried a table load with `method = beam.BigQueryIO.Write.Method.STORAGE_WRITE_API` (is that it?) and it worked.

So `JacksonNode` would replace `Json` or that's just an internal?

The one thing to check if the column is JSON, is it should *not* be stored escaped inside a JS string. AFAIK that can be observed only BigQuery.

PR #5525, opened 2024-11-18
Update beam to 2.61

RustedBones:
update beam to 2.61.0


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -385,7 +385,7 @@ object BigQueryTypedTable {
       beam.BigQueryIO.readTableRows(),
       beam.BigQueryIO.writeTableRows(),
       table,
-      BigQueryUtils.convertGenericRecordToTableRow(_, _)
+      (r, _) => BigQueryUtils.convertGenericRecordToTableRow(r)
```
RustedBones, line 388: Ideally we should change the APIs to remove the unused `TableRow` parameter. This is prob smth to do on next major upgrade

shnapz (approved):

PR #5529, opened 2024-11-27
Switch BQ typed underlying API to avro

RustedBones:
Leverages upstream [changes](https://github.com/apache/beam/pull/32482) available for BQ:
Annotated BQ typed avro translation leverages logical-type to have symmetric read/write. This fixes integration testfailure introduced in https://github.com/spotify/scio/pull/5523


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -180,17 +180,18 @@ private[types] object ConverterProvider {
...
 
-        case t if t =:= typeOf[Instant] => q"$tree.getMillis * 1000"
+        case t if t =:= typeOf[Instant] =>
+          q"_root_.com.spotify.scio.bigquery.Timestamp.micros($tree)"
         case t if t =:= typeOf[LocalDate] =>
-          q"_root_.com.spotify.scio.bigquery.Date($tree)"
+          q"_root_.com.spotify.scio.bigquery.Date.days($tree)"
         case t if t =:= typeOf[LocalTime] =>
-          q"_root_.com.spotify.scio.bigquery.Time($tree)"
+          q"_root_.com.spotify.scio.bigquery.Time.micros($tree)"
```
shnapz, line 194: so it was returning String before?

RustedBones: Indeed, by changing [this](https://github.com/spotify/scio/pull/5529/files#diff-39ac37a4ec2c09860762dae1c3467ca60ce4418840615f5fc0dac3426bffbd45R442), I created a behavior change.

When extracting from avro, we disabled the logical types -> those types are returned as STRING (mentioned [here](https://github.com/spotify/scio/blob/cc1e584206e70b3e88dffcad9ea7b6f4a0d7e8b7/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/ScioContextSyntax.scala#L81)). This is very annoying because the same record can't be written back in the table.

Setting `useAvroLogicalTypes` on the IO enables that. This will however be breaking for all pipelines reading BQ `DATE`, `TIME` and `DATETIME` that expects `string` and will get logical type.

This is also not in sync with `storage` API that uses logical types for such columns.

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/SCollectionSyntax.scala
```diff
@@ -203,7 +203,7 @@ final class SCollectionGenericRecordOps[T <: GenericRecord](private val self: SC
     self
       .covary[GenericRecord]
       .write(
-        BigQueryTypedTable(table, Format.GenericRecord)(
+        BigQueryTypedTable(table, Format.GenericRecordWithLogicalTypes)(
```
RustedBones, line 206: On write, logical types were enabled by default

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
RustedBones: Updated the PR to keep the same behavior.
Introduced new `Format.GenericRecordWithLogicalType` to enable BQ avro read with desired setup.
Change `BigQueryTyped.Table` to always use logical types

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/TypeProvider.scala
```diff
@@ -334,7 +334,7 @@ private[types] object TypeProvider {
           q"override def schema: ${p(c, GModel)}.TableSchema = ${p(c, SUtil)}.parseSchema(${schema.toString})"
         }
         val defAvroSchema =
-          q"override def avroSchema: org.apache.avro.Schema = ${p(c, BigQueryUtils)}.toGenericAvroSchema(${cName.toString}, this.schema.getFields)"
+          q"override def avroSchema: org.apache.avro.Schema = ${p(c, BigQueryUtils)}.toGenericAvroSchema(this.schema, true)"
```
RustedBones, line 337: Use logical types by default for typed BQ avro

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryIO.scala
```diff
@@ -738,12 +745,31 @@ object BigQueryTyped {
...
+    private val underlying = {
+      val readFn = Functions.serializableFn[SchemaAndRecord, T] { x =>
+        BigQueryType[T].fromAvro(x.getRecord)
+      }
+      val writeFn = Functions.serializableFn[AvroWriteRequest[T], GenericRecord] { x =>
+        BigQueryType[T].toAvro(x.getElement)
+      }
+      val schemaFactory = Functions.serializableFn[TableSchema, org.apache.avro.Schema] { _ =>
+        BigQueryType[T].avroSchema
+      }
```
RustedBones, line 757: use strict schema from the annotated class instead of relying on the table conversion

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/SCollectionSyntax.scala
kellen: This is aligned with the BQ storage API as alluded to in another comment?

kellen (approved):

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/BigQueryTypes.scala
```diff
@@ -250,6 +257,9 @@ object DateTime {
   /** Convert BigQuery `DATETIME` string to `LocalDateTime`. */
   def parse(datetime: String): LocalDateTime =
     Parser.parseLocalDateTime(datetime)
+
+  //  For BigQueryType macros only, do not use directly
+  def format(datetime: LocalDateTime): String = apply(datetime)
```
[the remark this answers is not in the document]
RustedBones: unfortunatelly no. This is used by code generated in macros, meaning it can be referenced from any package.

PR #5532, opened 2024-11-28
Report exception for unmatched batch request

RustedBones:
When getting an response for a batch request, return an UnmatchedRequestException for unmatched requests


RustedBones:
Integrated in #5521

PR #5534, opened 2024-12-11
Allow timeout for async DoFn

### LINKED ISSUE #4040: Support Future timeouts in AsyncLookupDoFn?
currently AsyncLookupDoFn will [block indefinitely](https://github.com/spotify/scio/blob/main/scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java#L38) waiting for Futures to complete in the `@FinishBundle` step. Should we support a user-specified timeout (`long timeout, TimeUnit timeUnit`) after which the Future will be interrupted and return as Failed? the current impl can be problematic if one of the futures in a bundle starts hanging (this has been seen with grpc calls, even with a deadline set).

The downside is that it would potentially be a lot of work to implement; both exposing it to the user as an option (with a default of indefinite timeout) and implementing it for all three Future types (guava, vanilla java, and vanilla scala). For Guava I believe you need to pass in a `ScheduledExecutorService` so that's another resource to manage within the `DoFn`, probably at the per-class or per-instance resource level.

any thoughts? I can see an argument being made that this belongs in user-code and should be handled by their lookup functions. It's just tricky/error-prone especially when ExecutorServices/wrapped futures get involved.

RustedBones:
Fix #4040

Should we set a sane timeout default ?

### DIFF
--- a/scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
@@ -42,7 +42,7 @@ public class FutureHandlers {
   public interface Base<F, V> {
 
     default Duration getTimeout() {
-      return null;
+      return Duration.ofMinutes(1);
     }
 
     void waitForFutures(Iterable<F> futures)
--- a/scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
+++ b/scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
@@ -18,7 +18,14 @@
 package com.spotify.scio.transforms;
 
 import com.google.common.util.concurrent.*;
-import java.util.concurrent.*;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 import java.util.stream.StreamSupport;
 import javax.annotation.Nullable;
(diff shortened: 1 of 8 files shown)


scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigtable/BigtableDoFnTest.scala
```diff
@@ -66,21 +68,26 @@ object BigtableDoFnTest {
   val queue: ConcurrentLinkedQueue[Int] = new ConcurrentLinkedQueue[Int]()
 }
 
-class TestBigtableDoFn extends BigtableDoFn[Int, String](null) {
+class TestBigtableDoFn extends BigtableDoFn[Int, String](BigtableOptions.getDefaultOptions) {
```
kellen, line 71: Did you mean to have these BT changes here?

scio-core/src/main/java/com/spotify/scio/transforms/FutureHandlers.java
```diff
@@ -53,10 +66,16 @@ default Executor getCallbackExecutor() {
...
     default void waitForFutures(Iterable<ListenableFuture<V>> futures)
-        throws InterruptedException, ExecutionException {
+        throws InterruptedException, ExecutionException, TimeoutException {
       // use Future#successfulAsList instead of Futures#allAsList which only works if all
       // futures succeed
-      Futures.successfulAsList(futures).get();
+      ListenableFuture<?> f = Futures.successfulAsList(futures);
+      Duration timeout = getTimeout();
+      if (timeout != null) {
+        f.get(timeout.toMillis(), TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
RustedBones: Yes!

scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigtable/BigtableDoFnTest.scala
RustedBones: Yes, in BT, the timeout is taken from the settings now, we can't pass null.

kellen (approved):
🎉

PR #5536, opened 2024-12-19
Stable TableRow converted from BQ types

RustedBones:
Coder[TableRow] is destructive (it is a dummy JSON serializer), we should make sure that the TableRow object converted from a BQ model is stable after serialization.

We currently have an issue with
- long that are serialized as string to avoid overflow
- float that are read back as double
- json that is read as nested TableRow

As side effect, I needed to avoid `toString` conversion when converting back a BQ typed model from a `TableRow`

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
@@ -270,7 +270,7 @@ private[types] object ConverterProvider {
     // Converter helpers
     // =======================================================================
     def cast(tree: Tree, tpe: Type): Tree = {
-      val msg = Constant(s"Cannot convert to ${tpe.typeSymbol.name}: ")
+      val msg = s"Cannot convert to ${tpe.typeSymbol.name}: "
       val fail = q"""throw new _root_.java.lang.IllegalArgumentException($msg + $tree)"""
 
       def readBase64(term: TermName) =
@@ -418,7 +418,7 @@ private[types] object ConverterProvider {
 
       val tree = q"$fn.get($name)"
       def nonNullTree(fType: String) = {
-        val msg = Constant(s"$fType field '$name' is null")
+        val msg = s"""$fType field "$name" is null"""
         q"""{
           val v = $fn.get($name)
           if (v == null) {
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
@@ -270,50 +270,136 @@ private[types] object ConverterProvider {
     // Converter helpers
     // =======================================================================
     def cast(tree: Tree, tpe: Type): Tree = {
+      val msg = Constant(s"Cannot convert to ${tpe.typeSymbol.name}: ")
+      val fail = q"""throw new _root_.java.lang.IllegalArgumentException($msg + $tree)"""
+
+      def readBase64(term: TermName) =
+        q"_root_.com.google.common.io.BaseEncoding.base64().decode($term)"
+
       val provider: OverrideTypeProvider =
         OverrideTypeProviderFinder.getProvider
-      val s = q"$tree.toString"
       tpe match {
         case t if provider.shouldOverrideType(c)(t) =>
           provider.createInstance(c)(t, q"$tree")
-        case t if t =:= typeOf[Boolean] => q"$s.toBoolean"
-        case t if t =:= typeOf[Int]     => q"$s.toInt"
-        case t if t =:= typeOf[Long]    => q"$s.toLong"
-        case t if t =:= typeOf[Float]   => q"$s.toFloat"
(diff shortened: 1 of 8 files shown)


scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala
```diff
@@ -270,58 +270,144 @@ private[types] object ConverterProvider {
...
     def cast(tree: Tree, tpe: Type): Tree = {
+      val msg = Constant(s"Cannot convert to ${tpe.typeSymbol.name}: ")
+      val fail = q"""throw new _root_.java.lang.IllegalArgumentException($msg + $tree)"""
+
+      def readBase64(term: TermName) =
+        s"_root_.com.google.common.io.BaseEncoding.base64().decode($term)"
+
       val provider: OverrideTypeProvider =
         OverrideTypeProviderFinder.getProvider
-      val s = q"$tree.toString"
```
RustedBones, line 275: we were forcing `toString` before converting back to desired type

scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigquery/types/ConverterProviderSpec.scala
```diff
@@ -53,7 +53,9 @@ final class ConverterProviderSpec
   }
   implicit val arbJson: Arbitrary[Json] = Arbitrary(
     for {
-      key <- Gen.alphaStr
+      // f is a field from TableRow.
+      // Jackson ObjectMapper will fail with such key
+      key <- Gen.alphaStr.retryUntil(_ != "f")
```
RustedBones, line 58: I'm wondering what happens if a BQ table a a field named `f`. It's probable that we can't use the `TableRow` API

RustedBones (commented):
Ideally all the `TableRow -> scala` cast logic should be shared with [TableRowOps](https://github.com/spotify/scio/blob/532e91504bb38577c8062300b87110b12220c49e/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/syntax/TableRowSyntax.scala#L29C13-L29C24)

scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigquery/types/ConverterProviderSpec.scala
RustedBones: Did a quick test. Reading a column named 'f' fails to create a `TableRow` item [here](https://github.com/apache/beam/blob/40151abbfdd3b1deb6a7b3de395562de91b818a6/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryAvroUtils.java#L277).

See https://github.com/apache/beam/issues/33531

scio-google-cloud-platform/src/test/scala/com/spotify/scio/bigquery/TableRowSyntaxTest.scala
```diff
@@ -17,21 +17,379 @@
...
 
 class TableRowSyntaxTest extends AnyFlatSpec with Matchers {
-  "TableRowOps" should "#2673: not throw on get record" in {
-    val json = """{"record":{"foo":"bar"}}"""
-    val row = com.spotify.scio.util.ScioUtil.jsonFactory.fromString(json, classOf[TableRow])
-    val expected = Map("foo" -> "bar")
-    row.getRecord("record") shouldBe expected
-  }
-
-  it should "#3378: not throw an NPE on a non-existent subrecord" in {
```
RustedBones, line 31: This is not very logical.
Added a `getRecordOpt` to be able to get `Option[TableRow]` instead

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
```diff
@@ -63,10 +65,15 @@ package object types {
    */
   case class Json(wkt: String)
   object Json {
+    // Use same mapper as the TableRowJsonCoder
     private lazy val mapper = new ObjectMapper()
+      .registerModule(new JavaTimeModule())
+      .registerModule(new JodaModule())
+      .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
```
clairemcginty, line 72: what's this needed for?

clairemcginty (approved):

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
RustedBones: this is the config from [TableRowJsonCoder](https://github.com/apache/beam/blob/d7c56914f80c6d0b817a3a69b58de0742ef8d235/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/TableRowJsonCoder.java#L80).

It forces serialization of dates as string
> Feature that determines whether Date (and date/ time) values (and Date-based things like java. util. Calendars) are to be serialized as numeric time stamps (true; the default), or as something else (usually textual representation).

PR #5537, opened 2024-12-19
Update default Dataflow add-open opts for Java 17+

clairemcginty:
Match all the add-opens required by Kryo

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowContext.scala
@@ -41,16 +41,22 @@ case object DataflowContext extends RunnerContext {
       .filesToStage(options, classLoader, localArtifacts, artifacts)
       .asJavaCollection
 
-    // Required for Kryo w/ Java 17
+    // Required for Kryo w/ Java 17+
     lazy val dataflowPipelineOpts = options.as(classOf[DataflowPipelineOptions])
-    if (
-      sys
-        .props("java.version")
-        .startsWith("17.") && dataflowPipelineOpts.getJdkAddOpenModules == null
-    ) {
-      dataflowPipelineOpts.setJdkAddOpenModules(
-        List("java.base/java.util=ALL-UNNAMED", "java.base/java.lang.invoke=ALL-UNNAMED").asJava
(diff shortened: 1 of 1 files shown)


scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowContext.scala
```diff
@@ -41,16 +41,22 @@ case object DataflowContext extends RunnerContext {
...
-        .startsWith("17.") && dataflowPipelineOpts.getJdkAddOpenModules == null
-    ) {
-      dataflowPipelineOpts.setJdkAddOpenModules(
-        List("java.base/java.util=ALL-UNNAMED", "java.base/java.lang.invoke=ALL-UNNAMED").asJava
-      )
+    val javaVersionRe = "^(\\d+)\\..*".r
+
+    sys.props("java.version") match {
+      case javaVersionRe(version)
+          if version.toInt >= 17 && dataflowPipelineOpts.getJdkAddOpenModules == null =>
```
clairemcginty, line 50: I suppose this is also a bugfix PR for Java 21 users 😬

RustedBones (approved):

scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowContext.scala
```diff
@@ -41,16 +41,22 @@ case object DataflowContext extends RunnerContext {
...
     lazy val dataflowPipelineOpts = options.as(classOf[DataflowPipelineOptions])
-    if (
-      sys
-        .props("java.version")
-        .startsWith("17.") && dataflowPipelineOpts.getJdkAddOpenModules == null
-    ) {
-      dataflowPipelineOpts.setJdkAddOpenModules(
-        List("java.base/java.util=ALL-UNNAMED", "java.base/java.lang.invoke=ALL-UNNAMED").asJava
-      )
+    val javaVersionRe = "^(\\d+)\\..*".r
```
RustedBones, line 46: This regex would give `1` for java 8 (version is set as `1.8....`).
Let's use the same def as for the build
```
System.getProperty("java.version").stripPrefix("1.").takeWhile(_.isDigit).toInt
```

PR #5539, opened 2025-01-13
Prepare for Scio 0.14.10

clairemcginty:


site/src/main/paradox/releases/Apache-Beam.md
```diff
@@ -33,6 +33,7 @@ Also check out the [SDK Version Support Status](https://cloud.google.com/dataflo
 
 | **Scio Version** | **Beam Version** | **Details**                                           |
 |:----------------:|:----------------:|:------------------------------------------------------|
+|      0.14.10     |      2.61.0      | This version will be deprecated on November 25, 2025. |
```
RustedBones, line 36: `0.13.6` and `0.14.0` could have been marked as deprecated.
We'll update that on next release.

PR #5544, opened 2025-01-16
Support BQ Json arrays and literals

### LINKED ISSUE #5542: BigQuery load fails with JSON array
When loading data to BigQuery with a JSON column, it fails when the actual data is a JS array `[ data1, data2 ]`:

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `com.google.api.services.bigquery.model.TableRow` from Array value (token `JsonToken.START_ARRAY`)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1]
	com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
	com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1767)
	com.fasterxml.jackson.databind.DeserializationContext.handleUnexpectedToken(DeserializationContext.java:1541)
	com.fasterxml.jackson.databind.deser.std.StdDeserializer._deserializeFromArray(StdDeserializer.java:222)
	com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:457)
	com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:32)
	com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342)
	com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4917)
	com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3860)
	com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3828)
	com.spotify.scio.bigquery.types.package$Json$.parse(package.scala:76)
	@BigQueryType.toTable <== line where there is the toTable annotation
```

It did work with `JacksonNode`, but I wonder if it is possible with `TableRow`.

RustedBones:
Json should simply be mapped to java `List` or `Map`

Fix #5542

### DIFF
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
@@ -27,7 +27,7 @@ import org.typelevel.scalaccompat.annotation.nowarn
 
 import java.math.MathContext
 import java.nio.ByteBuffer
-import scala.annotation.StaticAnnotation
+import scala.annotation.{unused, StaticAnnotation}
 import scala.util.Try
 
 package object types {
@@ -77,12 +77,15 @@ package object types {
       .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
       .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS);
 
+    // force to use the apply(AnyRef)
+    @unused private def apply(wkt: String): Json = new Json(wkt)
+
     def apply(value: AnyRef): Json = value match {
       case str: String if Try(mapper.readTree(str)).isSuccess =>
         // string formatted json vs string literal
-        Json(str)
+        new Json(str)
       case _ =>
-        Json(mapper.writeValueAsString(value))
+        new Json(mapper.writeValueAsString(value))
     }
     def parse(json: Json): AnyRef = mapper.readValue(json.wkt, classOf[Object])
   }
--- a/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
+++ b/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
@@ -72,8 +72,17 @@ package object types {
       .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
       .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS);
 
-    def apply(row: TableRow): Json = Json(mapper.writeValueAsString(row))
-    def parse(json: Json): TableRow = mapper.readValue(json.wkt, classOf[TableRow])
+    def apply(value: AnyRef): Json = Json(mapper.writeValueAsString(value))
+    def parse(json: Json): AnyRef = {
+      val node = mapper.readTree(json.wkt)
+      if (node.isObject) {
+        mapper.treeToValue(node, classOf[java.util.Map[_, _]])
+      } else if (node.isArray) {
+        mapper.treeToValue(node, classOf[java.util.List[_]])
+      } else {
+        throw new IllegalArgumentException(s"Invalid json ${json.wkt}")
+      }
+    }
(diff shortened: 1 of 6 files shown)


RustedBones:
parsed json fields are not stable through serialization. I feel we'll have to fallback to handling them as parsed string as before

turb:
Thanks for this!

I executed some loads to BQ based on 574db491449fda922c52d8024d502fed28031926 and it seems whole entries with JSON non-null columns are just filtered out, which is odd (jsarray or jsobject as well).

Tried to find the origin, but everything seems correct...

turb:
tl;dr: I think I found it (see at the end).

This problem is only when writing with **Storage API**.

Test code:
```scala
import com.spotify.scio.ContextAndArgs
import com.spotify.scio.bigquery._
import com.spotify.scio.bigquery.Table
import com.spotify.scio.bigquery.types.{BigQueryType, Json, description}
import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO
import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO.Write.WriteDisposition

import scala.concurrent.duration.Duration

object TestJob {

  @BigQueryType.toTable
  @description("Test")
  case class Test(id: String, json: Option[Json])

  def main(args: Array[String]): Unit = {
    implicit val (sc, _) = ContextAndArgs(args)

    val tests = sc.parallelize(Seq(
      Test("stored", None),
      Test("ignored", Some(Json("{\"key\":\"value\"}"))),
    ))

    tests.saveAsTypedBigQueryTable(
      Table.Spec(s"Test.test"),
      method = BigQueryIO.Write.Method.STORAGE_WRITE_API,
      writeDisposition = WriteDisposition.WRITE_APPEND,
    )

    sc.run().waitUntilFinish(Duration.Inf, cancelJob = false)
  }
}
```

Here the first `Test` is inserted in the `test` table, but the second one is just filtered out.

After some digging, the error can be found in `org.apache.beam.sdk.io.gcp.bigquery.WriteResult#getFailedStorageApiInserts` called by `BigQueryStorageWriteApiSchemaTransformProvider`:

`BigQueryStorageApiInsertError{row=GenericData{classInfo=[f], {id=ignored, json={key=value}}}, errorMessage='syntax error while parsing object key - invalid literal; last read: '{k'; expected string literal'}`.

It looks like at some point `{"key":"value"}` becomes `{key=value}`, so the insert is erroneous, then ignored.

## Found solution
It seems it comes from `com.spotify.scio.bigquery.Json`:

```scala
def parse(json: Json): AnyRef = mapper.readValue(json.wkt, classOf[Object])
```

I think since it's there only to encapsulate a JSON string it should be:

```scala
def parse(json: Json): AnyRef = json.wkt
```

It solves the problem for my use-case, but I wonder if there are side-effects I don't know.

(another thing is Beam silently dropping bad inserts...)

scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/package.scala
```diff
@@ -63,17 +65,29 @@ package object types {
...
+    @unused private def apply(wkt: String): Json = new Json(wkt)
+
+    def apply(value: AnyRef): Json = value match {
+      case str: String if Try(mapper.readTree(str)).isSuccess =>
+        // string formatted json vs string literal
+        new Json(str)
+      case _ =>
+        new Json(mapper.writeValueAsString(value))
+    }
+    def parse(json: Json): AnyRef = mapper.readValue(json.wkt, classOf[Object])
```
turb, line 90:
```suggestion
    def parse(json: Json): AnyRef = json.wkt
```

RustedBones:
Closing in favor of #5589

PR #5545, opened 2025-01-20
Update GHA

RustedBones:
Apply same changes from CI gha to other actions.
Drop release drafter.
Fix populate test data to work with JSON


RustedBones:
Successful run: https://github.com/spotify/scio/actions/runs/12887324004

PR #5549, opened 2025-01-22
Supply gcs-connector options from default Configuration

clairemcginty:
https://github.com/apache/beam/pull/32769 is released in Beam 2.62.0; this PR uses the new GcsOption to pass all available gcs-connector options, parsed from the default `Configuration` (i.e. core-site.xml).

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
@@ -48,6 +48,7 @@ import scala.collection.mutable.{Buffer => MBuffer}
 import scala.concurrent.duration.Duration
 import scala.io.Source
 import scala.reflect.ClassTag
+import scala.util.chaining._
 import scala.util.control.NoStackTrace
 import scala.util.{Failure, Success, Try}
 import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions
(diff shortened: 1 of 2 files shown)


scio-core/src/main/scala/com/spotify/scio/ScioContext.scala
```diff
@@ -454,6 +455,87 @@ class ScioContext private[scio] (
...
+    import org.apache.hadoop.conf.Configuration
+    import com.google.cloud.hadoop.fs.gcs.{GoogleHadoopFileSystemConfiguration => GfsConfig}
+    import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions
+
+    try {
+      // If Hadoop is on the classpath, try to parse default gcs-connector options
+      val config = new Configuration()
+      val o = optionsAs[GcsOptions]
+
+      o.setGoogleCloudStorageReadOptions(
```
clairemcginty, line 468: There's a method to parse all this for us but it's not public: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/v3.0.4/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L645-L660

I do have a PR up to make it public so hopefully it can be dropped in in a future release.

kellen: Link the PR in the code somewhere so we can find it later

kellen (approved):

PR #5550, opened 2025-01-23
Some missing options in SnowflakeIO

turb:


scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
```diff
@@ -186,7 +186,7 @@ final case class SnowflakeTable[T](connectionOptions: SnowflakeConnectionOptions
       .read[T]()
       .fromTable(table)
       .withDataSourceConfiguration(dataSourceConfiguration(connectionOptions))
-      .withStorageIntegrationName(params.storageIntegrationName)
+      .pipe(r => Option(params.storageIntegrationName).fold(r)(r.withStorageIntegrationName))
```
RustedBones, line 189: isn't this a required parameter on read ?

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
```diff
@@ -210,6 +210,7 @@ final case class SnowflakeTable[T](connectionOptions: SnowflakeConnectionOptions
       .pipe(w => Option(params.flushTimeLimit).fold(w)(w.withFlushTimeLimit))
       .pipe(w => Option(params.quotationMark).fold(w)(w.withQuotationMark))
       .pipe(w => Option(params.storageIntegrationName).fold(w)(w.withStorageIntegrationName))
+      .pipe(w => Option(params.tableSchema).fold(w)(w.withTableSchema))
```
RustedBones, line 213: Oops. Good catch

scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
turb: At one point I got an NPE on this one because unset then `null`, but can't find back how I managed to do that...

turb: Makes no difference, removing.

RustedBones (approved):

PR #5551, opened 2025-01-28
Snowflake write: escape backslash in CSV files

turb:
Unsure it is the right way to do that — can't find how to properly configure escapes chars with kantan.

### DIFF
--- a/scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
+++ b/scio-snowflake/src/main/scala/com/spotify/scio/snowflake/SnowflakeIO.scala
@@ -124,7 +124,8 @@ object SnowflakeIO {
   }
 
   private[snowflake] def userDataMapper[T: RowEncoder]: UserDataMapper[T] = { (element: T) =>
-    RowEncoder[T].encode(element).toArray
+    // Snowflake COPY uses backslash as escape char
+    RowEncoder[T].encode(element).toArray.map(_.replace("\\", "\\\\"))
   }
 }
 


RustedBones:
If the data contains `\`, shouldn't it be better to set `ESCAPE_UNENCLOSED_FIELD=NONE` as explained [here](https://community.snowflake.com/s/article/How-To-Load-data-with-backslash-in-Snowflake-table) ?

turb:
> If the data contains \, shouldn't it be better to set ESCAPE_UNENCLOSED_FIELD=NONE as explained [here](https://community.snowflake.com/s/article/How-To-Load-data-with-backslash-in-Snowflake-table) ?

I don't think it would work with data that may contains both comme `,` or simple quote `'` — which is escaped by `\'`.

We found this because we have data with a string like `something \` (or in Scala syntax `"something \\"`) that encodes in CSV as `'something \'`. Then the Snowflake parser interprets the quote as escaped.

With the fix it's now `'something \\'` (`"something \\\\"` in scala syntax).

It may be managed by kantan (with `kantan.csv.CsvConfiguration`), but it seems only at the kantan.csv level, not kantan.codecs.

turb:
FYI I also opened a PR in Apache Beam: https://github.com/apache/beam/pull/33803

I think it's difficult to have something really clean, since Beam does half a CSV serialization (doing a naive `fields.mkString("'", "','", "'")`, so providing the other half (array of serialized fields) can only be a hack around that...

kellen:
Should we punt on this until we integrate beam 2.63?

turb:
> Should we punt on this until we integrate beam 2.63?

No, there isn't a dependency. This one and the one in Beam fix distinct bugs, though in the same area.

turb:
Hello @kellen and @RustedBones, any chance this can be merged?

turb:
I opened https://github.com/apache/beam/pull/33948 on Apache Beam, that should take care of it. Better to wait a bit to see if it can be merged before closing this one.

turb:
https://github.com/apache/beam/pull/33948 has been merged, closing.

PR #5591, opened 2025-01-31
10 minute timeout for futures by default to split the difference between Inf and 1

kellen:
We've had several internal reports of "previously working" pipelines repeatedly failing due to timeout


RustedBones (approved):
Sounds good. I wasn't too sure about 1mn TBH. 10mn is definitly on the safe side.
Looks it also matches with the 1st dataflow log:
```
Operation ongoing in step XXX for at least 10m00s without outputting or completing in state XXX
```

PR #5592, opened 2025-01-31
Fix memory leak in avro SlowGenericRecordCoder

RustedBones:
- `AvroCoder` should not be created on every elements as it uses  ThreadLocal variables that do not get unregistered.
- avro `GenericDatumReader` uses a thread local `RESOLVER_CACHE` with schema keys to reuse decoders. Reuse parsed schemas to avoid growing the cache

### DIFF
--- a/scio-avro/src/main/scala/com/spotify/scio/coders/avro/AvroCoders.scala
+++ b/scio-avro/src/main/scala/com/spotify/scio/coders/avro/AvroCoders.scala
@@ -39,7 +39,12 @@ final private class SlowGenericRecordCoder extends CustomCoder[GenericRecord] {
   // Use String for 1.8 compat
   private val sc = StringUtf8Coder.of()
 
-  // Reuse parsed schemas because avro uses caching
+  /**
+   * Reuse parsed schemas because GenericDatumReader caches decoders based on schema reference
+   * equality.
+   * @see
+   *   [[org.apache.avro.generic.GenericDatumReader.getResolver]].
+   */
   @transient private lazy val schemaCache = new TrieMap[String, Schema]()
 
   private val encoder = new EmptyOnDeserializationThreadLocal[BinaryEncoder]()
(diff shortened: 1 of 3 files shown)


scio-avro/src/main/scala/com/spotify/scio/coders/avro/AvroCoders.scala
```diff
@@ -59,18 +71,6 @@ final private class SlowGenericRecordCoder extends AtomicCoder[GenericRecord] {
       this,
       "Coder[GenericRecord] without schema is non-deterministic"
     )
-  override def consistentWithEquals(): Boolean = false
-  override def structuralValue(value: GenericRecord): AnyRef =
```
kellen, line 63: Is this not important?

scio-avro/src/main/scala/com/spotify/scio/coders/avro/AvroCoders.scala
```diff
@@ -22,35 +22,47 @@ import com.spotify.scio.coders.{Coder, CoderGrammar}
...
-final private class SlowGenericRecordCoder extends AtomicCoder[GenericRecord] {
-  // TODO: can we find something more efficient than String ?
+final private class SlowGenericRecordCoder extends CustomCoder[GenericRecord] {
+  // Schema is serializable on avro 1.9+
+  // Use String for 1.8 compat
   private val sc = StringUtf8Coder.of()
 
-  private def genericCoder(schema: Schema): BCoder[GenericRecord] =
-    AvroCoder.of(GenericRecordDatumFactory, schema)
+  // Reuse parsed schemas because avro uses caching
```
clairemcginty, line 42: nit: can you expand this comment a bit to explain what specifically is cached in Avro? I don't want us to forget the context on this later 🙏

scio-avro/src/main/scala/com/spotify/scio/coders/avro/AvroCoders.scala
kellen: We discussed in fireside

PR #5602, opened 2025-02-10
Vendor chill

kellen:
I looked into whether importing some subset of chill was reasonable but most of the various chill deps are kryo serializers or registrars, so appear to be required for our usecase.


kellen:
Punting to 0.16

jto:
Wouldn't it be easier to just get rid of Chill ?

A lot of usage in Scio code are references to types which are just aliases:

```scala
type Kryo = com.esotericsoftware.kryo.Kryo
type KSerializer[T] = com.esotericsoftware.kryo.Serializer[T]
type Input = com.esotericsoftware.kryo.io.Input
type Output = com.esotericsoftware.kryo.io.Output
```

Then there's `ClosureCleaner` but it's has no dependency on the rest of chill and could be part of Scio or we could use another implementation.

We also rely on `Externalizer` but again that's 200 loc and potentially we could re-implement it.

The biggest issue perhaps would be the loss of Kryo coders for Scala types but since we barely use Kryo anymore maybe it's ok to not have them ?

Issue #5597, opened 2025-02-05
BigQueryType writes with nested record fields fail on 0.14.11


clairemcginty:
The following write operation fails on 0.14.11:

```scala
case class BqNested(id: Int)
@BigQueryType.toTable
case class BqOutput(id: Int, nestedField: Option[BqNested])

...

sc.parallelize(1 to 10)
  .map(i => BqOutput(i, Some(BqNested(i + 1))))
  .saveAsTypedBigQueryTable(
    table = Table.Spec(args("output")),
    writeDisposition = WRITE_TRUNCATE,
    createDisposition = CREATE_IF_NEEDED
  )
```

with the following worker error log:

```
Error message from worker: org.apache.beam.sdk.util.UserCodeException: org.apache.avro.file.DataFileWriter$AppendWriteException: org.apache.avro.UnresolvedUnionException: Not in union ["null",{"type":"record","name":"nestedField","namespace":"org.apache.beam.sdk.io.gcp.bigquery","doc":"Translated Avro Schema for nestedField","fields":[{"name":"id","type":"long"}]}]: {"id": 2} (field=nestedField)
	org.apache.beam.sdk.util.UserCodeException.wrap(UserCodeException.java:39)
	org.apache.beam.sdk.io.gcp.bigquery.WriteBundlesToFiles$DoFnInvoker.invokeProcessElement(Unknown Source)
	org.apache.beam.fn.harness.FnApiDoFnRunner.processElementForWindowObservingParDo(FnApiDoFnRunner.java:823)
	org.apache.beam.fn.harness.data.PCollectionConsumerRegistry$MetricTrackingFnDataReceiver.accept(PCollectionConsumerRegistry.java:348)
	org.apache.beam.fn.harness.data.PCollectionConsumerRegistry$MetricTrackingFnDataReceiver.accept(PCollectionConsumerRegistry.java:275)
	org.apache.beam.fn.harness.MapFnRunners$ExplodedWindowedValueMapperFactory.lambda$create$0(MapFnRunners.java:133)
	org.apache.beam.fn.harness.data.PCollectionConsumerRegistry$MetricTrackingFnDataReceiver.accept(PCollectionConsumerRegistry.java:348)
	org.apache.beam.fn.harness.data.PCollectionConsumerRegistry$MetricTrackingFnDataReceiver.accept(PCollectionConsumerRegistry.java:275)
[...]

PR #5611, opened 2025-02-18
Fix bq Avro format issue for nested records

clairemcginty:
A proper fix for https://github.com/spotify/scio/issues/5597.

tl;dr for Avro-formatted BQ loads, an Avro schema should always set the _type_ of a nested record type to match its _field name_, since BQ schemas do not support naming nested schemas. The behavior of Beam's `BigQueryUtils.toGenericAvroSchema` utility is correct in this regard (minus inconsistent namespace assignment); we just needed to properly set the right field names in the BigQueryType macro.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1050,6 +1050,7 @@ lazy val `scio-google-cloud-platform` = project
       moduleFilter("org.apache.arrow", "arrow-vector"),
       moduleFilter("com.fasterxml.jackson.datatype", "jackson-datatype-jsr310")
     ).reduce(_ | _),
+    Test / javaOptions += "-Doverride.type.provider=com.spotify.scio.bigquery.validation.SampleOverrideTypeProvider",
     libraryDependencies ++= Seq(
       // compile
       "com.esotericsoftware" % "kryo-shaded" % kryoVersion,
--- a/integration/src/test/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
+++ b/integration/src/test/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
@@ -21,11 +21,14 @@ import com.google.protobuf.ByteString
 import com.spotify.scio.avro._
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.bigquery.BigQueryTypedTable.Format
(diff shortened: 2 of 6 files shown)


integration/src/test/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
```diff
@@ -190,10 +224,42 @@ class TypedBigQueryIT extends PipelineSpec with BeforeAndAfterAll {
         )
     }.waitUntilFinish()
 
+    runWithRealContext(options) { sc =>
+      sc.typedBigQuery[Record](avroTable) should containInAnyOrder(records)
+    }
+
+    // Due to Beam bug with automatic schema detection, can't parse nested record types as GenericRecords yet
+    // Todo remove assertThrows after fixing in Beam
+    assertThrows[UnresolvedUnionException] {
```
clairemcginty, line 233: basically Beam attempts to infer an Avro schema the selected BQ table here, and it doesn't work correctly if you have a nested record (capitalization is wrong and record name is just "root" for some reason?). There's also no builder method available for us to manually set an Avro schema, either, which I think we'll have to add to Beam.

kellen (approved):

build.sbt
```diff
@@ -1050,6 +1050,7 @@ lazy val `scio-google-cloud-platform` = project
       moduleFilter("org.apache.arrow", "arrow-vector"),
       moduleFilter("com.fasterxml.jackson.datatype", "jackson-datatype-jsr310")
     ).reduce(_ | _),
+    Test / javaOptions += "-Doverride.type.provider=com.spotify.scio.bigquery.validation.SampleOverrideTypeProvider",
```
clairemcginty, line 1053: For context, BigQueryValidationTest was [failing](https://github.com/spotify/scio/actions/runs/13419010807/job/37486888149?pr=5611) because the system property `override.type.provider` used in tests was getting unset. Somehow this is happening in the test I added for `ConverterProviderTest` -- I set that test to `ignore` and the system property stayed set. Not sure why but let's just set it in the build?

kellen (approved):

PR #5613, opened 2025-02-19
Upgrade to Beam 2.63.0

clairemcginty:
Contains bugfixes for BQ Storage-API writes and BQ Avro-format reads.

### DIFF
--- a/integration/src/test/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
+++ b/integration/src/test/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
@@ -145,7 +145,7 @@ object TypedBigQueryIT {
   private val typedTableStorage = table("records_storage")
   private val tableRowTable = table("records_tablerow")
   private val avroTable = table("records_avro")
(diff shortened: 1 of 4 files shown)


integration/src/test/scala/com/spotify/scio/bigquery/TypedBigQueryIT.scala
```diff
@@ -186,6 +188,22 @@ class TypedBigQueryIT extends PipelineSpec with BeforeAndAfterAll {
...
+          method = WriteMethod.STORAGE_WRITE_API
+        )
+    }.waitUntilFinish()
+
+    runWithRealContext(options) { sc =>
+      val data = sc.typedBigQuery[Record](typedTableStorage)
+      data should containInAnyOrder(records)
+    }
+  }
+
```
clairemcginty, line 206: Both of these tests are failing. `LocalTime` fields are being written incorrectly using the Storage Write API. For example, for the same record, FileLoads will write the value `19:00:00` and StorageWrite will write the value `15:59:15.338944`. Not quite sure what's going on, needs investigation.

kellen (approved):

PR #5620, opened 2025-02-26
Upgrade GHA os to ubuntu-24.04

clairemcginty:


clairemcginty:
Looks like SBT is not supported; closing

PR #5623, opened 2025-03-03
Fall back to TableRow representation for typed BQ writes containing a LocalDateTime field

clairemcginty:
typed FILE_LOADS writes containing a LocalDateTime field fail with this error:

```json
  "status": {
    "errorResult": {
      "location": "query",
      "message": "Field month has incompatible types. Configured schema: datetime; Avro file: string.",
      "reason": "invalidQuery"
    },
    "errors": [
      {
        "location": "query",
        "message": "Field month has incompatible types. Configured schema: datetime; Avro file: string.",
        "reason": "invalidQuery"
      }
    ],
    "state": "DONE"
  },
```

as it turns out, Beam's TableRow-to-Avro converter writes LocalDateTimes as a _datetime string_, which is supported by the Storage Writes API. Unfortunately, FILE_LOADS expects a _micro timestamp as a Long_.


jto (approved):
LGTM. 
Hopefully this gets fixed in Beam and we can just remove the workaround.

PR #5641, opened 2025-03-11
Fix link in Scio-Unit-Tests.md

RickardZwahlen:
SCollectionMatchers doesn't have a companion object, which I think is what the $ syntax refers to.

The current link returns a 404, but https://spotify.github.io/scio/api/com/spotify/scio/testing/SCollectionMatchers.html works.


clairemcginty (approved):
thanks for the fix!

PR #5658, opened 2025-04-01
Upgrade to Beam 2.64.0

clairemcginty:


clairemcginty:
...running into some Protobuf compat issues in IT tests, investigating..

build.sbt
```diff
@@ -860,7 +861,7 @@ lazy val `scio-core` = project
       "com.google.guava" % "guava" % guavaVersion,
       "com.google.http-client" % "google-http-client" % gcpBom.key.value,
       "com.google.http-client" % "google-http-client-gson" % gcpBom.key.value,
-      "com.google.protobuf" % "protobuf-java" % gcpBom.key.value,
+      "com.google.protobuf" % "protobuf-java" % protobufVersion,
```
clairemcginty, line 864: these do not change the build since protobuf-java(-util) version is already being overridden in `depdencyOverrides`, but was needed to make `sbt unusedCompileDependenciesTest` pass (I guess it doesn't like `libraryDepdencies` pulling in 4.x and `dependencyOverrides` pulling 3.x).

Issue #5533, opened 2024-12-05
Add missing api to `SCollectionWithSideInput`


RustedBones:
`SCollectionWithSideInput` only provides the basic operations. It would be convenient foe the users to expose more of them (eg. `collect`, `mapWithResource`, ...)

PR #5666, opened 2025-04-03
Add collect method to SCollectionWithSideInput

charles-tan:
Relates to https://github.com/spotify/scio/issues/5533

Adds `collect` method to `SCollectionWithSideInput`.

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideInput.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideInput.scala
@@ -82,6 +82,16 @@ class SCollectionWithSideInput[T] private[values] (
     new SCollectionWithSideInput[U](o, sides)
   }
 
+  /** [[SCollection.map]] with an additional [[SideInputContext]] argument. */
+  def collect[U: Coder](
+    f: PartialFunction[(T, SideInputContext[T]), U]
+  ): SCollectionWithSideInput[U] = {
+    val o = coll
+      .pApply(parDo(FunctionsWithSideInput.partialFn(f)))
+      .setCoder(CoderMaterializer.beam(context, Coder[U]))
+    new SCollectionWithSideInput[U](o, sides)
+  }
+
   /**
    * Allows multiple outputs from [[SCollectionWithSideInput]].
    *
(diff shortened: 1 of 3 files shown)


scio-core/src/test/scala/com/spotify/scio/values/SCollectionWithSideInputTest.scala
```diff
@@ -379,4 +379,16 @@ class SCollectionWithSideInputTest extends PipelineSpec {
...
   }
+
+  it should "support collect" in {
+    runWithContext { sc =>
+      val p1 = sc.parallelize(Seq("a", "b", "d"))
+      val p2 = sc.parallelize(sideData).asMapSingletonSideInput
+      val pf: PartialFunction[(String, SideInputContext[String]), Int] = {
+        case (i, c) if c(p2).contains(i) => c(p2)(i) * 2
+      }
+      val s = p1.withSideInputs(p2).collect(pf).toSCollection
```
clairemcginty, line 390: 🎨 could be written in a slightly more idiomatic way, like:

```suggestion
      val s = p1
        .withSideInputs(p2)
        .collect {
          case (record, ctx) if ctx(p2).contains(record) => ctx(p2)(record) * 2
        }
        .toSCollection
```

scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideInput.scala
```diff
@@ -82,6 +90,101 @@ class SCollectionWithSideInput[T] private[values] (
...
+  def withWindow[W <: BoundedWindow: Coder]: SCollectionWithSideInput[(T, W)] =
+    new SCollectionWithSideInput(coll.withWindow, sides)
+
+  /** [[SCollection.withFixedWindows]] that retains [[SideInput]]. */
+  def withFixedWindows(
+    duration: Duration,
+    offset: Duration = Duration.ZERO,
+    options: WindowOptions = WindowOptions()
+  ): SCollectionWithSideInput[T] =
+    new SCollectionWithSideInput(coll.withFixedWindows(duration, offset, options), sides)
```
clairemcginty, line 186: I think we probably don't need these methods, because the user can just write `data.withPaneInfo.withSideInputs` rather than `data.withSideInputs.withPaneInfo` and end up with the same result, right? 🤔

clairemcginty (commented):
nice! appreciate the thorough tests!

scio-core/src/main/scala/com/spotify/scio/values/SCollectionWithSideInput.scala
charles-tan: yes you're right, i didn't know if there was a use case for someone who already has a SCollectionWithSIdeInput then later wanted to get pane info. I'll remove them.

clairemcginty (commented):
lgtm!

PR #5700, opened 2025-06-10
Expose Beam's Lineage

shnapz:
- Expose Beam's lineage in ScioResult
  - Beam's lineage is available as `Lineage.query(pipelineResult.metrics(), Lineage.Type.SOURCE)` (only available as pipeline result)
  - Expose as Java API in Scio result: `scioResult.getMetrics.getBeamLineage` 
  - Existing implementation will dump this to a file configured by `--metricsLocation`
  - Impove `--metricsLocation` option by accepting a directory path. If param value ends with `/` then it is treated as a directory and new file is generated inside. This is needed in case when a Java app has more *than one Scio pipelines* but shares the same cli args.
- Fix `JdbcIO` to force Beam to report CloudSQL lineage by adding `cloudSqlInstance=...` to `dataConfig.connectionProperties`

### DIFF
--- a/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
+++ b/scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
@@ -59,6 +59,9 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
   /** Get metrics of the finished pipeline. */
   def getMetrics: Metrics
 
+  /** Get lineage data of the finished pipeline. */
+  def getLineage: Lineage
+
   /** Whether this is the result of a test. */
   def isTest: Boolean = false
 
@@ -66,12 +69,19 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
   def state: State = Try(internal.getState).getOrElse(State.UNKNOWN)
 
   /** Save metrics of the finished pipeline to a file. */
-  def saveMetrics(filename: String): Unit = {
+  def saveMetrics(filename: String): Unit =
+    saveJsonFile(filename, getMetrics)
+
+  /** Save lineage of the finished pipeline to a file. */
+  def saveLineage(filename: String): Unit =
+    saveJsonFile(filename, getLineage)
+
+  private def saveJsonFile(filename: String, value: Object): Unit = {
     val mapper = ScioUtil.getScalaJsonMapper
     val resourceId = FileSystems.matchNewResource(filename, false)
     val out = FileSystems.create(resourceId, MimeTypes.TEXT)
     try {
-      out.write(ByteBuffer.wrap(mapper.writeValueAsBytes(getMetrics)))
+      out.write(ByteBuffer.wrap(mapper.writeValueAsBytes(value)))
     } finally {
       if (out != null) {
         out.close()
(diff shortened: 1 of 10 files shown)


scio-jdbc/src/main/scala/com/spotify/scio/jdbc/JdbcConnectionOptions.scala
```diff
@@ -35,5 +35,35 @@ final case class JdbcConnectionOptions(
   username: String,
   password: Option[String],
   connectionUrl: String,
+  cloudSqlInstanceConnectionName: Option[String],
   driverClass: Class[_ <: Driver]
 )
+
+object JdbcConnectionOptions {
+  def apply(
```
shnapz, line 43: This is going to make the code compatible (although binary incompatible)

scio-jdbc/src/main/scala/com/spotify/scio/jdbc/JdbcConnectionOptions.scala
```diff
@@ -35,5 +35,35 @@ final case class JdbcConnectionOptions(
   username: String,
   password: Option[String],
   connectionUrl: String,
+  cloudSqlInstanceConnectionName: Option[String],
```
shnapz, line 38: Added this field because:
- `com.spotify.scio.jdbc.CloudSqlOptions` has `--cloudSqlInstanceConnectionName` arg, so some users can use it
- want to avoid parsing URL if possible

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -65,19 +68,59 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
...
   def state: State = Try(internal.getState).getOrElse(State.UNKNOWN)
 
-  /** Save metrics of the finished pipeline to a file. */
-  def saveMetrics(filename: String): Unit = {
+  /** Save metrics of the finished pipeline to a file or directory. */
+  def saveMetrics(path: String): Unit = {
+    val isDirectory = path.endsWith(File.separator)
+
+    val resourceId = if (isDirectory) {
+      val randomFileName = f"${UUID.randomUUID()}.json"
```
clairemcginty, line 76: hm, do we actually need this to be random? mould just `metrics.json` work? or even `metrics-{jobId}.json` ?

scio-core/src/main/scala/com/spotify/scio/runners/dataflow/DataflowResult.scala
```diff
@@ -58,7 +57,7 @@ class DataflowResult(val internal: DataflowPipelineJob) extends RunnerResult {
   /** Get a generic [[ScioResult]]. */
   override def asScioResult: ScioResult = new DataflowScioResult(internal)
 
-  private class DataflowScioResult(internal: PipelineResult) extends ScioResult(internal) {
+  private class DataflowScioResult(internal: DataflowPipelineJob) extends ScioResult(internal) {
```
clairemcginty, line 60: is the type change needed here?

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
charles-tan: maybe appending timestamp can be good?

scio-core/src/main/scala/com/spotify/scio/ScioResult.scala
```diff
@@ -65,19 +68,59 @@ abstract class ScioResult private[scio] (val internal: PipelineResult) {
...
+        buffer += iterator.next()
+      }
+      buffer
+    }
+
+    def getBeamResource(metricValue: String): LineageResource = {
+      metricValue.split(':') match {
+        case Array(tp, id) => LineageResource(tp, id, metricValue)
+        case _             => LineageResource(null, null, metricValue)
+      }
```
charles-tan, line 113: What kind of values are returned in `case Array(tp, id)` vs `case _`?

charles-tan (approved):

PR #5714, opened 2025-06-25
(fix #5713) Fix BigQueryType#fromAvro on missing fields for Avro 1.11

### LINKED ISSUE #5713: BigQueryType#fromAvro throws an error populating empty Optional fields when using Avro 1.11
Given an Avro record with schema:

```avsc
[{
    "type": "record",
    "name": "MyRecord",
    "fields": [
        {
          "name": "foo",
          "type": "string"
        }
    ]
}]
```

On Avro 1.8, invoking `GenericRecord#get("bar")` returns null.

On Avro 1.11+, invoking `GenericRecord#get("bar")` throws an `org.apache.avro.AvroRuntimeException: Not a valid schema field: bar`.

This has a consequence for BigQueryType classes that include an optional field that's not guaranteed to exist in the BQ table. Typed BQ that specifies BigQueryIO's `GenericRecord` format and invokes `BigQueryType#fromAvro` on the resulting record will throw the `AvroRuntimeException` on Avro 1.11.

We could fix it by updating `BigQueryType#fromAvro` to check for the existence of the field in the schema before calling `GenericRecord#get`, if the corresponding case class field is an `Option`: https://github.com/spotify/scio/blob/v0.14.17/scio-google-cloud-platform/src/main/scala/com/spotify/scio/bigquery/types/ConverterProvider.scala#L134-L136

clairemcginty:


scio-jmh/src/test/scala/com/spotify/scio/jmh/BigQueryTypeBenchmark.scala
```diff
@@ -70,6 +78,12 @@ class BigQueryTypeBenchmark {
   @Benchmark
   def bqtSimpleFromAvro(): Unit = BqtSimple.fromAvro(GrSimple)
 
+  @Benchmark
```
clairemcginty, line 81: Baseline results:

```
Baseline:

Benchmark                                          Mode  Cnt    Score   Error  Units
BigQueryTypeBenchmark.bqtOptionalEmptyFromAvro     avgt   10   86.771 ± 3.388  ns/op
BigQueryTypeBenchmark.bqtOptionalNonEmptyFromAvro  avgt   10  103.570 ± 1.676  ns/op
```

Results after applying this PR branch:
```
Benchmark                                          Mode  Cnt    Score   Error  Units
BigQueryTypeBenchmark.bqtOptionalEmptyFromAvro     avgt   10   85.049 ± 1.070  ns/op
BigQueryTypeBenchmark.bqtOptionalNonEmptyFromAvro  avgt   10  109.330 ± 0.417  ns/op
```

shnapz (approved):

jto (approved):

PR #5717, opened 2025-07-02
Update sonatype host to central portal

clairemcginty:
see: https://central.sonatype.org/pages/ossrh-eol/


charles-tan (approved):

build.sbt
```diff
@@ -141,7 +141,6 @@ ThisBuild / organization := "com.spotify"
 ThisBuild / organizationName := "Spotify AB"
 ThisBuild / startYear := Some(2016)
 ThisBuild / licenses := Seq(License.Apache2)
-ThisBuild / sonatypeCredentialHost := Sonatype.sonatypeLegacy
```
clairemcginty, line 144: this option was removed in 0.8.0, looks like it defaults to Central

PR #5732, opened 2025-07-25
Add support for vectoredRead-compatible stream in scio-parquet

clairemcginty:
Not ready for merge!!

glue code to support readVectored-compatible streams for scio-parquet.


build.sbt
```diff
@@ -706,6 +707,23 @@ def beamRunnerSettings: Seq[Setting[_]] = Seq(
...
+  .map(_.toBoolean)
+  .collect { case true =>
+    Seq(
+      dependencyOverrides ++= Seq(
+        "com.google.cloud.bigdataoss" % "gcs-connector" % bigdataoss3Version,
+        "com.google.cloud.bigdataoss" % "gcsio" % bigdataoss3Version,
+        "com.google.cloud.bigdataoss" % "util-hadoop" % bigdataoss3Version,
+        "org.apache.hadoop" % "hadoop-common" % "3.3.6",
+        "org.apache.hadoop" % "hadoop-auth" % "3.3.6",
+        "org.apache.hadoop" % "hadoop-client" % "3.3.6"
```
charles-tan, line 721: nit: maybe move these hardcoded versions to some variable

charles-tan: wait I just realized we already have `hadoopVersion` variable set to `3.4.1` which is used for `org.apache.hadoop:hadoop-common` and `org.apache.hadoop:hadoop-client`. Will this be an issue? can we just bring everything down to `3.3.6` or does this cause issue with beam's hadoop version?

clairemcginty: I was on the fence about that a bit!! Ultimately I left the default version as 3.4 since it does contain some (mostly) transitive security fixes.

build.sbt
```diff
@@ -706,6 +707,23 @@ def beamRunnerSettings: Seq[Setting[_]] = Seq(
   libraryDependencies ++= beamRunnersEval.value
 )
 
+def vectoredReadSettings: Seq[Setting[_]] = sys.props
+  .get("vectoredReadsEnabled")
```
jto, line 711: So if I understand things correctly to use vectored reads you need to have the `vectoredReadsEnabled` set at build time and also have `parquet.hadoop.vectored.io.enabled=true` in conf ?
Could you add doc in this PR ?

jto (approved):
Maybe add basic doc. I think the setup is a bit complex. It'd be great if you could just pass `--useVectoredParquetReads`.

build.sbt
clairemcginty: the `vectoredReadsEnabled` setting isn't user-facing - it's only for enabling the overrides to run scio-examples with (similar to how we support `sbt -DbeamRunners=DataflowRunner scio-examples/runMain ...`

I realize I did not include the config setting in the doc though, will add!

clairemcginty:
> Maybe add basic doc. I think the setup is a bit complex. It'd be great if you could just pass `--useVectoredParquetReads`.

My hope is that eventually the dependency stuff gets resolved (in Scio 0.15, when we officially drop Java 8, we can just upgrade the project gcs-connector version) so that the eventual user experience is just adding the boolean flag to their config 🙏  Maybe until 0.14 we just consider it as an experimental feature (I kept the new classes package-private so the public facing API isn't changed)

PR #5811, opened 2025-11-12
Vendor Socco

kellen:
Vendor socco to avoid release issues blocking everything else

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,7 +301,7 @@ jobs:
       - name: Submit Dependencies
         uses: scalacenter/sbt-dependency-submission@v2
         with:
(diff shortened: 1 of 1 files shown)


project/Socco.scala
```diff
@@ -0,0 +1,205 @@
...
+import com.typesafe.sbt.site.SitePlugin
+import java.io.PrintWriter
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.util.stream.Collectors
+import scala.collection.JavaConverters._
+import scala.util.Try
+import scala.io.Source
+
+object SbtSoccoPlugin extends AutoPlugin {
```
clairemcginty: feel like we could go even simpler and just leave this as a plain object like:

```scala
object Socco {
  val settings = Seq(
     commands += enableSoccoCommand,
     soccoOnCompile.in(p) := onCompile,
      soccoPackage := Nil,
      soccoOut := target.value / "socco",
      soccoIndex := index.value,
      ...
     )
}
```

and just add `.settings(Socco.settings)` to the `site` module in build.sbt. wdyt?

PR #5821, opened 2025-12-02
Support Avro 1.11 java-class annotations in scio-parquet

clairemcginty:
parquet-avro requires a default configuration option to be set to read and write Avro classes using `java-class` property: https://github.com/clairemcginty/parquet-mr/blob/master/parquet-avro/README.md#configuration-for-reading


kellen (approved):
yolo

PR #5849, opened 2026-01-02
Enable Beam lineage for JdbcIO

shnapz:
Bringing back only a part of the old PR #5700 
Forces Beam to report CloudSQL lineage by adding cloudSqlInstance=... to dataConfig.connectionProperties

### DIFF
--- a/scio-jdbc/src/test/scala/com/spotify/scio/jdbc/JdbcIOTest.scala
+++ b/scio-jdbc/src/test/scala/com/spotify/scio/jdbc/JdbcIOTest.scala
@@ -160,7 +160,7 @@ class JdbcIOTest extends AnyFlatSpec with Matchers {
     connectionProperties should be(None)
   }
 
-  it must "be case-sensitive for cloudSqlInstance parameter name" in {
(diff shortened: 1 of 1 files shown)


scio-jdbc/src/main/scala/com/spotify/scio/jdbc/JdbcIO.scala
```diff
@@ -50,16 +50,32 @@ object JdbcIO {
...
-          .withUsername(opts.username)
+    BJdbcIO.DataSourceConfiguration
+      .create(opts.driverClass.getCanonicalName, opts.connectionUrl)
+      .withUsername(opts.username)
+      .pipe { c =>
+        opts.password.fold(c)(c.withPassword)
+      }
+      .pipe { c =>
+        getUrlParameters(opts.connectionUrl)
+          .get("cloudSqlInstance")
```
clairemcginty: do we need `getUrlParameters` to return a Map if we're only retrieving a single key?

clairemcginty (approved):

PR #5850, opened 2026-01-02
Report Beam Lineage from Parquet reads

shnapz:


scio-parquet/src/main/scala/com/spotify/scio/parquet/tensorflow/ParquetExampleIO.scala
```diff
@@ -100,7 +100,8 @@ final case class ParquetExampleIO(path: String) extends ScioIO[Example] {
     params: ReadP
   ): SCollection[Example] = {
     val job = Job.getInstance(conf)
-    GcsConnectorUtil.setInputPaths(sc, job, path)
+    val filePattern = ScioUtil.filePattern(path, params.suffix)
```
shnapz: I am surprised suffix was not used initially. Or was it intentional?

scio-parquet/src/main/scala/com/spotify/scio/parquet/HadoopParquet.scala
```diff
@@ -0,0 +1,119 @@
...
+      })
+
+    val withSkipClone = skipValueClone.fold(hadoop)(skip => hadoop.withSkipValueClone(skip))
+
+    val withValueTranslation = projectionFn.fold {
+      withSkipClone.withValueTranslation(
+        new SimpleFunction[T, T]() {
+          override def apply(input: T): T = input
+
+          override def getInputTypeDescriptor: TypeDescriptor[T] = TypeDescriptor.of(outputType)
```
shnapz:
```suggestion
          override def getInputTypeDescriptor = TypeDescriptor.of(outputType)
```

scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/ParquetAvroIO.scala
```diff
@@ -245,26 +246,14 @@ object ParquetAvroIO {
...
-
-      sc.applyTransform(transform).map(_.getValue)
+      HadoopParquet
+        .readHadoopFormatIO[A, T](
+          sc,
+          job.getConfiguration,
+          Some(projectionFn),
+          None
+        )
+        .parDo(new LineageReportDoFn(filePattern))
```
kellen: Isn't this going to result in a new node in the graph? Why are we doing this in sequence w/ the read if it's not actually using any of the read elements; we should be doing like the scio init metrics which is just its own distinct graph `create impulse -> submit parquet lineage`

scio-parquet/src/main/scala/com/spotify/scio/parquet/HadoopParquet.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+import org.apache.beam.sdk.transforms.{DoFn, SimpleFunction}
+import org.apache.beam.sdk.values.TypeDescriptor
+import org.apache.hadoop.conf.Configuration
+import org.slf4j.LoggerFactory
+
+import java.util.concurrent.atomic.AtomicBoolean
+import scala.reflect.ClassTag
+
+private[parquet] object HadoopParquet {
```
kellen: This is just to reduce duplication or is there a functional change here?

shnapz: just to reduce, no new functionality, except I noticed that in some cases Scio's derived coder was not set to HadoopFormatIO transformation. Probably Beam auto-derives the same coder, but anyway it is better to set explicitly

PR #5867, opened 2026-01-07
Upgrade to Beam 2.70

clairemcginty:


clairemcginty:
it seems that the libraries-bom bump somehow broke our AutoValue processor in scio-smb 

edit: isolated it to some change between libraries-bom [26.66.0 -> 26.67.0](https://github.com/googleapis/java-cloud-bom/compare/v26.66.0...v26.67.0)

scio-core/src/main/java/com/spotify/scio/transforms/BaseAsyncBatchLookupDoFn.java
```diff
@@ -127,6 +131,14 @@ public BaseAsyncBatchLookupDoFn(
 
   public abstract TryWrapper failure(Throwable throwable);
 
+  // Set to arbitrarily high value; required to preserve timestamp of original element
```
clairemcginty: missed this one in https://github.com/spotify/scio/pull/5806

PR #5918, opened 2026-04-03
feat(sparkey): Deduplicate SparkeyReader across DoFn clones

spkrka:
## Summary

Beam creates one DoFn clone per vCPU thread (e.g. 80 on an n4-standard-80 worker). Previously, each clone independently called `uri.getReader()`, which downloads files from GCS and opens new `SparkeyReader` instances — duplicating file I/O, file descriptors, and mmap regions.

This adds a `CompletableFuture`-based reader cache so the first thread loads the reader and all others wait on the same future. The reader is reused for the JVM lifetime, which is safe for Dataflow batch (one pipeline per JVM).

### Changes

- `SparkeySideInput`, `LargeMapSideInput`, and `LargeSetSideInput` now go through `SparkeySideInput.getOrCreateReader()` instead of calling `uri.getReader()` directly
- Reader cache is a `ConcurrentHashMap[String, CompletableFuture[SparkeyReader]]` keyed by URI path
- `checkMemory` is preserved but called once per reader (inside the cache loader) instead of on every `get()` call
- No API changes, no new dependencies, no version bumps

### Why this is safe

- **No behavior change for correct programs**: readers were already expected to be long-lived; this just ensures they're shared
- **Thread safety**: `SparkeyReader` from `Sparkey.open()` returns a `PooledSparkeyReader` which is already thread-safe
- **Cache semantics**: keyed by path, never evicted. Acceptable for Dataflow batch where one pipeline runs per JVM. For DirectRunner/REPL reuse, stale readers could be returned if the same path is rewritten — same limitation as the existing `checkMemory` which also uses a static reference

### What this does NOT include

This is intentionally minimal. The heap-backed reader fallback and memory-aware shard placement from #5903 / #5904 are separate concerns and will be proposed separately (likely as opt-in features).

## Test plan

- [ ] Existing sparkey tests pass (no behavior change for single-reader scenarios)
- [ ] Verified on production Dataflow pipeline (n4-standard-80, 80 DoFn clones) — readers are loaded once instead of 80 times

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
+++ b/scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
@@ -32,6 +32,7 @@ import org.apache.beam.sdk.util.CoderUtils
 import org.apache.beam.sdk.values.PCollectionView
 import org.slf4j.LoggerFactory
 
+import java.util.concurrent.{CompletableFuture, ConcurrentHashMap}
 import scala.util.hashing.MurmurHash3
 
 /**
@@ -545,12 +546,11 @@ package object sparkey extends SparkeyReaderInstances with SparkeyCoders {
     mapFn: SparkeyReader => T
   ) extends SideInput[T] {
     override def updateCacheOnGlobalWindow: Boolean = false
-    override def get[I, O](context: DoFn[I, O]#ProcessContext): T =
-      mapFn(
-        SparkeySideInput.checkMemory(
-          context.sideInput(view).getReader(RemoteFileUtil.create(context.getPipelineOptions))
-        )
-      )
+    override def get[I, O](context: DoFn[I, O]#ProcessContext): T = {
+      val uri = context.sideInput(view)
+      val rfu = RemoteFileUtil.create(context.getPipelineOptions)
+      mapFn(SparkeySideInput.getOrCreateReader(uri, rfu))
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -576,16 +578,42 @@ package object sparkey extends SparkeyReaderInstances with SparkeyCoders {
...
+
+    private val readerCache =
+      new ConcurrentHashMap[String, CompletableFuture[SparkeyReader]]()
+
+    def getOrCreateReader(uri: SparkeyUri, rfu: RemoteFileUtil): SparkeyReader =
+      readerCache
+        .computeIfAbsent(
+          uri.path,
+          _ =>
+            CompletableFuture.supplyAsync { () =>
```
spkrka, line 608: Note: using a future here to support loading multiple readers in parallel _if_ they are expensive to create - currently that's not true, but if we add loading/locking/load to heap later it might be.

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
```diff
@@ -576,15 +578,41 @@ package object sparkey extends SparkeyReaderInstances with SparkeyCoders {
...
+              val reader = uri.getReader(rfu)
+              checkMemory(reader)
+              reader
```
labianchin, line 611:
```suggestion
              checkMemory(uri.getReader(rfu))
```

Can this be simplified?

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
labianchin: 🤔 But is the CompletableFuture any relevant if its call is always followed by `.join()`? I wonder removing CompletableFuture might be simpler.. If keeping, then consider leaving a comment for its motivation.

labianchin (approved):

scio-extra/src/main/scala/com/spotify/scio/extra/sparkey/package.scala
spkrka: Hm, perhaps you're right and I reasoned incorrectly about this - my hope was that multiple threads loading the same sideinputs would still allow for loading multiple things in parallel, but if the file ordering is stable, that doesn't actually help. Will think more about this, and potentially simplify

PR #5923, opened 2026-04-09
Fix multi-path URI escaping in GcsConnectorUtil

charles-tan:
## Summary
- `StringUtils.escapeString` was applied to the entire comma-separated path string in `GcsConnectorUtil.setInputPaths`, escaping the commas that delimit individual URIs. This caused `FileInputFormat.getInputPaths` to return the full string as a single `Path`, where Hadoop's `Path.normalizePath` collapsed embedded `gs://` schemes to `gs:/` for URIs 2–N.
- Split paths before escaping so each URI is escaped individually and commas remain as proper delimiters.
- Added `GcsConnectorUtilTest` with 4 test cases covering single-path, multi-path (24 hourly URIs), multi-bucket, and round-trip correctness.

Fixes the `InvalidInputException` reported in [#scio-users](https://spotify.slack.com/archives/C047H3XMT/p1775765935618929) when using `readAllParquetAvro` with multiple GCS URIs after upgrading to sbt-spotify-data 0.15.4+.

## Test plan
- [x] `sbt "scio-parquet/testOnly com.spotify.scio.parquet.GcsConnectorUtilTest"` — 4/4 pass
- [x] `sbt "scio-parquet/test"` — 74/74 pass
- [x] Verified tests fail with the old `StringUtils.escapeString(path)` code (multi-path tests return length 1 instead of expected 24/2)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
+++ b/scio-parquet/src/main/scala/com/spotify/scio/parquet/GcsConnectorUtil.scala
@@ -71,7 +71,9 @@ private[parquet] object GcsConnectorUtil {
     // This is needed since `FileInputFormat.setInputPaths` validates paths locally and requires
     // the user's GCP credentials.
     GcsConnectorUtil.setCredentials(conf)
-    conf.set(FileInputFormat.INPUT_DIR, StringUtils.escapeString(path));
+    // Split comma-separated paths and escape each individually to preserve URI schemes (gs://)
+    val escaped = path.split(",").map(StringUtils.escapeString).mkString(",")
+    conf.set(FileInputFormat.INPUT_DIR, escaped)
 
     // It will interfere with credentials in Dataflow workers
     if (!ScioUtil.isLocalRunner(sc.options.getRunner)) {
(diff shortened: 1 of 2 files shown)


charles-tan:
**Context on the approach:** This fix mirrors exactly what `FileInputFormat.setInputPaths(Job, String)` does internally in Hadoop ([v0.15.3 code](https://github.com/spotify/scio/pull/5883/changes#r2795677371)) — it splits the comma-separated string into individual paths, escapes each one individually via `StringUtils.escapeString`, then rejoins with commas:

```java
// FileInputFormat.setInputPaths(Job, Path...)
Path path = inputPaths[0].getFileSystem(conf).makeQualified(inputPaths[0]);
StringBuilder str = new StringBuilder(StringUtils.escapeString(path.toString()));
for(int i = 1; i < inputPaths.length; i++) {
    str.append(StringUtils.COMMA_STR);
    path = inputPaths[i].getFileSystem(conf).makeQualified(inputPaths[i]);
    str.append(StringUtils.escapeString(path.toString()));
}
conf.set(INPUT_DIR, str.toString());
```

The previous code applied `StringUtils.escapeString` to the entire comma-separated string, which escaped the delimiter commas themselves — causing `FileInputFormat.getInputPaths` to return the whole string as a single `Path`. Hadoop's `Path.normalizePath` then collapsed `//` to `/` in the path component, turning embedded `gs://` into `gs:/` for all URIs after the first.

**Note:** Hadoop's `getPathStrings` is slightly smarter than a plain `split(",")` — it skips commas inside `{...}` glob braces (e.g. `gs://bucket/{a,b}/*.parquet`). This isn't needed for the reported use case (hourly GCS partition URIs), but worth noting if glob brace patterns are ever passed as input.

clairemcginty (approved):

Issue #5939, opened 2026-06-15
(Iceberg) Infer `keep` config from Magnolify RowType schema


clairemcginty:
Ideally the user shouldn't have to specify `keep`; it can be inferred from the Magnolify `RowType`.

To be even more opinionated, I'm not sure we need to support `drop` and `only`, since the case class already represents the desired projection.

clairemcginty:
fixed in #5942

PR #5942, opened 2026-06-15
(iceberg) Infer `keep` config from Magnolify RowType schema

clairemcginty:
Fix #5941 https://github.com/spotify/scio/issues/5939

Infers `keep` based on case class fields, since that presumably represents the desired projection.

It does mean we get rid of `drop`/`only` since they're mutually exclusive with `keep`.

Also updates Scaladoc

### DIFF
--- a/scio-managed/src/main/scala/com/spotify/scio/iceberg/IcebergIO.scala
+++ b/scio-managed/src/main/scala/com/spotify/scio/iceberg/IcebergIO.scala
@@ -110,26 +110,26 @@ final case class IcebergIO[T: RowType: Coder](table: String, catalogName: Option
 object IcebergIO {
   case class ReadParam private (
     catalogProperties: Map[String, String] = ReadParam.DefaultCatalogProperties,
-    configProperties: Map[String, String] = ReadParam.DefaultConfigProperties,
+    configProperties: Map[String, String] = ReadParam.DefaultHadoopConfigProperties,
     filter: String = ReadParam.DefaultFilter
   )
 
   object ReadParam {
     val DefaultCatalogProperties: Map[String, String] = null
-    val DefaultConfigProperties: Map[String, String] = null
+    val DefaultHadoopConfigProperties: Map[String, String] = null
     val DefaultFilter: String = null
 
     implicit val configMap: ConfigMap.ConfigMapType[ReadParam] = ConfigMap.gen[ReadParam]
   }
   case class WriteParam private (
     catalogProperties: Map[String, String] = WriteParam.DefaultCatalogProperties,
-    configProperties: Map[String, String] = WriteParam.DefaultConfigProperties,
+    configProperties: Map[String, String] = WriteParam.DefaultHadoopConfigProperties,
     triggeringFrequencySeconds: Option[Int] = None,
     directWriteByteLimit: Option[Int] = None
   )
   object WriteParam {
     val DefaultCatalogProperties: Map[String, String] = null
-    val DefaultConfigProperties: Map[String, String] = null
+    val DefaultHadoopConfigProperties: Map[String, String] = null
     val DefaultTriggeringFrequencySeconds: Int = -1
     val DefaultDirectWriteByteLimit: Int = -1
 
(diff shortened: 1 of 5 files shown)


scio-managed/src/main/scala/com/spotify/scio/iceberg/syntax/IcebergScioContextSyntax.scala
```diff
@@ -25,20 +25,32 @@ import magnolify.beam.RowType
...
+   * https://docs.cloud.google.com/dataflow/docs/guides/managed-io-iceberg
    */
   def iceberg[T: Coder](
     table: String,
     catalogName: String = null,
     catalogProperties: Map[String, String] = IcebergIO.ReadParam.DefaultCatalogProperties,
-    configProperties: Map[String, String] = IcebergIO.ReadParam.DefaultConfigProperties,
-    keep: List[String] = IcebergIO.ReadParam.DefaultKeep,
-    drop: List[String] = IcebergIO.ReadParam.DefaultDrop,
+    configProperties: Map[String, String] = IcebergIO.ReadParam.DefaultHadoopConfigProperties,
```
clairemcginty: this is very trivial but I renamed `DefaultConfigProperties` -> `DefaultHadoopConfigProperties` just for clarity, because I personally found this option very confusing in combination with `catalogProperties` 😅

build.sbt
```diff
@@ -1143,15 +1143,19 @@ lazy val `scio-managed` = project
...
       "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
       "org.apache.beam" % "beam-sdks-java-managed" % beamVersion,
-      "com.spotify" %% "magnolify-beam" % magnolifyVersion
+      "com.spotify" %% "magnolify-beam" % magnolifyVersion,
+      "com.spotify" %% "magnolify-shared" % magnolifyVersion,
+      "com.softwaremill.magnolia1_2" %% "magnolia" % magnoliaVersion
       // test
-    )
+    ),
+    mimaPreviousArtifacts := Set.empty // Todo remove after 0.15.6
```
clairemcginty: `scio-managed` has never been published so there are no previous artifacts to compare against - thus, we can make all the breaking changes we want :D

scio-managed/src/main/scala/com/spotify/scio/iceberg/syntax/IcebergScioContextSyntax.scala
regadas: I guess on scio we could rename `configProperties` as well to `hadoopConfigProperties`

scio-managed/src/main/scala/com/spotify/scio/iceberg/IcebergIO.scala
```diff
@@ -74,10 +76,21 @@ final case class IcebergIO[T: RowType: Coder](table: String, catalogName: Option
...
+        if (field.getType.getTypeName == BSchema.TypeName.ROW)
+          leafFieldNames(field.getType.getRowSchema, name)
+        else
+          List(name)
+      }
+
     val b = Map.newBuilder[String, AnyRef]
     b += ("table" -> table)
     catalogName.foreach(name => b += ("catalog_name" -> name))
+    b += ("keep" -> leafFieldNames(rowType.schema))
```
regadas: are nested fields `a.b.c` supported in writes as well?

clairemcginty: AFAIK it's the same behavior as reads, so it should... however I noticed that the spec for whether we need `keep` for writes at all is a bit unclear actually 

From https://docs.cloud.google.com/dataflow/docs/guides/managed-io-iceberg#iceberg-write:

```
keep | list[str] | A list of field names to keep in the input record. All other fields are dropped before writing. Is mutually exclusive with 'drop' and 'only'.
```

However, the Beam IcebergIO [spec](https://github.com/apache/beam/blob/v2.74.0/sdks/java/io/iceberg/src/main/java/org/apache/beam/sdk/io/iceberg/IcebergIO.java#L198) seems to regard it as optional for writes:

```
 * <h4>Pre-filtering Options</h4>
 *
 * <p>Some use cases may benefit from filtering record fields right before the write operation. For
 * example, you may want to provide meta-data to guide records to the right destination, but not
 * necessarily write that meta-data to your table. Some light-weight filtering options are provided
 * to accommodate such cases, allowing you to control what actually gets written (see <b>{@code
 * drop}</b>, <b>{@code keep}</b>, <b>{@code only}</b>}).
 ```

intuitively I would expect that if you don't specify a `keep` config it would just write all fields but.. I'll try to quickly test this out on some real tables and check

clairemcginty: okay turns out writes do _not_ support `keep`, and the default behavior is to write all fields - so, let's only set `keep` for reads 👍

kellen (approved):

PR #5951, opened 2026-06-25
(WIP) Test Parquet format interop with old and new list encodings

clairemcginty:
WIP: `be able to read data written with .saveAsTypedParquetFile with new list encoding` is currently failing.

### DIFF
--- a/scio-parquet/src/test/scala/com/spotify/scio/parquet/ParquetFormatInteropTest.scala
+++ b/scio-parquet/src/test/scala/com/spotify/scio/parquet/ParquetFormatInteropTest.scala
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2026 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


scio-parquet/src/test/scala/com/spotify/scio/parquet/ParquetFormatInteropTest.scala
```diff
@@ -0,0 +1,149 @@
...
+          .saveAsTypedParquetFile(dir.toString)
+      )
+
+      runWithRealContext()(
+        _.parquetAvroFile[GenericRecord](s"$dir/*.parquet", projection = pt.avroSchema)
+          .map(identity) should containInAnyOrder(genericRecords)
+      )
+  }
+
+  it should "be able to read data written with .saveAsTypedParquetFile with new list encoding" in withTempDir {
```
clairemcginty, line 130: this test is failing due to an issue with how `pt.avroSchema` is constructed (see https://github.com/spotify/scio/issues/5952#issuecomment-4801435212)

PR #5986, opened 2026-07-24
Expand IcebergIO write API to cover all write options

clairemcginty:
Expands iceberg write API to support

- sort order
- partition spec
- arbitrary other options you want to pass to the ManagedIO config map, ie. `distribution_mode`

+ Renames `configProperties` to `writeProperties` to match Iceberg convention https://iceberg.apache.org/docs/latest/configuration/#write-properties

### DIFF
--- a/integration/src/test/scala/com/spotify/scio/iceberg/IcebergIOIT.scala
+++ b/integration/src/test/scala/com/spotify/scio/iceberg/IcebergIOIT.scala
@@ -17,6 +17,7 @@
 package com.spotify.scio.iceberg
 
 import com.dimafeng.testcontainers.{ForAllTestContainer, GenericContainer}
+import com.spotify.scio.parquet.BeamInputFile
 import com.spotify.scio.testing.PipelineSpec
 import magnolify.beam._
 import org.apache.iceberg.catalog.{Namespace, TableIdentifier}
@@ -28,7 +29,15 @@ import org.apache.iceberg.types.Types.{
   StringType,
   StructType
 }
-import org.apache.iceberg.{CatalogProperties, CatalogUtil, PartitionSpec, Schema}
+import org.apache.iceberg.{
+  CatalogProperties,
+  CatalogUtil,
+  NullOrder,
+  PartitionSpec,
+  Schema,
+  SortDirection
+}
+import org.apache.parquet.hadoop.ParquetFileReader
 import org.testcontainers.containers.wait.strategy.HostPortWaitStrategy
 
 import java.time.Duration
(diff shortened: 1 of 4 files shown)


scio-managed/src/main/scala/com/spotify/scio/iceberg/syntax/IcebergSCollectionSyntax.scala
```diff
@@ -48,19 +55,24 @@ class IcebergSCollectionSyntax[T: RowType: Coder](self: SCollection[T]) {
     table: String,
     catalogName: String = null,
     catalogProperties: Map[String, String] = IcebergIO.WriteParam.DefaultCatalogProperties,
-    hadoopConfigProperties: Map[String, String] =
-      IcebergIO.WriteParam.DefaultHadoopConfigProperties,
+    writeProperties: Map[String, String] = IcebergIO.WriteParam.DefaultWriteProperties,
+    sortFields: List[String] = IcebergIO.WriteParam.DefaultSortFields,
+    partitionFields: List[String] = IcebergIO.WriteParam.DefaultPartitionFields,
+    extraConfigProperties: Map[String, AnyRef] = IcebergIO.WriteParam.DefaultExtraConfigProperties,
```
clairemcginty, line 62: I thought this was more flexible than adding a dedicated named param for every new option added to the Iceberg write api (i.e. `distribution_mode`, `autosharing`, etc...)

regadas: Yeah might be better this way indeed till the API stabilizes

regadas (approved):

integration/src/test/scala/com/spotify/scio/iceberg/IcebergIOIT.scala
```diff
@@ -104,4 +116,104 @@ class IcebergIOIT extends PipelineSpec with ForAllTestContainer {
...
+          reader.getFooter.getBlocks.asScala.foreach { block =>
+            block.getColumns.asScala.foreach { col =>
+              val hasBloom = col.getBloomFilterOffset > 0
+              col.getPath.toDotString match {
+                case "b" =>
+                  // flip assertion once https://github.com/apache/beam/pull/39250/ is release in Beam 2.76
+                  withClue(
+                    "Iceberg writeProperties are not supported in Beam versions <= 2.76. Once Beam is upgraded, flip this assertion to `true`."
+                  ) {
+                    hasBloom shouldBe false
```
clairemcginty: This won't work until the next Beam release w/ https://github.com/apache/beam/pull/39250/ ; I did validate that this test passes when using a locally built Beam snapshot 👍

regadas (approved):

scio-managed/src/main/scala/com/spotify/scio/iceberg/syntax/IcebergSCollectionSyntax.scala
```diff
@@ -48,19 +56,24 @@ class IcebergSCollectionSyntax[T: RowType: Coder](self: SCollection[T]) {
     table: String,
     catalogName: String = null,
     catalogProperties: Map[String, String] = IcebergIO.WriteParam.DefaultCatalogProperties,
-    hadoopConfigProperties: Map[String, String] =
```
shnapz, line 51: this is a breaking change. We assume that there are not many affected users? Should we add some doc though?

clairemcginty: yeah... not ideal, but AFAIK there are no users of this API. We weren't even publishing the scio-managed artifact until about a month ago, in [0.15.7](https://github.com/spotify/scio/releases/tag/v0.15.7) 😅

shnapz (approved):

PR #5987, opened 2026-07-24
Add wrapper for setRowSchema method

clairemcginty:
a few Beam transforms require a Schema to be set on the SCollection; though the method `SCollection#def setSchema(schema: com.spotify.scio.schemas.Schema[T])` already exists, if you're working with magnolify-beam, calling RowType#schema gives you a `org.apache.beam.sdk.schemas.Schema`, not a  `com.spotify.scio.schemas.Schema`.


shnapz (approved):
lgtm

PR #5993, opened 2026-08-11
Lower row group size default to 512MB

clairemcginty:
Lower row group size default to 512MB to relieve read-time memory pressure.

Note: Row group size is very much a tradeoff and there's no blanket one-size-fits-all best option. Each row group adds a bit of extra overhead to the file footer/row group dictionaries, however, for low-to-moderate cardinality datasets, the default compression algorithm (ZSTD) largely reduces any storage overhead. 

Possibly we could go even smaller--Iceberg defaults to [128 MB row groups](https://iceberg.apache.org/docs/latest/configuration/#write-properties), for example. But I think this PR is a good starting point. In practice, without explicit configuration, Parquet writes over Dataflow seems to produce files in the realm of 200-300MB compressed (and 2-3x that size uncompressed)--thus, this change would result in a modest increase from 1 row group per file to 2.

(Another option is to log a warning on scio-smb writes that a smaller row group size is recommended, especially if the schema is narrow.)


clairemcginty:
moved this over to my fork of Scio: https://github.com/spotify/scio/pull/5994

PR #6002, opened 2026-09-18
Fix fat jar merge failure from Beam 2.76 envoy proto collision

shnapz:
- Exclude `io.envoyproxy.controlplane:api` build-wide — pulled in transitively by Beam 2.76.0, it bundles **unshaded** `com.google.api.*` / `com.google.rpc.*` protobuf classes that collide with `proto-google-common-protos`
- Discard conflicting OSGi bundle manifests in the `scio-repl` assembly
- Pin `bcprov-jdk18on` to 1.82 so every module resolves one BouncyCastle version
- Fixes `Publish Artifacts` → `Package repl` on `main`: 18 merge errors → 0

## Root cause

#5996 upgraded Beam 2.74.0 → 2.76.0. In 2.76.0, [apache/beam#39021](https://github.com/apache/beam/pull/39021) made GCP IO depend on `beam-sdks-java-io-components`, which drags in `io.envoyproxy.controlplane:api`:

```
scio-google-cloud-platform
  └─ beam-sdks-java-io-google-cloud-platform:2.76.0
       └─ beam-sdks-java-io-components:2.76.0        ← new edge in 2.76.0
            └─ io.envoyproxy.controlplane:api:1.0.49  ← bundles google protos
```

That jar compiles `google/api/http.proto`, `google/api/annotations.proto` and `google/rpc/status.proto` into `com/google/api/**` and `com/google/rpc/**` instead of depending on `proto-google-common-protos` — duplicating 17 classes including `com.google.rpc.Status`, `HttpRule` and `AnnotationsProto`. The copies aren't interchangeable: envoy's are protobuf **3.x** gencode (`extends GeneratedMessageV3`), the real ones are **4.x** (`extends GeneratedMessage`). Since protobuf 4.33.2 still ships the `GeneratedMessageV3` shim, the wrong copy loads silently rather than failing loudly.

The 18th error is unrelated: `bcprov-jdk18on` and `jspecify` both ship `META-INF/versions/9/OSGI-INF/MANIFEST.MF`. These are OSGi bundle descriptors, meaningless in an uber jar, so they're discarded — matching the existing `module-info.class` handling. The rule covers any version directory since bcprov ships them for Java 9/11/15/21.

## bcprov version split

Noticed while debugging the above: `bcprov-jdk18on` resolved to 1.82 in `scio-core` / `scio-extra` / `scio-smb` but 1.80 in `scio-google-cloud-platform` / `scio-repl`. The cause is `hadoop-common`, which requires 1.82 and is `Provided` in `scio-core` — `Provided` doesn't propagate transitively, so modules that don't pull hadoop themselves fall back to `google-cloud-spanner`'s 1.80 (presence of `hadoop-common` correlates exactly with getting 1.82).

Only one version was ever on a given classpath, so this caused no collision — but a crypto provider that silently varies by module is worth pinning. Pinned to 1.82, the higher of the two, so no module is downgraded.

## Why exclusion rather than a merge strategy

Both jars are on the classpath of **scio-core, scio-google-cloud-platform, scio-extra, scio-smb and scio-repl** — so this is latent classpath shadowing for every Scio user, not just a packaging problem. The duplicated `com.google.rpc.Status` is referenced by BigQuery Storage, Bigtable, Spanner and `io.grpc.protobuf.StatusProto`, on error-handling and retry paths.

Beam needs envoy only for `EnvoyRateLimiterFactory`; GCP IO uses just `AdaptiveThrottler` from `io-components`, and nothing in Scio reaches the rate limiter — so the jar is dead weight. Excluding it fixes all five modules at once. A `com/google/**` merge strategy would only paper over `scio-repl` and would permanently silence this class of bug.

## Changes

1. **`project/Exclude.scala`** — add `envoyControlPlane` rule, following the existing `beamKafka` precedent
2. **`build.sbt`** — apply it via `excludeDependencies` in `commonSettings`
3. **`build.sbt`** — discard `META-INF/versions/*/OSGI-INF/MANIFEST.MF` in the `scio-repl` merge strategy
4. **`build.sbt`** — add `bouncyCastleVersion` and pin `bcprov-jdk18on` via `dependencyOverrides`

## Test plan

- [x] `scio-repl/assembly` succeeds — 18 merge errors → 0
- [x] envoy jar gone from all 5 affected module classpaths; `proto-google-common-protos` retained
- [x] Assembled jar ships the protobuf 4.x `com.google.rpc.Status` (gencode version guard present); no OSGi manifests
- [x] Runtime — REPL boots from the jar, GCP/BigQuery/`ScioContext` initialize, and `classOf[com.google.rpc.Status].getSuperclass` is `com.google.protobuf.GeneratedMessage`
- [x] `bcprov-jdk18on` resolves to 1.82 in all 5 modules (was 1.82/1.80 split)
- [x] `sbt compile` all modules — 0 errors
- [x] CI green

## Note

`io/envoyproxy/**` classes still appear in the jar (2980, from `grpc-xds`), which bundles envoy xDS protos for gRPC load balancing. That jar carries no `com/google/api` or `com/google/rpc` classes, so it's unrelated and expected.


ccamenares-spotify (approved):
LGTM

clairemcginty (approved):

